PR #22, opened 2026-02-11
fix: locate builtin skills directory

24f2006619:
## Summary
Fix builtin skills discovery so `skills install-builtin` and `skills list-builtin` work from a repo checkout or installed binary.

## Changes
- Added a helper to resolve builtin skills dir:
  - prefer `./skills` in a source checkout
  - fallback to `skills/` next to the executable

## Why
Current paths point to non-existent locations (`./picoclaw/skills`, `~/.picoclaw/picoclaw/skills`), so builtin skills are not found.

## Testing
Not run (path resolution only).

### DIFF
--- a/cmd/picoclaw/main.go
+++ b/cmd/picoclaw/main.go
@@ -102,11 +102,7 @@ func main() {
 
 		workspace := cfg.WorkspacePath()
 		installer := skills.NewSkillInstaller(workspace)
-		// 获取全局配置目录和内置 skills 目录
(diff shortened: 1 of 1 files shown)


lxowalle:
Thank you for your PR. This change still doesn't seem to resolve the issue of not finding the skills when only the executable file is downloaded. We have an idea to package the built-in skills files into the executable using embed, and when running picoclaw onboard, read the files and install them to the specified location. This idea will be implemented later.

PR #24, opened 2026-02-11
Fix concurrency races in session manager and stabilize service lifecycles

Esubaalew:
This PR tries to addres concurrency and sessions issuses inside  session manager , cron  , and heartbeat services .  the first  bug like fix was race condtions issue that might happen when two requests could create two different sessions for the same key at the same time.

Key Changes
1. pkg/session/manager: 
race isssue: changed GetOrCreate to use a single Lock during the check-and-set phase, so avoiding  multiple goroutines from creating and overwriting the same session.

Non-blocking I/O: Refactored Save to snapshot session data under a RLock and perform disk I/O outside of the critical section. This prevents the entire manager from hanging during slow disk writes.

Atomic Writes: (CreateTemp -> Sync -> Rename) this can prevent data corruption/partial writes during system crashes.

2. pkg/cron/service.go & pkg/heartbeat.go: 
Idempotent Stop/Start: Fixed a bug where restarting a service would cause an immediate exit or panic due to a closed stopChan.

Channel Management: Start() now re-initializes the stopChan on every call, and Stop() safely handles nil/closed states to ensure reliability during repeated toggle operations.

Granular Locking: Refactored checkJobs and executeJobByID to ensure thread-safe state updates without holding locks across external callback executions.

most od the issues i thing i fixed might not be urgent fixes thy are thing that could likely casuse bug


lxowalle:
Hi @Esubaalew , Thanks for your pr. I encountered some issues, there were too many conflicts when merging this pr. Could you please test it on the latest code and submit again?

Esubaalew:
> Hi @Esubaalew , Thanks for your pr. I encountered some issues, there were too many conflicts when merging this pr. Could you please test it on the latest code and submit again?

Hey @lxowalle  I just did, thanks.

lxowalle:
Thanks for the pr!

Orgmar:
@Esubaalew Nice work on the race condition fixes! The atomic write approach with temp/sync/rename is a solid pattern for crash safety, and separating the disk I/O from the lock is a good call for throughput.

By the way, we're building the **PicoClaw Dev Group** to bring contributors closer together and push the project forward. If you're interested, send an email to `support@sipeed.com` with the subject `[Join PicoClaw Dev Group] + Your GitHub account` and we'll get you a Discord invite!

Esubaalew:
Thanks,  I’ve just sent the email.

PR #26, opened 2026-02-11
Enforce workspace directory boundaries for system tools

RinZ27:
Integrated a `restrict_to_workspace` option into the configuration to allow tools to access external files when necessary, particularly in containerized environments. While the default remains `true` for security, flipping this toggle lets tools reach any system path. Updated the `validatePath` helper and all filesystem/shell tools to respect this new setting. Both modes are now fully covered by the expanded test suite.

Technical highlights:
- Added `restrict_to_workspace` toggle in `config.json` and `AgentDefaults`.
- Updated `validatePath` to skip prefix checks when restriction is disabled.
- Modified `read_file`, `write_file`, `list_dir`, `append_file`, `edit_file`, and `exec` to support the toggle.
- Included comprehensive test cases for restricted and unrestricted access.


lxowalle:
Hi, this is a feature worth supporting. However, it also needs to allow `picoclaw` to operate on files outside the workspace, such as modifying certain third-party configurations within a container where there's no concern about damaging the system. So this needs to be improved by making this feature an option that can be disabled via `config.json`.

yinwm:
picoclaw should support restrict_to_workspace / restrict option in `config.json`. 

If true, only works in workspace; otherwise can use tools in anywhere.

Could you please support this option ?  @RinZ27

RinZ27:
Added the `restrict_to_workspace` toggle in `config.json` per the feedback. Default is still `true` for safety, but flipping it to `false` lets tools reach anywhere now—which I noticed was a requirement for containerized setups. Updated the tests too so they cover both modes properly. @yinwm @lxowalle

lxowalle:
Thank you for this PR, I've tested it and it works well. However, I noticed the `filesystem_test.go` file, which seems irrelevant to the functionality and appears to be AI-generated. I removed this file after merging the code.

RinZ27:
@lxowalle I included the tests to verify the workspace restriction logic during development, but I understand if they don't fit the project's current scope. Glad the core feature is working as expected!

Zepan:

PR #30, opened 2026-02-11
fix(agent): use atomic.Bool for AgentLoop.running to prevent data race

DevEverything01:
## Summary

- Replace bare `bool` with `sync/atomic.Bool` for the `AgentLoop.running` field

## Problem

`Run()` reads `running` in a loop on one goroutine, while `Stop()` writes it from another. Without synchronization this is a data race per the Go memory model.

## Changes

| Before | After |
|--------|-------|
| `running bool` | `running atomic.Bool` |
| `al.running = true` | `al.running.Store(true)` |
| `for al.running {` | `for al.running.Load() {` |
| `al.running = false` | `al.running.Store(false)` |

1 file changed, 5 insertions(+), 5 deletions(-)

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `go test ./...` passes

### DIFF
--- a/pkg/agent/loop.go
+++ b/pkg/agent/loop.go
@@ -14,6 +14,7 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/sipeed/picoclaw/pkg/bus"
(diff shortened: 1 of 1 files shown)


Zepan:

PR #48, opened 2026-02-12
Add provider field support for explicit provider selection

victorhdchagas:
This is an automated message generated during a code contribution process.

## Issue/Demand
When using the `provider` field in the configuration file, the picoclaw agent was unable to use the configured API keys unless the model name included the provider prefix (e.g., `groq/llama-3.1-8b-instant`). This prevented users from using clean model names like `llama-3.1-8b-instant` even when explicitly setting the provider in the config.

## Problem Summary
The `agent.defaults.provider` field existed in the JSON configuration schema but was not being used by the code. The provider was only detected from the model name string, which caused issues because the prefix was being sent to the API as part of the model name, resulting in errors like:
```
no API key configured for model: llama-3.1-8b-instant
```

## Solution Implemented
1. Added `Provider` field to `AgentDefaults` struct in `pkg/config/config.go`
2. Modified `CreateProvider()` function in `pkg/providers/http_provider.go` to:
   - **First**: Check `cfg.Agents.Defaults.Provider` if explicitly set
   - **Then**: Fallback to existing model name detection logic (backward compatible)
   - Support aliases: `openai`/`gpt`, `anthropic`/`claude`, `google`/`gemini`, `zhipu`/`glm`

## Files Changed
- `pkg/config/config.go`: Added `Provider` field to struct and default config
- `pkg/providers/http_provider.go`: Modified `CreateProvider()` to use explicit provider field

## Testing
✅ Successfully reads API key from `providers.groq.api_key` when `provider: "groq"` is set
✅ Correct API base: `https://api.groq.com/openai/v1`
✅ Model name sent without prefix: `llama-3.1-8b-instant`
✅ Successfully makes API calls (only error was rate limit, not configuration)
✅ Existing tests continue to pass

## Benefits
- Users can now use the `provider` field from config as originally intended
- No need to prefix model names with provider in the config file
- Fully backward compatible with existing configurations
- Supports all providers: groq, openai, anthropic, openrouter, zhipu, gemini, vllm

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -24,6 +24,7 @@ type AgentsConfig struct {
 
 type AgentDefaults struct {
 	Workspace         string  `json:"workspace" env:"PICOCLAW_AGENTS_DEFAULTS_WORKSPACE"`
+	Provider          string  `json:"provider" env:"PICOCLAW_AGENTS_DEFAULTS_PROVIDER"`
 	Model             string  `json:"model" env:"PICOCLAW_AGENTS_DEFAULTS_MODEL"`
 	MaxTokens         int     `json:"max_tokens" env:"PICOCLAW_AGENTS_DEFAULTS_MAX_TOKENS"`
 	Temperature       float64 `json:"temperature" env:"PICOCLAW_AGENTS_DEFAULTS_TEMPERATURE"`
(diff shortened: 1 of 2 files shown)


lahcim:
Great change! Can we take this in?

victorhdchagas:
> Great change! Can we take this in?

Yes! It's my first contribution... :3rd_place_medal:

Zepan:

PR #52, opened 2026-02-12
feat: Telegram proxy support, flexible allow_from, and Windows shell fix

siciyuan404:
### 1. Telegram Proxy Support
Add `proxy` config field for Telegram channel, enabling HTTP/SOCKS proxy for users behind firewalls (e.g. in China).

```json
"telegram": {
  "enabled": true,
  "token": "...",
  "proxy": "http://127.0.0.1:7890",
  "allow_from": ["123456789"]
}
```

Uses `telego.WithHTTPClient` so all Telegram API requests route through the proxy. No proxy configured = original behavior unchanged.

### 2. Flexible `allow_from` Matching
**Problem**: `allow_from` only did exact string match, but Telegram senderID format is `123456|username`. Users had to know the exact compound format.

**Fix**: 
- `IsAllowed()` now matches against numeric ID, username, and @username
- New `FlexibleStringSlice` type allows both strings and numbers in JSON config

All these formats now work:
```json
"allow_from": [123456789]
"allow_from": ["123456789"]
"allow_from": ["username"]
"allow_from": ["@username"]
```

### 3. Windows Shell Compatibility
**Problem**: `exec` tool hardcoded `sh -c` which doesn't exist on Windows, causing all tool calls to fail silently.

**Fix**: Use `cmd /c` on Windows, `sh -c` on other platforms via `runtime.GOOS` check.

## Files Changed
- `config.example.json` - Add proxy field example
- `pkg/config/config.go` - Add `Proxy` field + `FlexibleStringSlice` type
- `pkg/channels/telegram.go` - Proxy support via `telego.WithHTTPClient`
- `pkg/channels/base.go` - Flexible `IsAllowed()` matching
- `pkg/tools/shell.go` - Windows `cmd /c` compatibility

### DIFF
--- a/config.example.json
+++ b/config.example.json
@@ -80,6 +80,15 @@
     "vllm": {
       "api_key": "",
       "api_base": ""
+    },
+    "nvidia": {
+      "api_key": "nvapi-xxx",
+      "api_base": "",
+      "proxy": "http://127.0.0.1:7890"
+    },
+    "moonshot": {
+      "api_key": "sk-xxx",
+      "api_base": ""
     }
   },
   "tools": {
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -125,9 +125,9 @@ type DingTalkConfig struct {
 }
 
 type SlackConfig struct {
-	Enabled  bool     `json:"enabled" env:"PICOCLAW_CHANNELS_SLACK_ENABLED"`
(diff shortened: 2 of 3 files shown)


lxowalle:
Thanks for the pr

lxowalle:
I merged and pushed locally, so I closed this pr

PR #64, opened 2026-02-12
fix(heartbeat): resolve bug where service could never start

DevEverything01:
## Summary

- **HeartbeatService.Start() never launched the goroutine** due to a logic error in state detection: `running()` checks whether `stopChan` is closed, but a newly created service has an open channel, so `running()` always returned `true`, causing `Start()` to exit early thinking it was already running.
- Introduced a `started bool` field to separate "has been started" from "has not been stopped", which also prevents a potential double-close panic when `Stop()` is called more than once.

| State | `started` | `stopChan` | Behavior |
|-------|-----------|------------|----------|
| New | `false` | open | `Start()` proceeds, launches goroutine |
| Running | `true` | open | `Start()` is idempotent (no-op) |
| Stopped | `false` | closed | `Stop()` is idempotent, `runLoop` exits |

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./pkg/heartbeat/...` passes
- [x] Verify heartbeat goroutine launches after `Start()` (enable heartbeat in config, observe periodic heartbeat logs)
- [x] Verify `Stop()` is safe to call multiple times without panic

### DIFF
--- a/pkg/heartbeat/service.go
+++ b/pkg/heartbeat/service.go
@@ -14,6 +14,7 @@ type HeartbeatService struct {
 	interval    time.Duration
 	enabled     bool
 	mu          sync.RWMutex
+	started     bool
 	stopChan    chan struct{}
 }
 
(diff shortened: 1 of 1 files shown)


lxowalle:
Thanks for the pr!

PR #87, opened 2026-02-12
Refactor provider resolution, add Zen provider, and gateway service install

euhenriquecosta:
## Summary
This PR refactors provider selection to a registry-based flow, integrates the Zen provider, hardens provider resolution rules, and adds an automated `systemd` installer command for the gateway.

## Problem
Provider creation was centralized in a large monolithic factory with duplicated condition branches and broad model matching heuristics. This increased change risk for each new provider and could route to the wrong backend in production. Gateway service installation was also fully manual.

## Root Cause
- Provider resolution logic was spread across explicit provider branches and model-name fallbacks in one function.
- Matching relied on ambiguous string checks, which could misclassify models.
- No first-class CLI flow existed to install and enable the gateway as a Linux service.

## Changes
- Replaced monolithic provider selection with a registry-based resolver.
- Added self-registration for built-in providers.
- Added Zen provider registration and config support.
- Hardened resolution behavior:
  - explicit `provider` now fails fast for unknown/unconfigured provider
  - model prefix matching tightened to reduce ambiguous auto-selection
  - removed unsafe Zen `opencode/` auto-prefix route
- Kept `CreateProvider` as a single delegated call to the registry.
- Added `picoclaw gateway install`:
  - validates Linux + root + `systemctl`
  - writes `/etc/systemd/system/picoclaw.service`
  - runs `daemon-reload` + `enable --now`
- Updated CLI help text for gateway command.
- Updated example config with `providers.zen`.

## User Impact
- Adding providers is now lower-risk and simpler.
- Production behavior is safer: explicit provider config errors are surfaced immediately.
- Zen usage is supported with explicit provider configuration and real model IDs.
- Gateway deployment on VPS is faster with one command.

## Validation
- `go test ./pkg/providers`
- `go test ./...`
- Live Zen integration checks:
  - direct `/models` and `/chat/completions` with a `-free` model
  - end-to-end `picoclaw agent -m ...` with `provider=zen` and `model=kimi-k2.5-free`

### DIFF
--- a/cmd/picoclaw/main.go
+++ b/cmd/picoclaw/main.go
@@ -882,6 +882,7 @@ func statusCmd() {
 		hasOpenAI := cfg.Providers.OpenAI.APIKey != ""
 		hasGemini := cfg.Providers.Gemini.APIKey != ""
 		hasZhipu := cfg.Providers.Zhipu.APIKey != ""
+		hasZAI := cfg.Providers.ZAI.APIKey != ""
 		hasGroq := cfg.Providers.Groq.APIKey != ""
 		hasVLLM := cfg.Providers.VLLM.APIBase != ""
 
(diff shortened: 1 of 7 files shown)


Leeaandrob:
@Zepan This PR refactors provider resolution and adds a Zen provider. However, PR #213 (Provider Refactor by Protocol — roadmap #283) is a more comprehensive approach to the same problem and is already In Progress on the board.

**Recommendation: Close in favor of #213.** The provider refactor in #213 supersedes this approach. The Zen provider can be added as a follow-up after #213 lands.

Leeaandrob:
@euhenriquecosta take look my comments.

Leeaandrob:
Please open another pr just adding the Zen as provider. We are merge the https://github.com/sipeed/picoclaw/pull/213 with the focus in to have a standard and package for providers. @euhenriquecosta

PR #114, opened 2026-02-13
feat: send rich text message to feishu instead of plain text

FortuneW:
## Summary
- Send rich text (post) message to feishu instead of plain text


mengzhuo:
Please resolve conflicts.

PR #130, opened 2026-02-13
fix: correct index bug in extractCodeBlocks and extractInlineCodes

MHCP000:
The previous implementation used len(codes)-1 in ReplaceAllStringFunc, which caused all placeholders to point to the same (last) index. This resulted in only the last code block being displayed correctly when multiple code blocks were present in a message.

Now using a proper counter variable that increments for each match, ensuring each code block gets a unique placeholder index.

### DIFF
--- a/pkg/channels/telegram.go
+++ b/pkg/channels/telegram.go
@@ -470,8 +470,11 @@ func extractCodeBlocks(text string) codeBlockMatch {
 		codes = append(codes, match[1])
 	}
 
+	i := 0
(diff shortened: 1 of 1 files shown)


Orgmar:
@MHCP000 That index bug was a sneaky one, all placeholders pointing to the last code block must have been confusing for users. Good find and clean fix with the counter variable!

By the way, we have the **PicoClaw Dev Group** on Discord for contributors. Want to join? Just email `support@sipeed.com` with the subject `[Join PicoClaw Dev Group] + Your GitHub account` and we'll send the invite your way.

PR #80, opened 2026-02-12
feat: add Codex CLI provider for subprocess integration

Leeaandrob:
## Summary

- Add `CodexCliProvider` that wraps `codex exec --json` as a subprocess, analogous to the existing `ClaudeCliProvider` pattern
- Add credential reader for `~/.codex/auth.json` with token expiry detection (based on file mtime + 1h)
- Integrate into provider factory: `provider: "codex-cli"` for subprocess mode, `auth_method: "codex-cli"` for API mode with local CLI tokens
- Fix tilde `~` expansion in workspace path for all CLI providers (`cfg.WorkspacePath()` instead of raw config value)

## New Files

| File | Description |
|------|-------------|
| `codex_cli_provider.go` | Subprocess wrapper: builds prompt, runs `codex exec --json`, parses JSONL events |
| `codex_cli_credentials.go` | Reads `~/.codex/auth.json`, estimates token expiry from file mtime |
| `codex_cli_provider_test.go` | 29 tests: JSONL parsing, prompt building, mock CLI, context cancel |
| `codex_cli_credentials_test.go` | 8 tests: valid/invalid/expired credentials, CODEX_HOME env |

## Configuration

```json
{
  "agents": {
    "defaults": {
      "provider": "codex-cli",
      "model": "codex-cli"
    }
  }
}
```

Or use Codex CLI tokens with the HTTP API provider:
```json
{
  "providers": {
    "openai": {
      "auth_method": "codex-cli"
    }
  }
}
```

## Test plan

- [x] All 37 Codex-related unit tests passing
- [x] `go vet ./...` clean
- [x] Tested locally: `picoclaw agent -m "test"` returns valid response with `provider: "codex-cli"`
- [x] Verified no regression on existing providers

PR #135, opened 2026-02-13
fix: allow claude CLI provider to run inside Claude Code sessions

Leeaandrob:
## Summary

- Remove `CLAUDECODE` env var from subprocess environment so Claude Code can be invoked as a subprocess from PicoClaw (or from within another Claude Code session)
- Parse stdout before checking exit code to handle stderr diagnostic noise without losing valid JSON output (same resilience pattern applied to codex CLI provider in #80)

## Problem

When PicoClaw's `claude-cli` provider is used inside a Claude Code session (or any environment where `CLAUDECODE` is set), the subprocess fails with:
```
Error: Claude Code cannot be launched inside another Claude Code session.
```

Additionally, if Claude Code writes diagnostic messages to stderr, the non-zero exit code would cause the provider to return an error even when valid JSON output exists on stdout.

## Changes

- `pkg/providers/claude_cli_provider.go`: Filter out `CLAUDECODE` env var + stdout-first parsing

## Test plan

- [x] All existing unit tests pass (`go test ./pkg/providers/... -run Claude`)
- [x] `gofmt` and `go vet` clean
- [x] Manually tested: PicoClaw → claude-cli → Claude Code → Context7 MCP → valid response

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -136,11 +136,11 @@ func TestDefaultConfig_WebTools(t *testing.T) {
 	cfg := DefaultConfig()
 
 	// Verify web tools defaults
-	if cfg.Tools.Web.Search.MaxResults != 5 {
-		t.Error("Expected MaxResults 5, got ", cfg.Tools.Web.Search.MaxResults)
+	if cfg.Tools.Web.Brave.MaxResults != 5 {
+		t.Error("Expected Brave MaxResults 5, got ", cfg.Tools.Web.Brave.MaxResults)
 	}
-	if cfg.Tools.Web.Search.APIKey != "" {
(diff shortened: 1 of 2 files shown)


Leeaandrob:
Closing this PR — the fixes here (CLAUDECODE env var removal + stderr resilience) will be included in an upcoming larger PR that refactors the CLI provider architecture entirely.

The new PR will:
- Replace `claude_cli_provider.go` with a unified `cli_runner.go` that handles both Claude CLI and Codex CLI
- Add proper `--session-id` + `--resume` support
- Add execution serialization queue
- Add `clearEnv` support (which covers the CLAUDECODE fix from this PR)
- Add image pipeline for CLI providers

All changes from this PR are preserved in the new implementation.

PR #141, opened 2026-02-13
Fix/legacy test scripts and readme

huaaudio:
There are some test scripts for web search still using the old class, edited to match the current handler.

The English README is also edited slightly to reflect the correct branding and website of Zhipu/Z.AI for the global market.


lxowalle:
Hi @Huaaudio ,thanks for the pr, this part has already been fixed.

PR #154, opened 2026-02-14
feat: add minimax provider support

Danieldd28:
feat: add minimax provider support

## Description

Adds support for MiniMax LLM provider.

### Changes
- Added `minimax` configuration to `pkg/config`.
- Implemented `MiniMaxProvider` in `pkg/providers` using the `/v1/text/chatcompletion_v2` endpoint.
- Updated `HTTPProvider` to support custom request suffixes.
- Added integration tests for MiniMax.

### Verification
- Verified manually with a live API key (handled valid and error responses correctly).

### DIFF
--- a/README.ja.md
+++ b/README.ja.md
@@ -223,7 +223,7 @@ picoclaw agent -m "What is 2+2?"
 
 ## 💬 チャットアプリ
 
-Telegram、Discord、QQ、DingTalk で PicoClaw と会話できます
+Telegram、Discord、QQ、DingTalk、LINE で PicoClaw と会話できます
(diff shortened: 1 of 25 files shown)


nikolasdehor:
There are now two PRs for MiniMax support: #416 (custom provider) and #633 (OpenAI-compatible endpoint). #633 is likely the simpler approach since MiniMax exposes an OpenAI-compatible API, reducing maintenance burden. Related: #416.

PixelTux:
@Danieldd28 could you check if this still makes sense after the refactor?

PR #155, opened 2026-02-14
refactor: remove unused min func

moveyield:
Go 1.21 adds built-in functions min and max. https://tip.golang.org/doc/go1.21
Therefore, we can directly remove our own implementations and use the built-in methods with the same names.


Orgmar:
@moveyield Clean refactor! Now that Go 1.21 has built in min/max there's no reason to keep custom implementations around. Less code to maintain is always a win.

We have the **PicoClaw Dev Group** on Discord for contributors. If you'd like to join, send an email to `support@sipeed.com` with the subject `[Join PicoClaw Dev Group] + Your GitHub account` and we'll send the invite!

Issue #149, opened 2026-02-14
在对话中增加快捷设置功能-使用特殊开头的消息进行配置设置


sdial:
现在的IM纯粹是交互功能，希望可以增加使用IM进行常用功能配置功能。
如：
/show model         查看当前使用model 
/show channel      查看当前使用channel
/list models           查看可用渠道
/switch model to model_name 切换当前渠道模型
/list channels         查看所有已配置的渠道
/switch channel to channel_name 切换渠道

nikolasdehor:
Good feature request. This overlaps with #162 (session management commands) and could share the same in-chat command dispatch infrastructure.

For the `/switch model` and `/switch channel` commands specifically, the current config is loaded once at startup in `loadConfig()` and passed to the agent loop. Supporting runtime model switching would require either:
1. Making the agent config mutable and adding a reload path, or
2. Spawning a new agent loop with the updated config (cleaner but heavier)

The `/show` read-only commands are simpler -- they just need access to the current config struct. These could be implemented as a lightweight middleware in `BaseChannel.handleMessage()` that intercepts `/`-prefixed messages before they reach the agent.

PR #164, opened 2026-02-14
Implemented Telegram Info Commands (/show, /list) #149

biisal:
# feat: Implemented Telegram Info Commands (/show, /list)

# Problem
Previously, all messages sent to the Telegram bot including commands like /start and /help  were passed directly to the model without any command handling. 
Issue #149 requested dedicated commands to inspect PicoClaw's current configuration.

# Changes
Added command handlers for the following:

- /start - checks if the bot is alive
- /help - lists all available commands
- /list [models|channels] - shows all configured models and channels
- /show [model|channel] - shows the currently active model and channel

# Improvements

- Replaced the manual for/select goroutine loop with telegohandler.BotHandler, which manages update dispatching internally
- Commands are now registered declaratively via th.CommandEqual(...) instead of manual message.Text comparisons
- Errors returned from handlers are surfaced instead of being silently ignored
- Graceful shutdown via bh.Stop() on context cancellation
- Defined a `TelegramCommander` interface for command handlers, making it easy to add new commands in the future
- removed unnecessary use of fmt.Sprintf from telegram message handler

### DIFF
--- a/pkg/channels/manager.go
+++ b/pkg/channels/manager.go
@@ -48,7 +48,7 @@ func (m *Manager) initChannels() error {
 
 	if m.config.Channels.Telegram.Enabled && m.config.Channels.Telegram.Token != "" {
 		logger.DebugC("channels", "Attempting to initialize Telegram channel")
-		telegram, err := NewTelegramChannel(m.config.Channels.Telegram, m.bus)
+		telegram, err := NewTelegramChannel(m.config, m.bus)
 		if err != nil {
 			logger.ErrorCF("channels", "Failed to initialize Telegram channel", map[string]interface{}{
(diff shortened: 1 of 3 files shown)


lxowalle:
Thanks for the pr.

Orgmar:
@biisal Great job implementing the /show and /list commands for Telegram! Having proper command handling instead of passing everything straight to the model makes the bot feel much more polished.

We have a **PicoClaw Dev Group** on Discord where contributors hang out and collaborate. If you'd like to join, email `support@sipeed.com` with the subject `[Join PicoClaw Dev Group] + Your GitHub account` and we'll send you the invite!

PR #172, opened 2026-02-14
feat(docker): Added curl for Docker image

mymmrac:
This PR adds `curl` into Docker image as it doesn't come with Alpine by default and agent may use it

Other small changes:
- Updated Alpine image from 3.21 to 3.23 (latest)
- Added IDEs in `.gitignore`
- Updated `github.com/openai/openai-go/v3`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,7 @@ coverage.html
 ralph/
 .ralph/
 tasks/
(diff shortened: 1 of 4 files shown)


mengzhuo:
LGTM, thanks

PR #173, opened 2026-02-14
fix: resolve multiple bugs from code review #116

is-Xiaoen:
Fixes four issues identified in the community code review:

- Session persistence broken on Windows: session keys like "telegram:123456" contain ':', which is illegal in Windows filenames. filepath.Base() strips drive-letter prefixes on Windows, causing Save() to silently fail. Added sanitizeFilename() to replace invalid chars in the filename while keeping the original key in the JSON payload.

- HTTP client with no timeout: HTTPProvider used Timeout: 0 (infinite wait), which can hang the entire agent if an API endpoint becomes unresponsive. Set a 120s safety timeout.

- Slack AllowFrom type mismatch: SlackConfig used plain []string while every other channel uses FlexibleStringSlice, so numeric user IDs in Slack config would fail to parse.

- Token estimation wrong for CJK: estimateTokens() divided byte length by 4, but CJK characters are 3 bytes each, causing ~3x overestimation and premature summarization. Switched to utf8.RuneCountInString() / 3 for better cross-language accuracy.

Also added unit tests for the session filename sanitization.

Ref #116

### DIFF
--- a/pkg/session/manager.go
+++ b/pkg/session/manager.go
@@ -145,13 +145,27 @@ func (sm *SessionManager) TruncateHistory(key string, keepLast int) {
 	session.Updated = time.Now()
 }
 
+// sanitizeFilename converts a session key into a cross-platform safe filename.
+// Session keys use "channel:chatID" (e.g. "telegram:123456") but ':' is the
+// volume separator on Windows, so filepath.Base would misinterpret the key.
+// We replace it with '_'. The original key is preserved inside the JSON file,
+// so loadSessions still maps back to the right in-memory key.
+func sanitizeFilename(key string) string {
+	return strings.ReplaceAll(key, ":", "_")
+}
+
 func (sm *SessionManager) Save(key string) error {
 	if sm.storage == "" {
 		return nil
 	}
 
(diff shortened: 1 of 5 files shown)


is-Xiaoen:
Fixes four issues identified in the community code review:

pkg/session/manager.go
```diff
@@ -145,13 +145,33 @@ func (sm *SessionManager) TruncateHistory(key string, keepLast int) {
 	session.Updated = time.Now()
 }
 
+// sanitizeFilename converts a session key into a cross-platform safe filename.
+// Characters like ':' are valid on Linux but illegal on Windows; we replace
+// them so the same key works everywhere.  The original key is preserved inside
+// the JSON file, so loadSessions still maps back to the right in-memory key.
+func sanitizeFilename(key string) string {
```
mengzhuo, line 153: can we use filepath in standard library?

is-Xiaoen: Simplified — now uses \ for the colon (the only char that actually appears in session keys) and \ for validation. Dropped the manual replacer. Thanks for the review\!

mengzhuo:
LGTM, thanks

PR #195, opened 2026-02-15
feat: add picoclaw doctor diagnostic command

### LINKED ISSUE #39: Feature Request: Add `picoclaw doctor` command
Implement a health check and diagnostic tool inspired by OpenClaw's doctor system to validate configuration, check system health, and provide guided repair workflows.

### Proposed Functionality
1. **Configuration validation**
2. **Provider health checks**
3. **System health**
4. **Channel health**

### Example Usage
```bash
picoclaw doctor                    # Run all health checks
picoclaw doctor --section config   # Check only configuration
picoclaw doctor --json

### PR DESCRIPTION
**Pull request by mrbeandev**
## Summary
- Adds a new `picoclaw doctor` command for health diagnostics.
- Checks config validity, workspace access, provider setup, connectivity, tools, and channels.
- Provides clear pass/warn/fail output with actionable guidance.

## Why
- Helps users quickly troubleshoot setup issues before running agents/channels.
- Improves onboarding and reduces support/debug friction.

## Included
- New package: `pkg/doctor`
- CLI wiring in `cmd/picoclaw/main.go`
- Docker/test support updates

## Validation
- Added tests under `pkg/doctor/doctor_test.go`
- Manual run via:
  - `picoclaw doctor`
  - `docker compose run --rm picoclaw-doctor`

Closes #39

### CODE REVIEW

File: pkg/doctor/doctor.go [hunk 1]
```diff
@@ -0,0 +1,420 @@
...
+
+	fmt.Println("----------")
+	fmt.Printf("✅ %d passed  ⚠️ %d warnings  ❌ %d errors\n", okCount, warningCount, errorCount)
+	fmt.Println()
+
+	if errorCount > 0 {
+		fmt.Println("❌ Please fix the errors above before using picoclaw")
+		os.Exit(1)
+	} else if warningCount > 0 {
+		fmt.Println("⚠️  PicoClaw should work, but consider addressing the warnings")
```
johngibb: nitpick: extraneous space:
```suggestion
		fmt.Println("⚠️ PicoClaw should work, but consider addressing the warnings")
```

pkg/doctor/doctor.go
```diff
@@ -0,0 +1,420 @@
...
+	case StatusOK:
+		return "✅"
+	case StatusWarning:
+		return "⚠️"
+	case StatusError:
+		return "❌"
+	default:
+		return "❓"
+	}
+}
```
johngibb, line 38: Consider just making a string.
Rationale: fewer lines of code + zero value is invalid (not ok)
```suggestion
// Status represents the status of a check
type Status string

const (
	StatusOK      Status = "✅"
	StatusWarning Status = "⚠️"
	StatusError   Status = "❌"
)

func (s Status) String() string {
	if s == "" {
		return "❓"
	}
	return string(s)
}
```

pkg/doctor/doctor.go
```diff
@@ -0,0 +1,420 @@
...
+)
+
+// Check represents a single diagnostic check
+type Check struct {
+	Name    string
+	Status  Status
+	Message string
+	Details []string
+}
+
```
johngibb, line 23: Suggestion: A simple helper would reduce significant boilerplate and aid readability
at call sites:
```suggestion

func (c *Check) addDetail(s string, args ...any) {
	c.Details = append(c.Details, fmt.Sprintf(s, args...))
}

```

Example:
```go
check.Details = append(check.Details, fmt.Sprintf("Error: %v", err))

// Versus:

check.addDetail("Error: %v", err)
```

pkg/doctor/doctor.go
```diff
@@ -0,0 +1,420 @@
...
+func (d *Doctor) Run() {
+	fmt.Println("🏥 PicoClaw Doctor")
+	fmt.Println("==================")
+	fmt.Println()
+
+	// Load configuration
+	d.checkConfig()
+
+	// If config loaded successfully, run more checks
+	if d.cfg != nil {
```
johngibb, line 64: I was thinking it might be nice to be able to do:
```go
cfgCheck := d.checkConfig() {
checks = append(checks, cfgCheck) // or a helper

if cfgCheck.Status == StatusOK {
	d.checkWorkspace()
...
```
Your check funcs do seem to all generate a single `Check` result object, so this might be a nice pattern. The rest could be:
```go
checks = append(checks, d.checkFoo())
```
it'd be awfully easy to forget to call the append at the end of a newly added check function as-is, so I'd consider it a maintainability improvement as well.

pkg/doctor/doctor.go
```diff
@@ -0,0 +1,420 @@
...
+
+	client := &http.Client{Timeout: 5 * time.Second}
+	passed := 0
+	failed := 0
+
+	for name, url := range testURLs {
+		// Only test if provider is configured
+		var configured bool
+		switch name {
+		case "OpenRouter":
```
johngibb, line 234: I don't like how easy it is to typo these names, forget to add the provider in both places, and that the order is randomized (i.e. loop iteration) but I couldn't think of a great way to clean it.

Considered a table-driven approach (like tests — `providers := []struct {Name, URL, APIKey string } {...}` but the URLs don't align. And any other approach requires reflection to get the provider object. Wish i had a better suggestion though!

pkg/doctor/doctor.go
```diff
@@ -0,0 +1,420 @@
...
+		case "Moonshot":
+			configured = d.cfg.Providers.Moonshot.APIKey != ""
+		}
+
+		if !configured {
+			continue
+		}
+
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		req, _ := http.NewRequestWithContext(ctx, "GET", url, nil)
```
johngibb: nitpick:
```suggestion
		req, _ := http.NewRequestWithContext(ctx, http.MethodGET, url, nil)
```

pkg/doctor/doctor.go
```diff
@@ -0,0 +1,420 @@
...
+		if err != nil {
+			check.Details = append(check.Details, fmt.Sprintf("❌ %s: %v", name, err))
+			failed++
+		} else {
+			resp.Body.Close()
+			if resp.StatusCode == 200 || resp.StatusCode == 401 { // 401 is OK - means API is reachable
+				check.Details = append(check.Details, fmt.Sprintf("✓ %s: Reachable", name))
+				passed++
+			} else {
+				check.Details = append(check.Details, fmt.Sprintf("⚠️ %s: HTTP %d", name, resp.StatusCode))
```
johngibb: I believe `resp.Status` gets you the name (e.g. "Not Found"):
```suggestion
				check.Details = append(check.Details, fmt.Sprintf("⚠️ %s: HTTP %d %s", name, resp.StatusCode, resp.Status))
```

pkg/doctor/doctor.go
```diff
@@ -0,0 +1,420 @@
...
+			errorCount++
+		}
+	}
+
+	fmt.Println("----------")
+	fmt.Printf("✅ %d passed  ⚠️ %d warnings  ❌ %d errors\n", okCount, warningCount, errorCount)
+	fmt.Println()
+
+	if errorCount > 0 {
+		fmt.Println("❌ Please fix the errors above before using picoclaw")
```
johngibb: nitpick: inconsistent casing:
```suggestion
		fmt.Println("❌ Please fix the errors above before using PicoClaw")
```

johngibb (approved):

pkg/doctor/doctor.go
mrbeandev: Thanks for the suggestion. Keeping the current explicit append style in this PR to keep scope small. I can revisit this in a follow up refactor.

pkg/doctor/doctor.go
mrbeandev: Good maintainability point. Deferring this function signature refactor to a separate PR so this one stays focused.

pkg/doctor/doctor.go
mrbeandev: Agreed on the concern. Leaving this as is for now and tracking a cleaner provider mapping approach in a follow up refactor.

Leeaandrob:
@Zepan `picoclaw doctor` diagnostic command — helps users debug their setup. At +776 lines, it's a substantial addition but diagnostics are valuable for a project with diverse hardware targets.

**Recommendation: Review.** Useful for DX (Developer Experience) which is a requested volunteer role. Check that it doesn't add heavy dependencies.

mrbeandev:
Checked Leeaandrob's recommendation about dependency weight.

Result: this PR does **not** introduce new Go module dependencies (`go.mod` / `go.sum` unchanged in the PR diff), so no heavy dependency concern from this change set.

I didn't apply code changes for that comment since it was review guidance, not a requested code modification.

PR #201, opened 2026-02-15
fix: only handle @bot message from slack once. This removes @bot call…

Pangstar:
…s from handleMessageEvent handler

The slack integration currently gives 2 replies for @slack-bot calls. This pr makes it so if you @slack-bot you will get a message in the slack thread and not both the slack thread and main conversation window.

### DIFF
--- a/pkg/channels/slack.go
+++ b/pkg/channels/slack.go
@@ -198,6 +198,11 @@ func (c *SlackChannel) handleMessageEvent(ev *slackevents.MessageEvent) {
 		return
 	}
 
+	// Check if this is an app mention (bot was mentioned) - if so, let handleAppMention handle it
+	if strings.Contains(ev.Text, fmt.Sprintf("<@%s>", c.botUserID)) {
+		return
+	}
+
 	// 检查白名单，避免为被拒绝的用户下载附件
 	if !c.IsAllowed(ev.User) {
 		logger.DebugCF("slack", "Message rejected by allowlist", map[string]interface{}{
@@ -227,7 +232,6 @@ func (c *SlackChannel) handleMessageEvent(ev *slackevents.MessageEvent) {
 	})
 
 	content := ev.Text
-	content = c.stripBotMention(content)
 
 	var mediaPaths []string
 	localFiles := []string{} // 跟踪需要清理的本地文件


Leeaandrob:
@Zepan Fixes Slack bot handling @bot mentions twice — causes duplicate responses.

**Recommendation: Merge.** +5/-1, simple deduplication fix.

pkg/channels/slack.go
```diff
@@ -198,6 +198,11 @@ func (c *SlackChannel) handleMessageEvent(ev *slackevents.MessageEvent) {
 		return
 	}
 
+	// Check if this is an app mention (bot was mentioned) - if so, let handleAppMention handle it
+	if strings.Contains(ev.Text, fmt.Sprintf("<@%s>", c.botUserID)) {
+		return
```
Leeaandrob, line 203: **[H2 — HIGH: DM Messages With @Mention Silently Dropped]** @Pangstar — In Slack, Direct Messages (DMs) with the bot only generate `message` events — **NOT** `app_mention` events (Slack docs: "app_mention events are only dispatched in channels and group conversations, not in DMs").

If a user types `@bot help` in a DM with the bot, the `message` event will contain `<@BOT_ID>`, this check will match, and `handleMessageEvent` returns early. But no `app_mention` event exists to pick it up → **the message is silently dropped**.

While typing `@bot` in a DM is unusual (you're already talking to the bot), it does happen — especially when users copy-paste from a channel or are testing.

**Suggested fix:** Also check the channel type before skipping:
```go
// Only skip in channels/groups where app_mention will fire.
// In DMs (channel prefix "D"), app_mention events are not sent.
if strings.Contains(ev.Text, fmt.Sprintf("<@%s>", c.botUserID)) && !strings.HasPrefix(ev.Channel, "D") {
    return
}
```

Alternatively, always strip the mention and process it in DMs:
```go
if strings.Contains(ev.Text, fmt.Sprintf("<@%s>", c.botUserID)) {
    if strings.HasPrefix(ev.Channel, "D") {
        // DM: process here since app_mention won't fire
        content = c.stripBotMention(ev.Text)
    } else {
        // Channel: let handleAppMention handle it
        return
    }
}
```

pkg/channels/slack.go
```diff
@@ -198,6 +198,11 @@ func (c *SlackChannel) handleMessageEvent(ev *slackevents.MessageEvent) {
 		return
 	}
 
+	// Check if this is an app mention (bot was mentioned) - if so, let handleAppMention handle it
+	if strings.Contains(ev.Text, fmt.Sprintf("<@%s>", c.botUserID)) {
```
Leeaandrob, line 202: **[H1 — HIGH: File/Media Attachments Lost for @Mentions]** @Pangstar — By routing @mentions to `handleAppMention` exclusively, **all file/media handling is lost**. Compare the two handlers:

**`handleMessageEvent`** (lines 236-275) has:
- File download via `downloadSlackFile()`
- Voice transcription via `c.transcriber.Transcribe()`
- Media paths passed to `HandleMessage(senderID, chatID, content, mediaPaths, metadata)`
- Temp file cleanup via `defer`

**`handleAppMention`** (line 339) has:
- `c.HandleMessage(senderID, chatID, content, nil, metadata)` — **`nil` for mediaPaths**
- No file handling whatsoever

**Regression scenario:** A user sends `@bot analyze this image` with a file attachment. Before this PR, `handleMessageEvent` would download the file, pass it to the agent, and the agent could analyze it. After this PR, the file is silently dropped and the agent only sees the text.

**Suggested fix:** Extract the file download logic into a shared helper method and call it from both handlers. Or, expand `handleAppMention` to also process `AppMentionEvent.Files` (though `AppMentionEvent` may have different file fields than `MessageEvent` — check the Slack SDK types).

pkg/channels/slack.go
```diff
@@ -227,7 +232,6 @@ func (c *SlackChannel) handleMessageEvent(ev *slackevents.MessageEvent) {
 	})
 
 	content := ev.Text
```
Leeaandrob, line 234: **[Positive]** @Pangstar — Correct removal of `c.stripBotMention(content)`. Since messages containing bot mentions now exit at line 203, this line is unreachable for @mention messages. Keeping it would be dead code. Clean.

pkg/channels/slack.go
Pangstar: What should the expected behaviour be in general - should picoclaw message back in the main channel or just in the thread?

Currently DM to the bot with @bot results in a message back in the main DM. In a channel does this still make sense? My intuitions says that in larger channels perhaps all replies should be in the thread, only triggers with @bot (handleAppMentions) and nothing in the main channel to alleviate clutter? If this is the case I would just filter by channel type in both the handleAppMentions and handleMessageEvent whether or not to post in the channel or thread.

PR #208, opened 2026-02-15
fix(web): ensure fresh URL fetch context reaches model

punishell:
Make web_fetch return the fetched payload to the LLM (not just metadata), fix HTML text extraction regex replacements, and add a prompt rule requiring fresh web_fetch calls for URL visit/check requests to prevent stale memory answers.

The bot was constantly hallucinating when asked e.g to visit the website

### DIFF
--- a/pkg/tools/web.go
+++ b/pkg/tools/web.go
@@ -477,7 +477,7 @@ func (t *WebFetchTool) Execute(ctx context.Context, args map[string]interface{})
 	resultJSON, _ := json.MarshalIndent(result, "", "  ")
 
 	return &ToolResult{
-		ForLLM:  fmt.Sprintf("Fetched %d bytes from %s (extractor: %s, truncated: %v)", len(text), urlStr, extractor, truncated),
+		ForLLM:  string(resultJSON),
 		ForUser: string(resultJSON),
 	}
 }


Leeaandrob:
@Zepan Ensures a fresh URL fetch context reaches the model — fixes web_fetch tool breaking silently when context is reused.

**Recommendation: Merge.** +8/-6, small correctness fix for web tools.

pkg/tools/web.go
```diff
@@ -404,23 +404,23 @@ func (t *WebFetchTool) Execute(ctx context.Context, args map[string]interface{})
 	resultJSON, _ := json.MarshalIndent(result, "", "  ")
 
 	return &ToolResult{
-		ForLLM:  fmt.Sprintf("Fetched %d bytes from %s (extractor: %s, truncated: %v)", len(text), urlStr, extractor, truncated),
+		ForLLM:  string(resultJSON),
```
Leeaandrob, line 480: **[Positive — Core Bug Fix]** @punishell — This is the key fix. The original code was:
```go
ForLLM: fmt.Sprintf("Fetched %d bytes from %s (extractor: %s, truncated: %v)", ...)
```
The LLM only saw "Fetched 5000 bytes from example.com" but **never** the actual page content. Since `ForLLM` is what gets injected into the conversation as the tool result (agent/loop.go:551-558), the model had zero information to work with and was forced to hallucinate.

Setting `ForLLM: string(resultJSON)` gives the LLM the full JSON with the `text` field — now it can actually analyze the fetched content. Correct fix.

pkg/tools/web.go
```diff
@@ -404,23 +404,23 @@ func (t *WebFetchTool) Execute(ctx context.Context, args map[string]interface{})
...
 		ForUser: string(resultJSON),
 	}
 }
 
 func (t *WebFetchTool) extractText(htmlContent string) string {
 	re := regexp.MustCompile(`<script[\s\S]*?</script>`)
-	result := re.ReplaceAllLiteralString(htmlContent, "")
+	result := re.ReplaceAllString(htmlContent, "")
```
Leeaandrob: **[M1 — MEDIUM: No-Op Change]** @punishell — All 4 `ReplaceAllLiteralString` → `ReplaceAllString` changes in `extractText()` are functionally identical. Both replace regex matches with `""` (empty string). The difference between the two:

- `ReplaceAllLiteralString(src, repl)` — treats `repl` as literal text
- `ReplaceAllString(src, repl)` — interprets `$` groups in `repl`

Since `repl` is always `""` here, there are no `$` characters to expand — both produce the exact same result. Additionally, none of the 4 regexes have capture groups.

`ReplaceAllLiteralString` was actually the more correct (and slightly faster) choice since it signals intent: "I want a literal replacement, no template expansion." Consider reverting these 4 lines to reduce diff noise, or if there is a reason for the change, add a comment explaining why.

pkg/agent/context.go
```diff
@@ -80,7 +80,9 @@ Your workspace is at: %s
 
 2. **Be helpful and accurate** - When using tools, briefly explain what you're doing.
 
-3. **Memory** - When remembering something, write to %s/memory/MEMORY.md`,
+3. **Fresh web checks are mandatory** - If the user asks to "visit", "check", "open", "fetch", or "get source/code/content" for a URL, you MUST call the web_fetch tool for that URL in the current turn. Never answer from memory for URL-visit requests.
```
Leeaandrob: **[L1 — LOW: English-Only Prompt Rule]** @punishell — This hardcoded keyword list ("visit", "check", "open", "fetch", "get source/code/content") only works for English-speaking users. Users interacting in Portuguese, Chinese, Spanish, etc. would not trigger this rule, and the model might still hallucinate for them.

Not a blocker — this is a reasonable first step and a common prompt engineering pattern. Just noting the limitation for future improvement.

**Note on co-authorship:** The commit is co-authored by Cursor. While there is nothing wrong with AI-assisted development, the system prompt rule in particular should be carefully validated since LLMs sometimes generate prompt rules that sound good but do not actually influence model behavior reliably. Consider testing with a few different phrasings to confirm the rule works in practice.

pkg/tools/web.go
punishell: so this is resolved or need more action on that?

pkg/agent/context.go
punishell: fixed ed0a2e8

pkg/tools/web.go
punishell: fixed ed0a2e8

pkg/agent/loop.go
```diff
@@ -12,6 +12,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"sync"
```
Leeaandrob, line 14: **[HIGH] Severe false positives — matches filenames and JS libraries**

The bare-domain pattern `\b[a-z0-9][a-z0-9.-]*\.[a-z]{2,}\b` matches any `word.extension` with 2+ char extension. Verified locally:

- `config.json` → MATCH (every config discussion)
- `MEMORY.md` → MATCH (PicoClaw's own workspace file!)
- `node.js`, `vue.js`, `three.js` → MATCH
- `my-app.service` → MATCH

This causes the model to attempt `web_fetch("config.json")` etc., wasting tool calls and degrading UX.

**Fix:** Only match explicit `https?://` URLs:
```suggestion
var userMessageURLRegex = regexp.MustCompile(`https?://[^\s<>"')\]]+`)
```

punishell:
removed everything just keeping: 1 line changed: ForLLM: string(resultJSON) instead of the metadata string

lxowalle:
The real reason for this issue is that the returned ForUser parameter was not handled correctly. I have tested this in the new version and it has been fixed. You can try it again.

nikolasdehor:
PRs #389 and #512 both address this. #512 has been reviewed and approved — it should be the one to land. Suggesting this issue reference #512 as the fix.

lxowalle:
https://github.com/sipeed/picoclaw/pull/833 fixed

PR #177, opened 2026-02-14
feat: Add Z.AI provider support

ghost:
## Summary

- **Add Z.AI (z.ai) as a new LLM provider**, the international AI platform by Zhipu AI.
- Automatic endpoint routing between the **standard API** and the **dedicated Coding API** based on model name.
- Auto-detection of Z.AI models by name (GLM, CogView, CogVideo, AutoGLM, Vidu) when the `zai` provider key is configured, with priority over the existing `zhipu` provider.

### Endpoints

| Type | Base URL |
|------|----------|
| Standard | `https://api.z.ai/api/paas/v4` |
| Coding | `https://api.z.ai/api/coding/paas/v4` |

The Coding endpoint is automatically selected when the model name contains `coding`. Users can also override this via the `api_base` config field.

### Supported models

| Category | Models |
|----------|--------|
| LLM | `glm-5`, `glm-4.7`, `glm-4.6`, `glm-4.5`, `glm-4-32b-0414-128k` |
| Vision/Multimodal | `glm-4.6v`, `glm-4.5v`, `glm-ocr` |
| Image generation | `glm-image`, `cogview-4` |
| Audio | `glm-asr-2512` |
| Video | `cogvideox-3`, `vidu-q1`, `vidu-2` |
| Agent | `autoglm-phone-multilingual` |

### Files changed

- `pkg/config/config.go` — Added `ZAI` field to `ProvidersConfig`, initialized in defaults, included in `GetAPIKey()` fallback chain.
- `pkg/providers/http_provider.go` — Added `case "zai", "z.ai"` in explicit provider switch; added model-name auto-detection fallback (Z.AI takes priority over Zhipu when configured); added `zai/` prefix stripping for model names.
- `config/config.example.json` — Added `zai` provider entry with placeholder API key.

### Configuration example

```json
{
  "providers": {
    "zai": {
      "api_key": "your-z.ai-api-key",
      "api_base": ""
    }
  },
  "agents": {
    "defaults": {
      "provider": "zai",
      "model": "glm-5"
    }
  }
}
```

## Test plan

- [ ] Configure `zai` provider with a valid API key and verify chat completion with `glm-5`
- [ ] Verify coding endpoint is auto-selected when model name contains `coding`
- [ ] Verify `zai/glm-4.7` prefix stripping works correctly
- [ ] Verify `zhipu` provider still works independently when `zai` is not configured
- [ ] Verify model auto-detection routes to Z.AI when both `zai` and `zhipu` keys are set

[...]

PR #110, opened 2026-02-13
feat(provider): add gemini google-generative-ai compatibility

fuhao009:
## Summary
- add provider compatibility groundwork for API mode and custom headers
- add Gemini (google-generative-ai) request/response compatibility
- support Gemini generateContent endpoint and tool-call mapping

## Validation
- go build -o ./build/picoclaw ./cmd/picoclaw
- debug run confirms request path: /v1beta/models/{model}:generateContent
- upstream service may return 503 when account pool unavailable

PR #82, opened 2026-02-12
feat(providers): add OpenAI-compatible endpoint support for GLM/Zai m…

glira:
…odels

Enables using GLM and Zai models (e.g. zai-org/GLM-4.6) with custom OpenAI-compatible API endpoints like Chutes.ai.

When model contains 'glm' or 'zai' and OpenAI provider has a custom api_base (not api.openai.com), the provider will route requests to the custom endpoint with the model name passed as-is.

Resolves configuration issues when using Chutes.ai with PicoClaw.

PR #213, opened 2026-02-15
Refactor providers by protocol family (discussion #122)

jmahotiedu:
## Summary
- move provider routing into a dedicated protocol-aware factory (`pkg/providers/factory.go`)
- extract OpenAI-compatible transport into `pkg/providers/openai_compat` and keep `http_provider` as a thin compatibility adapter
- move Anthropic protocol logic into `pkg/providers/anthropic` with compatibility wrappers in `claude_provider.go`
- add provider routing/compatibility tests and update docs + migration compatibility coverage

## Verification
- `go test ./pkg/providers/... ./pkg/migrate`
- `go test ./...` still has pre-existing failures on this Windows environment:
  - `cmd/picoclaw` embed path setup (`pattern workspace: no matching files found`)
  - `pkg/auth` file permission tests on Windows temp paths

Implements: https://github.com/sipeed/picoclaw/discussions/122

### DIFF
--- a/README.ja.md
+++ b/README.ja.md
@@ -12,7 +12,7 @@
 <img src="https://img.shields.io/badge/license-MIT-green" alt="License">
 </p>
 
-[中文](README.zh.md) | **日本語** | [Português](README.pt-br.md) | [English](README.md)
+[中文](README.zh.md) | **日本語** | [Português](README.pt-br.md) | [Tiếng Việt](README.vi.md) | [English](README.md)
 
 </div>
 
@@ -253,7 +253,7 @@ Telegram、Discord、QQ、DingTalk、LINE で PicoClaw と会話できます
     "telegram": {
       "enabled": true,
       "token": "YOUR_BOT_TOKEN",
-      "allowFrom": ["YOUR_USER_ID"]
+      "allow_from": ["YOUR_USER_ID"]
     }
   }
 }
@@ -293,7 +293,7 @@ picoclaw gateway
     "discord": {
       "enabled": true,
       "token": "YOUR_BOT_TOKEN",
-      "allowFrom": ["YOUR_USER_ID"]
+      "allow_from": ["YOUR_USER_ID"]
     }
   }
 }
@@ -676,7 +676,7 @@ HEARTBEAT_OK 応答         ユーザーが直接結果を受け取る
     "telegram": {
       "enabled": true,
       "token": "123456:ABC...",
-      "allowFrom": ["123456789"]
+      "allow_from": ["123456789"]
     },
     "discord": {
       "enabled": true,
@@ -692,7 +692,7 @@ HEARTBEAT_OK 応答         ユーザーが直接結果を受け取る
       "appSecret": "xxx",
       "encryptKey": "",
       "verificationToken": "",
-      "allowFrom": []
+      "allow_from": []
     }
   },
   "tools": {
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@
     <a href="https://x.com/SipeedIO"><img src="https://img.shields.io/badge/X_(Twitter)-SipeedIO-black?style=flat&logo=x&logoColor=white" alt="Twitter"></a>
   </p>
 
- [中文](README.zh.md) | [日本語](README.ja.md) | [Português](README.pt-br.md) | **English**
+ [中文](README.zh.md) | [日本語](README.ja.md) | [Português](README.pt-br.md) | [Tiếng Việt](README.vi.md) | **English**
 </div>
 
 ---
@@ -283,7 +283,7 @@ Talk to your picoclaw through Telegram, Discord, DingTalk, or LINE
     "telegram": {
       "enabled": true,
       "token": "YOUR_BOT_TOKEN",
-      "allowFrom": ["YOUR_USER_ID"]
+      "allow_from": ["YOUR_USER_ID"]
     }
   }
 }
@@ -326,7 +326,7 @@ picoclaw gateway
     "discord": {
       "enabled": true,
       "token": "YOUR_BOT_TOKEN",
-      "allowFrom": ["YOUR_USER_ID"]
+      "allow_from": ["YOUR_USER_ID"]
     }
   }
 }
--- a/README.vi.md
+++ b/README.vi.md
@@ -1,0 +1,859 @@
+<div align="center">
+<img src="assets/logo.jpg" alt="PicoClaw" width="512">
+
+<h1>PicoClaw: Trợ lý AI Siêu Nhẹ viết bằng Go</h1>
+
+<h3>Phần cứng $10 · RAM 10MB · Khởi động 1 giây · 皮皮虾，我们走！</h3>
+
+  <p>
+    <img src="https://img.shields.io/badge/Go-1.21+-00ADD8?style=flat&logo=go&logoColor=white" alt="Go">
+    <img src="https://img.shields.io/badge/Arch-x86__64%2C%20ARM64%2C%20RISC--V-blue" alt="Hardware">
+    <img src="https://img.shields.io/badge/license-MIT-green" alt="License">
+    <br>
+    <a href="https://picoclaw.io"><img src="https://img.shields.io/badge/Website-picoclaw.io-blue?style=flat&logo=google-chrome&logoColor=white" alt="Website"></a>
+    <a href="https://x.com/SipeedIO"><img src="https://img.shields.io/badge/X_(Twitter)-SipeedIO-black?style=flat&logo=x&logoColor=white" alt="Twitter"></a>
+  </p>
+
+**Tiếng Việt** | [中文](README.zh.md) | [日本語](README.ja.md) | [English](README.md)
+</div>
+
+---
+
+🦐 **PicoClaw** là trợ lý AI cá nhân siêu nhẹ, lấy cảm hứng từ [nanobot](https://github.com/HKUDS/nanobot), được viết lại hoàn toàn bằng **Go** thông qua quá trình "tự khởi tạo" (self-bootstrapping) — nơi chính AI Agent đã tự dẫn dắt toàn bộ quá trình chuyển đổi kiến trúc và tối ưu hóa mã nguồn.
+
+⚡️ **Cực kỳ nhẹ:** Chạy trên phần cứng chỉ **$10** với RAM **<10MB**. Tiết kiệm 99% bộ nhớ so với OpenClaw và rẻ hơn 98% so với Mac mini!
+
+<table align="center">
+<tr align="center">
+<td align="center" valign="top">
+<p align="center">
+<img src="assets/picoclaw_mem.gif" width="360" height="240">
+</p>
+</td>
+<td align="center" valign="top">
+<p align="center">
+<img src="assets/licheervnano.png" width="400" height="240">
+</p>
+</td>
+</tr>
+</table>
+
+> [!CAUTION]
+> **🚨 TUYÊN BỐ BẢO MẬT & KÊNH CHÍNH THỨC**
+>
+> * **KHÔNG CÓ CRYPTO:** PicoClaw **KHÔNG** có bất kỳ token/coin chính thức nào. Mọi thông tin trên `pump.fun` hoặc các sàn giao dịch khác đều là **LỪA ĐẢO**.
+> * **DOMAIN CHÍNH THỨC:** Website chính thức **DUY NHẤT** là **[picoclaw.io](https://picoclaw.io)**, website công ty là **[sipeed.com](https://sipeed.com)**.
+> * **Cảnh báo:** Nhiều tên miền `.ai/.org/.com/.net/...` đã bị bên thứ ba đăng ký, không phải của chúng tôi.
+> * **Cảnh báo:** PicoClaw đang trong giai đoạn phát triển sớm và có thể còn các vấn đề bảo mật mạng chưa được giải quyết. Không nên triển khai lên môi trường production trước phiên bản v1.0.
+> * **Lưu ý:** PicoClaw gần đây đã merge nhiều PR, dẫn đến bộ nhớ sử dụng có thể lớn hơn (10–20MB) ở các phiên bản mới nhất. Chúng tôi sẽ ưu tiên tối ưu tài nguyên khi bộ tính năng đã ổn định.
+
+
+## 📢 Tin tức
+
+2026-02-16 🎉 PicoClaw đạt 12K stars chỉ trong một tuần! Cảm ơn tất cả mọi người! PicoClaw đang phát triển nhanh hơn chúng tôi tưởng tượng. Do số lượng PR tăng cao, chúng tôi cấp thiết cần maintainer từ cộng đồng. Các vai trò tình nguyện viên và roadmap đã được công bố [tại đây](docs/picoclaw_community_roadmap_260216.md) — rất mong đón nhận sự tham gia của bạn!
+
+2026-02-13 🎉 PicoClaw đạt 5000 stars trong 4 ngày! Cảm ơn cộng đồng! Chúng tôi đang hoàn thiện **Lộ trình dự án (Roadmap)** và thiết lập **Nhóm phát triển** để đẩy nhanh tốc độ phát triển PicoClaw.  
+🚀 **Kêu gọi hành động:** Vui lòng gửi yêu cầu tính năng tại GitHub Discussions. Chúng tôi sẽ xem xét và ưu tiên trong cuộc họp hàng tuần.
+
+2026-02-09 🎉 PicoClaw chính thức ra mắt! Được xây dựng trong 1 ngày để mang AI Agent đến phần cứng $10 với RAM <10MB. 🦐 PicoClaw, Lên Đường!
+
+## ✨ Tính năng nổi bật
+
+🪶 **Siêu nhẹ**: Bộ nhớ sử dụng <10MB — nhỏ hơn 99% so với Clawdbot (chức năng cốt lõi).
+
+💰 **Chi phí tối thiểu**: Đủ hiệu quả để chạy trên phần cứng $10 — rẻ hơn 98% so với Mac mini.
+
+⚡️ **Khởi động siêu nhanh**: Nhanh gấp 400 lần, khởi động trong 1 giây ngay cả trên CPU đơn nhân 0.6GHz.
+
+🌍 **Di động thực sự**: Một file binary duy nhất chạy trên RISC-V, ARM và x86. Một click là chạy!
+
+🤖 **AI tự xây dựng**: Triển khai Go-native tự động — 95% mã nguồn cốt lõi được Agent tạo ra, với sự tinh chỉnh của con người.
+
+|                               | OpenClaw      | NanoBot                  | **PicoClaw**                              |
+| ----------------------------- | ------------- | ------------------------ | ----------------------------------------- |
+| **Ngôn ngữ**                  | TypeScript    | Python                   | **Go**                                    |
+| **RAM**                       | >1GB          | >100MB                   | **< 10MB**                                |
+| **Thời gian khởi động**</br>(CPU 0.8GHz) | >500s         | >30s                     | **<1s**                                   |
+| **Chi phí**                   | Mac Mini $599 | Hầu hết SBC Linux ~$50  | **Mọi bo mạch Linux**</br>**Chỉ từ $10** |
+
+<img src="assets/compare.jpg" alt="PicoClaw" width="512">
+
+## 🦾 Demo
+
+### 🛠️ Quy trình trợ lý tiêu chuẩn
+
+<table align="center">
+<tr align="center">
+<th><p align="center">🧩 Lập trình Full-Stack</p></th>
+<th><p align="center">🗂️ Quản lý Nhật ký & Kế hoạch</p></th>
+<th><p align="center">🔎 Tìm kiếm Web & Học hỏi</p></th>
+</tr>
+<tr>
+<td align="center"><p align="center"><img src="assets/picoclaw_code.gif" width="240" height="180"></p></td>
+<td align="center"><p align="center"><img src="assets/picoclaw_memory.gif" width="240" height="180"></p></td>
+<td align="center"><p align="center"><img src="assets/picoclaw_search.gif" width="240" height="180"></p></td>
+</tr>
+<tr>
+<td align="center">Phát triển • Triển khai • Mở rộng</td>
+<td align="center">Lên lịch • Tự động hóa • Ghi nhớ</td>
+<td align="center">Khám phá • Phân tích • Xu hướng</td>
+</tr>
+</table>
+
+### 🐜 Triển khai sáng tạo trên phần cứng tối thiểu
+
+PicoClaw có thể triển khai trên hầu hết mọi thiết bị Linux!
+
+* $9.9 [LicheeRV-Nano](https://www.aliexpress.com/item/1005006519668532.html) phiên bản E (Ethernet) hoặc W (WiFi6), dùng làm Trợ lý Gia đình tối giản.
+* $30~50 [NanoKVM](https://www.aliexpress.com/item/1005007369816019.html), hoặc $100 [NanoKVM-Pro](https://www.aliexpress.com/item/1005010048471263.html), dùng cho quản trị Server tự động.
+* $50 [MaixCAM](https://www.aliexpress.com/item/1005008053333693.html) hoặc $100 [MaixCAM2](https://www.kickstarter.com/projects/zepan/maixcam2-build-your-next-gen-4k-ai-camera), dùng cho Giám sát thông minh.
+
+https://private-user-images.githubusercontent.com/83055338/547056448-e7b031ff-d6f5-4468-bcca-5726b6fecb5c.mp4
+
+🌟 Nhiều hình thức triển khai hơn đang chờ bạn khám phá!
+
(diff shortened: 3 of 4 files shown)


Leeaandrob:
@Zepan This PR directly addresses **roadmap issue #283** (Refactor Provider Architecture by Protocol — priority: high, status: **In Progress**). It reorganizes providers by protocol family instead of vendor, which is the exact refactor described in the board.

+1484/-676 across 11 files — significant but well-scoped. This is foundational work that will simplify adding new providers going forward.

**Recommendation: Prioritize review and merge.** This is the highest-priority In Progress item on the board. Other provider PRs (#246 Mistral, #252 Bedrock, #177 Z.AI, #110 Gemini, #82 OpenAI-compat) should ideally wait for or build on top of this refactor.

Leeaandrob:
@jmahotiedu please solve the conflicts i'll prioritize this pr for now!

pkg/providers/anthropic/provider.go
```diff
@@ -0,0 +1,241 @@
...
+	client      *anthropic.Client
+	tokenSource func() (string, error)
+}
+
+func NewProvider(token string) *Provider {
+	client := anthropic.NewClient(
+		option.WithAuthToken(token),
+		option.WithBaseURL("https://api.anthropic.com"),
+	)
+	return &Provider{client: &client}
```
Leeaandrob: **[C2 — CRITICAL]** @jmahotiedu — This base URL is hardcoded to `https://api.anthropic.com`, but `factory.go:98-102` resolves `cfg.Providers.Anthropic.APIBase` from the user's config (defaulting to `https://api.anthropic.com/v1`).

Two problems here:
1. The `apiBase` from config is **never passed** to `NewProvider()`. Users who configure custom Anthropic endpoints (e.g., a corporate proxy, AWS Bedrock adapter, or regional endpoint) will have their config silently ignored — this is a **functional regression**.
2. There's a subtle inconsistency: the factory resolves to `https://api.anthropic.com/v1` (with `/v1` suffix) while this constructor uses `https://api.anthropic.com` (without). The Anthropic SDK appends `/v1/messages` internally so it happens to work, but the mismatch is suspicious and fragile.

**Suggested fix:**
```go
func NewProvider(token string, opts ...ProviderOption) *Provider {
    baseURL := "https://api.anthropic.com"
    // apply opts for custom baseURL
    ...
}
```
Or simply add an `apiBase` parameter:
```go
func NewProvider(token, apiBase string) *Provider {
    if apiBase == "" {
        apiBase = "https://api.anthropic.com"
    }
    ...
}
```

pkg/providers/anthropic/provider.go
```diff
@@ -0,0 +1,241 @@
...
+	"github.com/anthropics/anthropic-sdk-go/option"
+)
+
+type ToolCall struct {
+	ID        string                 `json:"id"`
+	Type      string                 `json:"type,omitempty"`
+	Function  *FunctionCall          `json:"function,omitempty"`
+	Name      string                 `json:"name,omitempty"`
+	Arguments map[string]interface{} `json:"arguments,omitempty"`
+}
```
Leeaandrob: **[C1 — CRITICAL]** @jmahotiedu — These types (`ToolCall`, `FunctionCall`, `LLMResponse`, `UsageInfo`, `Message`, `ToolDefinition`, `ToolFunctionDefinition`) are defined **three times** in this PR:
1. `pkg/providers/types.go` (the canonical package-level types)
2. `pkg/providers/openai_compat/provider.go` (duplicated)
3. Here in `pkg/providers/anthropic/provider.go` (duplicated)

This triple duplication creates:
- **~180 lines of pure type-conversion boilerplate** in `claude_provider.go` and `http_provider.go` (the `toAnthropicProviderMessages`, `fromOpenAICompatToolCalls`, etc. functions)
- **Shotgun surgery risk**: any future change to these types (e.g., adding a `metadata` field to `ToolCall`) requires updating 3 type definitions + 10+ conversion functions
- The adapter files (`claude_provider.go`, `http_provider.go`) have **zero business logic** — they exist solely to convert between identical type definitions

This violates the DRY principle and would not pass code review at Google, Apple, or Anthropic. The refactoring should **reduce** duplication, not multiply it.

**Suggested approaches:**

**Option A** — Type aliases in sub-packages (zero allocation, zero conversion):
```go
// openai_compat/types.go
package openai_compat
import "github.com/sipeed/picoclaw/pkg/providers"
type Message = providers.Message
type ToolCall = providers.ToolCall
// ... etc
```

**Option B** — Shared types package:
```go
// pkg/providers/types/ — imported by all three packages
```

**Option C** — Keep types only in the parent package and have sub-packages accept/return them via interfaces.

pkg/providers/openai_compat/provider.go
```diff
@@ -0,0 +1,230 @@
...
+		return nil, fmt.Errorf("API base not configured")
+	}
+
+	// Strip provider prefix for OpenAI-compatible backends.
+	if idx := strings.Index(model, "/"); idx != -1 {
+		prefix := model[:idx]
+		if prefix == "moonshot" || prefix == "nvidia" || prefix == "groq" || prefix == "ollama" {
+			model = model[idx+1:]
+		}
+	}
```
Leeaandrob: **[H2 — HIGH]** @jmahotiedu — This prefix stripping logic is hardcoded to only 4 providers (`moonshot`, `nvidia`, `groq`, `ollama`), but the factory in `factory.go:191-196` routes models with prefixes like `openrouter/`, `anthropic/`, `openai/`, `meta-llama/`, `deepseek/`, and `google/` — none of which get stripped here.

**Concrete bug scenario:** If a user sets `provider: "deepseek"` with `model: "deepseek/deepseek-chat"`, the factory routes to `providerTypeHTTPCompat` but the model string `"deepseek/deepseek-chat"` is sent as-is to the DeepSeek API, which likely won't recognize its own prefix.

**Suggested fixes:**
1. Make this a configurable allow-list or use a more generic heuristic
2. Move prefix stripping to the factory layer (closer to where routing decisions are made) rather than the protocol layer
3. At minimum, add `deepseek` to the list

Also consider: should this be the provider's responsibility at all? The factory already knows the prefix — it could strip it before passing to the provider.

pkg/providers/openai_compat/provider.go
```diff
@@ -0,0 +1,230 @@
...
+				}
+			}
+		} else if tc.Function != nil {
+			name = tc.Function.Name
+			if tc.Function.Arguments != "" {
+				if err := json.Unmarshal([]byte(tc.Function.Arguments), &arguments); err != nil {
+					arguments["raw"] = tc.Function.Arguments
+				}
+			}
+		}
```
Leeaandrob, line 168: **[M1 — MEDIUM]** @jmahotiedu — These two branches (lines 201-206 and 208-214) are **identical code**. The `if tc.Type == "function"` check is redundant because the `else if tc.Function != nil` branch does exactly the same thing.

Simplify to:
```go
if tc.Function != nil {
    name = tc.Function.Name
    if tc.Function.Arguments != "" {
        if err := json.Unmarshal([]byte(tc.Function.Arguments), &arguments); err != nil {
            arguments["raw"] = tc.Function.Arguments
        }
    }
}
```

pkg/providers/openai_compat/provider.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+
+	if proxyURL != "" {
+		parsed, err := url.Parse(proxyURL)
+		if err == nil {
+			client.Transport = &http.Transport{
+				Proxy: http.ProxyURL(parsed),
+			}
+		}
+	}
```
Leeaandrob, line 46: **[M3 — MEDIUM]** @jmahotiedu — If `url.Parse(proxyURL)` fails, the error is silently swallowed and the provider operates without a proxy. The user has explicitly configured a proxy, so this silent fallback could be a security concern (traffic going unproxied when the user expects it to be proxied).

At minimum, consider returning an error from `NewProvider` or logging a warning.

pkg/providers/http_provider.go
```diff
@@ -7,444 +7,124 @@
...
-	apiKey     string
-	apiBase    string
-	httpClient *http.Client
+	delegate *openai_compat.Provider
 }
 
-func NewHTTPProvider(apiKey, apiBase, proxy string) *HTTPProvider {
-	client := &http.Client{
-		Timeout: 120 * time.Second,
+func NewHTTPProvider(apiKey, apiBase string, proxy ...string) *HTTPProvider {
```
Leeaandrob: **[H1 — HIGH]** @jmahotiedu — This signature changed from `func NewHTTPProvider(apiKey, apiBase, proxy string)` (3 positional args) to `func NewHTTPProvider(apiKey, apiBase string, proxy ...string)` (variadic).

This is a **breaking API change** for any external consumer of this package. While internal callers work fine (factory.go passes `sel.proxy` as a single string), any code that was calling `NewHTTPProvider("key", "base", "")` will still compile, but the semantics changed subtly: an empty string is now passed as `proxy[0]` inside the variadic, whereas before it was a direct positional arg.

If this is intentional to allow calling without proxy (`NewHTTPProvider("key", "base")`), please document the change. If not, consider keeping the 3-arg signature for backward compatibility.

pkg/providers/claude_provider.go
```diff
@@ -2,206 +2,151 @@ package providers
...
-		params.System = system
-	}
-
-	if temp, ok := options["temperature"].(float64); ok {
-		params.Temperature = anthropic.Float(temp)
-	}
-
-	if len(tools) > 0 {
-		params.Tools = translateToolsForClaude(tools)
+func toAnthropicProviderMessages(messages []Message) []anthropicprovider.Message {
```
Leeaandrob: **[C1 — Related]** @jmahotiedu — This entire block (lines 62-152) is ~90 lines of pure type-conversion boilerplate: `toAnthropicProviderMessages`, `toAnthropicProviderTools`, `toAnthropicProviderToolCalls`, `fromAnthropicProviderResponse`, `fromAnthropicProviderToolCalls`.

The same pattern exists in `http_provider.go` for `openai_compat` types (~90 more lines).

Combined, that's **~180 lines of code that exist solely because the same struct definitions were duplicated** into the sub-packages. If the types were shared (via aliases or a shared package), these entire files would shrink to ~15-20 lines each — just the constructor and a direct delegate call.

pkg/providers/anthropic/provider.go
```diff
@@ -0,0 +1,241 @@
...
+		switch block.Type {
+		case "text":
+			tb := block.AsText()
+			content += tb.Text
+		case "tool_use":
+			tu := block.AsToolUse()
+			var args map[string]interface{}
+			if err := json.Unmarshal(tu.Input, &args); err != nil {
+				args = map[string]interface{}{"raw": string(tu.Input)}
+			}
```
Leeaandrob, line 202: **[M2 — MEDIUM]** @jmahotiedu — When `json.Unmarshal` fails for tool call input, the error is silently converted to `{"raw": ...}`. This makes it very hard to debug tool call failures in production — there's no log entry, no metric, no indication that parsing failed.

Consider at minimum adding a `log.Printf` or returning the error as a diagnostic field. The same pattern exists in `openai_compat/provider.go:204-205`.

pkg/providers/factory_test.go
```diff
@@ -0,0 +1,204 @@
...
+			}
+			if tt.wantAPIBase != "" && got.apiBase != tt.wantAPIBase {
+				t.Fatalf("apiBase = %q, want %q", got.apiBase, tt.wantAPIBase)
+			}
+			if tt.wantProxy != "" && got.proxy != tt.wantProxy {
+				t.Fatalf("proxy = %q, want %q", got.proxy, tt.wantProxy)
+			}
+		})
+	}
+}
```
Leeaandrob, line 195: **[H3 — HIGH]** @jmahotiedu — Good coverage for `resolveProviderSelection` (13 cases), but several important paths in `factory.go` are untested:

1. **`createClaudeAuthProvider()`** and **`createCodexAuthProvider()`** (lines 33-53) — these are the OAuth paths, zero test coverage
2. **`shengsuanyun`**, **`deepseek`**, **`nvidia`** provider routing in the explicit provider switch — no test cases
3. **Token refresh** via `tokenSource` in `anthropic/provider.go` — no test
4. **Proxy configuration** in `openai_compat/provider.go` — no test that proxy is actually applied to the HTTP transport
5. **Edge cases**: nil options map, wrong type assertions (e.g., `max_tokens` as `float64` instead of `int`)

The auth provider tests would need mocking `auth.GetCredential`, which is understandable, but at least the routing decision (`sel.providerType == providerTypeClaudeAuth`) should be verified.

pkg/providers/factory.go
```diff
@@ -0,0 +1,320 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("loading auth credentials: %w", err)
+	}
+	if cred == nil {
+		return nil, fmt.Errorf("no credentials for openai. Run: picoclaw auth login --provider openai")
+	}
+	return NewCodexProviderWithTokenSource(cred.AccessToken, cred.AccountID, createCodexTokenSource()), nil
+}
+
+func resolveProviderSelection(cfg *config.Config) (providerSelection, error) {
```
Leeaandrob, line 65: **[Positive]** @jmahotiedu — Clean separation between `resolveProviderSelection` (pure decision logic, easily testable) and `CreateProvider` (side-effectful instantiation). The `providerSelection` struct as an intermediate representation is a good pattern. The typed enum via `iota` is much more maintainable than string matching.

One minor suggestion: consider adding a `String()` method to `providerType` for better error messages and debugging.

jmahotiedu:
Addressed the requested changes on top of `refactor/provider-protocol-122` (latest commit: `c4cbb5f`).

### Mapping: Review Items -> Fixes

- **C1 (type duplication)**
  - Canonical shared protocol types moved to `pkg/providers/protocoltypes/types.go`.
  - Compatibility aliases retained in `pkg/providers/types.go`.
  - Protocol providers consume aliases from shared package in:
    - `pkg/providers/openai_compat/provider.go`
    - `pkg/providers/anthropic/provider.go`

- **C2 (Anthropic `api_base` propagation)**
  - Factory now carries Anthropic `api_base` through auth path:
    - `pkg/providers/factory.go` (`resolveProviderSelection`, `createClaudeAuthProvider(apiBase string)`)
  - Claude adapter forwards token source + base URL:
    - `pkg/providers/claude_provider.go` (`NewClaudeProviderWithTokenSourceAndBaseURL`)
  - Anthropic provider normalizes/applies base URL:
    - `pkg/providers/anthropic/provider.go` (`NewProviderWithTokenSourceAndBaseURL`, `normalizeBaseURL`)

- **H1 (HTTPProvider constructor compatibility)**
  - Constructor kept stable as `(apiKey, apiBase, proxy string)` in `pkg/providers/http_provider.go`.

- **H2 (prefix/model normalization hardening)**
  - Centralized normalization with endpoint-aware behavior in `pkg/providers/openai_compat/provider.go` (`normalizeModel`):
    - Preserves namespaced IDs for OpenRouter endpoints.
    - Strips vendor prefixes for vendor-native endpoints (including DeepSeek/OpenRouter/Google-prefixed inputs where applicable).

- **H3 (coverage gaps: auth/provider/token/proxy)**
  - Auth/provider route coverage in `pkg/providers/factory_test.go`:
    - `TestResolveProviderSelection`
    - `TestCreateProviderReturnsClaudeProviderForAnthropicOAuth`
    - `TestCreateProviderReturnsCodexProviderForOpenAIOAuth`
    - `TestCreateProviderReturnsCodexProviderForCodexCliAuthMethod`
  - Proxy transport coverage in `pkg/providers/openai_compat/provider_test.go`:
    - `TestProvider_ProxyConfigured`
  - Token-source refresh + base URL behavior in `pkg/providers/anthropic/provider_test.go`:
    - `TestProvider_ChatUsesTokenSource`
    - `TestProvider_NewProviderWithBaseURL_NormalizesV1Suffix`

- **M1 (duplicate parse branch)**
  - Removed duplicate tool-call parsing branch in `pkg/providers/openai_compat/provider.go` (`parseResponse`).

- **M2 (silent tool-JSON errors)**
  - Added warning logging on decode failure and preserved raw payload for debugging in:
    - `pkg/providers/openai_compat/provider.go`
    - `pkg/providers/anthropic/provider.go`

- **M3 (silent proxy parse failure)**
  - Added warning logging for invalid proxy URL parsing in `pkg/providers/openai_compat/provider.go` (`NewProvider`).

### Verification Commands + Outcomes

- `rg "type ToolCall struct|type Message struct|type LLMResponse struct" pkg/providers`
  - Result: matches only in `pkg/providers/protocoltypes/types.go`.
- `go test ./pkg/providers -run TestResolveProviderSelection`
  - Result: pass.
- `go test ./pkg/providers/openai_compat -run TestProvider_`
  - Result: pass.
- `go test ./pkg/providers/anthropic -run TestProvider_`
  - Result: pass.
- `docker run --rm -v c:/projects/picoclaw:/src -w /src golang:1.25.7 gofmt -w ...`
  - Result: pass.
- `docker run --rm -v c:/projects/picoclaw:/src -w /src golang:1.25.7 go test ./pkg/providers/... ./pkg/migrate`
  - Result: pass.
- `docker run --rm -v c:/projects/picoclaw:/src -w /src golang:1.25.7 go test ./...`
  - Result: pass.

Note: `golang:1.25.7` was required because `go.mod` now requires `go >= 1.25.7`.

Requesting re-review.

jmahotiedu:
@Leeaandrob @Zepan

Follow-up is pushed on `refactor/provider-protocol-122` (latest commit: `c4cbb5f`) and addresses the requested C1/C2/H1/H2/H3/M1/M2/M3 items.

Detailed item-by-item mapping and verification results are here:
https://github.com/sipeed/picoclaw/pull/213#issuecomment-3915631352

Could you please take another look when convenient? Thanks.

pkg/providers/protocoltypes/types.go
```diff
@@ -0,0 +1,45 @@
+package protocoltypes
```
Leeaandrob, line 1: **[POSITIVE]** Clean solution to the C1 triple-duplication finding. `protocoltypes` as the single source of truth with type aliases in consuming packages is the correct Go idiom — zero runtime cost, no conversions needed, and each package preserves its own public API.

This eliminated ~180 lines of field-by-field conversion code from `claude_provider.go` and `http_provider.go`.

pkg/providers/anthropic/provider.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+}
+
+func NewProviderWithClient(client *anthropic.Client) *Provider {
+	return &Provider{
+		client:  client,
+		baseURL: defaultBaseURL,
+	}
+}
+
```
Leeaandrob, line 53: **[POSITIVE]** Good resolution of C2 — the `NewProviderWithTokenSourceAndBaseURL` constructor cleanly chains through `NewProviderWithBaseURL`, which calls `normalizeBaseURL` to handle the `/v1` config/SDK mismatch.

The `BaseURL()` getter is also a nice addition — makes the flow testable end-to-end as proven by `TestCreateProviderReturnsClaudeProviderForAnthropicOAuth`.

pkg/providers/openai_compat/provider.go
```diff
@@ -0,0 +1,232 @@
...
+
+	return parseResponse(body)
+}
+
+func parseResponse(body []byte) (*LLMResponse, error) {
+	var apiResponse struct {
+		Choices []struct {
+			Message struct {
+				Content   string `json:"content"`
```
Leeaandrob, line 128: **[POSITIVE]** `normalizeModel()` now covers all 8 routed prefixes plus the smart OpenRouter passthrough. The separation of the stripping logic into its own function with `apiBase` awareness is well-designed — it correctly preserves `openrouter/model-name` when targeting OpenRouter's API, while stripping `deepseek/deepseek-chat` for DeepSeek's API.

pkg/providers/factory.go
```diff
@@ -0,0 +1,353 @@
+package providers
+
+import (
+	"fmt"
+	"strings"
+
```
Leeaandrob, line 6: **[POSITIVE]** `var getCredential = auth.GetCredential` is a textbook Go testability pattern. Previously `createClaudeAuthProvider` and `createCodexAuthProvider` called `auth.GetCredential` directly — making them untestable. Now the factory tests can inject mock credential functions. Well done.

Minor note: `defaultAnthropicAPIBase` as a package-level constant avoids magic strings scattered across the factory — good practice.

Leeaandrob (approved):
## Re-Review: PR #213 — Refactor providers by protocol family

@jmahotiedu Excellent work on the fix commit. All 8 findings from my previous review have been addressed comprehensively, plus you made additional improvements I didn't ask for. Let me trace each one.

### Local Verification

| Check | Result |
|-------|--------|
| `go vet ./...` | PASS |
| `go build ./...` | PASS |
| `go test -race ./...` | PASS (all packages) |
| `go test -race ./pkg/providers/...` | PASS (4 packages) |
| CI checks | 3/3 PASS (fmt-check, vet, test) |

### Previous Findings — Resolution Status

**[C1 — Triple type duplication] → FIXED**

New `protocoltypes` package is the single source of truth. Both `openai_compat` and `anthropic` packages use Go type aliases (`type ToolCall = protocoltypes.ToolCall`), which is the correct Go pattern — zero runtime overhead, no conversion needed, preserves each package's public API. This eliminated ~180 lines of conversion boilerplate from `claude_provider.go` and `http_provider.go`.

**[C2 — Anthropic base URL hardcoded] → FIXED**

- `NewProviderWithBaseURL()` and `NewProviderWithTokenSourceAndBaseURL()` added to anthropic package
- `normalizeBaseURL()` correctly strips `/v1` suffix (SDK appends it internally)
- `createClaudeAuthProvider(apiBase)` now accepts and forwards the base URL
- Test `TestCreateProviderReturnsClaudeProviderForAnthropicOAuth` verifies the full flow: `config APIBase → factory → anthropic provider → BaseURL()` with `/v1` normalization

**[H1 — Breaking API change (variadic proxy)] → FIXED**

`NewHTTPProvider(apiKey, apiBase, proxy string)` restored to 3 positional args. No variadic. Same for `openai_compat.NewProvider`. API stability preserved.

**[H2 — Model prefix stripping incomplete] → FIXED**

New `normalizeModel()` function covers all routed providers: `moonshot`, `nvidia`, `groq`, `ollama`, `deepseek`, `google`, `openrouter`, `zhipu`. Plus the smart OpenRouter check preserves full model names when `apiBase` contains `openrouter.ai`. Tests verify both prefix stripping and OpenRouter passthrough.

**[H3 — Missing test coverage] → LARGELY FIXED**

Significant test additions:
- Factory integration tests: `CreateProvider` returns correct types for OpenRouter, CodexCLI, CodexCLIAuth, AnthropicOAuth, OpenAIOAuth
- `resolveProviderSelection` now has 16 test cases covering deepseek, nvidia, shengsuanyun, ollama, groq, moonshot, codex-code, copilot, error paths
- Auth testability via `var getCredential = auth.GetCredential` dependency injection
- `anthropic` package: round-trip Chat, token source refresh, base URL normalization, stop reasons
- `openai_compat` package: tool call parsing, HTTP errors, prefix stripping, proxy config, numeric option coercion, GLM max_completion_tokens

**[M1 — Identical code branches] → FIXED**

Tool call parsing in `openai_compat/provider.go` simplified to single `if tc.Function != nil` check.

**[M2 — Silent error swallowing] → FIXED**

Both `anthropic/provider.go:200` and `openai_compat/provider.go:164` now log with `log.Printf` before falling back to `{"raw": ...}`.

**[M3 — Silent proxy parse failure] → FIXED**

`openai_compat/provider.go:44` now logs `log.Printf("openai_compat: invalid proxy URL...")`. Appropriate for library code.

### Additional Improvements (Author Initiative)

These were NOT requested but significantly improve the PR:

- **[POSITIVE]** Proxy config was missing from multiple explicit providers (Groq, OpenAI, OpenRouter, Zhipu, Gemini, VLLM, ShengSuanYun, DeepSeek) — now all wire through `sel.proxy`. Bug prevented users from using proxies with these providers.
- **[POSITIVE]** `nvidia` explicit provider case added to factory switch (was only in model-inference fallback before)
- **[POSITIVE]** `asInt()` / `asFloat()` helpers handle numeric type coercion (`float64`, `int`, `int64`, `float32`) — JSON deserialization often yields `float64` where callers expect `int`
- **[POSITIVE]** `var getCredential = auth.GetCredential` makes factory testable via dependency injection — previously untestable auth paths now have full test coverage
- **[POSITIVE]** `normalizeBaseURL()` handles the config/SDK mismatch cleanly (config uses `/v1`, SDK adds it internally)

### Architecture Assessment (Post-Fix)

The final architecture is clean:

```
protocoltypes/types.go     ← Single source of truth for shared types
├── anthropic/provider.go   ← Anthropic-native protocol (SDK-based)
├── openai_compat/provider.go ← OpenAI chat/completions protocol
providers/
├── types.go               ← Type aliases + LLMProvider interface
├── factory.go             ← Protocol-aware routing (pure logic + DI)
├── claude_provider.go     ← Thin adapter → anthropic package
└── http_provider.go       ← Thin adapter → openai_compat package
```

This is textbook Strategy + Factory pattern. Each protocol handler is independently testable, the routing logic is pure (no side effects), and the adapter layer is minimal (no conversion boilerplate).

### Verdict: **APPROVE**

All critical, high, and medium findings resolved. Tests pass. Architecture is sound. The fix commit demonstrates strong attention to reviewer feedback — not just fixing what was flagged, but also finding related issues (proxy wiring, nvidia routing) and fixing those too.

Well done, @jmahotiedu.

PR #214, opened 2026-02-15
[refactor] Refactor main.go with cobra

xiaket:
Refactor main.go to use cobra, making it much more readable and maintainable.

It is to be noted that some cli behaviour has changed, for example, for some flags, main.go before refactor will do a `return` in the sub-command(for example, [here](https://github.com/sipeed/picoclaw/blob/214b201bfaf2577a6ea53896c1bf626a63af41c6/cmd/picoclaw/main.go#L681)), resulting in a exit code of 0. I would argue that this is wrong, and this PR will correct this behaviour.

### DIFF
--- a/cmd/picoclaw/agent.go
+++ b/cmd/picoclaw/agent.go
@@ -30,9 +30,9 @@ var agentCmd = &cobra.Command{
 }
 
 var (
-	agentDebug    bool
-	agentMessage  string
-	agentSession  string
+	agentDebug   bool
+	agentMessage string
+	agentSession string
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


xiaket:
Happy to resolve the conflicts, but would want some confirmation that this is where we want to be.

Leeaandrob:
@Zepan This PR refactors `main.go` using Cobra for CLI management. At +1861/-1403 lines, it's a significant restructuring of the entry point.

While Cobra is a solid choice for Go CLIs, this is a high-risk refactor that touches the core startup path. With 68+ open PRs in flight, merging this now would create conflicts for virtually every other PR that touches `main.go` or `cmd/`.

**Recommendation: Defer to a stabilization window.** This is good work but the timing is challenging — it should ideally land when the PR queue is smaller to minimize merge conflicts across the board.

xiaket:
Closing this as provider work is going in.

PR #221, opened 2026-02-15
Remove premature defer client.Stop() in GitHub Copilot provider

mattn:
GitHubCopilotProvider does not work because client is always closed immediately.


mattn:
I'm not sure, but I believe we must keep the client alive for the duration of the session.

Leeaandrob:
@Zepan This PR by @mattn fixes a premature `defer client.Stop()` in the GitHub Copilot provider — the client gets closed before it's used. Classic Go defer-ordering bug.

**Recommendation: Merge.** Small, targeted fix that prevents the Copilot provider from breaking.

nikolasdehor (approved):
Important bug fix. The `defer client.Stop()` inside a constructor was tearing down the Copilot connection as soon as `NewGitHubCopilotProvider` returned, making the provider immediately non-functional.

Good addition of error handling for `CreateSession` too — the previous `session, _ = ...` was silently swallowing errors.

One question (non-blocking): with the `defer` removed, is there a cleanup path for `client.Stop()` in the normal (non-error) lifecycle? Presumably the provider or channel manager calls `Close()`/`Stop()` on shutdown.

LGTM.

PR #222, opened 2026-02-15
security: fix workspace path restriction bypass in validatePath

mattn:
Fix a path validation bypass where paths outside the workspace could be accessed when the workspace path was a prefix of the target path.

Example: workspace `/tmp/abc` would allow access to `/tmp/abcdef/secret`.


mattn:
main branch had fixed this issue in https://github.com/sipeed/picoclaw/commit/e3f65fc3d62aa3a5577c0a59e6417d056ff14c02

PR #227, opened 2026-02-15
Fix shadowing field runnnig

mattn:
`running` field is shadowing in BaseChannel.running. So running field always be false.

### DIFF
--- a/pkg/channels/maixcam.go
+++ b/pkg/channels/maixcam.go
@@ -18,7 +18,6 @@ type MaixCamChannel struct {
 	listener   net.Listener
 	clients    map[net.Conn]bool
 	clientsMux sync.RWMutex
-	running    bool
 }
 
 type MaixCamMessage struct {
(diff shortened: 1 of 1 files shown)


Leeaandrob:
@Zepan Trivial but correct fix by @mattn — removes a shadowed field. Clean code hygiene.

**Recommendation: Merge.** 2-line fix, zero risk.

pkg/channels/maixcam.go
```diff
@@ -18,7 +18,6 @@ type MaixCamChannel struct {
 	listener   net.Listener
 	clients    map[net.Conn]bool
 	clientsMux sync.RWMutex
```
Leeaandrob, line 20: **[Positive — Correct Shadowing Fix]** @mattn — Clean removal. The embedded `*BaseChannel` already provides `running bool` (base.go:23). The duplicate field here caused `c.running` (direct access at line 73) to always read `false`, since `setRunning()` operates on the base struct field.

**Minor observation (not blocking):** Line 73 is the only place across all 12 channels that uses direct field access (`c.running`) instead of the `IsRunning()` method. While it works correctly now through Go embedding promotion, using `c.IsRunning()` would be more consistent with the rest of the codebase.

PR #238, opened 2026-02-15
fix(claude): resolve empty tool name in Claude API requests

pHequals7:
## Summary

- Fix 400 Bad Request errors when using Claude/Anthropic as the LLM provider
- The `ToolCall` struct stores tool names in `Function.Name` (OpenAI format), but `claude_provider.go` was reading `tc.Name` (top-level), which is always empty
- Also handles `Arguments` stored as a JSON string in `Function.Arguments` rather than the top-level map

## Root Cause

In `pkg/agent/loop.go`, tool calls are stored in OpenAI's format:
```go
ToolCall{
    Function: &FunctionCall{
        Name:      tc.Name,      // name goes here
        Arguments: string(JSON),  // args go here as string
    },
    // tc.Name is empty
    // tc.Arguments is nil
}
```

But `pkg/providers/claude_provider.go` reads the top-level fields:
```go
anthropic.NewToolUseBlock(tc.ID, tc.Arguments, tc.Name)  // both empty/nil
```

This causes every multi-turn conversation with tool calls to fail with:
```
messages.N.content.0.tool_use.name: String should have at least 1 character
```

## Fix

Fall back to `Function.Name` and `Function.Arguments` when top-level fields are empty:
```go
name := tc.Name
if name == "" && tc.Function != nil {
    name = tc.Function.Name
}
args := tc.Arguments
if args == nil && tc.Function != nil && tc.Function.Arguments != "" {
    json.Unmarshal([]byte(tc.Function.Arguments), &args)
}
blocks = append(blocks, anthropic.NewToolUseBlock(tc.ID, args, name))
```

## Test plan

- [x] `go build ./...` compiles
- [x] Tested with Claude Sonnet 4.5 via setup-token auth — multi-turn conversations with tool calls work correctly
- [x] Verified on Telegram channel with real tool executions (web search, file read)

### DIFF
--- a/pkg/providers/claude_provider.go
+++ b/pkg/providers/claude_provider.go
@@ -81,7 +81,15 @@ func buildClaudeParams(messages []Message, tools []ToolDefinition, model string,
 					blocks = append(blocks, anthropic.NewTextBlock(msg.Content))
 				}
 				for _, tc := range msg.ToolCalls {
-					blocks = append(blocks, anthropic.NewToolUseBlock(tc.ID, tc.Arguments, tc.Name))
+					name := tc.Name
+					if name == "" && tc.Function != nil {
+						name = tc.Function.Name
+					}
+					args := tc.Arguments
+					if args == nil && tc.Function != nil && tc.Function.Arguments != "" {
+						json.Unmarshal([]byte(tc.Function.Arguments), &args)
+					}
+					blocks = append(blocks, anthropic.NewToolUseBlock(tc.ID, args, name))
 				}
 				anthropicMessages = append(anthropicMessages, anthropic.NewAssistantMessage(blocks...))
 			} else {


Leeaandrob:
@Zepan Fixes empty tool name in Claude API requests — this causes 400 errors from Anthropic's API. One-line targeted fix.

**Recommendation: Merge.** +9/-1, prevents Claude provider from breaking.

nuc134r:
> [07.03.2026 21:57] Sergey Tikhonov: Hello there!
> [07.03.2026 21:57] Assistant: Error processing message: LLM call failed after retries: claude API call: POST "https://api.anthropic.com/v1/messages": 400 Bad Request (Request-ID: req011CYpEQ8bZZNcWzeotK6xTW) {"type":"error","error":{"type":"invalidrequesterror","message":"messages.5.content.1.tooluse.name: String should have at least 1 character"},"requestid":"req011CYpEQ8bZZNcWzeotK6xTW"}


I still have this error. On first launch it was ok, now it fails. Is this not merged?

nuc134r:
Fixed like this locally using Claude, seem to work well 

```
diff --git a/pkg/providers/anthropic/provider.go b/pkg/providers/anthropic/provider.go
index 242ded1..1b4195d 100644
--- a/pkg/providers/anthropic/provider.go
+++ b/pkg/providers/anthropic/provider.go
@@ -189,7 +189,11 @@ func buildParams(
                                        if args == nil {
                                                args = map[string]any{}
                                        }
-                                       blocks = append(blocks, anthropic.NewToolUseBlock(tc.ID, args, tc.Name))
+                                       name := tc.Name
+                               if name == "" && tc.Function != nil {
+                                       name = tc.Function.Name
+                               }
+                               blocks = append(blocks, anthropic.NewToolUseBlock(tc.ID, args, name))
                                }
                                anthropicMessages = append(anthropicMessages, anthropic.NewAssistantMessage(blocks...))
                        } else {
```

PR #249, opened 2026-02-15
feat: add REST API with OpenAPI 3.1 spec

pkonowrocki:
## Summary
- Add REST API endpoints: `/api/chat` (POST), `/api/sessions` (GET), `/api/health` (GET), `/api/openapi.yaml` (GET)
- Bearer token authentication via `Authorization` header with configurable API key
- OpenAPI 3.1 specification embedded and served at `/api/openapi.yaml`
- Add `APIConfig` struct, `GetSessionManager()` accessor, and `ListSessions()` method

## Test plan
- [x] Auth middleware tests (valid/invalid/missing token, public paths, open mode)
- [x] Health endpoint returns status and version
- [x] Spec endpoint returns valid YAML with correct content-type
- [x] Chat endpoint validates required message field
- [x] Chat endpoint rejects non-POST methods
- [x] Chat endpoint rejects invalid JSON
- [ ] Manual: `curl -H "Authorization: Bearer <key>" -d '{"message":"hi"}' localhost:18790/api/chat`

### DIFF
--- a/api/openapi.yaml
+++ b/api/openapi.yaml
@@ -1,0 +1,89 @@
+openapi: "3.1.0"
+info:
+  title: PicoClaw API
+  version: "1.0.0"
+  description: REST API for PicoClaw AI Gateway
+servers:
+  - url: http://localhost:18790
+security:
+  - ApiKeyAuth: []
+components:
+  securitySchemes:
+    ApiKeyAuth:
+      type: http
+      scheme: bearer
+  schemas:
+    ChatRequest:
+      type: object
+      required: [message]
+      properties:
+        session_id: { type: string, description: "Session ID (auto-generated if omitted)" }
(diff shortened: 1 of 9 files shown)


Leeaandrob:
@Zepan This PR adds a REST API with OpenAPI 3.1 spec (+1542/-114). While a REST API could be valuable for programmatic access, it's not currently on the roadmap board.

**Consideration:** This adds an HTTP server to a project designed for <10MB RAM on $10 hardware. The memory and attack surface implications should be evaluated carefully. Also, MCP support (#290, priority: high) may provide a more standard integration protocol for the same use cases.

**Recommendation: Defer.** Not on the roadmap, and MCP (#282) may be a better fit for the extensibility story. Worth revisiting after MCP lands.

PR #261, opened 2026-02-16
feat(skills): add retry for HTTP requests in skill installer

yumosx:
**Change**
- Introduced a new lightweight retry utility: `utils/http_retry.go`
     Linear Incremental Retry with Context Support
    - Retry up to 3 times with linearly increasing delays (1s, 2s, 3s)
    - Retry only on 429 or 5xx status codes
    - Respect context cancellation during retry delays
 
- Integrated retry logic into the skill installer
  - Applied to key GitHub fetch/clone operations
  - Default: up to 3 retries, starting at ~1s, with reasonable max backoff

**Tests**
- Added comprehensive unit tests in `utils/http_retry_test.go` covering:

### DIFF
--- a/pkg/utils/http_retry_test.go
+++ b/pkg/utils/http_retry_test.go
@@ -29,22 +29,6 @@ func TestDoRequestWithRetry(t *testing.T) {
 			wantAttempts: 1,
 		},
 		{
-			name: "success-on-second-attempt",
-			serverBehavior: func(server *httptest.Server) int {
-				return 1
-			},
-			wantSuccess:  true,
-			wantAttempts: 2,
-		},
-		{
-			name: "success-on-third-attempt",
-			serverBehavior: func(server *httptest.Server) int {
-				return 2
-			},
-			wantSuccess:  true,
-			wantAttempts: 3,
-		},
-		{
 			name: "fail-all-attempts",
 			serverBehavior: func(server *httptest.Server) int {
 				return 4


afjcjsbx:
Hi, the retry idea is good but the way the function is structured it does a blind retry on anything that is not a 200 OK. It might be convenient to define a list of status codes worth retries for example (5XX, 429) so that resources and time are not wasted. There is also a problem with closing resp.Body, otherwise it keeps the TCP connection blocked. This loop overwrites resp with each iteration without closing previous failed responses.

yumosx:
You're correct. I need to refine the retry logic to be more specific and ensure the response body is properly closed to avoid blocking connections.

yumosx:
@xiaket @afjcjsbx all done. please check.

xiaket:
Heya @yumosx thanks for your PR! Do you mean to implement exponential backoff? Asking because the current implementation does not look like that.

xiaket:
Also the title may be misleading, this is only scoped to the retry in the skill installer, but the title reads like we are adding retry to all the HTTP things in skill space.

pkg/skills/installer.go
```diff
@@ -23,6 +23,45 @@ type AvailableSkill struct {
 	Tags        []string `json:"tags"`
 }
 
+type BuiltinSkill struct {
```
xiaket: Why do we need this struct?

pkg/skills/installer.go
```diff
@@ -23,6 +23,45 @@ type AvailableSkill struct {
...
+}
+
+const maxRetries = 3
+
+func shouldRetry(statusCode int) bool {
+	return statusCode == http.StatusTooManyRequests ||
+		statusCode >= 500
+}
+
+func doRequestWithRetry(client *http.Client, req *http.Request) (*http.Response, error) {
```
xiaket: should probably receive a context object in the args so context cancel can be correctly handled.

pkg/skills/installer_test.go
```diff
@@ -0,0 +1,133 @@
...
+			name: "fail-all-attempts",
+			serverBehavior: func(server *httptest.Server) int {
+				return 4
+			},
+			wantSuccess:  false,
+			wantAttempts: 3,
+		},
+		{
+			name: "non-ok-status-code",
+			serverBehavior: func(server *httptest.Server) int {
```
xiaket: How come this is a non-ok-status-code?

pkg/utils/http_retry_test.go
xiaket, line 1: I also have a concern that these tests will slow down our tests process by a lot.

yumosx:
> Heya @yumosx thanks for your PR! Do you mean to implement exponential backoff? Asking because the current implementation does not look like that.

Implementing backoff was indeed my original plan, but I was worried it might introduce extra dependencies and potentially cause memory overhead, so I opted for this simple retry function instead.

pkg/skills/installer.go
yumosx: ooh, missing add, i will remove it

xiaket:
> Implementing backoff was indeed my original plan, but I was worried it might introduce extra dependencies and potentially cause memory overhead, so I opted for this simple retry function instead.

To be clear I think the current mechanism looks good to me, but my concern is the PR description is not aligned with the implementation.

yumosx:
> > Implementing backoff was indeed my original plan, but I was worried it might introduce extra dependencies and potentially cause memory overhead, so I opted for this simple retry function instead.
> 
> To be clear I think the current mechanism looks good to me, but my concern is the PR description is not aligned with the implementation.

Thanks for your suggestion. I will improve it later as well.

pkg/skills/installer.go
yumosx: hello, @xiaket 
The context is already embedded in the request object. When we create the request using `http.NewRequestWithContext(ctx, ...)` , the context is bound to the request. The `client.Do(req)` call automatically uses this context for timeout and cancellation control.

yumosx:
> Also the title may be misleading, this is only scoped to the retry in the skill installer, but the title reads like we are adding retry to all the HTTP things in skill space.

We can wait until the code is stable or the PR is merged, and then I will create a PR to add this retry mechanism to other HTTP requests as well. WDYT ？

pkg/skills/installer.go
yumosx: But has a bug, I using `time.sleep` here doesn't perceive the outside context. I've fixed it.

pkg/skills/installer_test.go
yumosx: i've removed it.

pkg/utils/http_retry_test.go
yumosx: I extracted the unit of the retry waiting time and adjusted it to ms in the test, which can speed up the testing.

yumosx:
@xiaket  all done, please check.

xiaket:
In the PR description, it is still mentioning exponential backoff.

pkg/utils/http_retry_test.go
```diff
@@ -0,0 +1,134 @@
...
+		{
+			name: "success-on-first-attempt",
+			serverBehavior: func(server *httptest.Server) int {
+				return 0
+			},
+			wantSuccess:  true,
+			wantAttempts: 1,
+		},
+		{
+			name: "success-on-second-attempt",
```
xiaket: Please remove this test case.

yumosx:
thanks for your CR, all done, please check.

PR #239, opened 2026-02-15
feat: Coolify deployment support and Telegram UI/UX enhancements

mrbeandev:
This PR introduces comprehensive support for Coolify deployment and several improvements to the Telegram channel to enhance the self-hosting experience.

### ☁️ Coolify Deployment
- Adds [COOLIFY.md](cci:7://file:///mnt/main_disk/mrbeandev_server/picoclaw/picoclaw/COOLIFY.md:0:0-0:0) guide with 3 configuration methods (JSON, Env Vars, File).
- Adds [entrypoint-coolify.sh](cci:7://file:///mnt/main_disk/mrbeandev_server/picoclaw/picoclaw/entrypoint-coolify.sh:0:0-0:0) to dynamically generate [config.json](cci:7://file:///mnt/main_disk/mrbeandev_server/picoclaw/picoclaw/config/config.json:0:0-0:0) from environment variables at runtime.
- Adds `Dockerfile.coolify` and [docker-compose-coolify.yml](cci:7://file:///mnt/main_disk/mrbeandev_server/picoclaw/picoclaw/docker-compose-coolify.yml:0:0-0:0) optimized for Coolify's environment.
- Support for full JSON configuration via `PICOCLAW_CONFIG_JSON` environment variable.

### 🤖 Telegram Enhancements
- **Persistent Typing Indicator**: Implemented a repeating goroutine that sends the 'typing' status every 4s while the AI is thinking, preventing the status from expiring (fixing the 5s Telegram API limit).
- **Auto-Command Registration**: The bot now automatically registers its commands (`/model`, `/models`) with Telegram on startup using `setMyCommands`.
- **Unified /model Command**: Supports a new `provider/model` syntax (e.g., `/model vllm/qwen3-coder-next:cloud`) to switch both the backend API and the model name in a single message.
- **Dynamic /models List**: The `/models` command now dynamically reads the loaded configuration to show actually available providers and the currently active model.

### ⚙️ Core Improvements
- Refactored [AgentLoop](cci:2://file:///mnt/main_disk/mrbeandev_server/picoclaw/picoclaw/pkg/agent/loop.go:31:0-45:1) to store a reference to the configuration, allowing for hot-switching of providers and models without needing to restart the container.

These changes make PicoClaw significantly easier to deploy on cloud platforms and improve the interactability of the Telegram bot.

### 🔗 Related Issues
- Closes #17: Provides an explicit way to select providers via `/model provider/model`.
[...]

PR #270, opened 2026-02-16
fix: resolve provider env vars not being applied (#66)

### LINKED ISSUE #66: PICOCLAW_PROVIDERS_* env vars not applied because {{.Name}} is unsupported by caarlos0/env
### Summary
Environment variables like `PICOCLAW_PROVIDERS_ANTHROPIC_API_KEY` and other provider keys are never read. The application looks for `PICOCLAW_PROVIDERS_{{.Name}}_API_KEY` literally instead of substituting the provider name.

### Environment
- PicoClaw version: v0.0.1 (Docker)
- OS: Any
- Config: JSON with env var overrides

### Steps to reproduce

1. Set `PICOCLAW_PROVIDERS_ANTHROPIC_API_KEY` in the environment or `.env`.
2. Set `PICOCLAW_AGENTS_DEFAULTS_MODEL=claude-opus-4-6`.
3. Use a config that only defines other providers (e.g. `openai`), not anthropic.
4. Run `picoclaw gateway` (or equivalent).
**Result:** Error such as `Error creating provider: no API key configured for provider (model: claude-opus-4-6)`.

### Expected behavior
The value of `PICOCLAW_PROVIDERS_ANTHROPIC_API_KEY` should be used for the Anthropic provider (and similarly for other providers).

### Root cause
In `pkg/config/config.go`, `ProviderConfig` uses:
```
type ProviderConfig struct {
    APIKey   string `json:"api_key" env:"PICOCLAW_PROVIDERS_{{.Name}}_API_KEY"`
    APIBase  string `json:"api_base" env:"PICOCLAW_PROVIDERS_{{.Name}}_API_BASE"`
    AuthMethod string `json:"auth_method,omitempty" env:"PICOCLAW_PROVIDERS_{{.Name}}_AUTH_METHOD"`
}
```
`github.com/caarlos0/env/v11` treats the env tag as a literal string. It does not support `{{.Name}}` or similar placeholders. It therefore looks for the env var `PICOCLAW_PROVIDERS_{{.Name}}_API_KEY` (literally), which is never set.

boengai:
## Summary
  - Fix #66: `PICOCLAW_PROVIDERS_*` env vars (e.g. `PICOCLAW_PROVIDERS_ANTHROPIC_API_KEY`) were not being read because `caarlos0/env` does not support `{{.Name}}` template syntax in env tags
  - Replace broken template tags with `envPrefix` on each provider field in `ProvidersConfig`, which correctly prepends the provider name to each env key
  - Add test verifying env vars are properly parsed for multiple providers

  ## Test plan
  - [x] `go fmt` — no formatting changes needed
  - [x] `go vet ./...` — no issues
  - [x] `go test ./...` — all packages pass
  - [x] New `TestProviderConfig_EnvVars` test confirms env vars like `PICOCLAW_PROVIDERS_ANTHROPIC_API_KEY` are correctly read

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -119,7 +119,7 @@ clean:
 	@rm -rf $(BUILD_DIR)
 	@echo "Clean complete"
 
-## fmt: Format Go code
+## vet: Run go vet for static analysis
 vet:
 	@$(GO) vet ./...
 
@@ -131,10 +131,18 @@ test:
 fmt:
 	@$(GO) fmt ./...
 
-## deps: Update dependencies
+## deps: Download dependencies
 deps:
+	@$(GO) mod download
+	@$(GO) mod verify
+
+## update-deps: Update dependencies
+update-deps:
 	@$(GO) get -u ./...
 	@$(GO) mod tidy
+
+## check: Run vet, fmt, and verify dependencies
+check: deps fmt vet test
 
 ## run: Build and run picoclaw
 run: build
--- a/cmd/picoclaw/main.go
+++ b/cmd/picoclaw/main.go
@@ -562,7 +562,7 @@ func gatewayCmd() {
 		})
 
 	// Setup cron tool and service
-	cronService := setupCronTool(agentLoop, msgBus, cfg.WorkspacePath())
+	cronService := setupCronTool(agentLoop, msgBus, cfg.WorkspacePath(), cfg.Agents.Defaults.RestrictToWorkspace)
 
 	heartbeatService := heartbeat.NewHeartbeatService(
 		cfg.WorkspacePath(),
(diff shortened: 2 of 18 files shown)


mrbeandev:
Great fix for those env-var tags! This correction is essential for anyone running without a static config. In **PR #239**, I've implemented a related dynamic provider discovery system using reflection that relies on these exact fields. Your fix will help ensure those providers are detected correctly. Looking forward to getting these consistent!

Leeaandrob:
@Zepan Fixes provider environment variables not being applied when config.json exists — this means `OPENAI_API_KEY` env var gets ignored if you have a config file. Real usability bug.

**Recommendation: Merge.** +46/-18, fixes a configuration precedence issue that affects Docker/container deployments where env vars are the standard config mechanism.

PR #271, opened 2026-02-16
fix: env overrides when config.json is missing and add regression test

tbeaudouin05:
- **Problem**: When config.json is absent (common in Fly deployments using secrets/env only), LoadConfig returned defaults without applying env overrides, so the app still used the default model (glm-4.7) and failed due to missing credentials.
- **Fix**: Always run env.Parse(cfg) even when config.json is missing, so secrets/env values take effect in env-only deployments.
- **Test**: Added TestLoadConfig_EnvOverridesWithoutFile to verify env vars (e.g., Fly secrets for model/provider) are honored without a config file.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -340,6 +340,9 @@ func LoadConfig(path string) (*Config, error) {
 	data, err := os.ReadFile(path)
 	if err != nil {
 		if os.IsNotExist(err) {
+			if err := env.Parse(cfg); err != nil {
+				return nil, err
+			}
 			return cfg, nil
 		}
 		return nil, err
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"path/filepath"
 	"testing"
 )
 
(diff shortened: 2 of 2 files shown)


mrbeandev:
Good catch on the LoadConfig logic! It seems like we're both looking at the parsing order of environment variables. I've actually just updated **PR #239** which handles this by unsetting conflicting env vars at the entrypoint level (ensuring the JSON remains the source of truth) while also adding full Coolify/Docker support. It might be worth checking how we can unify these so they don't fight each other!

Leeaandrob:
@Zepan Fixes env overrides when `config.json` is missing entirely, plus regression test. Companion to #270.

**Recommendation: Merge.** +19 lines with a test. Fixes the bootstrapping experience for new users who haven't created a config file yet.

nikolasdehor (approved):
Clean, minimal fix for an important edge case. In file-less deployments (Fly.io, Docker with env-only config), `LoadConfig` was returning `DefaultConfig()` without calling `env.Parse(cfg)`, silently ignoring all environment variable overrides.

The regression test is well-written — uses `t.Setenv` for proper cleanup and `filepath.Join(t.TempDir(), "missing.json")` to guarantee a non-existent path. LGTM.

PR #272, opened 2026-02-16
Support building for linux/loong64

xen0n:
Happy Chinese New Year!

This platform has a growing desktop and embedded user base, and is fully supported by Go. The only necessary change is the mapping between `uname -m` output and GOARCH.

Due to [non-technical reasons][1], there is currently no Docker official image that provides linux/loong64 support, so Docker-based builds are not included in this commit for now.

[1]: https://github.com/docker-library/official-images/issues/16404

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,8 @@ ifeq ($(UNAME_S),Linux)
 		ARCH=amd64
 	else ifeq ($(UNAME_M),aarch64)
 		ARCH=arm64
+	else ifeq ($(UNAME_M),loongarch64)
+		ARCH=loong64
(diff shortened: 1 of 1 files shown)


lxowalle:
Thanks for the pr~

Orgmar:
@xen0n Happy (belated) Chinese New Year! loong64 support is a welcome addition, especially with the growing user base on that platform. Clean change too, just the uname to GOARCH mapping.

We have the **PicoClaw Dev Group** on Discord for contributors. If you're interested in joining, email `support@sipeed.com` with the subject `[Join PicoClaw Dev Group] + Your GitHub account` and we'll send the invite link!

PR #277, opened 2026-02-16
feat: update the `make deps` logic to prevent the project from freque…

lxowalle:
…ntly updating dependency package versions

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,8 +39,6 @@ ifeq ($(UNAME_S),Linux)
 		ARCH=amd64
 	else ifeq ($(UNAME_M),aarch64)
 		ARCH=arm64
-	else ifeq ($(UNAME_M),loongarch64)
-		ARCH=loong64
 	else ifeq ($(UNAME_M),riscv64)
 		ARCH=riscv64
 	else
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -39,8 +39,6 @@ ifeq ($(UNAME_S),Linux)
 		ARCH=amd64
 	else ifeq ($(UNAME_M),aarch64)
 		ARCH=arm64
-	else ifeq ($(UNAME_M),loongarch64)
-		ARCH=loong64
```
xen0n, line 43: #272 got inadvertently reverted in this PR? I just got bit when rebasing my work on top of latest master...

lxowalle: I fixed it here https://github.com/sipeed/picoclaw/pull/361

bogdanovich:
Addressed the admission wire-contract review finding in 8865dd68: the Go result now matches `node-auth.v1`, and conformance coverage validates the actual result returned by `Admit`. Local verification: race tests for `./pkg/nodes/...`, targeted golangci-lint, and docs lint all pass.

bogdanovich:
Addressed the registry namespace finding in 9e34bd27. Node IDs and aliases are now collision-free in both insertion orders, with the same invariant checked on reload. Race tests for `./pkg/nodes/...`, targeted golangci-lint, and docs lint pass.

PR #299, opened 2026-02-16
Replace \s+ with [^\S\n]+ to preserve newlines

mattn:
`extractText` in `pkg/tools/web.go` replaces all whitespace (including newlines) with a single space using `\s+`, then tries to split by `\n` to filter empty lines. Since all newlines are already gone, the output is always a single line.

### DIFF
--- a/pkg/tools/web.go
+++ b/pkg/tools/web.go
@@ -419,8 +419,10 @@ func (t *WebFetchTool) extractText(htmlContent string) string {
 
 	result = strings.TrimSpace(result)
 
-	re = regexp.MustCompile(`\s+`)
-	result = re.ReplaceAllLiteralString(result, " ")
+	re = regexp.MustCompile(`[^\S\n]+`)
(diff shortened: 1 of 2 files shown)


Leeaandrob:
@Zepan This PR by @mattn replaces `\s+` with `[^\S\n]+` in regex patterns to preserve newlines. This is a subtle but important fix — without it, multi-line content gets collapsed into single lines.

**Recommendation: Merge.** Correct regex behavior that preserves formatting. Well-tested by a contributor who clearly understands Go regex edge cases.

kuator:
> Great bug fix. The `[^\S\n]+` idiom is the correct regex for horizontal-whitespace-only collapsing, and the `\n{3,}` → `\n\n` pass prevents excessive blank lines. The old `\s+` was destroying all paragraph structure by collapsing newlines.
> 
> Tests are thorough — table-driven with edge cases for script/style removal, blank line collapsing, and empty input. LGTM.

chatgpt ahh response

PR #304, opened 2026-02-16
feat(linters): Added golangci-lint config & CI job

mymmrac:
This PR adds [golangci-lint](https://golangci-lint.run) config, it's not final and needs tweaking, but it should be a good starting point,

For now most linters are disabled (because we have 2600+ issues currently) and we should enable them as we fix them. 

> Note: This PR doesn't include auto formatting, will be done in later PRs.

### DIFF
--- a/pkg/providers/claude_provider.go
+++ b/pkg/providers/claude_provider.go
@@ -3,6 +3,7 @@ package providers
 import (
 	"context"
 	"fmt"
+
 	anthropicprovider "github.com/sipeed/picoclaw/pkg/providers/anthropic"
 )
 
--- a/pkg/providers/http_provider.go
+++ b/pkg/providers/http_provider.go
@@ -8,6 +8,7 @@ package providers
 
 import (
 	"context"
+
 	"github.com/sipeed/picoclaw/pkg/providers/openai_compat"
 )
 


Leeaandrob:
@Zepan This PR adds `golangci-lint` configuration and CI job. With 68+ open PRs and rapid growth, automated linting is essential for maintaining code quality at scale.

**Recommendation: Merge.** This is infrastructure that pays dividends on every future PR. The roadmap explicitly calls for better CI/CD (volunteer role #5: "AI-Powered CI/CD Optimization").

.github/workflows/pr.yml
```diff
@@ -1,17 +1,39 @@
...
+
+      - name: Setup Go
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: go.mod
+
+      - name: Run go generate
+        run: go generate ./...
+
+      - name: Golangci Lint
```
Leeaandrob, line 22: **[MEDIUM]** `version: latest` combined with `default: all` in `.golangci.yaml` means any new linter introduced in a golangci-lint release will be auto-enabled and could break all PRs.

Suggestion: Pin to a specific version:
```yaml
version: v2.1
```

This way CI upgrades are intentional, not accidental.

.golangci.yaml
```diff
@@ -0,0 +1,184 @@
...
+    - funlen
+    - gochecknoinits
+    - gocognit
+    - goconst
+    - gocritic
+    - gocyclo
+    - godox
+    - goprintffuncname
+    - gosec
+    - govet
```
Leeaandrob, line 50: **[MEDIUM]** `govet` is disabled here but already runs successfully as a separate CI job (`vet` in pr.yml). Consider enabling it now to get immediate value from golangci-lint — it's a zero-risk enable since the separate `vet` job already proves it passes.

Once enabled here, the separate `vet` job can be removed (as noted in the TODO).

.golangci.yaml
```diff
@@ -0,0 +1,184 @@
+version: "2"
```
Leeaandrob, line 1: **[POSITIVE]** Using golangci-lint v2 config format with `default: all` is the right call. The two-block disable strategy (permanent project preferences vs temporary failures) makes the improvement roadmap very clear.

Nice config overall — the revive rules, test exclusions, and formatter setup are all well-tuned.

Leeaandrob:
@mymmrac take look. After that we can accept without problem.

mymmrac:
Thanks for the review, from the comment all apart from fixed version is by design, if I will enable all linters and formatters right away we will have a PR that touches almost all files in the project

I will make linter version fixed, if you see a need for change in anything else let's discuss

PR #315, opened 2026-02-16
fix: retry with nudge when LLM returns empty response after tool use

elseasama:
## Summary

- Some models (e.g. `kimi-k2.5` via NVIDIA API) occasionally return empty content after processing tool call results, causing the bot to reply with "I've completed processing but have no response to give."
- When the LLM returns empty content after a tool use iteration (`iteration > 1`), the agent now appends a nudge message and retries, giving the model another chance to generate a proper response
- The retry only happens once and respects `maxIterations` to avoid infinite loops

## Test plan

- [ ] Send a message that triggers a tool call (e.g. web search) and verify the bot responds with actual content
- [ ] Verify that direct answers (no tool calls) are unaffected
- [ ] Verify that the nudge retry only fires once per request (check logs for "Empty response after tool use, retrying with nudge")

### DIFF
--- a/pkg/agent/loop.go
+++ b/pkg/agent/loop.go
@@ -468,6 +468,19 @@ func (al *AgentLoop) runLLMIteration(ctx context.Context, messages []providers.M
 					"iteration":     iteration,
 					"content_chars": len(finalContent),
 				})
+			// If content is empty and we just processed tool results (iteration > 1),
+			// retry once by adding a nudge message to prompt the model to respond.
+			if strings.TrimSpace(finalContent) == "" && iteration > 1 && iteration < al.maxIterations {
+				logger.InfoCF("agent", "Empty response after tool use, retrying with nudge",
+					map[string]interface{}{
+						"iteration": iteration,
+					})
+				messages = append(messages, providers.Message{
+					Role:    "user",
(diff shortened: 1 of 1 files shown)


Leeaandrob:
@Zepan When an LLM returns an empty response after tool use, the conversation stalls. This PR adds a retry with nudge — a practical resilience fix that improves user experience across all providers.

**Recommendation: Merge.** +13 lines, zero risk, solves a real UX problem.

nikolasdehor (changes requested):
The intent is right — empty responses after tool use are a real problem with some models. However:

1. **Not actually "retry once"**: The comment says "retry once" but there's no flag to prevent retrying on every subsequent iteration where content is empty. Consider adding a `nudgeRetried` bool to enforce single-retry semantics.

2. **Message history pollution**: The nudge message permanently stays in the conversation. Subsequent iterations see this synthetic user message, which could confuse the model or cause unexpected behavior.

3. **Code placement**: Please verify the scoping — the `if` block sits between a structured log call's closing `})` and the `break` statement. Make sure the `break` is still reached for non-empty content.

Suggestion: add a `nudgeRetried` flag, and consider removing the nudge message from history after a successful retry.

PR #308, opened 2026-02-16
feat: add ActionBook browser automation tools

Danieldd28:
## Goal

Enable PicoClaw agents to perform browser automation (opening pages, clicking elements, filling forms, taking screenshots, extracting text) via native tool calls.

## Changes

Three new tools wrapping the ActionBook CLI:

| Tool | Purpose | CLI Command |
|------|---------|-------------|
| `browser_search` | Find action manuals for a website/task | `actionbook search` |
| `browser_get` | Get CSS selectors and page structure by ID | `actionbook get` |
| `browser` | Execute browser commands (open, click, fill, text, etc.) | `actionbook browser *` |

Workflow: `browser_search` -> `browser_get` -> `browser` (open -> interact -> close)

### New Files
- `pkg/tools/browser.go` - Three tool structs implementing the Tool interface
- `pkg/tools/browser_test.go` - 18 unit tests + 1 integration test

### Modified Files
- `pkg/config/config.go` - Added BrowserConfig struct (enabled, headless)
- `pkg/agent/loop.go` - Registered tools in createToolRegistry()

## Design Decisions

- **Zero new dependencies** - thin CLI wrapper using os/exec, same pattern as ExecTool
- **Action allowlist** - only whitelisted browser actions are permitted (no arbitrary CLI injection)
- **Config-gated** - `tools.browser.enabled` (default: true), `tools.browser.headless` (default: true)
- **Lean** - ~310 lines, lighter than web.go (436 lines)

## Testing

All 18 unit tests pass. Integration test with live actionbook search passes.
Verified end-to-end with `picoclaw agent` - the LLM successfully used the browser tool to open example.com, extract page text, and close the browser.

## Requirements

Requires `actionbook` CLI (v0.7.3+) in PATH. If not found, tools return a clear error without crashing.

Issue #293, opened 2026-02-16
Feature: Autonomous Browser Operations


Zepan:
### **Overview**

To expand PicoClaw's operational reach to the web, we are seeking to implement browser automation capabilities. This will allow the AI to interact with websites directly—navigating pages, extracting data, and performing actions like a human user. We are considering two primary paths for this: **Chrome DevTools Protocol (CDP)** for low-level, high-speed control, or **ActionBook** for a more abstracted, action-oriented approach, or **webmcp** for chrome.

### **Core Requirements**

* **Browser Driver Integration**: Implement a lightweight browser controller (e.g., using `chromedp` for Go) to manage Chromium-based instances.
* **Action Primitives**: Support for fundamental web interactions:
* `click(selector)`
* `type(selector, text)`
* `scroll(direction/amount)`
* `screenshot()` for VLM-based visual reasoning.

* **Headless Support**: Ensure the browser can run in headless mode to minimize resource usage, staying true to PicoClaw’s "small and fast" philosophy.
* **ActionBook Compatibility**: Explore integration with ActionBook to leverage pre-defined web action sequences and improve the reliability of complex workflows.

### **Use Case Examples**

1. **Information Retrieval**: "Search for the latest news on Go 1.24 and summarize the top 3 articles."
2. **Task Automation**: "Log in to my dashboard and export the weekly report as a PDF."
3. **Dynamic Interaction**: "Find the cheapest flight from Shanghai to Tokyo on this travel site."

nikolasdehor:
Good feature request. For the Go implementation, `chromedp` is the natural choice — it's pure Go, well-maintained, and has excellent headless support.

**Architecture suggestion:**

This should be implemented as a tool (not a provider), fitting picoclaw's existing pattern:

```go
// pkg/tools/browser.go
type BrowserTool struct {
    allocCtx context.Context
    cancel   context.CancelFunc
    timeout  time.Duration
}

func (b *BrowserTool) Definition() ToolDef {
    return ToolDef{
        Name: "browser",
        Description: "Navigate web pages, extract content, take screenshots",
        Parameters: map[string]Param{
            "action":   {Type: "string", Enum: []string{"navigate", "click", "type", "screenshot", "extract"}},
            "url":      {Type: "string", Description: "URL to navigate to"},
            "selector": {Type: "string", Description: "CSS selector for click/type actions"},
            "text":     {Type: "string", Description: "Text to type"},
        },
    }
}
```

**Key implementation considerations:**

[...]

PR #187, opened 2026-02-15
feat: add browser tool powered by playwright-go

strelov1:
## Summary

- Add browser automation tool using `playwright-go`
- Support two connection protocols:
  - **CDP** (`ConnectOverCDP`) for Chromium/Browserless
  - **Playwright Wire Protocol** (`Firefox.Connect`) for Firefox/Camoufox
- 13 browser actions: navigate, click, type, screenshot, get_text, evaluate, wait, scroll, hover, select, pdf, cookies, close
- Add `BrowserConfig` to config with protocol, URLs, token, stealth, timeouts
- Unit tests (10 tests) + integration test scaffolding for both protocols

## Configuration

```json
"tools": {
  "browser": {
    "enabled": true,
    "protocol": "cdp",
    "cdp_url": "ws://localhost:3000",
    "ws_url": "",
    "token": "YOUR_TOKEN",
    "stealth": true,
    "launch_timeout": 120000,
    "action_timeout": 30000
  }
}
```

## New dependency

- `github.com/playwright-community/playwright-go` v0.5200.1

## Test plan

- [x] `go build ./pkg/...` — compiles without errors
- [x] `go test ./pkg/...` — all existing tests pass
- [x] 10 unit tests for browser tool pass
- [x] Integration test with Browserless (CDP): `BROWSER_TEST_CDP_URL=ws://... go test ./pkg/tools/ -run TestBrowserTool_Integration_CDP -v`
- [x] Integration test with Camoufox (Playwright): `BROWSER_TEST_PW_URL=ws://... go test ./pkg/tools/ -run TestBrowserTool_Integration_Playwright -v`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #318, opened 2026-02-16
feat: add browser automation tool via agent-browser CLI

Danieldd28:
## Summary

Integrate the agent-browser CLI as a lightweight browser automation tool for PicoClaw. This replaces the previous PR #308 (ActionBook approach) with a much leaner design that wraps an external CLI binary instead of embedding browser dependencies.

## Design

Instead of pulling in heavy Go browser libraries (chromedp, etc.), this delegates all browser complexity to the external agent-browser binary via exec.Command. PicoClaw stays lean:

- Zero new Go dependencies (all stdlib: os/exec, bytes, fmt, strings, time)
- ~200 bytes RAM overhead when enabled, zero when disabled
- Browser engine runs in a separate process (OS-level memory isolation)
- Binary size increase: ~5KB

### New files
- `pkg/tools/browser.go` - BrowserTool wrapping agent-browser CLI
- `pkg/tools/browser_test.go` - 11 unit tests

### Modified files
- `pkg/config/config.go` - Add BrowserConfig (enabled, session, headless, timeout, cdp_port)
- `pkg/agent/loop.go` - Register browser tool conditionally

## Configuration

```json
{
  "tools": {
    "browser": {
      "enabled": true,
      "headless": false,
      "timeout": 60,
      "cdp_port": 9222
    }
  }
}
```

## How it works

The tool exposes a single `browser` tool with a `command` parameter. The LLM constructs agent-browser subcommands directly:

- `browser open https://example.com`
- `browser snapshot -i`
- `browser click @e2`
- `browser fill @e3 "text"`
- `browser close`

Global flags (--cdp, --headed, --session, --json) are added automatically based on config.

## Testing

- All 11 unit tests pass
- Full test suite passes with zero regressions
- Manual integration tested with headed browser on CDP port 9222

## Prerequisite

Requires agent-browser CLI installed: `npm install -g @anthropic/agent-browser`

### DIFF
--- a/pkg/agent/loop.go
+++ b/pkg/agent/loop.go
@@ -84,6 +84,16 @@ func createToolRegistry(workspace string, restrict bool, cfg *config.Config, msg
 	}
 	registry.Register(tools.NewWebFetchTool(50000))
 
+	// Browser automation tool (agent-browser CLI)
+	if cfg.Tools.Browser.Enabled {
+		registry.Register(tools.NewBrowserTool(tools.BrowserToolOptions{
+			Session:  cfg.Tools.Browser.Session,
+			Headless: cfg.Tools.Browser.Headless,
+			Timeout:  cfg.Tools.Browser.Timeout,
+			CDPPort:  cfg.Tools.Browser.CDPPort,
+		}))
+	}
+
 	// Hardware tools (I2C, SPI) - Linux only, returns error on other platforms
 	registry.Register(tools.NewI2CTool())
 	registry.Register(tools.NewSPITool())
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -211,8 +211,17 @@ type WebToolsConfig struct {
 	DuckDuckGo DuckDuckGoConfig `json:"duckduckgo"`
 }
 
+type BrowserConfig struct {
+	Enabled  bool   `json:"enabled" env:"PICOCLAW_TOOLS_BROWSER_ENABLED"`
+	Session  string `json:"session" env:"PICOCLAW_TOOLS_BROWSER_SESSION"`
+	Headless bool   `json:"headless" env:"PICOCLAW_TOOLS_BROWSER_HEADLESS"`
+	Timeout  int    `json:"timeout" env:"PICOCLAW_TOOLS_BROWSER_TIMEOUT"`
+	CDPPort  int    `json:"cdp_port" env:"PICOCLAW_TOOLS_BROWSER_CDP_PORT"`
+}
+
 type ToolsConfig struct {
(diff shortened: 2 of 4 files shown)


Leeaandrob:
@Zepan This PR addresses **roadmap issue #293** (Autonomous Browser Operations — priority: high). It uses `agent-browser` CLI as a subprocess — lightweight approach that avoids embedding a full browser engine.

Note: PR #187 also targets browser automation but uses `playwright-go` (+1055 lines, heavier dependency). For PicoClaw's ultra-lightweight philosophy, the CLI subprocess approach in this PR is arguably more aligned.

**Recommendation: Review both #318 and #187, pick one approach.** The CLI subprocess model (this PR) is more consistent with PicoClaw's existing pattern (see: Codex CLI provider, Claude CLI provider). Playwright-go would add significant binary size.

nikolasdehor (commented):
Well-designed browser automation tool. The exec.Command delegation to agent-browser keeps PicoClaw lean (zero Go dependencies for browser) while providing full browser control. The splitCommand parser, timeout handling, and output truncation are all solid.

Issues and observations:

1. Command injection via splitCommand: The splitCommand function splits user input into arguments but does not handle all shell escaping edge cases. For example, a command like 'eval "rm -rf /"' would be split and passed to agent-browser as-is. While agent-browser is the actual binary being invoked (not a shell), the user-provided command string is fully trusted. Since the LLM constructs these commands, and agent-browser is a separate binary, the real risk is low, but worth noting.

2. splitCommand does not handle escaped quotes: Input like 'fill @e3 "she said \"hello\""' will not parse correctly because there is no backslash-escape handling inside quoted strings. This could cause issues with URLs containing quotes or form fields with special characters.

3. No binary existence check: If agent-browser is not installed, every Execute call will fail with a cryptic exec error. Consider checking for the binary in NewBrowserTool (via exec.LookPath) and returning a helpful error message in the tool description or at registration time.

4. Session isolation: The session flag is configurable but defaults to empty string, meaning all concurrent chats share the same browser state. For multi-user gateway deployments, this could cause cross-session contamination. Consider using the chat ID as a session identifier by default.

5. The test for buildArgs uses Headless: false (default from BrowserToolOptions{}) which adds --headed. But the default in config is Headless: true. The tests should mirror the default config to catch regressions.

Good feature. The design decision to wrap an external CLI is the right one for keeping the binary small.

amirmamaghani:
I made a simplier version which only needs a skill file. tested and working smoothly. 

https://github.com/sipeed/picoclaw/pull/1861

afjcjsbx:
close in favor of: https://github.com/sipeed/picoclaw/pull/1861

PR #319, opened 2026-02-16
Feature: Add Pushover notification channel and tool

vernonstinebaker:
Add Pushover as an outbound notification channel with a pushover tool that the agent can use to send push notifications to your phone.

### DIFF
--- a/pkg/agent/loop.go
+++ b/pkg/agent/loop.go
@@ -19,7 +19,6 @@ import (
 	"unicode/utf8"
 
 	"github.com/sipeed/picoclaw/pkg/bus"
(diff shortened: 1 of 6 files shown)


Leeaandrob:
@Zepan Pushover notification channel and tool. Pushover is a niche notification service — not on the roadmap.

**Recommendation: Defer.** Focus should be on stabilizing existing channels (Telegram, Discord, Slack, OneBot) rather than adding new niche ones. The +245/-304 scope also suggests significant refactoring beyond just the new channel.

vernonstinebaker:
Superseded by #520 which is a clean, minimal implementation rebased on current main with only the Pushover-related changes.

PR #143, opened 2026-02-13
Feat: Discord message length check and auto split

huaaudio:
## Description

Discord has a strict message length limit (2000 characters per message). This PR automatically splits long messages into multiple chunks when sending via Discord, preventing API errors when messages exceed the character limit.

## Changes

- Modified Discord message handler to check message length before sending
- Added chunking logic to split long messages intelligently
- Maintains markdown formatting integrity across chunks
- Tested with:
  - Plain text
  - Markdown-formatted documents (headers, tables, code blocks, lists)

## Notes

- Edge cases to consider: Long code blocks, single long words (e.g., URLs)

PR #323, opened 2026-02-16
fix(discord): handle character limits and maintain typing status

Andyi955:
# Overview
This PR improves the reliability and user experience of the Discord channel by fixing 400 errors caused by message length limits and ensuring the bot's activity status is accurately reflected during processing.

# Key Changes

## Discord Message Handling
* Rune-Based Message Splitting: Replaced the previous byte-based splitting logic with a more robust rune-based system. This ensures that the bot stays strictly within Discord's 2000-character limit regardless of multi-byte characters (emojis, etc.) while preserving the integrity of Markdown code blocks.
* Chunk Delivery: Added a reliable multi-chunk delivery system that ensures long responses are broken into coherent segments rather than being cut off.

## User Experience Improvements
* Persistent Typing Heartbeat: Implemented a background routine to maintain the "Bot is typing..." indicator during long-running tasks. This prevents the status from timing out while the agent is iterating through tools or performing complex reasoning, providing better feedback to the user.
* Threshold Notification: Softened the memory threshold warning message to be more user-friendly.

## Tool Stability
* Improved Path Validation: Updated the safety guard logic in pkg/tools/shell.go to correctly handle scoped package names (e.g. @mastra/core) during command execution. This fixes a bug where legitimate npm commands were incorrectly flagged as path traversal attempts.

# Technical Details
* Modified Components:
    * pkg/channels/discord.go: Logic for message chunking and typing status synchronization.
    * pkg/tools/shell.go: Refined regex for safe path validation.
    * pkg/agent/loop.go: Updated notification messaging.

# Verification
* Successfully tested with responses over 4000 characters via Discord.
* Verified that the typing indicator remains active throughout the entire agent loop duration.
* Confirmed that commands using modern package management syntax are correctly allowed while system security remains intact.


Leeaandrob:
@Zepan Discord character limit handling and auto-split, plus typing indicator improvements. PR #143 (merged) already addressed message splitting — this PR improves on that implementation.

**Recommendation: Merge.** Solid UX improvement for Discord users. +194/-95 is reasonable scope.

Issue #311, opened 2026-02-16
[BUG] Shell process need kill


liangxiwei:
Quick Summary
Failed shell commands are not being terminated. When the agent executes a command that fails or times out, the process remains active in the shell/system background instead of being killed, leading to resource leaks.

Environment & Tools

Operating System: [armbian]

Channels: [telegram]

📸 Steps to Reproduce
Start the PicoClaw agent.,tell him use playright to do something like open google.com and search some keyword.

<img width="1782" height="744" alt="Image" src="https://github.com/user-attachments/assets/d17a4415-a162-4647-a747-95670e82ff17" />

❌ Actual Behavior
The failed or timed-out command continues to run in the background. The shell session does not send a SIGKILL or SIGTERM to the child process, resulting in multiple "zombie" or orphaned processes over time.

✅ Expected Behavior
When a command execution fails or reaches a timeout, the agent should ensure the associated process tree is completely terminated before proceeding.

💬 Additional Context
This might be related to how exec.Command is handled in Go. We may need to use Context with timeout ( CommandContext ) and ensure that the process group is killed to prevent orphaned child processes.

Aditya7880900936:
hi maintainers , 
i would like to work on this issue , please assign it to me if it is available , 
Regards,

mattn:
Addressed in https://github.com/sipeed/picoclaw/pull/358

nikolasdehor:
PR #378 by @lunareed720 addresses this with the canonical solution: process groups on Unix (`Setpgid: true` + `syscall.Kill(-pid, SIGKILL)`) and `taskkill /T /F /PID` on Windows. I reviewed it and it looks solid — approved with minor suggestions.

The key issue is that Go's `exec.CommandContext` only sends `os.Kill` to the direct process (the `sh -c` shell), not to its children. PR #378 fixes this by:
1. Setting `SysProcAttr.Setpgid = true` to create a new process group
2. On timeout, killing the entire process group with `syscall.Kill(-pid, SIGKILL)`
3. Including a regression test that spawns `sleep 60` and verifies it gets killed

PR #324, opened 2026-02-16
fix: ensure shell processes are properly terminated

zenchantlive:
Fixes Issue #311 - Shell process need kill

## Root Cause
The ExecTool ran shell commands via  but only killed the parent shell process on timeout/cancellation. Child processes spawned by the shell survived, causing resource leaks.

## Fix
- Set process group on non-Windows systems to kill entire process tree
- Use Start/Wait goroutine pattern for proper cleanup
- Kill process group on context timeout/cancellation

## Changes
- : 27 insertions, 1 deletion

## Testing
- All pkg/tools tests pass (80 tests)
- No functional changes to existing behavior

### DIFF
--- a/pkg/tools/shell.go
+++ b/pkg/tools/shell.go
@@ -10,6 +10,7 @@ import (
 	"regexp"
 	"runtime"
 	"strings"
+	"syscall"
 	"time"
 )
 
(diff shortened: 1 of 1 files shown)


Leeaandrob:
@Zepan This PR ensures shell processes are properly terminated — prevents zombie processes that could leak resources on constrained devices.

**Recommendation: Merge.** Small fix (+27/-1) that directly supports PicoClaw's ultra-lightweight runtime goal. Zombie processes on a $10 RISC-V board would be devastating.

Issue #310, opened 2026-02-16
[BUG] Telegram messages stuck on "Thinking..." due to allowlist mismatch in HandleMessage


AbdelkaderBah:
## Quick Summary
Telegram messages get stuck on "Thinking..." and never receive a response when the allowlist uses username format (@username).

## Environment & Tools
PicoClaw Version: latest main
Go Version: go 1.26
AI Model & Provider: OpenRouter
Operating System: Linux (Docker on Ubuntu)
Channels: Telegram

## 📸 Steps to Reproduce
Configure PICOCLAW_CHANNELS_TELEGRAM_ALLOW_FROM=@myusername
Send any message to the Telegram bot
Bot replies "Thinking..." but never sends a response

## ❌  Actual Behavior
The bot sends "Thinking..." then goes silent. The message is silently dropped and never reaches the agent loop.

Root cause: In pkg/channels/telegram.go:358, HandleMessage is called with only the numeric user ID:

`c.HandleMessage(fmt.Sprintf("%d", user.ID), ...)  // passes "123456789"`

But BaseChannel.HandleMessage (pkg/channels/base.go:86) performs a second allowlist check before publishing to the message bus. When the allowlist contains @myusername, the numeric-only ID "123456789" fails the match and the message is silently dropped.

The first allowlist check in handleMessage (line 205) passes because it uses the compound format "123456789|myusername", but the second check never sees the username.

##  ✅ Expected Behavior
The bot should process the message and reply after "Thinking...".

## 💬 Additional Context
Fix: Pass the compound senderID (which includes the username) to HandleMessage:

`c.HandleMessage(senderID, fmt.Sprintf("%d", chatID), content, mediaPaths, metadata)`
This ensures both allowlist checks receive the same compound `"123456789|myusername"` format that matches `@myusername`.

Aditya7880900936:
hi maintainers , 
i would like to work on this issue , please assign it to me if it is available , 
Regards,

kang0024:
I would like to add that showing Thinking actually blocks the response from sending another notification when completed. I would prefer getting one meaningful notification after my message.

blib:
pr #371

nikolasdehor:
This looks like a timing issue between the allowlist check and the typing indicator. The fix in PR #391 by @JexLau addresses part of this — it moves `startTyping` to after all early returns (including the allowlist check), so rejected messages no longer trigger a typing indicator.

[...]

PR #320, opened 2026-02-16
fix: improve message handling by using senderID for identification

cointem:
修复了pkg/channels/telegram.go文件夹下handleMessage方法校验senderID不一致的bug。
** 第一次校验senderID **
image

** 第二次调用HandleMessage也应该传senderID进行校验 **
旧代码：
image

新代码：
image

PR #325, opened 2026-02-16
fix: pass compound senderID to HandleMessage for Telegram allowlist

zenchantlive:
Fixes Issue #310 - Telegram messages stuck on Thinking

## Root Cause
In pkg/channels/telegram.go:358, HandleMessage was called with only the numeric user ID (e.g., "123456789"). However, the allowlist check in BaseChannel.HandleMessage expects the compound format ("123456789|myusername") to match @username entries.

When the allowlist contained @myusername, the numeric-only ID failed the match and the message was silently dropped after showing "Thinking...".

## Fix
Use the pre-computed compound senderID instead of just the numeric ID:
- Before: c.HandleMessage(fmt.Sprintf("%d", user.ID), ...)
- After: c.HandleMessage(senderID, ...)

## Changes
- pkg/channels/telegram.go: 1 line changed

## Testing
- All pkg/channels tests pass (6 tests)
- Build compiles successfully

### DIFF
--- a/pkg/channels/telegram.go
+++ b/pkg/channels/telegram.go
@@ -355,7 +355,7 @@ func (c *TelegramChannel) handleMessage(ctx context.Context, message *telego.Mes
 		"is_group":   fmt.Sprintf("%t", message.Chat.Type != "private"),
 	}
 
-	c.HandleMessage(fmt.Sprintf("%d", user.ID), fmt.Sprintf("%d", chatID), content, mediaPaths, metadata)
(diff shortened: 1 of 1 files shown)


nikolasdehor:
This fixes the same Telegram senderID routing bug as #320 (opened earlier) and #371. All three PRs address the same root cause — `HandleMessage` receiving the wrong identifier for allowlist matching. Suggesting the maintainer pick one (likely #320 as the earliest) to reduce review burden and avoid merge conflicts.

xiaket:
dup of #371 , closing. Thanks team!

PR #327, opened 2026-02-16
docs: add missing Chinese language link to Japanese README

humaid0x:


Orgmar:
@humaid0x Thanks for catching the missing Chinese language link in the Japanese README! Small doc fixes like this make a real difference for discoverability.

We're building the **PicoClaw Dev Group** on Discord for contributors to connect. If you'd like to join, send an email to `support@sipeed.com` with the subject `[Join PicoClaw Dev Group] + Your GitHub account` and we'll send the invite!

PR #334, opened 2026-02-16
feat(http_provider): enhance K2-Think model support 

shauryr:
Added reasoning_effort parameter and handling reasoning_content in responses

### DIFF
--- a/pkg/providers/http_provider.go
+++ b/pkg/providers/http_provider.go
@@ -198,6 +198,15 @@ func (p *HTTPProvider) parseResponse(body []byte) (*LLMResponse, error) {
 		content = choice.Message.ReasoningContent
 	}
 
+	// Strip special tokens that some models include in their output
+	content = strings.TrimSpace(content)
+	specialTokens := []string{"<|im_end|>", "<|endoftext|>", "</think_fast>", "<|end|>"}
+	for _, token := range specialTokens {
+		content = strings.TrimSuffix(content, token)
+		content = strings.TrimPrefix(content, token)
+		content = strings.TrimSpace(content)
+	}
+
 	return &LLMResponse{
 		Content:      content,
 		ToolCalls:    toolCalls,


lukemilby:
Tech document referencing the need for https://platform.moonshot.ai/docs/guide/use-kimi-k2-thinking-model#accessing-the-reasoning-content

@shauryr have you test this with other models that are used by the HTTP Provider?

shauryr:
Thanks for reviewing! Quick clarification; LLM360/K2-Think-V2  https://huggingface.co/LLM360/K2-Think-V2 what I tested and Moonshot Kimi K2 are different models. The changes are generic and support any reasoning model.

Here are the specific change I made which should be generic (all models):
- Add `reasoning_content` field parsing
- Fallback to `reasoning_content` if `content` is empty
- Strip special tokens

K2-Think specific:
- `reasoning_effort: "high"` parameter

Also the changes are backward compatible
- Standard models: `content` has data so it uses `content` 
- Reasoning models: `content` is empty so it  uses `reasoning_content` 

I have done the testing for - K2-Think-V2 via vLLM (basic queries + tool calling) as well as multi-step reasoning with tools. Also this should Work With DeepSeek R1, QwQ, and other models using `reasoning_content`

Happy to test with specific models if you have recommendations!

nikolasdehor (changes requested):
Good addition for reasoning model support. The `reasoning_content` fallback benefits K2-Think, DeepSeek R1, QwQ, and other reasoning models.

However:

1. **`reasoning_effort` is hardcoded to `"high"`** with no user control. Consider a config field defaulting to `"high"` for K2-Think.

2. **Special token stripping is ungated** — it runs for ALL models using the HTTP provider, not just reasoning ones. A model discussing tokenization (e.g., "The tokenizer uses `<|endoftext|>` as...") could have legitimate content stripped. Consider gating behind a model-name check.

3. **No tests.** A unit test for `reasoning_content` fallback + special token stripping would be valuable for this critical code path.

nikolasdehor (commented):
The K2-Think model support and special token stripping are useful additions. A few concerns:

1. Hardcoded reasoning_effort: The reasoning_effort is hardcoded to 'high' for all K2-Think models. This should be configurable (e.g., via LLM options or a config field). Some users may want 'low' or 'medium' for faster, cheaper responses.

2. Reasoning content fallback: Using reasoning_content as a fallback when content is empty makes sense for K2-Think, but this applies to ALL models going through HTTPProvider. If a non-K2 model returns empty content with a non-empty reasoning_content for some other reason, the user would get internal chain-of-thought leaked into their response. Consider gating this fallback behind a model name check (strings.Contains(model, 'k2-think')).

3. Special token stripping scope: The special token stripping (im_end, endoftext, think_fast, end) is applied to all model responses, not just K2-Think. This is probably fine for most tokens, but if a legitimate response contains the literal string '<|im_end|>' (e.g., in a discussion about tokenization), it would be silently stripped. The risk is low but worth noting.

4. No tests: Adding a test for parseResponse with a K2-Think style response (reasoning_content populated, content empty, special tokens present) would prevent regressions.

Overall, a practical fix for K2-Think compatibility. The hardcoded reasoning_effort (point 1) is the main concern.

PR #182, opened 2026-02-15
feat: add slash command support (e.g., /show model, /help) 

Tuzfucius:
### Description
Based on the suggestion in issue #149, this PR implements slash command support to enhance user interaction. 

### Key Changes
- Added support for `/help` to display available commands.
- Added support for `/show model` to display the current model configuration.
- Improved command parsing logic to handle slash prefix.

### Related Issue
 Fixes #149

### How to test
I have tested the commands locally:
- Entering `/help` correctly shows the help message.
- Entering `/show model` correctly returns the current model info.

Issue #162, opened 2026-02-14
建议增加会话功能,提高上下文质量


btkawai:
建议增加控制上下文的指令如

/session new 新话题
/session ls 话题列表
/session switch 切换话题
/session remove 删除话题

让不同 channel ,不同 uid 用户可以随时切换话题的上下文.

(使用**软连接**的话会不会比较简单实现?)

nikolasdehor:
This is a solid proposal. The session management in picoclaw currently uses a `channel:chatID` key pattern (see `pkg/session/manager.go`), which means each channel+user combination gets one session. Multi-session support per user would need a session index or namespace layer on top of that.

A few implementation notes:
- The symlink approach mentioned in the original issue could work for sharing context across sessions, but the current `session.Key` uses `:` which is an illegal filename char on Windows (tracked in #116 as BUG-003). That should be fixed first.
- The `/session` slash-command pattern aligns well with #149 (in-chat configuration commands). Both could share a common command dispatch layer in the channel message handler.
- Cron jobs (#63) also create sessions, so the session lifecycle would need to account for both user-initiated and system-initiated sessions.

PR #335, opened 2026-02-17
feat: implement session context switching via /session command

Tuzfucius:
## Description
This PR introduces the ability for users to maintain and seamlessly switch between multiple isolated conversation contexts. 

It builds upon the slash command infrastructure introduced in my previous PR (#182) and addresses the feature request discussed in #162.

- Add SessionCommand with subcommands: `new`, `switch`, `ls`, `remove`
- Implement persistent user-session mapping in `state.Manager`
- Update SessionManager to support listing and deleting session files
- Update AgentLoop to dynamically resolve session keys based on active user context
- Enable isolated conversation contexts per user across all channels

## Related Issues
- **Resolves:** #162 
- **Depends on:** #182 (Built on top of the newly added slash command support)


nikolasdehor (commented):
Good architecture -- extracting commands into a registry and adding session context switching is useful. A few things to address:

**1. Data race on `al.model` (SetModel/GetModel)**

`AgentLoop.model` is a plain string field. `SetModel` writes it and `GetModel` reads it, but `AgentLoop` processes messages concurrently (the bus handler runs in a goroutine). There is no synchronization protecting `model`. A user running `/switch model to X` from one channel while another message is being processed on a different channel will race. Use `sync.RWMutex` or `atomic.Value` for the model field.

**2. Session key construction inconsistency**

In `Run()`:
```go
baseKey := fmt.Sprintf("%s:%s", msg.Channel, msg.ChatID)
```
But `msg.SessionKey` is already set by the bus layer (typically `channel:chatID`). If the bus ever changes its format, the hand-constructed `baseKey` will diverge from the actual session key prefix, causing `ListSessions(prefix)` to return nothing. Consider deriving `baseKey` from `msg.SessionKey` or the bus's canonical format.

**3. Duplicate session resolution logic**

The session key resolution (check `activeSession`, append to `baseKey`) is duplicated verbatim in `Run()` and `ProcessDirectWithChannel()`. This should be extracted into a helper like `resolveSessionKey(channel, chatID, sessionKey)`.

**4. `GetSessionManager()` and `GetStateManager()` return `interface{}`**

These methods return `interface{}` to avoid circular imports, but then `SessionCommand` defines ad-hoc interfaces (`SessionManager`, `StateManager`) and does type assertions at runtime. If the underlying types change their method signatures, this will silently break. Consider defining these interfaces in the `command` package as exported types and having the actual managers satisfy them, or use a shared interface package.

**5. `ListCommands()` iteration order is non-deterministic**

`HelpCommand.Execute()` iterates over `r.commands` (a map), so `/help` output order will vary between runs. Sort the command names before displaying.

**6. No tests for session command**

The session command has significant logic (list, switch, remove) but no tests. Given it involves persistent state and file deletion (`DeleteSession` removes from disk), test coverage is important.

**7. Telegram-specific commands removed without migration path**

The PR deletes `telegram_commands.go`, which had Telegram-specific bot command handlers (`/help`, `/show`, `/list`, `/start`). These were registered with the Telegram bot handler framework. The new generic `/` command handling goes through the agent loop, which means Telegram users will no longer see command autocompletion in the Telegram UI. This is a UX regression worth noting in the PR description.

Solid direction overall. The race condition on `model` and the duplicate session resolution logic are the most important items.

PR #339, opened 2026-02-17
Added Email Tool, Calender Integration and System Stats Overview Tool

udbhav-44:
## 📝 Description
This PR integrates Google Calendar support, enhances the Email channel with better polling and content fetching, and adds new developer tools (GitHub, System Stats) to PicoClaw.

- **Google Calendar**: Added a new [CalendarTool](cci:2://file:///Users/udbhavagarwal/Documents/Code/picoclaw/pkg/tools/calendar.go:16:0-19:1) that allows the agent to list and add events. Supports authentication via Service Account credentials.
- **Email Improvements**:
    - Fixed an issue where restarting the agent would re-process old unread emails. Now filters emails older than 1 hour.
    - Updated `check_email` tool to fetch visible content, making "check mail" commands actionable.
    - Added `calendar_id` to configuration to allow specifying a target calendar (e.g., user's primary email) different from the Service Account's default.
- **Developer Tools**:
    - Added `GitHubTool` for listing repos and issues.
    - Added `SystemTool` for monitoring system stats (CPU, Memory).

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🔗 Linked Issue
N/A

* **Reference:** Google Calendar API v3
* **Reasoning:**
    - The Service Account approach was chosen for headless agent operation.
    - `go-imap` was used for email polling.
    - Time-based filtering in email polling was necessary to prevent "spamming" the user with backlog processing on every restart.

## 🧪 Test Environment & Hardware
- **Hardware:** Raspberry Pi 5, Mac (Apple Silicon)
- **OS:** Debian 12 (RPi), macOS (Sonoma)
- **Model/Provider:** OpenAI GPT-4o
- **Channels:** Telegram, Email

## 📸 Proof of Work (Optional for Docs)
<details>

**Calendar Event Creation:**

</details>

## ☑️ Checklist


nikolasdehor (commented):
This PR adds a lot of surface area -- email channel, email tools, calendar, GitHub tool, and system stats -- all in one PR. I would recommend splitting this into separate PRs per feature for easier review. That said, here are the issues I found:

**1. JSON syntax error in config.example.json**

The diff shows:
```json
    "github": {
      "enabled": false,
      "token": "YOUR_GITHUB_TOKEN"
    "cron": {
```

There is a missing closing brace `}` and comma before the `"cron"` key. This will cause a JSON parse error and break anyone who copies the example config.

**2. Security: GitHub token and IMAP/SMTP passwords in config.json**

The email config stores IMAP and SMTP passwords in plain text in `config.json`. The GitHub tool stores the API token the same way. There is no warning in the documentation about file permissions, and `config.json` might be world-readable by default. Consider:
- Documenting that the config file should be `chmod 600`
- Supporting env var overrides for secrets (the env tags are present for some fields but not all)
- At minimum, adding a comment in the example config about security

**3. Email: no TLS verification option**

The IMAP client uses `client.DialTLS(addr, nil)` with default TLS config. The SMTP handler for port 465 creates a TLS config with only `ServerName` set. There is no option to skip TLS verification for self-hosted mail servers, which is a common need for picoclaw's target audience (home labs, embedded devices).

**4. Calendar: `filepathJoin` helper is broken**

```go
func filepathJoin(elem ...string) string {
    for i, e := range elem {
        if e != "" {
            return strings.Join(elem[i:], string(os.PathSeparator))
        }
    }
    return ""
}
```

This function skips empty leading elements and joins the rest with `os.PathSeparator`, but it does NOT normalize paths. `filepathJoin("/home/user", "creds.json")` produces `/home/user/creds.json` on Unix (correct), but `strings.Join` does not deduplicate separators. Just use `filepath.Join` from the standard library -- it is already imported transitively.

**5. Calendar: hardcoded UTC timezone**

```go
Start: &calendar.EventDateTime{
    DateTime: startTimeStr,
    TimeZone: "UTC",
},
```

The timezone is hardcoded to UTC. For a personal calendar tool, events should use the user's local timezone (or at least make it configurable). An event created at "2025-03-01T10:00:00" in UTC will show at the wrong time for most users.

**6. System tool: `gopsutil` dependency is heavy for a lightweight project**

`gopsutil/v3` pulls in a significant dependency tree including `golang.org/x/sys` and platform-specific code. For a project that targets embedded systems (NanoKVM, OpenWrt), this adds binary size. The same info can be obtained from `/proc/stat`, `/proc/meminfo`, and `df` on Linux, or via `exec.Command("top", "-l", "1")` on macOS, without any dependencies.

**7. Email channel: no rate limiting on polling**

The email channel polls IMAP every `poll_interval` seconds. There is no rate limiting or backoff on connection failures. If the IMAP server is down, picoclaw will hammer it with connection attempts every 30 seconds indefinitely.

**8. No tests for any of the new tools**

1700+ lines of new code with zero tests. The calendar, email, GitHub, and system tools all have no test coverage. Even basic unit tests for input validation would help.

**9. Dependency footprint is large**

This PR adds `go-imap`, `go-message`, `go-github/v60`, `gopsutil/v3`, and `google.golang.org/api` (Google Calendar). That is a lot of new dependencies for a project that emphasizes being lightweight. Consider making these tools optional (behind build tags) or splitting them into separate installable plugins.

I would strongly recommend splitting this into at least 3 PRs: (1) Email channel + tools, (2) Calendar, (3) GitHub + System. The JSON syntax error needs fixing regardless.

udbhav-44:
Thanks for the thorough review!, this has been my first contribution. As suggested, I have split this massive PR up into three atomic pieces. I also addressed all of your feedback from this thread (JSON syntax, security warnings, TLS skipping, fixing the calendar timezone/filepath, writing tests, and removing the heavy gopsutil dependency) inside the new PRs.

Here are the split PRs for review:
 - Email Enhancements: #716
- Calendar Integration: #717 
- GitHub and System Tools: #718 

I'll go ahead and close this original PR so we can continue the review process on those specific threads."

Issue #312, opened 2026-02-16
[Feature] Add custom timout for slow LLM self-hosted


leadvic:
## 🎯 The Goal / Use Case

Allow users to configure HTTP client timeout for LLM requests when using slow models or hardware (when self-hosting with Ollama for example). Currently, PicoClaw appears to have a hardcoded ~2 minute timeout that terminates requests before slower LLM backends (self-hosted Ollama with large models) can complete their inference.

**Use case**: Running PicoClaw with self-hosted Ollama on AMD integrated GPUs (Radeon Renoir) using Vulkan backend, or any scenario where model inference legitimately takes 3-10 minutes per response.

## 💡 Proposed Solution

Add a configurable environment variable (e.g., `PICOCLAW_HTTP_TIMEOUT` or `PICOCLAW_REQUEST_TIMEOUT`) that allows users to override the default HTTP client timeout. This would enable support for:

- Self-hosted LLMs on consumer hardware (AMD/Intel GPUs via Vulkan)
- Large models (70B+) running on limited resources
- Complex multi-turn reasoning tasks that require extended inference time

**Suggested default**: Keep 2 minutes for API providers, but allow override for self-hosted scenarios.

## 🛠 Potential Implementation (Optional)

In the HTTP client initialization (likely in Go code):

```go
func newHTTPClient() *http.Client {
    timeout := 2 * time.Minute // default
    
    if envTimeout := os.Getenv("PICOCLAW_HTTP_TIMEOUT"); envTimeout != "" {
        if d, err := time.ParseDuration(envTimeout); err == nil {
            timeout = d
        }
    }
    
    return &http.Client{
        Timeout: timeout,
    }
}
```
Then document in config.json or environment variables section. Or something like that I guess.

💬 Additional Context
Evidence from logs:

```text
# PicoClaw logs - timeout at client side
2026/02/16 13:30:35 [ERROR] agent: LLM call failed {error=context deadline exceeded (Client.Timeout exceeded while awaiting headers)}

# Ollama logs - request processing for ~2 minutes before 500 error
[GIN] 2026/02/16 - 13:28:29 | 500 |         1m55s |      172.27.0.1 | POST     "/v1/chat/completions"
[GIN] 2026/02/16 - 13:30:35 | 500 |         1m59s |      172.27.0.1 | POST     "/v1/chat/completions"
time=2026-02-16T13:30:35.056Z level=INFO source=runner.go:916 msg="aborting completion request due to client closing the connection"
[...]

Issue #197, opened 2026-02-15
Bot stuck on running express server.


Akshay-Rana-Gujjar:
I asked bot to create a express server and run it. After that it stuck on running command and it is not replying back to my message on telegram. It seems like when I ask bot to start any service it waits for the service to finish to proceed further.

Same thing I noticed when I asked it to install and run ollama on my system.  Ollama started but bot stuck and waiting for ollama to finish.

To solve this what I am think to add something to run or execute commands on separate process so that bot never blocks.

My setup:
Device: samsung m21 android
Running ubuntu in termux.

Installed picoclaw from source.

mrbeandev:
I've implemented a fix for this! 🚀

The `exec` tool now supports an optional `background` parameter. When set to `true`, PicoClaw will spawn the command in a separate process and return immediately, allowing the agent to continue chatting while your server (or Ollama) runs in the background.

It will report the final output back to the chat once the background process eventually finishes. 🦞🦾

mrbeandev:
Also, worth noting for your setup: You actually don't need to run Ubuntu inside Termux for PicoClaw anymore! You can install it directly in Termux, which is much faster and uses fewer resources. 

Our new [PicoClaw Termux Manager](https://github.com/sipeed/picoclaw/blob/main/assets/scripts/picoclaw-manager.sh) handles everything for you natively in Termux. 🚀📱🦞

mrbeandev:
Quick correction: The manager script and diagnostics mentioned above are currently in **PR #298** and will be available in main once merged! 🚀🦞

Issue #79, opened 2026-02-12
Gemini 3 Flash Preview: Function call fails due to missing thought_signature


gyuyeol:
When using the Gemini 3 Flash Preview model with picoclaw, function calls are consistently failing. The error messages indicate that a thought_signature is missing in the functionCall parts, which is stated as required for tools to work correctly and leads to degraded model performance.

This prevents the agent from properly utilizing available tools via function calling.

Error Message:

```
2026/02/12 15:56:24 [2026-02-12T15:56:24Z] [ERROR] agent: LLM call failed {iteration=2, error=API error: [{
  "error": {
    "code": 400,
    "message": "Function call is missing a thought_signature in functionCall parts. This is required for tools to work correctly, and missing thought_signature may lead to degraded model performance. Additional data, function call default_api:list_dir , position 10. Please refer to https://ai.google.dev/gemini-api/docs/thought-signatures for more details.",
    "status": "INVALID_ARGUMENT"
  "error": {
    "code": 400,
    "message": "Function call is missing a thought_signature in functionCall parts. This is required for tools to work correctly, and missing thought_signature may lead to degraded model performance. Additional data, function call default_api:list_dir , position 10. Please refer to https://ai.google.dev/gemini-api/docs/thought-signatures for more details.",
    "status": "INVALID_ARGUMENT"
  }
}
]}
Error: LLM call failed: API error: [{
  "error": {
    "code": 400,
    "message": "Function call is missing a thought_signature in functionCall parts. This is required for tools to work correctly, and missing thought_signature may lead to degraded model performance. Additional data, function call default_api:list_dir , position 10. Please refer to https://ai.google.dev/gemini-api/docs/thought-signatures for more details.",
    "status": "INVALID_ARGUMENT"
  }
}
]

```

Expected behavior:
Function calls should execute correctly without requiring manual intervention for thought_signature when using the Gemini 3 Flash Preview model, similar to how other models handle function calling.

Impact:
This issue severely limits the agent's ability to use tools and perform tasks when configured with the Gemini 3 Flash Preview model.

SatyamDevv:
### 🚀 Solution found

I managed to resolve the `thought_signature` and tool execution errors with Gemini.

The root cause is that PicoClaw expects **OpenAI API compatibility**, but the default Gemini endpoint uses Google's native format. Switching to the Google standard OpenAI-compatible endpoint fixes the issue.

### ✅ Configuration
[...]

Issue #161, opened 2026-02-14
Incompatibility with Gemini 3 Pro/Flash due to Mandatory Thought Signatures


yenteng25:
## Issue Summary:
When using the latest Gemini 3 models (gemini-3-pro-preview, gemini-3-flash-preview), multi-turn conversations with tool use fail with a 400 INVALID_ARGUMENT error.

## Technical Detail:
Google recently introduced Mandatory Thought Signatures for the Gemini 3 series. According to the [official documentation](https://www.google.com/url?sa=E&q=https%3A%2F%2Fai.google.dev%2Fgemini-api%2Fdocs%2Fthought-signatures), Gemini 3 requires that any function call or thought generated in previous turns must be returned to the model with its associated encrypted signature in the next turn.

Currently, PicoClaw (whether via the OpenAI compatibility layer or direct SDK) seems to drop these signatures during session history management, causing Google to reject subsequent requests.

## Error Message observed:
`"message": "Function call is missing a thought signature in functionCall parts. This is required for tools to work correctly..."`

## Steps to Reproduce:

1. Configure model to gemini-3-pro-preview.
2. Turn 1: Send a message that triggers a tool (e.g., exec or web_search). Turn 1 completes successfully.
3. Turn 2: Send a follow-up question.
4. Observation: PicoClaw returns a 400 error because the history passed back to Google lacks the required thought signatures from Turn 1.

## Request:
Please update the Go backend to properly store and echo back the thought_signature fields in the conversation history to support the next generation of Gemini models.

biisal:
Which base url are you using?

yentengchang:
> Which base url are you using?

I've tried leave it blank and https://generativelanguage.googleapis.com/v1beta/openai/, both doesn't work for Gemini-3 series.
Now I've tested other Gemini models w/o thought signatures required with a blank base url value and it worked.

taesiri:
I am facing the same issue with `gemini-3-flash-preview` and getting ""Function call is missing a thoughtsignature in functionCall parts".

kang0024:
+1

taesiri:
I asked Gemini 3.0 Flash to take a look and fix it, and 5 minutes later, a new compiled binary that works.

Here is the diff:


```diff
diff --git i/pkg/agent/loop.go w/pkg/agent/loop.go
index f3dd940..18a2595 100644
--- i/pkg/agent/loop.go
+++ w/pkg/agent/loop.go
[...]

PR #343, opened 2026-02-17
[...]

mrbeandev:
## Summary
- add full Google Antigravity (Cloud Code Assist) provider support with OAuth login flow, model handling, docs, and CLI wiring
- fix multiple Antigravity tool-call compatibility issues discovered during real Telegram/Coolify testing (missing function names, invalid tool-call history ordering, thought signature preservation)
- include deployment and runtime improvements already on this branch (Coolify support, Telegram command improvements, health endpoints, provider reliability/test updates)

## Validation
- go test ./pkg/providers ./pkg/agent ./pkg/tools
- go build -o build/picoclaw ./cmd/picoclaw
- manual verification: `picoclaw agent` with Antigravity + tool calls (web_fetch flow)

## Security Check
- scanned this branch for obvious secrets before opening the PR (`access_token`, `refresh_token`, `ya29.*`, private keys) and found no committed credential artifacts.

### DIFF
--- a/pkg/providers/antigravity_provider.go
+++ b/pkg/providers/antigravity_provider.go
@@ -159,9 +159,11 @@ type antigravityContent struct {
 }
 
 type antigravityPart struct {
-	Text             string                       `json:"text,omitempty"`
-	FunctionCall     *antigravityFunctionCall     `json:"functionCall,omitempty"`
-	FunctionResponse *antigravityFunctionResponse `json:"functionResponse,omitempty"`
+	Text                  string                       `json:"text,omitempty"`
+	ThoughtSignature      string                       `json:"thoughtSignature,omitempty"`
+	ThoughtSignatureSnake string                       `json:"thought_signature,omitempty"`
+	FunctionCall          *antigravityFunctionCall     `json:"functionCall,omitempty"`
+	FunctionResponse      *antigravityFunctionResponse `json:"functionResponse,omitempty"`
 }
 
 type antigravityFunctionCall struct {
@@ -233,7 +235,7 @@ func (p *AntigravityProvider) buildRequest(messages []Message, tools []ToolDefin
 				content.Parts = append(content.Parts, antigravityPart{Text: msg.Content})
 			}
 			for _, tc := range msg.ToolCalls {
-				toolName, toolArgs := normalizeStoredToolCall(tc)
+				toolName, toolArgs, thoughtSignature := normalizeStoredToolCall(tc)
 				if toolName == "" {
 					logger.WarnCF("provider.antigravity", "Skipping tool call with empty name in history", map[string]interface{}{
 						"tool_call_id": tc.ID,
@@ -244,6 +246,8 @@ func (p *AntigravityProvider) buildRequest(messages []Message, tools []ToolDefin
 					toolCallNames[tc.ID] = toolName
 				}
 				content.Parts = append(content.Parts, antigravityPart{
+					ThoughtSignature:      thoughtSignature,
+					ThoughtSignatureSnake: thoughtSignature,
 					FunctionCall: &antigravityFunctionCall{
 						Name: toolName,
 						Args: toolArgs,
@@ -307,12 +311,16 @@ func (p *AntigravityProvider) buildRequest(messages []Message, tools []ToolDefin
 	return req
 }
 
-func normalizeStoredToolCall(tc ToolCall) (string, map[string]interface{}) {
+func normalizeStoredToolCall(tc ToolCall) (string, map[string]interface{}, string) {
 	name := tc.Name
 	args := tc.Arguments
+	thoughtSignature := ""
 
 	if name == "" && tc.Function != nil {
 		name = tc.Function.Name
+		thoughtSignature = tc.Function.ThoughtSignature
+	} else if tc.Function != nil {
+		thoughtSignature = tc.Function.ThoughtSignature
 	}
 
 	if args == nil {
@@ -326,7 +334,7 @@ func normalizeStoredToolCall(tc ToolCall) (string, map[string]interface{}) {
 		}
 	}
 
-	return name, args
+	return name, args, thoughtSignature
 }
 
 func resolveToolResponseName(toolCallID string, toolCallNames map[string]string) string {
@@ -363,8 +371,10 @@ type antigravityJSONResponse struct {
 	Candidates []struct {
 		Content struct {
 			Parts []struct {
-				Text         string                   `json:"text,omitempty"`
-				FunctionCall *antigravityFunctionCall `json:"functionCall,omitempty"`
+				Text                  string                   `json:"text,omitempty"`
+				ThoughtSignature      string                   `json:"thoughtSignature,omitempty"`
+				ThoughtSignatureSnake string                   `json:"thought_signature,omitempty"`
+				FunctionCall          *antigravityFunctionCall `json:"functionCall,omitempty"`
 			} `json:"parts"`
 			Role string `json:"role"`
 		} `json:"content"`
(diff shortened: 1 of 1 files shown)


pkg/tools/shell.go
```diff
@@ -30,9 +31,7 @@ func NewExecTool(workingDir string, restrict bool) *ExecTool {
 		regexp.MustCompile(`\bdd\s+if=`),
 		regexp.MustCompile(`>\s*/dev/sd[a-z]\b`), // Block writes to disk devices (but allow /dev/null)
 		regexp.MustCompile(`\b(shutdown|reboot|poweroff)\b`),
-		regexp.MustCompile(`:\(\)\s*\{.*\};\s*:`),
```
lukemilby, line 33: Why was this removed?

docs/ANTIGRAVITY_AUTH.md
lukemilby, line 1: Please just provide a link to Antigravity Auth doc vs having AI rewrite them.

docker-compose-coolify.yml
lukemilby, line 1: picoclaw already has docker files please remove

pkg/tools/toolloop.go
```diff
@@ -152,3 +159,42 @@ func RunToolLoop(ctx context.Context, config ToolLoopConfig, messages []provider
 		Iterations: iteration,
 	}, nil
 }
+
+func normalizeProviderToolCall(tc providers.ToolCall) providers.ToolCall {
```
lukemilby, line 163: Why are we normalizing the tool call? This should be done when calling the tool. Is your an LLM please do not answer this question its for the PR creator

pkg/providers/http_provider.go
```diff
@@ -132,8 +141,9 @@ func (p *HTTPProvider) parseResponse(body []byte) (*LLMResponse, error) {
 					ID       string `json:"id"`
 					Type     string `json:"type"`
 					Function *struct {
-						Name      string `json:"name"`
-						Arguments string `json:"arguments"`
+						Name             string `json:"name"`
+						Arguments        string `json:"arguments"`
+						ThoughtSignature string `json:"thought_signature"`
```
lukemilby, line 137: Where at is `thought signature` in a function tool call? Please provide link to confirm Thought Signature needs to be in the `Function` struct.

cmd/picoclaw/main.go
```diff
@@ -976,12 +1079,70 @@ func authStatusCmd() {
...
+		if cred.ProjectID != "" {
+			fmt.Printf("    Project: %s\n", cred.ProjectID)
+		}
 		if !cred.ExpiresAt.IsZero() {
 			fmt.Printf("    Expires: %s\n", cred.ExpiresAt.Format("2006-01-02 15:04"))
 		}
 	}
 }
 
+func authModelsCmd() {
```
lukemilby, line 1095: Generic Name. This looks to only handle Antigravity auth only. Move auth to provider.

pkg/providers/http_provider.go
```diff
@@ -159,18 +169,11 @@ func (p *HTTPProvider) parseResponse(body []byte) (*LLMResponse, error) {
...
 
-		// Handle OpenAI format with nested function object
-		if tc.Type == "function" && tc.Function != nil {
-			name = tc.Function.Name
-			if tc.Function.Arguments != "" {
-				if err := json.Unmarshal([]byte(tc.Function.Arguments), &arguments); err != nil {
-					arguments["raw"] = tc.Function.Arguments
-				}
-			}
-		} else if tc.Function != nil {
```
lukemilby, line 171: Do we not need to handle nested function object anymore?

mrbeandev:
Thanks for the thorough review Luke. You're right — this PR accumulated too many concerns. I'll split it into separate PRs:

1. **Coolify deployment** (docker-compose, Dockerfile, entrypoint)
2. **Telegram command improvements** (slash commands, model switching)
3. **Antigravity provider** (oauth, provider, docs)
4. **Gemini thought_signature fix** (this one addresses #79 and #161 independently)

Addressing each inline comment below.

pkg/tools/shell.go
mrbeandev: Accidental removal — it got lost during a rebase when I added the `background` execution parameter. The fork bomb pattern needs to stay. Will restore it in the next push.

docs/ANTIGRAVITY_AUTH.md
mrbeandev: Fair point. Will replace the doc with a link to the official Cloud Code Assist setup guide: https://cloud.google.com/code/docs/vscode/get-started and keep just the picoclaw-specific config bits (the OAuth scopes and CLI flags).

docker-compose-coolify.yml
mrbeandev: Agreed. The Coolify-specific docker files were for my own deployment and should not be in upstream. Will remove docker-compose-coolify.yml, Dockerfile.coolify, and entrypoint-coolify.sh from this PR. If the Coolify setup is useful to anyone, I can document it separately in a wiki or discussion post.

pkg/tools/toolloop.go
mrbeandev: The Antigravity endpoint (cloudcode-pa.googleapis.com) wraps the Gemini response in an extra envelope. When I parse that envelope, the function name sometimes lands only in ToolCall.Name and not in ToolCall.Function.Name, or vice versa. The normalize function copies the name and args into both locations so the downstream tool dispatch in toolloop.go always finds them.

You are right that this should live closer to where the tool is actually called rather than being a generic post-processing step. I will move this logic into the Antigravity provider's response parser where the envelope is unwrapped, so it does not touch the common path.

pkg/providers/http_provider.go
mrbeandev: Gemini 3 (and 2.5 with thinking enabled) returns a `thought_signature` alongside each `functionCall` part. If you strip it out when rebuilding session history, the next API call fails with a 400: "Function call is missing a thought_signature."

This is documented at:
- https://ai.google.dev/gemini-api/docs/thought-signatures — see "Signatures in function calling parts" section
- https://ai.google.dev/gemini-api/docs/function-calling\#thought-signatures — see "Managing conversation history manually"

For the OpenAI-compatible endpoint (which picoclaw uses), the signature appears in `extra_content.google.thought_signature` on the tool call object. For the native Gemini endpoint, it is `thoughtSignature` on the `Part` alongside `functionCall`.

Since picoclaw uses the OpenAI-compatible format, the thought_signature field on the Function struct is where I store it after parsing so it can be re-serialized on the next turn. Without it, Gemini 3 Flash and Pro reject the request. This fixes #79 and #161.

cmd/picoclaw/main.go
mrbeandev: You are right — `authModelsCmd` is misleading since it only handles the Antigravity OAuth flow. I will rename it to something specific like `antigravityAuthCmd` and move the auth logic into the Antigravity provider package so it follows the same pattern as the other providers. The CLI entry point will just call the provider method.

pkg/providers/http_provider.go
mrbeandev: Yes we still need it. In the new code, the nested function object is still handled — the `if tc.Function != nil` check at line 181 covers both the `type == "function"` case and the legacy case without a type field. The difference is I collapsed the two branches (which had identical logic) into one since both did the same thing: extract `Name` and `Arguments` from `tc.Function`.

The old code:
```go
if tc.Type == "function" && tc.Function != nil {
    // extract name + args
} else if tc.Function != nil {
    // exact same extraction
}
```

The new code:
```go
if tc.Function != nil {
    // extract name + args (covers both cases)
}
```

The only actual change is adding `ThoughtSignature` extraction and populating both `ToolCall.Function` and `ToolCall.Name`/`ToolCall.Arguments` for consistent access downstream. The nested function struct is still used.

mrbeandev:
Force-pushed a clean branch. Rebased fresh from `upstream/main` and cherry-picked only the 10 Antigravity-related commits.

**Removed** (will be separate PRs if needed):
- Coolify docker files (docker-compose-coolify.yml, Dockerfile.coolify, entrypoint-coolify.sh)
- Telegram UI/UX enhancements and slash command persistence
- Background shell execution (fixes #197)
- PICOCLAW_HTTP_TIMEOUT env var (fixes #312)
- Deploy env override fixes

**What remains** (13 files, ~2400 lines):
- Antigravity provider (`pkg/providers/antigravity_provider.go` + tests)
- OAuth flow with headless support (`pkg/auth/oauth.go`)
- CLI wiring (`cmd/picoclaw/main.go`)
- Thought signature preservation for Gemini 3 (`pkg/providers/http_provider.go`, `pkg/agent/loop.go`)
- Tool call normalization for Antigravity envelope (`pkg/tools/toolloop.go`)
- Docs (`docs/ANTIGRAVITY_AUTH.md`, `docs/ANTIGRAVITY_USAGE.md`)

Build and tests pass: `go build` + `go test ./pkg/providers ./pkg/agent ./pkg/tools`

PR #245, opened 2026-02-15
(feat): Add support for multiple transcriber providers

sysradium:
Instead of using free Groq version I prefer having an option to select various providers.
So in this PR I added possibility to define and select various transcribers. I started with the OpenRouter one, since it unlocks virtually any model you want.

This adds a new configuration for the voice:
```json
{
  "voice": {
    "provider": "openrouter", // PICOCLAW_VOICE_PROVIDER
    "model": "google/gemini-2.5-flash-lite"  // PICOCLAW_VOICE_MODEL
  },
```

PR #353, opened 2026-02-17
refactor(voice): introduce Transcriber interface for pluggable STT

emadomedher:
## Summary

Currently Discord, Slack, and Telegram all hardcode `*voice.GroqTranscriber` as their transcription dependency. This makes it impossible to swap in a different STT backend without modifying every channel file.

This PR adds a `Transcriber` interface to `pkg/voice/transcriber.go` and updates the three channels to depend on it instead of the concrete type.

## Changes

- `pkg/voice/transcriber.go`: add `Transcriber` interface (`Transcribe` + `IsAvailable`)
- `pkg/channels/discord.go`: `*voice.GroqTranscriber` → `voice.Transcriber`
- `pkg/channels/slack.go`: same
- `pkg/channels/telegram.go`: same

`GroqTranscriber` already satisfies the interface — **zero behaviour change**.

## Why

This unblocks adding alternative STT providers (e.g. local Whisper — see follow-up PR) without touching channel code. It also aligns with how the `Synthesizer` interface works for TTS.

## Tests

All existing tests pass (`go test ./...`).

### DIFF
--- a/pkg/channels/discord.go
+++ b/pkg/channels/discord.go
@@ -24,7 +24,7 @@ type DiscordChannel struct {
 	*BaseChannel
 	session     *discordgo.Session
 	config      config.DiscordConfig
-	transcriber *voice.GroqTranscriber
+	transcriber voice.Transcriber
 	ctx         context.Context
 }
 
@@ -45,7 +45,7 @@ func NewDiscordChannel(cfg config.DiscordConfig, bus *bus.MessageBus) (*DiscordC
 	}, nil
 }
 
-func (c *DiscordChannel) SetTranscriber(transcriber *voice.GroqTranscriber) {
+func (c *DiscordChannel) SetTranscriber(transcriber voice.Transcriber) {
 	c.transcriber = transcriber
 }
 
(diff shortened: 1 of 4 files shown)


nikolasdehor:
This introduces the pluggable STT interface that #354 (local Whisper) and #635 (OpenAI-compatible Whisper) build on. All three address #245. Note that #635 takes a different approach (generalized OpenAI-compatible endpoint) which may conflict with the interface defined here. Worth coordinating to avoid duplicated abstraction layers.

nikolasdehor (approved):
Clean, minimal refactoring. Extracting the Transcriber interface from the concrete GroqTranscriber type is textbook Go interface design -- define the interface at the consumer site and keep the concrete implementation separate.

The interface is well-scoped (Transcribe + IsAvailable), the change touches only type signatures without altering behavior, and it unblocks PR #354 (local Whisper provider) and any future STT backends.

LGTM.

afjcjsbx:
implemented in: https://github.com/sipeed/picoclaw/pull/947

PR #355, opened 2026-02-17
feat(voice/tts): add TTS synthesis and voice parameter on message tool

emadomedher:
## Summary

Adds outbound voice capability: the agent can reply with an audio message by passing `voice=true` to the `message` tool. This is useful when the user sends a voice message and expects a voice reply, or explicitly asks for audio.

The implementation is provider-agnostic — any server that speaks OpenAI's `/v1/audio/speech` API works (Kokoro, Piper, Coqui, OpenAI, etc.).

## Changes

- `pkg/voice/synthesizer.go`: `Synthesizer` interface (`Synthesize`, `IsAvailable`)
- `pkg/voice/kokoro.go`: `KokoroSynthesizer` — OpenAI-compatible `/v1/audio/speech` client. Returns a temp `.mp3` path; caller cleans up.
- `pkg/bus/types.go`: add `Media []string` to `OutboundMessage` (backward-compatible, `omitempty`)
- `pkg/channels/manager.go`: add `SendFileToChannel()` — routes local file paths through a channel's `Send()`
- `pkg/tools/message.go`: add `voice=true` parameter + `SynthesizeCallback` + `SendMediaCallback`. Graceful fallback to text if TTS unavailable. `HasSentInRound` fires for both text and voice.
- `pkg/agent/loop.go`: add `SetVoiceCallbacks()` to attach TTS to the message tool after channel manager init
- `cmd/picoclaw/main.go`: wire Kokoro after channels init

## Config

```json
"tools": {
  "tts": {
    "enabled": true,
    "api_base": "http://localhost:8100",
    "voice": "en_us-lessac-medium"
  }
}
```

## Agent usage

The model simply sets `voice=true` in the message tool call:
```json
{ "tool": "message", "content": "Sure, here's your answer.", "voice": true }
```

## Depends on

➡️ #353 + #354 (Transcriber interface + Whisper) — shares the same voice pipeline foundation.

## Tests

All existing tests pass (`go test ./...`).

### DIFF
--- a/pkg/voice/kokoro.go
+++ b/pkg/voice/kokoro.go
@@ -201,19 +201,25 @@ func (s *KokoroSynthesizer) Synthesize(ctx context.Context, text string) (string
 	return tmpFile.Name(), nil
 }
 
-// IsAvailable checks if the Kokoro TTS server is reachable.
+// IsAvailable checks if the TTS server is reachable.
+// Chatterbox uses /health; all others use /v1/models.
 func (s *KokoroSynthesizer) IsAvailable() bool {
 	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
 	defer cancel()
 
-	req, err := http.NewRequestWithContext(ctx, "GET", s.apiBase+"/v1/models", nil)
+	endpoint := "/v1/models"
+	if s.isChatterbox() {
+		endpoint = "/health"
+	}
+
+	req, err := http.NewRequestWithContext(ctx, "GET", s.apiBase+endpoint, nil)
 	if err != nil {
 		return false
 	}
 
 	resp, err := s.httpClient.Do(req)
 	if err != nil {
-		logger.DebugCF("voice", "Kokoro TTS health check failed", map[string]interface{}{
+		logger.DebugCF("voice", "TTS health check failed", map[string]interface{}{
 			"error": err.Error(),
 		})
 		return false
@@ -221,9 +227,10 @@ func (s *KokoroSynthesizer) IsAvailable() bool {
 	defer resp.Body.Close()
 
 	available := resp.StatusCode == http.StatusOK
-	logger.DebugCF("voice", "Kokoro TTS availability", map[string]interface{}{
+	logger.DebugCF("voice", "TTS availability", map[string]interface{}{
 		"available":   available,
 		"status_code": resp.StatusCode,
+		"endpoint":    endpoint,
 	})
 	return available
 }


nikolasdehor (commented):
Well-designed feature. The provider-agnostic approach (any OpenAI-compatible TTS server) and the Chatterbox native endpoint support are both thoughtful. Notes:

**1. Same API endpoint concerns as PR #354**

The Whisper transcriber in this PR (same code) uses `/transcribe` and `/health` instead of the OpenAI-compatible `/v1/audio/transcriptions` and `/v1/models`. See my review on #354 for details.

**2. TTS fallback behavior is silent and confusing**

When `voice=true` but TTS is not available, the tool returns:
```go
return &ToolResult{ForLLM: "TTS not available — sending as text instead", IsError: false}
```

But it does NOT actually send the text. The LLM receives a success-like message saying it fell back to text, but the user gets nothing. The tool should either (a) actually fall back and send the text via `sendCallback`, or (b) set `IsError: true` so the LLM knows to retry with regular text.

Same issue with TTS synthesis failure:
```go
return &ToolResult{ForLLM: fmt.Sprintf("TTS synthesis failed: %v — falling back to text", err), IsError: false}
```
Again, the text is not actually sent.

**3. Zero-value ambiguity in TTSProfile defaults**

```go
if p.Exaggeration == 0 {
    p.Exaggeration = 0.5
}
if p.CFGWeight == 0 {
    p.CFGWeight = 0.5
}
```

If a user explicitly sets `exaggeration: 0.0` or `cfg_weight: 0.0` in config (both valid values for Chatterbox), the code overrides them to 0.5. Use a pointer type, a separate "is set" flag, or document that 0 is not a valid value.

**4. Temp file cleanup responsibility**

`Synthesize()` creates a temp file and the caller (`MessageTool.Execute`) is responsible for cleanup via `defer os.Remove(audioPath)`. However, if `sendMediaCallback` is asynchronous or buffers the file path for later delivery, the file might be deleted before it is sent. The current code appears synchronous, but this is a fragile contract that should be documented or enforced.

**5. Missing newline at end of config.example.json**

The diff removes the trailing newline from `config.example.json`. Minor, but it will cause a noisy diff for the next PR that touches this file.

**6. No tests for the voice pipeline**

The TTS synthesizer, message tool voice path, and channel media send path have no test coverage. Even a test for the fallback behavior (TTS unavailable) would catch the bug in point 2.

Good feature overall. The silent fallback bug (point 2) is the most important fix needed.

PR #372, opened 2026-02-17
fix: add proper Ollama provider support without requiring API key

ashishsharma349:
📝 Description
Fixed Ollama provider support which was broken due to two bugs in http_provider.go:

The fallback case required APIKey != "" for Ollama, but Ollama doesn't use an API key — so it never matched
Model name detection only worked if the model name contained "ollama", meaning any local model name (like lfm2.5-thinking:latest) would fail
The apiKey == "" validation was rejecting Ollama connections even when a valid api_base was configured

🗣️ Type of Change

🐞 Bug fix

🤖 AI Code Generation

🛠️ Mostly AI-generated (AI draft, Human verified/modified)

🧪 Test Environment & Hardware

Hardware: Intel i3-1215U laptop, Windows 11
OS: Windows 11
Model/Provider: Ollama with lfm2.5-thinking:latest
Channels: CLI

📸 Proof of Work
Before fix: Error: no API key configured for model: lfm2.5-thinking
After fix: Agent initializes and responds successfully via Ollama

### DIFF
--- a/pkg/providers/http_provider.go
+++ b/pkg/providers/http_provider.go
@@ -178,6 +178,11 @@ func (p *HTTPProvider) parseResponse(body []byte) (*LLMResponse, error) {
 			}
 		}
 
+		// Fix: skip tool calls with empty names to avoid downstream errors
+		if name == "" {
+			continue
+		}
+
 		toolCalls = append(toolCalls, ToolCall{
 			ID:        tc.ID,
 			Name:      name,
@@ -285,7 +290,8 @@ func CreateProvider(cfg *config.Config) (LLMProvider, error) {
 				apiKey = cfg.Providers.Gemini.APIKey
 				apiBase = cfg.Providers.Gemini.APIBase
 				if apiBase == "" {
-					apiBase = "https://generativelanguage.googleapis.com/v1beta"
+					// Use OpenAI-compatible endpoint for tool calling support
+					apiBase = "https://generativelanguage.googleapis.com/v1beta/openai"
 				}
 			}
 		case "vllm":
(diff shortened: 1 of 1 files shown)


nikolasdehor:
This overlaps significantly with #384 and #412. Of the three, #412 is the most comprehensive — it handles keyless Ollama support and also hardens the exec working_dir safety check. Suggesting this be closed in favor of #412 to consolidate reviewer effort.

nikolasdehor (changes requested):
The Ollama support improvements (no API key required, OpenAI-compatible Gemini endpoint) are welcome, but there are issues:

1. CRITICAL - Port-based API key bypass: The check 'strings.Contains(apiBase, "11434")' at line ~450 is fragile and insecure. It means any provider with '11434' anywhere in their URL (e.g., a malicious API base like 'https://evil.com/path/11434/steal') would bypass the API key requirement. The correct approach is to check for the 'ollama' provider name explicitly, not a port number in the URL string.

2. Dead code path in fallback detection: The old Ollama detection used 'strings.Contains(lowerModel, "ollama") || strings.HasPrefix(model, "ollama/")' which is model-name-based. The new code uses 'cfg.Providers.Ollama.APIBase \!= ""' which is config-based. This means if a user has Ollama configured but uses a model name like 'llama3', the explicit provider case works, but the fallback case now triggers for ANY configured Ollama.APIBase regardless of model name. This could incorrectly route non-Ollama models to Ollama.

3. Removed debug prints: Good -- the fmt.Println('Ollama provider selected...') and fmt.Println('Ollama apiBase:...') removals are correct.

4. Skipping empty tool call names: The 'if name == "" { continue }' guard in parseResponse is a reasonable defensive fix but could mask real bugs. Consider logging a warning when this happens so users can report it.

5. Gemini endpoint change: Switching from v1beta to v1beta/openai for ALL Gemini users is a breaking change if anyone depends on the native Gemini API format. This should be documented prominently.

Please fix the port-string-based API key bypass (point 1) -- it is a security concern.

PR #378, opened 2026-02-17
fix(exec): kill child process tree on timeout to prevent orphaned tasks

### LINKED ISSUE #311: [BUG] Shell process need kill

lunareed720:
## Summary
This fixes orphaned background processes when `exec` commands time out.

### What changed
- Run shell commands in a dedicated process group on Unix (`Setpgid: true`)
- On timeout/cancel, terminate the full process tree instead of only the parent shell
  - Unix: kill process group (`kill -SIGKILL -pgid`)
  - Windows: `taskkill /T /F /PID` fallback
- Switch `ExecTool` execution flow to `Start + Wait` with explicit timeout path and cleanup
- Add regression test (Unix): verifies timed-out command with spawned child (`sleep`) does not leave child process alive

## Why
Issue #311 reports that failed/timed-out shell commands keep running in background. This can leak resources and degrade long-running agent stability.

## Testing
- `go test ./pkg/tools -run TestShellTool -count=1`

Closes #311

### DIFF
--- a/pkg/tools/shell.go
+++ b/pkg/tools/shell.go
@@ -3,6 +3,7 @@ package tools
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
@@ -109,18 +110,43 @@ func (t *ExecTool) Execute(ctx context.Context, args map[string]interface{}) *To
 		cmd.Dir = cwd
 	}
 
+	prepareCommandForTermination(cmd)
+
 	var stdout, stderr bytes.Buffer
 	cmd.Stdout = &stdout
 	cmd.Stderr = &stderr
 
-	err := cmd.Run()
+	if err := cmd.Start(); err != nil {
+		return ErrorResult(fmt.Sprintf("failed to start command: %v", err))
+	}
+
+	done := make(chan error, 1)
+	go func() {
+		done <- cmd.Wait()
+	}()
+
+	var err error
+	select {
+	case err = <-done:
+	case <-cmdCtx.Done():
+		_ = terminateProcessTree(cmd)
+		select {
+		case err = <-done:
+		case <-time.After(2 * time.Second):
+			if cmd.Process != nil {
+				_ = cmd.Process.Kill()
+			}
+			err = <-done
+		}
+	}
+
 	output := stdout.String()
 	if stderr.Len() > 0 {
 		output += "\nSTDERR:\n" + stderr.String()
 	}
 
 	if err != nil {
-		if cmdCtx.Err() == context.DeadlineExceeded {
+		if errors.Is(cmdCtx.Err(), context.DeadlineExceeded) {
 			msg := fmt.Sprintf("Command timed out after %v", t.timeout)
 			return &ToolResult{
 				ForLLM:  msg,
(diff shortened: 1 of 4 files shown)


nikolasdehor (approved):
Solid PR — the approach (process groups on Unix, `taskkill /T` on Windows) is the canonical solution for orphaned child processes.

**What I liked:**
- `Setpgid: true` + `syscall.Kill(-pid, SIGKILL)` is the textbook Go approach for process group cleanup ✓
- Build tags (`!windows` / `windows`) for platform separation are idiomatic ✓
- The regression test spawns `sleep 60` and verifies it gets killed — well-structured ✓
- Already-exited process handling is correct (nil checks, discarded `ESRCH` errors) ✓
- Race between normal completion and timeout is handled correctly by the `select` ✓
- `errors.Is` improvement over `==` for `context.DeadlineExceeded` ✓

**Suggestions (non-blocking):**

1. **Double-kill interaction:** The command is created with `exec.CommandContext`, so Go's internal goroutine will _also_ try to kill the direct child when the context expires. This is benign (killing an already-killed process is harmless), but slightly confusing. On Go 1.20+, you could use `cmd.Cancel` to hook `terminateProcessTree` directly into the `CommandContext` machinery, eliminating the hybrid approach:
   ```go
   cmd.Cancel = func() error {
       return terminateProcessTree(cmd)
   }
   cmd.WaitDelay = 2 * time.Second
   ```

2. **`terminateProcessTree` always returns `nil`** — either return meaningful errors or change to no return value, since callers always discard it with `_ =`.

3. **Consider adding a normal-completion test** alongside the timeout test, to verify the `Start + Wait` refactor doesn't break the happy path.

4. **SIGTERM before SIGKILL:** Currently goes straight to SIGKILL. For an exec tool where the command has already timed out this is defensible, but a brief SIGTERM grace period (e.g., 500ms) before SIGKILL would allow children to flush buffers and clean up temp files.

Approving — these are all improvements, not blockers.

Orgmar:
@lunareed720 Solid fix! Killing the entire process group instead of just the parent shell is the right call to avoid orphaned tasks. The cross-platform handling with Setpgid on Unix and taskkill on Windows is thoughtful too.

We're building the **PicoClaw Dev Group** on Discord for contributors to connect. If you're interested, drop an email to `support@sipeed.com` with the subject `[Join PicoClaw Dev Group] lunareed720` and we'll send you an invite!

PR #379, opened 2026-02-17
ci: add PR memory budget guardrail (20MB startup RSS)

### LINKED ISSUE #346: [Optimization] Prevent memory bloat
### **Current Status**

* **Current Baseline**: The core process footprint is approximately **10MB**.
* **Trend**: Recent rapid PR merges show a trend of increasing memory usage.
* **Target Hardware**: embedded linux board DDR as low as 64MB RAM.

### **The Challenge**

As PicoClaw gains features, we risk "death by a thousand cuts" where small memory increases in each PR eventually make the agent unusable on 64MB RAM devices. Memory optimization must remain a first-class citizen over storage size.

### **Proposed Guardrails**

* **PR Memory Audit**: Require memory impact assessment for any PR adding new dependencies or complex data structures.
* **Strict Ceiling**: Establish a **20MB "Hard Red Line"**. Any PR pushing the footprint toward or over this limit must undergo mandatory optimization or be rejected.
* **Automated Tracking**: Implement a basic memory profiling step in the CI pipeline (or manual check-list) to compare footprint before and after merges.
* **Dependency Sanity Check**: Review recently merged PRs to identify and prune redundant libraries that contributed to the recent growth.

### **Definition of Done**

* [ ] Memory footprint stabilized near the **10MB** baseline.
* [ ] A documented process or tool for monitoring memory growth in new PRs.
* [ ] All core features verified to stay within the **20MB** limit on target low-end hardware.

lunareed720:
## Summary
Implements an initial memory guardrail for roadmap issue #346 by adding an automated startup RSS budget check in PR CI.

### What changed
- Added `scripts/memory_budget_check.sh`:
  - Builds binary if missing
  - Measures peak RSS using `/usr/bin/time`
  - Supports Linux and macOS parsing
  - Fails when RSS exceeds budget (default: `20480 KiB`, i.e. 20MB)
- Added `make memory-check` target
- Added `memory-budget` job in `.github/workflows/pr.yml` so every PR runs the check

## Why
Issue #346 asks for:
- memory monitoring in PR flow
- a hard red line guardrail

This PR provides a concrete and low-friction first guardrail while keeping the threshold configurable via `PICOCLAW_MEMORY_BUDGET_KB`.

## Local validation
- `make memory-check`
  - Example output on my machine: `peak RSS: ~12.5MB`, `PASS`.

## Follow-ups
- Extend to multi-command profile matrix (e.g., gateway startup path)
- Persist trend metrics as CI artifacts for regression history

Closes #346

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -56,3 +56,18 @@ jobs:
       - name: Run go test
         run: go test ./...
 
+  memory-budget:
+    runs-on: ubuntu-latest
+    needs: fmt-check
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: go.mod
+
+      - name: Check startup memory budget
+        run: make memory-check
+
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: all build install uninstall clean help test
+.PHONY: all build install uninstall clean help test memory-check
 
 # Build variables
 BINARY_NAME=picoclaw
@@ -126,9 +126,13 @@ clean:
 vet:
 	@$(GO) vet ./...
 
-## fmt: Format Go code
+## test: Run unit tests
 test:
 	@$(GO) test ./...
+
+## memory-check: Validate startup memory stays under budget (default 20MB RSS)
+memory-check:
+	@bash ./scripts/memory_budget_check.sh
 
 ## fmt: Format Go code
 fmt:
--- a/scripts/memory_budget_check.sh
+++ b/scripts/memory_budget_check.sh
@@ -1,0 +1,38 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+ROOT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+BINARY_PATH="${PICOCLAW_BINARY_PATH:-${ROOT_DIR}/build/picoclaw}"
+BUDGET_KB="${PICOCLAW_MEMORY_BUDGET_KB:-20480}"
+
+if [[ ! -x "${BINARY_PATH}" ]]; then
+  echo "[memory-check] binary not found at ${BINARY_PATH}; building..."
+  make -C "${ROOT_DIR}" build >/dev/null
+fi
+
+tmp_time_out="$(mktemp)"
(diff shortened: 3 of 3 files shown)


harshbansal7:
Nice work there mate! 👍 
Do you think this is sufficient to ensure we are well under 20MB limit? 
There could be multiple new features introduced in a PR that could cause memory increase at runtime:
- An implementation of a map in any sort, maintaining states for longer than a task's duration. 
- An HTTP call loading enormous amounts of data, possibly downloading a file. 

Yes these are second order impacts, but happy to discuss if we could maybe add more checks and measures to ensure that at a "PR feature" level, this doesn't break. Nonetheless, this is pretty good to begin with, but won't be sufficient to close https://github.com/sipeed/picoclaw/issues/346. 

@Zepan what do you think?

nikolasdehor (approved):
Clean CI guardrail implementation. The script is well-structured with proper `set -euo pipefail`, temp file cleanup via `trap`, and handles both macOS (bytes) and Linux (KB) `/usr/bin/time` output formats correctly.

The 20 MB budget with ~60% headroom over the current ~12.5 MB baseline is reasonable — tight enough to catch regressions, loose enough to avoid flaky failures.

One suggestion (non-blocking): Consider adding a brief comment in the script explaining that macOS reports bytes while Linux reports KB, to help future maintainers understand the division difference.

Also noted: the Makefile change fixes the `test:` target comment from "fmt" to "test" — good catch.

LGTM.

nikolasdehor (approved):
Nice CI addition. A memory budget guardrail for startup RSS is useful for catching accidental dependency bloat in a tool that advertises itself as 'ultra-lightweight'.

Highlights:
- Cross-platform support (Darwin vs Linux /usr/bin/time output parsing)
- Configurable via env vars (PICOCLAW_BINARY_PATH, PICOCLAW_MEMORY_BUDGET_KB)
- Auto-builds binary if missing
- Clean temp file handling with trap
- 20MB budget is reasonable for a Go binary with no heavy dependencies

Minor notes (non-blocking):
1. The CI job depends on fmt-check but not on the test job. If the test job fails but fmt passes, the memory check still runs. Consider adding 'needs: [fmt-check, test]' or making it independent.
2. The comment '## fmt: Format Go code' appears twice in the Makefile -- the first one (above test:) should say '## test: Run unit tests' which it was corrected to. Good catch.
3. On CI (ubuntu-latest), /usr/bin/time should be available, but some minimal containers may not have it. The 'set -euo pipefail' will correctly fail in that case.

LGTM.

PR #383, opened 2026-02-17
feat: integrate Yahoo finance stock price tool

afjcjsbx:
## 📝 Description

Introduced a new tool, StockTool (get_stock_price), which allows the agent to retrieve real-time stock prices and currency information directly from Yahoo Finance.

## 🗣️ Type of Change
- [x] 📖 Documentation update

## 🤖 AI Code Generation

Reference: Yahoo Finance API V8 (https://query1.finance.yahoo.com/v8/finance/chart/{ticker}?interval=1d&range=1d)

Reasoning: Initially, the LLM tried to write bash scripts via the exec tool to scrape Yahoo Finance, which was blocked by the safety guards. By introducing a dedicated, native tool, we save tokens, drastically reduce latency, and provide a deterministic output. The testing approach uses httptest to mock the Yahoo JSON response, ensuring tests pass reliably in CI/CD pipelines without internet dependency.

## 🧪 Test Environment & Hardware
- **Hardware:** Apple Silicon (Mac M-series)
- **OS:** macOS Tahoe 26.2
- **Model/Provider:** GLM-4.7-Flash / OpenAI / OpenRouter 
- **Channels:** CLI / Telegram

## 📸 Proof of Work (Optional for Docs)
<details>

</details>

## ☑️ Checklist


lukemilby:
Thanks for your contribution. This would be better utilized as a skill you can create in your own instance.

afjcjsbx:
yes absolutely, I introduced it only for those who do not want to start the agent without safety guards.

nikolasdehor (commented):
Clean implementation with good test coverage via httptest mocking. A few observations:

**Should fix**

1. **SSRF via ticker parameter**: The ticker value is interpolated directly into the URL without sanitization. A malicious LLM-generated ticker like AAPL?range=5y could manipulate query parameters. More critically, values containing ../ or encoded slashes could alter the URL path. Use url.PathEscape(ticker) instead of raw string interpolation.

2. **New http.Client per call**: A new http.Client is created on every Execute call. This means no connection reuse (TCP connection pool). Move the client to a struct field similar to how other tools in the codebase work. The httptest approach in tests already handles this cleanly via baseURL injection.

3. **User-Agent spoofing**: The fake browser User-Agent string works around Yahoo bot detection but is fragile. Yahoo has been known to change their API or block automated access. This tool will silently break when that happens. Consider documenting this fragility.

**Design consideration**

4. **Always-on registration**: The tool is unconditionally registered in createToolRegistry. For a lightweight project targeting Raspberry Pi, every tool adds to the tool definitions sent to the LLM (which costs tokens). Consider making it opt-in via config.

5. Tests are thorough - success, HTTP error, API-level error, invalid JSON, and missing parameter cases are all covered. The httptest pattern is well-applied.

Overall: solid implementation, but please address the SSRF issue (item 1) before merge.

PR #387, opened 2026-02-17
fix: exec tool should not treat URLs as file paths

### LINKED ISSUE #386: Bug: exec tool blocks commands containing URLs due to path guard misidentifying URLs as file paths
## Description

The `guardCommand` function in `pkg/tools/shell.go` incorrectly blocks shell commands that contain URLs (e.g. `curl https://example.com`). The path guard regex `/[^\s\"']+` matches URL paths like `//example.com` from `https://example.com` and treats them as filesystem paths, then rejects the command because the "path" is outside the working directory.

## Steps to Reproduce

1. Set `restrict_to_workspace: true` in config
2. Send a message to the bot asking it to fetch a webpage
3. The bot tries to run: `curl -L -s https://catellect.com`
4. Command is blocked with: `Command blocked by safety guard (path outside working dir)`

## Expected Behavior

URLs in shell commands should not be treated as file paths. Only actual filesystem paths should be checked against the workspace restriction.

## Actual Behavior

```
[ERROR] tool: Tool execution failed {tool=exec, error=Command blocked by safety guard (path outside working dir)}
```

## Root Cause

In `pkg/tools/shell.go` line 191, the regex `[A-Za-z]:\\[^\\\"']+|/[^\s\"']+` matches URL components as file paths. For `https://example.com`, it matches `//example.com` which then fails the workspace path check.

## Proposed Fix

Strip URLs from the command string before running path checks:

```go
urlPattern := regexp.MustCompile(`[a-zA-Z][a-zA-Z0-9+.-]*://[^\s\"']+`)
stripped := urlPattern.ReplaceAllString(cmd, "")
// then run path checks on `stripped` instead of `cmd`
```

JexLau:
## Summary

- Fix `guardCommand` in `pkg/tools/shell.go` misidentifying URLs as filesystem paths
- Commands like `curl https://example.com` were blocked with "path outside working dir" when `restrict_to_workspace` is enabled
- Strip URLs from command string before running path checks

Fixes #386

## Test plan

- [x] All existing shell tool tests pass (`TestShellTool_RestrictToWorkspace` etc.)
- [ ] Verify `exec` tool can run `curl https://example.com` with `restrict_to_workspace: true`
- [ ] Verify `exec` tool still blocks `cat /etc/passwd` with `restrict_to_workspace: true`

### DIFF
--- a/pkg/tools/shell.go
+++ b/pkg/tools/shell.go
@@ -188,8 +188,12 @@ func (t *ExecTool) guardCommand(command, cwd string) string {
 			return ""
 		}
 
+		// Strip URLs before path checking so they don't get misidentified as file paths
+		urlPattern := regexp.MustCompile(`[a-zA-Z][a-zA-Z0-9+.-]*://[^\s\"']+`)
+		stripped := urlPattern.ReplaceAllString(cmd, "")
+
 		pathPattern := regexp.MustCompile(`[A-Za-z]:\\[^\\\"']+|/[^\s\"']+`)
-		matches := pathPattern.FindAllString(cmd, -1)
+		matches := pathPattern.FindAllString(stripped, -1)
 
 		for _, raw := range matches {
 			p, err := filepath.Abs(raw)


JexLau:
Closing this PR — after further analysis, the exec tool should NOT allow network access by design. The web_fetch tool is the proper channel for external requests. The path guard blocking URLs was an unintended but correct security behavior.

PR #396, opened 2026-02-18
Add XMPP channel support/添加 XMPP 通道支持

toughworm:
## Description

- Add XMPP channel support to PicoClaw
- Support basic text messages and file uploads over XMPP
- Add XMPP config fields and documentation for configuration

## Notes

- Tested locally on Raspberry Pi (xmpp session + file upload)
- This PR only adds XMPP channel; session TTL and other features are kept separate

### DIFF
--- a/cmd/picoclaw/internal/onboard/workspace/AGENT.md
+++ b/cmd/picoclaw/internal/onboard/workspace/AGENT.md
@@ -1,0 +1,12 @@
+# Agent Instructions
+
+You are a helpful AI assistant. Be concise, accurate, and friendly.
+
+## Guidelines
+
+- Always explain what you're doing before taking actions
+- Ask for clarification when request is ambiguous
+- Use tools to help accomplish tasks
+- Remember important information in your memory files
+- Be proactive and helpful
+- Learn from user feedback
--- a/cmd/picoclaw/internal/onboard/workspace/IDENTITY.md
+++ b/cmd/picoclaw/internal/onboard/workspace/IDENTITY.md
@@ -1,0 +1,56 @@
+# Identity
+
+## Name
+PicoClaw 🦞
+
+## Description
+Ultra-lightweight personal AI assistant written in Go, inspired by nanobot.
+
+## Version
+0.1.0
+
+## Purpose
+- Provide intelligent AI assistance with minimal resource usage
+- Support multiple LLM providers (OpenAI, Anthropic, Zhipu, etc.)
+- Enable easy customization through skills system
+- Run on minimal hardware ($10 boards, <10MB RAM)
+
+## Capabilities
+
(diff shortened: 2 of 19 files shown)


nikolasdehor (changes requested):
The XMPP channel implementation looks functional and the XEP-0085 (chat states) and XEP-0184 (receipts) support is a nice touch. However, there are several issues:

**Must fix**

1. **Accidental file committed**: A file called `tream` (298 lines) is included in this PR. It is the output of the `less` command (a `SUMMARY OF LESS COMMANDS` pager dump). This must be removed from the PR.

2. **Unrelated deletions**: `workspace/skills/tmux/scripts/find-sessions.sh` and `workspace/skills/tmux/scripts/wait-for-text.sh` are deleted but unrelated to XMPP support. These should not be in this PR.

3. **Scope creep**: The PR adds session TTL cleanup logic (`startSessionCleanup`, `cleanupExpiredSessions`, `initChannelTTLs`) and per-channel `SessionTTL` config fields for ALL channels (not just XMPP). The session TTL feature should be a separate PR. It also adds `media` field support to `bus.OutboundMessage` and changes the `MessageTool` callback signature, which conflicts with PR #457 that adds `Attachments` to the same struct.

4. **Goroutine leak in `startSessionCleanup`**: The goroutine runs `for { <-ticker.C; ... }` with no exit condition. When the AgentLoop is stopped, this goroutine leaks forever. It needs to select on a context done channel or a stop signal.

5. **No test coverage for XMPP**: 656 lines of channel code with no tests. The message dedup logic (lastMsg check), `parseTTL`, and cleanup logic are all testable.

**Security**

6. **File upload without size validation**: `uploadFile` reads the entire file without checking size. XEP-0363 servers have upload limits (typically 10MB). Reading a large file into memory before discovering the server rejects it wastes resources. Check file size before reading.

7. **SSRF potential in `uploadFile`**: The PUT URL comes from the XMPP server's slot response. If a compromised XMPP server returns an internal URL (e.g. http://169.254.169.254/...), the client will PUT data to it. This is inherent to XEP-0363 but worth documenting.

**Code quality**

8. The `parseTTL` function with "d" suffix handling has a subtle bug: `time.ParseDuration("7h")` returns 7 hours, then multiplying by 24 gives 168 hours = 7 days. This is correct but confusing -- the input is "7d" which is parsed as "7h" then multiplied. A clearer approach: parse the numeric part with strconv.Atoi and multiply by 24*time.Hour.

9. The dedup window (2 seconds, last single message) is fragile. If two different contacts send identical messages within 2 seconds, the second is silently dropped because `lastFromBare` is overwritten by the first. The dedup should be per-sender, not global.

Please fix items 1-5 and 9 before merge.

alexhoshina:

PR #402, opened 2026-02-18
feat: mail channels ; support long-lived connection and attachment

blingmoon:
# Pull Request: Add Email Channel

### 🎯 The Goal / Use Case

Use an IMAP inbox as an input channel so PicoClaw can receive user messages via email and optionally send replies via SMTP. New emails are treated as user messages; sender email is the chat/session identity. Enables email-based agent interaction (e.g. QQ / 163 mailboxes) with optional attachment handling.

### 💡 Proposed Solution

- **Receive (IMAP):** Use IDLE by default — keep a long-lived connection and let the server push mailbox updates when new mail arrives (RFC 2177). If the server does not support IDLE, fall back to NOOP-based polling; for servers with poor IDLE/NOOP support, set `forced_polling` to use application-level polling at `check_interval`.
- **Send (optional):** Use SMTP when configured. Sender email is the chat/session identity; `allow_from` limits which senders can trigger the agent.
- **Attachments:** Stored under `attachment_dir`; file paths are passed to the agent.

### 🛠 Potential Implementation (Optional)

- IMAP client with IDLE / NOOP fallback and optional `forced_polling`.
- Optional SMTP send; session keyed by sender email.
- README updated with Email setup: QQ / 163 authorization links and config example.

### 🚦 Impact & Roadmap Alignment

- [ ] This is a Core Feature

### 🔄 Alternatives Considered

- Polling-only: rejected in favor of IDLE-first for lower latency and less server load when supported.
- No `forced_polling`: added for servers that do not implement IDLE/NOOP per spec.

### 💬 Additional Context

**Reference:**

- [RFC 3501 IMAP](https://www.rfc-editor.org/rfc/rfc3501), [RFC 2177 IMAP IDLE](https://www.rfc-editor.org/rfc/rfc2177)
- QQ 邮箱授权码: https://wx.mail.qq.com/list/readtemplate?name=app_intro.html#/agreement/authorizationCode
- 网易 163 客户端协议/授权码: https://help.mail.163.com/faq.do?m=list&categoryID=90

### DIFF
--- a/pkg/channels/email/email.go
+++ b/pkg/channels/email/email.go
@@ -49,9 +49,13 @@ const (
 
 type EmailChannel struct {
 	*channels.BaseChannel
-	config      config.EmailConfig
-	imapClient  *client.Client
-	lastUID     uint32
+	config     config.EmailConfig
+	imapClient *client.Client
+	lastUID    uint32
+	// checkEmailMutex
+	// maybe multiple goroutine check email at the same time, use mutex to avoid duplicate check
+	checkEmailMutex sync.Mutex
+
 	mu          sync.Mutex
 	cancel      context.CancelFunc
 	checkTicker *time.Ticker
@@ -580,6 +584,9 @@ func (c *EmailChannel) runIdleLoop(ctx context.Context, pollInterval time.Durati
 }
 
 func (c *EmailChannel) CheckNewEmails(ctx context.Context) {
+	// the lock is to avoid duplicate check email, maybe multiple goroutine check email at the same time
+	c.checkEmailMutex.Lock()
+	defer c.checkEmailMutex.Unlock()
 	for {
 		if err := ctx.Err(); err != nil {
 			return


blingmoon:
# PR #402 – Review feedback addressed

Hi @nikolasdehor,

Thanks for the review. I’ve pushed new commits that address your comments. Summary and main code changes are below. Please take another look when you have time.

---

## 1. Unbounded attachment write (disk exhaustion) and io.ReadAll on email body parts

**Change:** Cap attachment writes with `io.LimitReader` (configurable `attachment_max_bytes`, default 25MB) and cap body part reads with `io.LimitReader` (configurable `body_part_max_bytes`, default 1MB). In both places we use **limit+1** so we can detect when the stream exceeds the limit and reject/skip instead of accepting unbounded data.

**Config constants and defaults** (`pkg/channels/email.go`):

```go
const (
	reconnectBackoffInitial = 1 * time.Second
	reconnectBackoffMax     = 10 * time.Minute
	defaultAttachmentMaxBytes = 25 * 1024 * 1024 // 25MB
	defaultBodyPartMaxBytes  = 1 * 1024 * 1024  // 1MB
)
```

**Body part limit** – `extractEmailBodyAndAttachments` (read limit+1, then check `len(body) > limit`):

```go
limit := int64(c.config.BodyPartMaxBytes)
if limit <= 0 {
	limit = int64(defaultBodyPartMaxBytes)
}
limitedBody := io.LimitReader(p.Body, limit+1)
body, err := io.ReadAll(limitedBody)
if err != nil || len(body) == 0 {
	continue
}
if len(body) > int(limit) {
	textParts = append(textParts, fmt.Sprintf("[body part exceeds size limit (max %d bytes), you can check body_part_max_bytes in config]", limit))
	continue
}
```

**Attachment size limit** – `saveAttachmentToLocal`:

```go
limit := int64(c.config.AttachmentMaxBytes)
if limit <= 0 {
	limit = defaultAttachmentMaxBytes
}
safeName := utils.SanitizeFilename(filename)
// ...
// +1 to detect if the attachment exceeds the limit
limited := io.LimitReader(r, limit+1)
n, err := io.Copy(f, limited)
// ...
if n > limit {
	_ = os.Remove(localPath)
	logger.DebugCF("email", "Attachment exceeds size limit, skipped", ...)
	return ""
}
```

---

## 2. Path traversal via filename

**Change:** Use the **sanitized** filename for `filepath.Ext()` and for building the local name, so extension and path are derived only from the safe name.

**Code** – `saveAttachmentToLocal`:

```go
safeName := utils.SanitizeFilename(filename)
if safeName == "" {
	safeName = "attachment"
}
ext := filepath.Ext(safeName)   // ext from sanitized name, not raw filename
localName := fmt.Sprintf("%d_%d_%s%s", uid, index, strings.TrimSuffix(safeName, ext), ext)
localPath := filepath.Join(dir, localName)
```

---

## 3. SMTP header injection

**Change:** Strip CR/LF from header values via `sanitizeHeaderValue`, and build the outgoing message with `go-message/mail` (RFC-compliant headers via textproto).

**Sanitize and Send** – `email.go`:

```go
// sanitizeHeaderValue removes CR/LF from s to prevent SMTP header injection.
func sanitizeHeaderValue(s string) string {
	return strings.NewReplacer("\r", "", "\n", "").Replace(s)
}

// In Send():
fromRaw := sanitizeHeaderValue(c.config.Username)
toRaw := sanitizeHeaderValue(strings.TrimSpace(msg.ChatID))
// ...
// Build message with go-message/mail: RFC-compliant headers via textproto
var h mail.Header
if fromAddrs, err := mail.ParseAddressList(fromRaw); err == nil && len(fromAddrs) > 0 {
	h.SetAddressList("From", fromAddrs)
} else {
	h.Set("From", fromRaw)
}
if toAddrs, err := mail.ParseAddressList(toRaw); err == nil && len(toAddrs) > 0 {
	h.SetAddressList("To", toAddrs)
} else {
	h.Set("To", toRaw)
}
h.SetSubject(sanitizeHeaderValue("Reply from PicoClaw"))
h.Set("Content-Type", "text/plain; charset=utf-8")
// ... mail.CreateSingleInlineWriter(&buf, h), etc.
```

---

## 4. IDLE loop exits on any error (reliability)

**Change:** On IMAP errors (IDLE, search, fetch), we no longer exit the loop; we close the client and call **reconnect with exponential backoff** (1s → 2s → … up to 10min, then retry every 10min until success or context done). Only one goroutine performs reconnect at a time (guarded by `reconnectMutex`). I tested locally by cutting the network after connect and confirming reconnection behavior.

**Reconnect with backoff** – `reconnectWithBackoff`:

```go
// reconnectWithBackoff: only one goroutine can reconnect at a time
func (c *EmailChannel) reconnectWithBackoff(ctx context.Context) error {
	currentClientVersion := c.reconnectClientVersion
	c.reconnectMutex.Lock()
	defer c.reconnectMutex.Unlock()
	if currentClientVersion != c.reconnectClientVersion {
		// another goroutine already reconnected
		if ctx.Err() != nil {
			return ctx.Err()
		}
		c.mu.Lock()
		isOk := c.imapClient != nil && c.imapClient.State() == imap.SelectedState
		c.mu.Unlock()
		if isOk {
			return nil
		}
	}
	c.reconnectClientVersion++
	c.closeIMAPClient()
	backoff := reconnectBackoffInitial  // 1 * time.Second
	for {
		if err := ctx.Err(); err != nil {
			return err
		}
		err := c.connect()
		if err == nil {
			return nil
		}
		logger.ErrorCF("email", "IMAP reconnect failed, retrying with backoff", ...)
		timer := time.NewTimer(backoff)
		select {
		case <-ctx.Done():
			timer.Stop()
			return ctx.Err()
		case <-timer.C:
			if backoff < reconnectBackoffMax {
				backoff *= 2
				if backoff > reconnectBackoffMax {
					backoff = reconnectBackoffMax  // 10 * time.Minute
				}
			}
		}
	}
}
```

Reconnect is used after IDLE errors, search errors, and fetch errors (see `runIdleLoop` and `checkNewEmails`).

---

## 5. STARTTLS failure warning

**Change:** When STARTTLS fails on port 587 (or similar), we log a warning so users know the connection may be unencrypted and credentials could be sent in plaintext.

**Code** – in `Send()` for non–implicit-TLS path:

```go
if err = client.StartTLS(&tls.Config{ServerName: host}); err != nil {
	// Some servers on 587 do not require STARTTLS; continue anyway
	logger.WarnCF("email", "STARTTLS failed, connection may be unencrypted; credentials could be sent in plaintext", map[string]interface{}{
		"error": err.Error(),
	})
	_ = err
}
```

---

## 6. No test coverage

**Change:** Added `pkg/channels/email_test.go` with unit tests for:

- **sanitizeHeaderValue** – empty, plain, CR, LF, CRLF, and mixed CR/LF cases
- **parseFilenameFromDisposition** – e.g. `inline; filename="pico.png"`
- **decodeRFC2047Filename** – plain and GB2312/GBK RFC 2047 encoded filenames
- **extractEmailBodyAndAttachments** – nil message, plain text body, HTML body, multipart with attachment and text, and **body part max bytes** (limit enforced, placeholder message when over limit)
- **extractTextFromHTML** – script/style stripping and tag stripping
- **saveAttachmentToLocal** – within-limit save and over-limit (returns empty, file not kept)

**Snippet** – body part limit and attachment size tests:

```go
// Body part max bytes
newLimitClient := &EmailChannel{
	config: config.EmailConfig{
		BodyPartMaxBytes:   1,
		AttachmentDir:      t.TempDir(),
		AttachmentMaxBytes: 1024,
	},
}
t.Run("body part max bytes", func(t *testing.T) {
	// ... msg with multipart body ...
	content, paths := newLimitClient.extractEmailBodyAndAttachments(msg)
	assert.Contains(t, content, "[body part exceeds size limit (max 1 bytes), you can check body_part_max_bytes in config]")
	// ...
})

// saveAttachmentToLocal: at limit vs over limit
c.config.AttachmentMaxBytes = len([]byte(content))
path := c.saveAttachmentToLocal(1, 1, "test.txt", body)
assert.NotEmpty(t, path)
// ...
c.config.AttachmentMaxBytes = len([]byte(content)) - 1
path = c.saveAttachmentToLocal(1, 1, "test.txt", body)
assert.Empty(t, path)
```

---

Thanks again for the thorough review. If anything still looks off, happy to adjust.

blingmoon:
#### Why `github.com/bytedance/mockey` was added

The Email channel depends on `*client.Client` (the concrete go-imap type), which cannot be swapped via an interface or a hand-written fake. That made it hard to unit-test logic like `runIdleLoop` and `CheckNewEmails` without a real IMAP server. **mockey** is used to **patch `connect()`** in tests so it returns `nil` instead of performing a real connection. This allows:

- No real TCP connections or dependency on an IMAP server;
- Tests to still run `runIdleLoop` and `CheckNewEmails` and assert branches and error handling.

So mockey addresses the need to **unit-test code that strongly depends on a third-party concrete type (`*client.Client`)**.

#### What this fixes for unit tests

- **Before:** Paths that use IMAP (e.g. `checkNewEmails`, `runIdleLoop`) could only be tested by connecting to a real mailbox or by skipping tests.
- **After:** Tests patch `(*EmailChannel).connect` to return `nil`, then either inject mock IMAP behaviour or only assert that the code does not panic / reconnects when not connected, so these paths are covered without a real IMAP server.

#### Two new unit tests

| Test | Description |
|------|-------------|
| **runIdleLoop** | With `connect` patched, verifies IDLE loop behaviour when the context is cancelled or when an “update” is received, without a real connection. |
| **CheckNewEmails** | With `connect` patched, verifies fetching new mail, processing, and lastUID update logic (or at least that the code does not panic when there is no real connection). |

Both tests rely on mockey patching `connect` so they do not require a real IMAP server.

blingmoon:
Lifecycle compliance from #621 is done: Email Stop() now waits for checkLoop (WaitGroup) and the test verifies it. Ready for review.

blingmoon:
## Summary

### Format & checks
- Code was reformatted and now passes **golangci-lint** and **go vet**.

### New file: `pkg/channels/email_mockey_test.go`
- This file was added for tests that use [github.com/bytedance/mockey](https://github.com/bytedance/mockey). Mockey requires **`-gcflags="all=-N -l"`** to work.
- The PR workflow runs **`go test ./...`** without `-gcflags`, so these mockey-based tests were moved into `email_mockey_test.go` behind **`//go:build mockey`**. As a result:
  - **`go test ./...`** does not build or run these tests by default.
  - To run them locally: **`go test -tags=mockey -gcflags="all=-N -l" ./pkg/channels/...`**.
- **Later:** If the PR workflow is updated to **`go test ./... -gcflags="all=-N -l"`**, the **`//go:build mockey`** tag can be removed and these tests can be merged back into `email_test.go` (or the tag can be dropped and the file kept as-is).

alexhoshina:
Hi, @blingmoon  perhaps you could change the merge target to the [refactor branch](https://github.com/sipeed/picoclaw/tree/refactor/channel-system). Thanks your

alexhoshina:
The linter checks use `golangci-lint run`. Perhaps you could run `golangci-lint run --fix`

blingmoon:
@alexhoshina  Thanks for the feedback! All three points addressed:
Merge target updated — changed the base branch to refactor/channel-system.
golangci-lint clean — ran golangci-lint run ./... with Go 1.25.7; 0 issues reported.
Data race fixed — reconnectClientVersion was a plain int read without a lock before reconnectMutex.Lock(), while a concurrent goroutine could write it under the mutex — a classic read/write race the Go race detector would flag. Fixed by changing the field to atomic.Int64: the pre-lock snapshot uses Load() (always race-free), and the version is incremented with Add(1) only on a successful reconnect, so waiting goroutines can reliably distinguish "reconnected" from "still broken".

blingmoon:
1. Adds WangYi mailbox support: IMAP ID is sent before SELECT/EXAMINE to satisfy 163/126 security checks and prevent “Unsafe Login” errors; 
2. IDLE behavior (multiple updates per new mail) is accounted for. A follow-up commit adds checkEmailMutex to serialize CheckNewEmails, so multiple goroutines cannot run the check at once and duplicate processing is avoided.

pkg/channels/email/email.go
```diff
@@ -0,0 +1,1140 @@
...
+
+	// SenderInfo for allow-list and routing (canonical format: email:addr)
+	senderInfo := bus.SenderInfo{
+		Platform:    "email",
+		PlatformID:  senderID,
+		CanonicalID: identity.BuildCanonicalID("email", senderID),
+	}
+
+	// Check allowlist (HandleMessage will also check; we avoid duplicate work by passing SenderInfo)
+	if !c.IsAllowedSender(senderInfo) {
```
alexhoshina, line 743: This is a redundant check; the whitelist is automatically checked in HandleMessage, so the check here is redundant.

pkg/channels/email/email.go
```diff
@@ -0,0 +1,1140 @@
...
+	if c.cancel != nil {
+		c.cancel()
+		c.cancel = nil
+	}
+	if c.checkTicker != nil {
+		c.checkTicker.Stop()
+		c.checkTicker = nil
+	}
+	if c.imapClient != nil {
+		c.imapClient.Logout()
```
alexhoshina, line 129: `Logout()`Network I/O being blocked while holding a lock may cause a deadlock

pkg/channels/email/email.go
```diff
@@ -0,0 +1,1140 @@
...
+					)
+					return
+				}
+			}
+			c.CheckNewEmails(ctx)
+		}
+	}
+}
+
+func (c *EmailChannel) CheckNewEmails(ctx context.Context) {
```
alexhoshina, line 586: Internal methods should not be made public

pkg/channels/email/email.go
```diff
@@ -0,0 +1,1140 @@
...
+	// +1 to detect if the attachment exceeds the limit
+	limited := io.LimitReader(r, limit+1)
+	n, err := io.Copy(f, limited)
+	if err != nil {
+		_ = os.Remove(localPath)
+		logger.DebugCF("email", "Failed to write attachment", map[string]any{"error": err.Error(), "path": localPath})
+		return ""
+	}
+	if n > limit {
+		_ = os.Remove(localPath)
```
alexhoshina, line 955: The logic here will cause attachments larger than the limit to be written first and then removed. It should check the size before writing.

pkg/channels/email/email.go
```diff
@@ -0,0 +1,1140 @@
...
+	// Check allowlist (HandleMessage will also check; we avoid duplicate work by passing SenderInfo)
+	if !c.IsAllowedSender(senderInfo) {
+		logger.DebugCF("email", "Email from unauthorized sender", map[string]any{
+			"sender": senderID,
+		})
+		return
+	}
+
+	// Extract body and attachments (attachments saved to AttachmentDir, paths in mediaPaths)
+	content, mediaPaths := c.extractEmailBodyAndAttachments(msg)
```
alexhoshina, line 751: In the `Body` map of `*imap.Message`, `imap.Literal` is stored. Perhaps a deeper analysis of memory usage is needed, as memory could be exhausted when handling large emails.

pkg/channels/email/email.go
```diff
@@ -0,0 +1,1140 @@
...
+			continue
+		}
+		bodyStr := strings.TrimSpace(string(body))
+		if bodyStr == "" {
+			continue
+		}
+
+		switch {
+		case strings.HasPrefix(contentType, "text/plain"):
+			textParts = append(textParts, bodyStr)
```
alexhoshina, line 882: Each `text part` has a size limit, but the number of parts is not restricted. If there are a large number of `text parts` whose sizes are close to the limit, this string will consume a significant amount of memory.

pkg/channels/email/email.go
```diff
@@ -0,0 +1,1140 @@
...
+		if isAttachment {
+			filename := getPartFilename(p.Header)
+			if filename == "" {
+				filename = fmt.Sprintf("attachment_%d", attachmentIndex)
+			}
+			attachmentIndex++
+
+			var localPath string
+			if saveDir != "" {
+				localPath = c.saveAttachmentToLocal(msg.Uid, attachmentIndex, filename, p.Body)
```
alexhoshina, line 842: Attachments should be processed using `mediaStore.Store(localPath, meta, scope)`.

alexhoshina:

blingmoon:
@yinwm  
Thank you for your reply!
Initially, I chose to implement email as a channel because email is a universal protocol that is not tied to any specific app. It can work like an open, internet-wide service that anyone can call.
I fully agree that sending emails should be implemented as a skill. At the same time, if email could also run as a channel, it would be very useful for monitoring purposes.
By the way, I have already fixed all the issues mentioned in the last review.
If the Pico team needs to implement the email channel in the future, I would be happy to continue supporting it.

PR #414, opened 2026-02-18
feat(providers): add anthropic-beta header for OAuth token support

gh-xj:
## 📝 Description

Add OAuth token (`sk-ant-oat01-*`) support to the Anthropic provider. The Anthropic API requires the `anthropic-beta: oauth-2025-04-20` header for OAuth authentication — without it, requests return 401: "OAuth authentication is currently not supported."

**Changes:**
- Add `isOAuth` field to `Provider` struct, set via `NewProviderWithTokenSource*` constructors (the only OAuth entry points)
- Add `oauthOpts()` helper method that injects the `anthropic-beta` header only when targeting the official Anthropic API (`baseURL == defaultBaseURL`)
- `Chat()` calls `oauthOpts()` to prepend OAuth headers when applicable
- Custom API endpoints (Vertex AI, Bedrock, LiteLLM via `apiBase` override) are guarded — the header is NOT sent to avoid 400 errors

## 🗣️ Type of Change

## 🤖 AI Code Generation

## 🔗 Related Issue

N/A — discovered while setting up PicoClaw with Claude Code OAuth tokens.

## 📚 Technical Context
- **Reference URL:** https://docs.anthropic.com/en/docs/build-with-claude/token-counting#oauth-tokens
- **Reasoning:** OAuth tokens use Bearer auth instead of x-api-key, and the SDK needs the beta header to enable this code path on the Anthropic API side.

## 🧪 Test Environment
- **Hardware:** Apple M4 MacBook Pro
- **OS:** macOS 15.3.1 (Darwin 25.2.0)
- **Model/Provider:** Anthropic Claude Opus 4.6 via OAuth token
- **Channels:** CLI

### ✅ Checklist
- [x] Code follows the project's style guidelines
- [x] Self-reviewed
- [x] Tests added (3 test cases: OAuth→header sent, regular key→no header, OAuth+custom baseURL→no header)
- [x] No breaking changes
- [x] `make fmt` clean
- [x] `go vet ./...` clean
- [x] `go test ./...` all pass

### DIFF
--- a/pkg/providers/anthropic/provider.go
+++ b/pkg/providers/anthropic/provider.go
@@ -63,7 +63,7 @@ func NewProviderWithTokenSource(token string, tokenSource func() (string, error)
 func NewProviderWithTokenSourceAndBaseURL(token string, tokenSource func() (string, error), apiBase string) *Provider {
 	p := NewProviderWithBaseURL(token, apiBase)
 	p.tokenSource = tokenSource
-	p.isOAuth = true
+	p.isOAuth = strings.HasPrefix(token, "sk-ant-oat01-")
 	return p
 }
 


nikolasdehor (changes requested):
Good intent — OAuth token detection via `sk-ant-oat01-` prefix and the `anthropic-beta: oauth-2025-04-20` header value are both correct. However, there are blocking issues:

### 1. Stale base — needs rebase (BLOCKING)

This PR was branched from an older `main` where `claude_provider.go` contained all Anthropic logic inline. Since then, `main` has been refactored: `claude_provider.go` is now a thin ~65-line delegate wrapper, and all logic lives in `pkg/providers/anthropic/provider.go`. The PR's diff cannot merge cleanly.

**Action:** Rebase onto current `main` and apply the OAuth header logic to `pkg/providers/anthropic/provider.go` instead.

### 2. Missing apiBase guard

The `anthropic-beta` header is only recognized by the official Anthropic API (`api.anthropic.com`). Sending it to custom endpoints (Vertex AI, Bedrock, LiteLLM, etc. via `apiBase` override) can cause 400 errors. The header should only be injected when the request targets the official Anthropic API.

### 3. Missing tests

Please add test cases verifying:
- OAuth token (`sk-ant-oat01-...`) → header is sent
- Regular API key (`sk-ant-api03-...`) → header is NOT sent
- Token refresh via `tokenSource` returning an OAuth token → header is sent

### 4. PR description / code mismatch

The PR description mentions an `isOAuth` field and `oauthOpts()` helper, but the actual committed code uses inline `strings.HasPrefix` checks. Please update the description to match the implementation.

The core approach is sound — looking forward to the updated version!

gh-xj:
Thanks for the thorough review @nikolasdehor! All 4 items addressed in the force-push:

1. **Rebased onto current main** — Logic now lives in `pkg/providers/anthropic/provider.go` (the new refactored structure). `claude_provider.go` is untouched.

2. **apiBase guard added** — `oauthOpts()` only injects the header when `baseURL == defaultBaseURL` (`https://api.anthropic.com`). Custom endpoints (Vertex AI, Bedrock, LiteLLM) are skipped.

3. **3 test cases added:**
   - OAuth token with default API → header IS sent
   - Regular API key → header is NOT sent
   - OAuth token with custom baseURL → header is NOT sent

4. **PR description updated** to match the actual implementation (`isOAuth` field + `oauthOpts()` helper).

All checks pass locally: `make fmt`, `go vet ./...`, `go test ./...`.

PR #422, opened 2026-02-18
Use DefaultTransport for http.Client

### LINKED ISSUE #413: [Feature] allow web_fetch tool to use proxy
## 🎯 The Goal / Use Case

Some environment requires proxy to access the Internet. `web_fetch` should honor `HTTP_PROXY` or other proxy configuration so the agent can properly browse the web.

## 💡 Proposed Solution

Let `web_fetch` tool honor `HTTP_PROXY` environment variable, or add tool config sections for that.

None.

- [x] This is a Core Feature

## 🔄 Alternatives Considered

I asked the agent to fetch an reorganize picoclaw changelogs for me, but GitHub release note URL is not accessible without network proxy. So I installed GitHub cli skill and asked the agent to use `gh api` command to query GitHub API, but that did not work either, because commands not installed in workspace directory are always blocked (error is `Command blocked by safety guard (path outside working dir)`).

## 💬 Additional Context

See above.

mattn:
Fixes #413

## 📝 Description

web.go set nil to http.Client.Transport. This will cause the http.Client to use the default transport, this disable HTTP_PROXY use.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

- **Reference URL:**
- **Reasoning:**

## 🧪 Test Environment
- **Hardware:** 
- **OS:** 
- **Model/Provider:** 
- **Channels:** 

## 📸 Evidence (Optional)
<details>

</details>

## ☑️ Checklist

### DIFF
--- a/pkg/tools/web.go
+++ b/pkg/tools/web.go
@@ -36,7 +36,8 @@ func (p *BraveSearchProvider) Search(ctx context.Context, query string, count in
 	req.Header.Set("Accept", "application/json")
 	req.Header.Set("X-Subscription-Token", p.apiKey)
 
-	client := &http.Client{Timeout: 10 * time.Second}
+	client := http.NewClient()
+	client.Timeout = 10 * time.Second
 	resp, err := client.Do(req)
 	if err != nil {
 		return "", fmt.Errorf("request failed: %w", err)
@@ -96,7 +97,8 @@ func (p *DuckDuckGoSearchProvider) Search(ctx context.Context, query string, cou
 
 	req.Header.Set("User-Agent", userAgent)
 
-	client := &http.Client{Timeout: 10 * time.Second}
+	client := http.NewClient()
+	client.Timeout = 10 * time.Second
 	resp, err := client.Do(req)
 	if err != nil {
 		return "", fmt.Errorf("request failed: %w", err)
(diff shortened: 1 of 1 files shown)


nikolasdehor (commented):
Thanks for the PR, mattn! The intent is right, but I believe this change doesn't actually address issue #413.

When `Transport` is `nil` on `http.Client`, Go already uses `http.DefaultTransport` (which includes `Proxy: http.ProxyFromEnvironment`). So the Brave, DuckDuckGo, and Perplexity search providers **already respect** `HTTP_PROXY`/`HTTPS_PROXY` — this line is effectively a no-op.

The actual problem is in `WebFetchTool.Execute()` (~line 416 of `web.go`), which constructs a **custom `http.Transport{}`** without setting the `Proxy` field:

```go
Transport: &http.Transport{
    MaxIdleConns:        10,
    IdleConnTimeout:     30 * time.Second,
    DisableCompression:  false,
    TLSHandshakeTimeout: 15 * time.Second,
},
```

When you manually construct an `http.Transport{}`, the `Proxy` field defaults to `nil` (no proxy) — unlike `http.DefaultTransport` which sets `Proxy: http.ProxyFromEnvironment`.

**The fix for #413 should be:**

```go
Transport: &http.Transport{
    Proxy:               http.ProxyFromEnvironment,  // ADD THIS
    MaxIdleConns:        10,
    IdleConnTimeout:     30 * time.Second,
    DisableCompression:  false,
    TLSHandshakeTimeout: 15 * time.Second,
},
```

Would you be open to updating the PR to fix the `WebFetchTool` transport instead? That's the code path that issue #413 is specifically about. The Brave search client change here can be kept for explicitness, or dropped since it's a no-op — your call.

mattn:
I changed PR to use http.NewClient().

Initializing `&http.Client{}` or `&http.Transport{}` directly may set default values that differ from the official defaults or omit required field settings.

PR #316, opened 2026-02-16
feat: add configurable summary and compression trigger thresholds

Candlest:
- Add CompressionTriggerRatio, SummaryTriggerMessages, SummaryTriggerRatio to config
- Pass maxTokens and temperature to subagent for consistent LLM settings
- Update ToolLoop to support configurable MaxTokens and Temperature
- Update default config values and documentation
- Fix hardcoded values (8192, 0.7) to use configurable parameters

PR #428, opened 2026-02-18
feat(session): add configurable memory summarization thresholds

### LINKED ISSUE #420: What does "Memory threshold reached. Optimizing conversation history..." mean?
OS: Arch Linux
Installation Method: precompiled binary
Platform: Telegram bot
LLM: deepseek/deepseek-chat

I often encounter the above prompts while chatting with picoclaw. Is this a bug or a feature? It appears too frequently, often disrupting my train of thought. Perhaps it could be hidden away.

easyzoom:
## 📝 Description

- Add memory_message_limit, memory_token_percent, memory_notify_user to SessionConfig
- Don't notify user by default; log only when memory_notify_user is false
- Add `context_window`, `summary_max_tokens`, `summary_temperature` to config
- Use `agent.MaxTokens` and `agent.Temperature` in LLM calls (replace hardcoded values)
- Backward compat: migrate derives `context_window` from `max_tokens` when absent

Ref: @harshbansal7 b226e68
 
## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue

Closes #420 

## 🧪 Test Environment
- **Hardware:** PC
- **OS:** Ubuntu 22.04
- **Model/Provider:** GLM-4
- **Channels:** Telegram

## ☑️ Checklist

### DIFF
--- a/pkg/agent/loop.go
+++ b/pkg/agent/loop.go
@@ -977,8 +977,8 @@ func (al *AgentLoop) summarizeSession(agent *AgentInstance, sessionKey string) {
 			nil,
 			agent.Model,
 			map[string]any{
-				"max_tokens":  1024,
-				"temperature": 0.3,
+				"max_tokens":  agent.SummaryMaxTokens,
+				"temperature": agent.SummaryTemperature,
 			},
 		)
 		if err == nil {
@@ -1027,8 +1027,8 @@ func (al *AgentLoop) summarizeBatch(
 		nil,
 		agent.Model,
 		map[string]any{
-			"max_tokens":  1024,
-			"temperature": 0.3,
+			"max_tokens":  agent.SummaryMaxTokens,
+			"temperature": agent.SummaryTemperature,
 		},
 	)
 	if err != nil {
--- a/config/config.example.json
+++ b/config/config.example.json
@@ -4,8 +4,11 @@
       "workspace": "~/.picoclaw/workspace",
       "restrict_to_workspace": true,
       "model": "gpt4",
+      "context_window": 8192,
       "max_tokens": 8192,
       "temperature": 0.7,
+      "summary_max_tokens": 1024,
+      "summary_temperature": 0.3,
       "max_tool_iterations": 20
     }
   },
--- a/pkg/agent/instance.go
+++ b/pkg/agent/instance.go
@@ -15,22 +15,24 @@ import (
 // AgentInstance represents a fully configured agent with its own workspace,
 // session manager, context builder, and tool registry.
 type AgentInstance struct {
-	ID             string
-	Name           string
-	Model          string
-	Fallbacks      []string
-	Workspace      string
-	MaxIterations  int
-	MaxTokens      int
-	Temperature    float64
-	ContextWindow  int
-	Provider       providers.LLMProvider
-	Sessions       *session.SessionManager
-	ContextBuilder *ContextBuilder
-	Tools          *tools.ToolRegistry
-	Subagents      *config.SubagentsConfig
-	SkillsFilter   []string
-	Candidates     []providers.FallbackCandidate
+	ID                 string
+	Name               string
+	Model              string
+	Fallbacks          []string
+	Workspace          string
+	MaxIterations      int
+	ContextWindow      int // Token limit used for summarization threshold and message-size guard (not the LLM response limit).
+	MaxTokens          int // Max tokens allowed in a single LLM response (passed to provider Chat).
+	Temperature        float64
+	SummaryMaxTokens   int
+	SummaryTemperature float64
+	Provider           providers.LLMProvider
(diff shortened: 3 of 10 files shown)


pkg/agent/loop.go
```diff
@@ -697,19 +697,35 @@ func (al *AgentLoop) updateToolContexts(agent *AgentInstance, channel, chatID st
 func (al *AgentLoop) maybeSummarize(agent *AgentInstance, sessionKey, channel, chatID string) {
 	newHistory := agent.Sessions.GetHistory(sessionKey)
 	tokenEstimate := al.estimateTokens(newHistory)
-	threshold := agent.ContextWindow * 75 / 100
 
-	if len(newHistory) > 20 || tokenEstimate > threshold {
+	msgLimit := 20
+	tokenPercent := 75
```
harshbansal7: NIT: Please declare these as constants on the top of the file using const (...)

pkg/agent/loop.go
```diff
@@ -697,19 +697,35 @@ func (al *AgentLoop) updateToolContexts(agent *AgentInstance, channel, chatID st
...
+	if al.cfg != nil {
+		if al.cfg.Session.MemoryMessageLimit > 0 {
+			msgLimit = al.cfg.Session.MemoryMessageLimit
+		}
+		if al.cfg.Session.MemoryTokenPercent > 0 {
+			tokenPercent = al.cfg.Session.MemoryTokenPercent
+		}
+		notifyUser = al.cfg.Session.MemoryNotifyUser
+	}
+	threshold := agent.ContextWindow * tokenPercent / 100
```
harshbansal7, line 768: The origin of agent.ContextWindow is the max_tokens setting in the config.json. (In old code)
Ideally `max_tokens` itself is generally used in terms of limiting response tokens, and also exploring the code a bit, I found we never really use this config for the right purpose. 

@easyzoom Would you be willing to see how "max_tokens" config impacts this flow, and possibly decide the way on how to fix its implementation in code? 

Happy to hear thoughts on this once you see how we have hardcoded max_token configs in the code.

harshbansal7: @easyzoom Asked Cursor to find and resolve the ambiguity, here's the commit for your reference. 
https://github.com/harshbansal7/picoclaw/pull/1/changes/b226e68d9c70c489383e83988b63a00bb26496a2

easyzoom: Done. I've applied the approach from @harshbansal7's b226e68: added context_window (for summarization threshold), kept max_tokens for response limit, and added summary_max_tokens / summary_temperature for summarization calls.

pkg/agent/loop.go
easyzoom: Done. I've extracted them as defaultMemoryMessageLimit and defaultMemoryTokenPercent at the top of the file.

easyzoom:
> Good feature addition. Making the summarization thresholds configurable addresses a real need (see also issue #420 where a user was confused by the hardcoded "Memory threshold reached" message), and defaulting `memory_notify_user` to `false` is the right UX call — most users don't want to see internal optimization messages in their chat.
> 
> The fallback logic (defaulting to 20 messages / 75% tokens when config values are 0) is correct and backwards-compatible.
> 
> Two small suggestions (non-blocking):
> 
> 1. The `else` branch (logging when `notifyUser` is false) doesn't check `!constants.IsInternalChannel(channel)` like the original code did. Not harmful (it's just a log line), but for consistency you might want to keep the internal-channel guard.
> 2. This is a new feature (configurable thresholds), not a bug fix — minor, but the PR type checkbox should reflect that.
> 
> Approving as-is.

Thanks for the review! I'll add the !constants.IsInternalChannel(channel) check in the else branch for consistency, and update the PR type to "feature".

pkg/agent/loop.go
harshbansal7: @easyzoom Thanks mate. Just to be sure, please test it once, as these changes can have great consequences. If you are unable to test, let me know and I'll try to help.

easyzoom: I've tested it on Ubuntu 22.04 with Telegram and didn't run into any issues. It would be great if you could run a quick test on your side too, just to be sure. Thanks.

harshbansal7:
@easyzoom Please resolve the conflicts, I'll get it merged then.

nikolasdehor:
This is near-identical to #316 — both add configurable memory summarization thresholds. #316 was opened earlier. Suggesting this be closed in favor of #316 unless this version has meaningful improvements over it.

easyzoom:
> This is near-identical to #316 — both add configurable memory summarization thresholds. #316 was opened earlier. Suggesting this be closed in favor of #316 unless this version has meaningful improvements over it.

Makes sense. I'll close this in favor of #316.

PR #455, opened 2026-02-19
fix: correct invalid model ID in config example JSON

penzhan8451:
For default model id,  update default model ID to "z-ai/glm-4.7" in config.example.json.  Tried on my local, It works.


nikolasdehor (approved):
Straightforward fix. The model ID format in picoclaw uses `provider/model` notation, so `z-ai/glm-4.7` is the correct format vs the bare `glm-4.7`.

LGTM. Small but useful — example configs with invalid values are a common source of confusion for new users.

PR #467, opened 2026-02-19
build: pin base images and improve Docker setup

pixel365:
## 📝 Description

Improve Docker image reproducibility and reduce runtime size.

- Pin `golang` and `alpine` base images by sha256 to ensure deterministic builds.
- Remove explicit `wget` installation (use BusyBox-provided wget).
- Reduce final image size.
- Add optional `PICOCLAW_CONFIG_FILE` variable for docker-compose to allow flexible config mounting.
- Minor Dockerfile cleanup and simplification.

No application logic changes.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue

N/A

- **Reasoning:** Improve build reproducibility, reduce image size, and simplify Docker configuration without changing runtime behavior.

## 🧪 Test Environment
- **Hardware:** PC
- **OS:** Fedora

## 📸 Evidence (Optional)
<details>

- Image size reduced from ~48 MB to ~39 MB
- Healthcheck verified using BusyBox wget
- docker-compose config override tested with `PICOCLAW_CONFIG_FILE`

</details>

## ☑️ Checklist

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -32,6 +32,8 @@ COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/ca-certifi
 
 RUN mkdir -p /home/picoclaw && chown picoclaw:picoclaw /home/picoclaw
 
+WORKDIR /home/picoclaw
+
 # Health check
 # BusyBox (Alpine default) already provides wget, no extra package needed
 # Consider replacing with application-level health command in the future.


nikolasdehor (approved):
Solid improvement to the Docker setup. The sha256 pinning, non-root user, and image size reduction are all good practices.

A few observations:

1. **`/etc/passwd` overwrite** — `COPY --from=builder /etc/passwd_picoclaw /etc/passwd` replaces the entire `/etc/passwd` in the runtime image, removing `root`, `nobody`, and other system users. This can break processes that need to resolve UIDs (e.g., `crond`, `syslogd`, or anything `su`/`setuid`). A safer approach is to `COPY --from=builder` the passwd/group files under a different name and then `cat >> /etc/passwd` to append, or just run the `adduser` in Stage 2 directly (it's a single layer either way).

2. **`vendor/` in `.gitignore`** — This is fine if the project doesn't vendor dependencies, but worth noting in the PR description since it's a separate concern from Docker changes.

3. The `PICOCLAW_CONFIG_FILE` env var in docker-compose is a nice touch for flexibility.

Minor point on (1) aside, LGTM. The overall direction is correct and the size reduction from ~48MB to ~39MB is meaningful.

fabgolast:
PicoClaw need somehow a host system? Like alpine? if it is serving http, you should not need alpine at all, I would adopt scratch, second option distroless.

pixel365:
> PicoClaw need somehow a host system? Like alpine? if it is serving http, you should not need alpine at all, I would adopt scratch, second option distroless.

Alpine is only required because the image currently runs picoclaw onboard at build time to initialize config and directories.
If initialization is moved to application startup in the future, we can safely switch the runtime stage to scratch (or distroless), since the service itself does not require a full Alpine base.

PR #471, opened 2026-02-19
Add custom OpenAI-compatible provider support

cuandada:
### Changes
- **Configuration** (`pkg/config/config.go`)
  - Added `Custom` field to `ProvidersConfig` struct
  - Initialized `Custom` default value in `DefaultConfig()` function

- **Provider Factory** (`pkg/providers/factory.go`)
  - Added explicit `"custom"` provider configuration support in `resolveProviderSelection` function
  - Added custom provider auto-detection in fallback logic

### Usage
Add the following to your configuration file:

```json
{
  "agents": {
    "defaults": {
      "provider": "custom",
      "model": "your-model-name"
    }
  },
  "providers": {
    "custom": {
      "api_key": "your-api-key",
      "api_base": "https://api.your-provider.com/v1",
      "proxy": ""
    }
  }
}

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -263,6 +263,7 @@ type ProvidersConfig struct {
 	ShengSuanYun  ProviderConfig       `json:"shengsuanyun"`
 	DeepSeek      ProviderConfig       `json:"deepseek"`
 	GitHubCopilot ProviderConfig       `json:"github_copilot"`
+	Custom        ProviderConfig       `json:"custom"`
 }
 
 type ProviderConfig struct {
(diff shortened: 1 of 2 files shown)


nikolasdehor (changes requested):
The explicit `case "custom":` path is fine, but the auto-detection fallback has a **precedence bug**:

The `custom` case is checked **first** in the fallback switch and only requires `api_base != ""` (no model name heuristic). This means it shadows ALL subsequent auto-detection. If a user has `custom.api_base` set alongside other providers, custom always wins — even for models like `gpt-4o` or `claude-opus-4-6`.

Fix: either move `custom` to the **end** of the fallback chain, or remove it from auto-detection entirely (only trigger when `provider: "custom"` is explicitly set).

Also:
- How does this differ from the existing `vllm` provider, which already supports custom OpenAI-compatible endpoints?
- Missing tests and example config entry

PR #477, opened 2026-02-19
fix: use Anthropic SDK when api_key is configured

horsley:
When provider=anthropic and api_key is set in config, use the Anthropic SDK instead of generic HTTP provider to avoid 404 errors on /v1/chat/completions.


horsley:
#444 is better

PR #478, opened 2026-02-19
fix (security): ExecTool `working_dir` sandbox escape

GoCeylan:
@Leeaandrob

ExecTool's `working_dir` parameter bypassed `restrict_to_workspace`, allowing a prompt injection attacker to shift the sandbox boundary to an arbitrary path (e.g. `/etc`, `/`). This patch validates `working_dir` against the workspace via logical path checks and symlink resolution before accepting it.                                                        

  🗣️  Type of Change

  🐞 Bug fix (non-breaking change which fixes an issue)

  🤖 AI Code Generation

  👨‍💻 Mostly Human-written (Human lead, AI assisted or none)

  🔗 Related Issue

  Security audit finding, no existing issue.

  📚 Technical Context (Skip for Docs)

  - Reference URL: N/A
  - Reasoning: ExecTool.Execute() in pkg/tools/shell.go accepts an optional working_dir argument from the LLM. The existing guardCommand() checks paths relative to cwd, but since cwd itself was attacker-controlled, all relative-path commands became exploitable. The fix reuses the existing isWithinWorkspace() function (from filesystem.go, same package) and adds filepath.EvalSymlinks() to also catch symlink-based escapes.

  🧪 Test Environment

  - Hardware: MacBook (Apple Silicon)
  - OS: macOS (Darwin 23.3.0)
  - Model/Provider: N/A (code-level fix, no LLM provider needed)
  - Channels: N/A (affects all channels , the vulnerability is in the shared exec tool)

  📸 Evidence (Optional)

  Before fix: exec with working_dir: "/etc" and command: "cat passwd" succeeds — guardCommand() sees passwd as relative to
  /etc and allows it.

  After fix: Same call returns "Command blocked by safety guard (working_dir outside workspace)".

  ☑️  Checklist

  - My code/docs follow the style of this project.
  - I have performed a self-review of my own changes.
  - I have updated the documentation accordingly.

### DIFF
--- a/pkg/tools/shell.go
+++ b/pkg/tools/shell.go
@@ -144,29 +144,14 @@ func (t *ExecTool) Execute(ctx context.Context, args map[string]interface{}) *To
 	cwd := t.workingDir
 	if wd, ok := args["working_dir"].(string); ok && wd != "" {
 		if t.restrictToWorkspace && t.workingDir != "" {
-			absWD, err := filepath.Abs(wd)
+			resolvedWD, err := validatePath(wd, t.workingDir, true)
 			if err != nil {
-				return ErrorResult("invalid working_dir path")
+				return ErrorResult("Command blocked by safety guard (" + err.Error() + ")")
 			}
-			absWorkspace, err := filepath.Abs(t.workingDir)
-			if err != nil {
-				return ErrorResult("failed to resolve workspace path")
-			}
-			if !isWithinWorkspace(absWD, absWorkspace) {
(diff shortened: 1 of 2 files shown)


mengzhuo:
I think we need a test to show whether is it true that file.link escaped

GoCeylan:
@mengzhuo  Thank you. Added tests to verify that working_dir cannot escape the workspace directly.

Also added  ycclffs. suggestion in discord:  reusing an internal security package for validating paths

PR #479, opened 2026-02-19
fix(security): ensure custom deny patterns extend defaults instead of replacing them

GoCeylan:
@Leeaandrob 

 Custom deny patterns in ExecTool replace the default deny patterns instead of supplementing them. When a user configures  even one custom deny pattern, all 40+ default safety guards (rm -rf, sudo, eval, fork bombs, etc.) are silently dropped. 

This patch makes custom patterns additive — defaults are always loaded first, custom patterns are appended on top.         
                                                                                                                           
  🗣️  Type of Change

  - 🐞 Bug fix (non-breaking change which fixes an issue)

  🤖 AI Code Generation

  - 👨‍💻 Mostly Human-written (Human lead, AI assisted or none)

  🔗 Related Issue

  Security audit finding, no existing issue.

  📚 Technical Context (Skip for Docs)

  - Reference URL: N/A
  - Reasoning: In NewExecToolWithConfig() (pkg/tools/shell.go), the if len(execConfig.CustomDenyPatterns) > 0 branch built the deny list from only the custom patterns, with the else branch loading defaults. This meant the two were mutually exclusive. The fix loads defaults unconditionally when deny patterns are enabled, then appends any custom patterns after.

  🧪 Test Environment

  - Hardware: MacBook (Apple Silicon)
  - OS: macOS (Darwin 23.3.0)
  - Model/Provider: N/A (code-level fix, no LLM provider needed)
  - Channels: N/A (affects all channels — the vulnerability is in the shared exec tool)

  📸 Evidence (Optional)

  Before fix: Setting custom_deny_patterns: ["\bmy-command\b"] results in only 1 deny pattern — all defaults removed.

  After fix: Same config results in 41 deny patterns — 40 defaults + 1 custom.

  ☑️  Checklist

  - My code/docs follow the style of this project.
  - I have performed a self-review of my own changes.
  - I have updated the documentation accordingly.

### DIFF
--- a/pkg/tools/shell.go
+++ b/pkg/tools/shell.go
@@ -80,7 +80,7 @@ func NewExecToolWithConfig(workingDir string, restrict bool, config *config.Conf
 		execConfig := config.Tools.Exec
 		enableDenyPatterns = execConfig.EnableDenyPatterns
 		if enableDenyPatterns {
-			denyPatterns = append(denyPatterns, defaultDenyPatterns...)  
+			denyPatterns = append(denyPatterns, defaultDenyPatterns...)
 			if len(execConfig.CustomDenyPatterns) > 0 {
 				fmt.Printf("Using custom deny patterns: %v\n", execConfig.CustomDenyPatterns)
 				for _, pattern := range execConfig.CustomDenyPatterns {


nikolasdehor (approved):
Real security bug, clean fix. Custom deny patterns silently replacing defaults is a dangerous footgun. Defaults are now unconditionally loaded when EnableDenyPatterns is true, and custom patterns append on top. LGTM.

PixelTux (approved):
Precise fix  and doesn't introduce unnecessary complexity. LGTM

xiaket (approved):
LGTM, thanks for the fix!

Issue #425, opened 2026-02-18
I think i ported PicoClaw to Windows-x86? (I know, this is not an issue, but hear me out)


venkatram-s:
> ⚠️ **CAUTION** ⚠️ - This modified version of PicoClaw was vibe-coded, so be careful.

Thanks to vibe coding, some curiosity, and patience, I managed to port PicoClaw to a low-end laptop, and it works (I guess).

I initially tried compiling directly on the target machine, which turned out to be *really* slow, so most of the heavy lifting was done on a faster computer.

---

### Target System
- **OS:** Windows 10 x86 (Build 1803)
- **Architecture:** 32-bit (x86)
- **Storage:** 32 GB
- **RAM:** 2 GB

---

### What I Used
- Latest PicoClaw source code release — **v0.1.2**
- Go 1.26.0 (x86)
- gcc-15.2.0-mingw-w64ucrt-13.0.0-r5 (from winlibs.org)
- A faster computer to compile PicoClaw
- **For coding:** Claude, GitHub Copilot, and ChatGPT

---

### Attachments
If you want to try this yourself, I’ve attached the following:
- **Source Code** (ZIP, modified) - [Link](https://github.com/venkatram-s/picoclaw/releases/download/latest/picoclaw-0.1.2.zip)
- **EXE** (x86) - [Link](https://github.com/venkatram-s/picoclaw/releases/download/latest/picoclaw.exe)

---

### Usage Notes
- To run this properly, add `picoclaw.exe` to your **PATH** under **System Variables**.
- After that, you can run it by typing `./picoclaw` from **any directory** in a terminal.

---

### Notes
- This is an **experimental** port
- No extensive testing was done
- Stability and performance are not guaranteed
- Proceed at your own risk

Finally, thank you devs for putting that termux guide.

venkatram-s:
To @prahladyeri :

> This runs on that one laptop which you wrote about in your githib blog, check it out!

sky5454:
u cloud refer to https://github.com/sipeed/picoclaw/pull/484 ,the build.bat it the windows script to build x64 program and no need any other environment(no need makefile)

nikolasdehor:
[...]

PR #484, opened 2026-02-19
feat(config): support viper multi-format config loading and CLI confi…

sky5454:
…g override

- replace json-only load/save with viper-based multi-format support (json/yaml/toml/ini)

- add config path resolution for ~/.picoclaw with extension fallback and explicit --config/-c override

- keep agent model backward compatibility (string/object forms) via decoder hooks

- add regression tests for yaml/json loading, provider api_key decode, and config path resolution

- add config examples and update docs across README translations

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

- **Reference URL:**
- **Reasoning:**

## 🧪 Test Environment
- **Hardware:** x86 machine 
- **OS:** Windows11  
- **Model/Provider:** GPT5.3-Codex
- **Channels:** 

## 📸 Evidence (Optional)
<details>

</details>

## ☑️ Checklist

### DIFF
--- a/README.zh.md
+++ b/README.zh.md
@@ -3,7 +3,7 @@
 
 <h1>PicoClaw: 基于Go语言的超高效 AI 助手</h1>
 
-<h3>10$硬件 · 10MB内存 · 1秒启�?· 皮皮虾，我们走！</h3>
+<h3>10$硬件 · 10MB内存 · 1秒启动 · 皮皮虾，我们走！</h3>
 
   <p>
     <img src="https://img.shields.io/badge/Go-1.21+-00ADD8?style=flat&logo=go&logoColor=white" alt="Go">
@@ -19,9 +19,9 @@
 
 ---
 
-🦐 **PicoClaw** 是一个受 [nanobot](https://github.com/HKUDS/nanobot) 启发的超轻量级个�?AI 助手。它采用 **Go 语言** 从零重构，经历了一个“自举”过程——即�?AI Agent 自身驱动了整个架构迁移和代码优化�?
+🦐 **PicoClaw** 是一个受 [nanobot](https://github.com/HKUDS/nanobot) 启发的超轻量级个人 AI 助手。它采用 **Go 语言** 从零重构，经历了一个“自举”过程——即由 AI Agent 自身驱动了整个架构迁移和代码优化。
 
-⚡️ **极致轻量**：可�?**10 美元** 的硬件上运行，内存占�?**<10MB**。这意味着�?OpenClaw 节省 99% 的内存，�?Mac mini 便宜 98%�?
+⚡️ **极致轻量**：可在 **10 美元** 的硬件上运行，内存占用 **<10MB**。这意味着比 OpenClaw 节省 99% 的内存，比 Mac mini 便宜 98%！
 
 <table align="center">
 <tr align="center">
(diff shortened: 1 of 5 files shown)


sky5454:
This is the way the command line specifies configuration files for the time being, and it is better for the owner to support better uniform parameter parsing in the future.

build.bat
```diff
@@ -0,0 +1,55 @@
+@echo off
```
lesichkovm, line 1: This is Windows only, and serves no real purpose.
There is already a OS agnostic Makefile which does this job and much more.

lesichkovm (changes requested):

build.bat
sky5454: https://github.com/sipeed/picoclaw/issues/425
In fact, not everyone wants to build in a Linux or MINGW environment. golang don't need makefile.

sky5454:
> Interesting approach using viper for multi-format config support. A few observations:
> 
> **Positives:**
> 
> * The `ResolveConfigPath` function with extension fallback is well-designed — it tries json/yaml/toml/ini in order.
> * The `--config/-c` CLI flag is a good addition for explicit path override.
> * Backward compatibility is maintained via the `AgentModelConfig` decoder hook that handles both string and object forms.
> * Good test coverage: `loadconfig_agentmodel_test.go`, `path_resolution_test.go`, `viper_config_test.go`.
> * The `build.bat` for Windows is a nice bonus.
> 
> **Concerns:**
> 
> 1. **New dependency weight**: Adding `spf13/viper` + `go-viper/mapstructure/v2` is a significant dependency addition for a project that aims to be ultra-lightweight (<10MB). Viper pulls in a large transitive dependency tree (pflag, fsnotify, cast, etc.). Has the binary size impact been measured?
> 2. **`MarshalJSON` removal**: The `AgentModelConfig.MarshalJSON` method was deleted, but `SaveConfig` still needs to produce valid output. With viper handling the write path, does saving a config that was loaded from YAML produce correct JSON output if the user has a `config.json`? The format detection in `SaveConfig` based on `filepath.Ext(path)` looks correct, but this is worth testing explicitly.
> 3. **INI format limitations**: INI does not natively support nested structures or arrays. How does viper handle `allow_from` lists and nested `agents.defaults` in INI format? The example file uses flat keys like `allow_from=YOUR_USER_ID` but the real config expects `[]string`. This might silently produce incorrect config.
> 4. **`mapstructure:"model_fallbacks,omitempty"`**: The `omitempty` tag on mapstructure fields is not standard — mapstructure does not support `omitempty`. This will be silently ignored but is misleading.
> 5. **Deleted `config_test.go` tests**: 24 lines of tests were deleted. Are the replacement tests in the new test files covering the same cases?
> 
> Would like to see the INI array handling tested explicitly before merging.

1. I compiled it smaller than release. `D:\0Path\picoclaw-windows-amd64.exe` is downloaded from release, `build\picoclaw-windows-amd64.exe` is mine.
image

2. SaveConfig golang test is added, TestSaveConfig_AllSupportedFormats
3. ini handle by viper, by usine go-ini to parse, the format like `datas = 1,2,3` Split , but need external library go-ini, Considering that ini is rarely used in modern component scenarios, I would remove it.  
4. omitempty is need by viper, docs: [hdr-Omit_Empty_Values](https://pkg.go.dev/github.com/go-viper/mapstructure/v2@v2.4.0#hdr-Omit_Empty_Values)
5. the deleted line has cover by TestLoadConfig_AgentModelConfigStringFormat and TestLoadConfig_AgentModelConfigObjectFormat at pkg\config\loadconfig_agentmodel_test.go
new commit 6a5e1bea6d81ce2be214de7cb6130e87878e5d75.

build.bat
lesichkovm: Makefile works on Windows, and already does what the bat file does.
If you do no know how here is guide: https://stackoverflow.com/questions/2532234/how-to-run-a-makefile-in-windows

sky5454: I know, but you should reply to him, not to me. a lot of people don't want to see complicate things, and the go build actually supports multi-module, multi-configuration builds. Go doesn't need a makefile at all. a better change is go-task.

You could remove build.bat while merge, I have no objection.

sky5454:
I have merged the main branch of the code and resolved the Viper configuration read and write and all related issues such as code conflicts, please join as soon as possible to avoid config conflict again.

lxowalle:
Hi @sky5454, thanks for the pr. I tested it and the firmware size increased by about 100KB, and memory consumption increased by about 7KB (tested with /usr/bin/time -v). From a resource usage perspective, supporting multiple configuration file formats doesn't seem necessary. What do you think?

lesichkovm (approved):

edouard-claude:
Interesting discussion! Following @lxowalle's point — would something like `gopkg.in/yaml.v3` with a simple format detector be enough to cover the YAML use case, without the full Viper dependency? Or are there other Viper features (env var binding, hot reload) that the community actually needs?

build.bat
venkatram-s: Hi @sky5454, OP of #425 here, thanks for mentioning my issue here, while i was aware of the fact that Go can indeed build exe's on it's own without a makefile. To be honest, infact it was the first method i tried, and i was more obsessed with the idea of building with mingw, i will update it ASAP.

sky5454:
According to the vote of the community on x, json has the most votes but still does not exceed half, followed by yaml, toml, and all support. Personally, I believe that if there is no majority, it means that other formats are still important. In the web era, YAML/toml is the mainstream configuration, and YAML is the de facto standard in Java enterprise projects. In order for picoclaw to develop in the future, as an embedded infrastructure system in the cloud, I think it is necessary to support multiple formats.


And I think we should learn more about users' usage scenarios and expectations.

lxowalle:
The result of the last developer meeting was to keep it as JSON, so I will close this PR for now. Of course, if you think it is still necessary to add other formats, we can discuss it further.

PR #490, opened 2026-02-19
feat(observability): add ToolObserver hook for structured tool tracing

SebastianBoehler:
## Summary

Adds a lightweight observer pattern to `ToolRegistry` that fires after every tool execution. This enables external observability integrations (W&B Weave, OpenTelemetry, custom metrics) without modifying individual tools.

### `pkg/tools/registry.go`
- Add `ToolObserver` func type: `func(name string, args map[string]interface{}, result *ToolResult, durationMs int64)`
- Add `observer` field to `ToolRegistry`
- Add `SetObserver()` method
- Fire observer in `ExecuteWithContext` after every tool call

### `pkg/agent/loop.go`
- Add `SetToolObserver()` to `AgentLoop` — propagates observer to all agent tool registries

### `cmd/picoclaw/main.go`
- Wire observer in `agentCmd` when `PICOCLAW_WEAVE_OBSERVE=1`
- Emits `WEAVE_TOOL_EVENT:` JSON lines to stderr for downstream consumption

## Design

The observer is a **zero-cost abstraction** when not set — a single nil check before the call. No allocations, no goroutines.

The `WEAVE_TOOL_EVENT:` line format emitted to stderr:
```
WEAVE_TOOL_EVENT:{"tool":"exec","duration_ms":42,"is_error":false,"args":{...}}
```

This is intentionally transport-agnostic — any process reading the agent's stderr can parse these lines. The W&B Weave integration reads them from the agent subprocess and logs structured traces per run, but the hook itself has no Weave dependency.

## Usage

```go
// Custom observer — log to any backend
agentLoop.SetToolObserver(func(name string, args map[string]interface{}, result *tools.ToolResult, durationMs int64) {
    myTracer.RecordToolCall(name, durationMs, result.IsError)
})
```

Or via env var for the built-in stderr emitter:
```bash
PICOCLAW_WEAVE_OBSERVE=1 picoclaw agent -m "your prompt"
```

### DIFF
--- a/cmd/picoclaw/cmd_agent.go
+++ b/cmd/picoclaw/cmd_agent.go
@@ -6,6 +6,7 @@ package main
 import (
 	"bufio"
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"os"
@@ -18,6 +19,7 @@ import (
 	"github.com/sipeed/picoclaw/pkg/bus"
 	"github.com/sipeed/picoclaw/pkg/logger"
 	"github.com/sipeed/picoclaw/pkg/providers"
+	"github.com/sipeed/picoclaw/pkg/tools"
 )
 
 func agentCmd() {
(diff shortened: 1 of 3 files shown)


SebastianBoehler:
Removed conflicting features

PR #497, opened 2026-02-19
fix: harden tool call extraction scanner to handle braces in strings

RahulChand028:
## 📝 Description

This PR hardens the tool call extraction logic by implementing a robust, state-aware findMatchingBrace scanner.

The Problem: The previous logic used a simple brace counter that would fail if a closing brace } appeared inside a JSON string or was escaped (common in complex tool arguments like grep commands or regex).
The Solution: Implemented a new scanner that correctly handles strings and escapes.
Centralization: Moved the scanner to a shared utility in pkg/providers to ensure all current and future CLI-based providers benefit from these improvements.

## 🗣️ Type of Change
- [X] 🐞 Bug fix (non-breaking change which fixes an issue)
- [ ] 📖 Documentation update

## 🤖 AI Code Generation
- [X] 🛠️ Mostly AI-generated (AI draft, Human verified/modified)

- **Reference URL:**
- **Reasoning:** Autonomous agents frequently process code or shell commands containing braces. A naive brace counter causes tool call truncation, leading to JSON parsing errors and agent execution loops. This fix ensures reliability for complex tasks.

## 🧪 Test Environment
- **Hardware:** PC
- **OS:** Linux (Ubuntu/Debian)
- **Model/Provider:** Verified via Go Unit Tests and Claude CLI mock
- **Channels:** CLI

## 📸 Evidence (Optional)
<details>
image

</details>

## ☑️ Checklist
- [X] I have updated the documentation accordingly.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,7 +2,7 @@ name: build
 
 on:
   push:
-    branches: ["main"]
+    branches: [ "main" ]
 
 jobs:
   build:
@@ -15,11 +15,6 @@ jobs:
         uses: actions/setup-go@v6
         with:
           go-version-file: go.mod
-
-      - name: fmt
-        run: |
-          make fmt
-          git diff --exit-code || (echo "::error::Code is not formatted. Run 'make fmt' and commit the changes." && exit 1)
 
       - name: Build
         run: make build-all
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -24,48 +24,9 @@ jobs:
         with:
           version: v2.10.1
 
-  # TODO: Remove once linter is properly configured
-  fmt-check:
-    name: Formatting
-    runs-on: ubuntu-latest
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v6
-
-      - name: Setup Go
-        uses: actions/setup-go@v6
-        with:
-          go-version-file: go.mod
-
-      - name: Check formatting
-        run: |
-          make fmt
-          git diff --exit-code || (echo "::error::Code is not formatted. Run 'make fmt' and commit the changes." && exit 1)
-
-  # TODO: Remove once linter is properly configured
-  vet:
-    name: Vet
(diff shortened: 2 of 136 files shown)


nikolasdehor (approved):
Good fix for a real problem. The naive brace counter was a known weakness — LLM outputs containing braces in strings (grep patterns, regex, code snippets) would truncate the tool call JSON, causing parse failures and agent loops.

The new `findMatchingBrace` scanner correctly handles:
- Braces inside JSON strings (`inString` flag)
- Escaped characters including escaped quotes (`escaped` flag)
- Nested objects (depth counter)

The test cases cover the key scenarios including braces in strings and escaped quotes.

A few observations:

1. **Backslash handling outside strings** — The `escaped` flag triggers on any `\\` character, even outside strings. If the text surrounding the JSON has a bare backslash before a `{`, it would incorrectly skip that brace. In practice this is unlikely since the text is LLM output, but technically the escape handling should only apply inside strings.

2. **Performance change** — The old `extractToolCallsFromText` did a single `strings.Index` to find `{"tool_calls"`, then found the matching brace. The new version scans every `{` in the text, calls `findMatchingBrace` (which re-scans forward), extracts the substring, and does a `strings.Contains` check. For long outputs with many `{` characters (e.g., code blocks), this is O(n * k) where k is the number of opening braces. The old approach was O(n). For typical LLM output lengths this is fine, but worth noting.

3. **`stripToolCallsFromText` duplicated logic** — The `extractToolCallsFromText` and `stripToolCallsFromText` functions share nearly identical scanning logic. Consider extracting a common `findToolCallJSON(text string) (start, end int, found bool)` helper to DRY it up.

4. **Test for the old deleted test case** — The old `TestFindMatchingBrace` test had a case for `{"a":"b{c}d"}` (braces in strings) which returned 13 with a comment "simplified matcher". The new scanner should still pass this — it would be good to have it in the new test suite to explicitly verify the fix.

Overall this is a meaningful correctness improvement. LGTM.

PR #501, opened 2026-02-19
feat: support Gemini 3.x thought_signature in tool calls

talhamsajid:
Gemini 3.x models require a thought_signature to be returned in subsequent conversation turns when using tool calls. This patch adds the necessary fields to ToolCall protocol types and updates the OpenAI-compatible provider and agent loop to capture and preserve these signatures. This fixes the 400 Bad Request errors when using models like gemini-3.1-pro-preview with tools.


nikolasdehor (approved):
Targeted fix for Gemini 3.x tool call compatibility. The changes are minimal and well-scoped.

The approach of threading `ExtraContent` through the protocol types is clean — it avoids Gemini-specific hacks in the generic code path.

A few notes:

1. **Serialization round-trip** — The `extra_content` field is added to the `ToolCall` struct in protocoltypes with `json:"extra_content,omitempty"`. This means when the assistant message with tool calls is serialized back to the provider (in subsequent turns), the `extra_content` with the `thought_signature` will be included. This is exactly what Gemini 3.x requires — the thought signature must be echoed back.

2. **Other providers unaffected** — Since `ExtraContent` is `omitempty` and will be `nil` for non-Gemini providers, this adds no overhead or behavioral change for OpenAI/Anthropic/etc.

3. **No tests** — Understandable for a protocol fix like this (would require mocking Gemini 3.x responses), but a simple unit test for `parseResponse` with an `extra_content` field in the JSON would strengthen confidence.

4. **Streaming path** — This only modifies `parseResponse` (non-streaming). If the project also has a streaming response parser for OpenAI-compat, the same `extra_content` extraction would need to be added there too. Worth checking.

LGTM for the non-streaming path.

PR #512, opened 2026-02-20
fix(web_fetch): include fetched page content in ForLLM response

### LINKED ISSUE #388: web_fetch tool: ForLLM response missing actual page content
## Bug Description

The `web_fetch` tool's `ForLLM` field only returns a metadata summary line, not the actual fetched content. This means the LLM can never see the web page content it fetched, making the tool effectively useless.

## Current Behavior

`ForLLM` returns:
```
Fetched 5000 bytes from https://example.com (extractor: text, truncated: false)
```

The LLM sees only this summary — **no actual page content**.

Meanwhile, `ForUser` contains the full JSON result including the extracted text, which is displayed to the end user but invisible to the LLM.

## Expected Behavior

`ForLLM` should include the extracted text content so the LLM can read and reason about the fetched page.

## Impact

- The LLM repeatedly calls `web_fetch` on the same URL because it never receives the content, causing a tool call loop until `max_tool_iterations` is hit.
- The agent cannot answer any question that requires reading web content.

## Root Cause

In `pkg/tools/web.go`, line ~479:

```go
return &ToolResult{
    ForLLM:  fmt.Sprintf("Fetched %d bytes from %s (extractor: %s, truncated: %v)", len(text), urlStr, extractor, truncated),
    ForUser: string(resultJSON),
}
```

`ForLLM` only formats the metadata, omitting the `text` variable that contains the actual extracted content.

## Fix

Include the extracted text in `ForLLM`:

```go
forLLM := fmt.Sprintf("Fetched %d bytes from %s (extractor: %s, truncated: %v)\n\n%s", len(text), urlStr, extractor, truncated, text)

return &ToolResult{
    ForLLM:  forLLM,
    ForUser: string(resultJSON),
}
```

CrisisAlpha:
## 📝 Description

The `web_fetch` tool's `ForLLM` field only returned a metadata summary line (byte count, extractor type, truncation status), omitting the actual extracted page content. This meant the LLM could never read fetched web pages, causing it to loop calling `web_fetch` repeatedly until `max_tool_iterations` was hit.

The fix appends the extracted text content to the `ForLLM` response so the LLM can read and reason about fetched pages.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue

Fixes #388

- **Reference URL:** https://github.com/sipeed/picoclaw/issues/388
- **Reasoning:** `ForLLM` is what the LLM sees as the tool result. Without the actual content, the tool provides no value and causes the LLM to loop indefinitely.

## 🧪 Test Environment
- **Hardware:** Mac (x86_64)
- **OS:** macOS
- **Model/Provider:** N/A (unit tests only, no API keys needed)
- **Channels:** N/A

## 📸 Evidence (Optional)
<details>

**Before (broken):** `ForLLM` returns only:
```
Fetched 5000 bytes from https://example.com (extractor: text, truncated: false)
```

**After (fixed):** `ForLLM` returns metadata + actual content:
```
Fetched 5000 bytes from https://example.com (extractor: text, truncated: false)

[actual page content here]
```

All tests pass: `make fmt`, `make vet`, `make test` ✓

</details>

## ☑️ Checklist

Made with [Cursor](https://cursor.com)

### DIFF
--- a/pkg/tools/web.go
+++ b/pkg/tools/web.go
@@ -477,7 +477,7 @@ func (t *WebFetchTool) Execute(ctx context.Context, args map[string]interface{})
 	resultJSON, _ := json.MarshalIndent(result, "", "  ")
 
 	return &ToolResult{
-		ForLLM:  fmt.Sprintf("Fetched %d bytes from %s (extractor: %s, truncated: %v)", len(text), urlStr, extractor, truncated),
+		ForLLM:  fmt.Sprintf("Fetched %d bytes from %s (extractor: %s, truncated: %v)\n\n%s", len(text), urlStr, extractor, truncated, text),
 		ForUser: string(resultJSON),
 	}
 }
--- a/pkg/tools/web_test.go
+++ b/pkg/tools/web_test.go
@@ -36,9 +36,14 @@ func TestWebTool_WebFetch_Success(t *testing.T) {
 		t.Errorf("Expected ForUser to contain 'Test Page', got: %s", result.ForUser)
 	}
 
-	// ForLLM should contain summary
-	if !strings.Contains(result.ForLLM, "bytes") && !strings.Contains(result.ForLLM, "extractor") {
-		t.Errorf("Expected ForLLM to contain summary, got: %s", result.ForLLM)
+	// ForLLM should contain summary metadata
+	if !strings.Contains(result.ForLLM, "bytes") || !strings.Contains(result.ForLLM, "extractor") {
+		t.Errorf("Expected ForLLM to contain summary metadata, got: %s", result.ForLLM)
+	}
+
+	// ForLLM should contain the actual fetched page content
+	if !strings.Contains(result.ForLLM, "Test Page") {
+		t.Errorf("Expected ForLLM to contain actual page content 'Test Page', got: %s", result.ForLLM)
 	}
 }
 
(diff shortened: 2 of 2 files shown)


lxowalle:
Hi @CrisisAlpha, could you try again on the latest branch? I've tested and confirmed that this issue has been fixed on the latest branch.

CrisisAlpha:
Hey @lxowalle, thanks for checking! Just synced with the latest main (cb0c870) and the issue appears to be present. In [`pkg/tools/web.go` L486-492](https://github.com/sipeed/picoclaw/blob/main/pkg/tools/web.go#L486-L492), `ForLLM` only contains the metadata summary (byte count, extractor, truncated flag). The actual fetched `text` is only included in ForUser as JSON.

Since the [agent loop](https://github.com/sipeed/picoclaw/blob/main/pkg/agent/loop.go#L706-L713) sends ForLLM (not ForUser) to the LLM as the tool result, the LLM never receives the page content. The user sees it in the chat, but the LLM can't reason about it.

Could you double-check? Happy to rebase if needed!

nikolasdehor:
@lxowalle I can confirm CrisisAlpha's analysis is correct -- the bug is still present on main.

Looking at the code path:

1. In pkg/tools/web.go L486-492, ForLLM is set to only the metadata summary line. The actual text content is only serialized into ForUser as part of the JSON result.

2. In pkg/agent/loop.go L706-713, the agent loop sends ForLLM (not ForUser) to the LLM as the tool result. This is the correct design -- ForUser is for display to the human, ForLLM is what the model reasons about.

The net effect is that the LLM receives a metadata-only string with zero page content. The LLM has no way to read the fetched page, which causes it to loop calling web_fetch repeatedly until max_tool_iterations is exhausted.

This PR's fix (appending text to ForLLM) is the correct approach. The ForUser/ForLLM split is intentional architecture, and the content simply needs to be in both.

lxowalle:
Okay, I'll double-check.

lxowalle:
Hi @CrisisAlpha , I tested and confirmed that the message for user can be passed to the channel, but it cannot be passed to the LLM. I suggest modifying the method to:
```go
	return &ToolResult{
		ForLLM: fmt.Sprintf(
			"Fetched %d bytes from %s (extractor: %s, truncated: %v): %s",
			len(text),
			urlStr,
			extractor,
			truncated,
			string(resultJSON),
		),
		ForUser: "",
	}
```
This ensures that the message will not be sent to the channel repeatedly.

lxowalle:
https://github.com/sipeed/picoclaw/pull/833 fixed

PR #524, opened 2026-02-20
Use strings.Builder instead of += concatenation in loops

mattn:
## 📝 Description

Replace string concatenation with strings.Builder in:
- LoadBootstrapFiles (context.go)
- GetRecentDailyNotes, GetMemoryContext (memory.go)
- formatMessagesForLog, formatToolsForLog, summarizeBatch (loop.go)

Reduces allocations from O(n^2) to O(n) for repeated appends.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

- **Reference URL:**
- **Reasoning:**

## 🧪 Test Environment
- **Hardware:** 
- **OS:** 
- **Model/Provider:** 
- **Channels:** 

## 📸 Evidence (Optional)
<details>

</details>

## ☑️ Checklist

### DIFF
--- a/pkg/agent/memory.go
+++ b/pkg/agent/memory.go
@@ -133,7 +133,6 @@ func (ms *MemoryStore) GetMemoryContext() string {
 	}
 
 	var sb strings.Builder
-	sb.WriteString("# Memory\n\n")
 
 	if longTerm != "" {
 		sb.WriteString("## Long-term Memory\n\n")


nikolasdehor (approved):
Clean refactor. The string concatenation → strings.Builder conversions are correct and well-scoped.

A few notes:

1. **context.go L152** — `fmt.Fprintf(&sb, "...", filename, data)` where `data` is `[]byte`: this works because `%s` handles `[]byte` correctly in Go, so this is fine and avoids the `string(data)` allocation. Good catch.

2. **memory.go GetMemoryContext()** — The refactored version has a subtle behavior change: the old code only prepended `# Memory\n\n` at the end via `fmt.Sprintf`, but the new code writes it to the builder first. Functionally equivalent, just noting it for review.

3. **loop.go** — The `tools` → `toolDefs` rename in `formatToolsForLog` is a nice bonus to avoid shadowing the parameter name with the range variable, though Go wouldn't shadow it here since the types differ. Still cleaner.

LGTM — straightforward performance improvement with no functional changes.

PR #543, opened 2026-02-20
fix: remove premature media file cleanup in channel handlers

DevEverything01:
## Problem

Media files (photos, voice messages, audio, documents) downloaded by channel handlers are immediately deleted via `defer` when `handleMessage` returns. However, `HandleMessage` publishes the message (including media file paths) to an **async message bus** (`PublishInbound`), and the agent consumes it in a separate goroutine.

**Result:** By the time the agent goroutine tries to access the media files, they have already been deleted. Voice/image/document processing is effectively non-functional.

## Root Cause

```go
// handleMessage downloads files to temp paths...
defer func() {
    for _, file := range localFiles {
        os.Remove(file)  // ← deleted BEFORE agent consumes them
    }
}()
// ...then publishes to async bus
c.HandleMessage(...)  // → PublishInbound(msg) → channel send
return nil            // ← defer fires here, files gone
```

## Fix

Remove the premature `defer` cleanup from all 4 affected channels:
- `pkg/channels/telegram.go`
- `pkg/channels/discord.go`
- `pkg/channels/slack.go`
- `pkg/channels/line.go`

Media files live in `os.TempDir()/picoclaw_media/` with UUID-prefixed names, managed by the OS temp directory lifecycle.

## Verification

- `go build ./...` ✅
- `go test ./pkg/channels/...` ✅
- `go vet ./...` ✅

### DIFF
--- a/pkg/channels/onebot.go
+++ b/pkg/channels/onebot.go
@@ -569,8 +569,8 @@ func parseJSONString(raw json.RawMessage) string {
 type parseMessageResult struct {
 	Text           string
 	IsBotMentioned bool
-	Media   []string
-	ReplyTo string
+	Media          []string
+	ReplyTo        string
 }
 
 func (c *OneBotChannel) parseMessageSegments(raw json.RawMessage, selfID int64) parseMessageResult {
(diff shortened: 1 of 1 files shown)


ex-takashima:
Thanks for identifying this race condition — the analysis in the PR description is spot on. 👍

I ran into the same issue and built on your findings in #620, which takes a slightly different approach: instead of only removing the `defer` cleanup, it also adds a background `MediaCleaner` that periodically removes files older than 30 minutes from the temp directory. This way we avoid the race condition without risking unbounded temp file accumulation over time.

Would love to hear your thoughts!

alexhoshina:
fixed

PR #554, opened 2026-02-20
Allow vllm provider to work without API key

Loule95450:
Skip the API key requirement check when using the vllm provider, enabling use of OpenAI-compatible endpoints that don't require authentication (e.g. opencode.ai/zen free models).


Loule95450:
Wrong branch, closing to fix on the correct branch

Issue #101, opened 2026-02-13
Web Socket or REST API connecting to external admin/dashboard


Interpoolx:
Any idea whether the current implementation does this? or plan to add this in future

I want to be able to connect to picoclaw gateway instance through external dashboard app to manage the instance config settings, start/stop/restart/update picoclaw, etc... basically, keep the unmodified version installed on one system and the entire dashboard UI hosted on another domain/hosting and able to communicate bidirectionally.

Let me know

Thanks
Rajesh

heyarviind:
I second this

raine:
I'm looking to build an iOS/Apple Watch app that connects to picoclaw over a local network. Is there any plan to add a WebSocket or REST API for external clients?

alexhoshina:
This feature is planned! Thanks for the suggestion.

alexhoshina:
Added

nikolasdehor:
Good to see this was implemented. For anyone looking at external dashboard integration, the gateway already exposes HTTP on the configured host:port. A few notes for dashboard developers:

- The gateway currently uses the message bus (`pkg/bus/bus.go`) with buffered channels (100 messages). For real-time dashboard updates, a WebSocket upgrade endpoint would avoid polling overhead.
- Consider authentication for the management API -- currently the gateway HTTP endpoint has no auth layer, which is fine for localhost but risky if exposed on `0.0.0.0`.
- The `allow_from` enforcement (fixed in recent PRs for channels) should extend to the management API as well.

PR #598, opened 2026-02-21
feat: add HTMX dashboard with auth and CRUD

edouard-claude:
## Summary

Ref #101

Naive attempt at a built-in web dashboard — no strong opinion on this, just a proposal to explore the idea. Happy to hear thoughts or close if this isn't the right direction.

### Dashboard
dashboard-main

### Login
dashboard-login

### Settings
dashboard-settings

## What it does
- Optional dashboard at `/dashboard` (`dashboard.enabled: false` to disable)
- Password-protected with HMAC-SHA256 signed cookie sessions (24h expiry)
- Auto-generates a 16-char password on first launch, saved to config.json
- HTMX-powered with real-time SSE status updates
- CRUD for Agents (with AGENT.md instructions editing), Models, Channels, Settings
- Dark monospace theme, responsive
- HTMX loaded from CDN, no vendored JS
- Restart banner warns when changes need a gateway restart

## Performance

Measured on macOS (Apple Silicon), 3 agents, no channels:

| | Without dashboard | With dashboard | Delta |
|---|---|---|---|
| Binary | 25 MB | 61 MB | +36 MB |
| RSS at idle | 21 MB | 30 MB | +9 MB |

Binary increase comes from the Go linker retaining more SDK symbols when the dashboard references AgentLoop — not from the dashboard code itself (169 symbols, ~20KB static). Fully optional.

## E2E tested
- Login/logout flow, cookie persistence across restarts
- Agent CRUD with AGENT.md written to correct workspace
- Config persistence, restart banner

## Test plan
- [ ] `go test ./pkg/dashboard/...` (55 tests)
- [ ] `go vet ./...`
- [ ] Manual: login, CRUD, disable with config flag

### DIFF
--- a/cmd/picoclaw/cmd_gateway.go
+++ b/cmd/picoclaw/cmd_gateway.go
@@ -193,20 +193,32 @@ func gatewayCmd() {
 			cfg.Dashboard.Password = dashboard.GeneratePassword()
 			configPath := getConfigPath()
 			if saveErr := config.SaveConfig(configPath, cfg); saveErr != nil {
-				logger.ErrorCF("dashboard", "Failed to save generated password", map[string]any{"error": saveErr.Error()})
+				logger.ErrorCF(
+					"dashboard",
+					"Failed to save generated password",
+					map[string]any{"error": saveErr.Error()},
+				)
 			}
 			fmt.Printf("🔑 Dashboard password generated: %s\n", cfg.Dashboard.Password)
 		}
 		dashConfigPath := getConfigPath()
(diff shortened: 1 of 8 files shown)


xiaket:
Thanks for bringing this idea into PicoClaw. At this point in time, we do strive to keep the memory usage under 20M, as you can see [here](https://github.com/sipeed/picoclaw/blob/main/ROADMAP.md#-1-core-optimization-extreme-lightweight). This PR will bring a lot of memory consumption to PicoClaw, hence I'm going to close this.

I do appreciate the effort to bring this to life. Thank you!

PR #600, opened 2026-02-21
feat: bubbletea TUI for picoclaw agent

rbansal42:
## Summary
- Replace the readline-based REPL in picoclaw agent with a full bubbletea TUI
- Add AgentEventListener to the agent loop for real-time tool call/thinking notifications
- New pkg/tui/ package with chat viewport, markdown rendering (glamour), tool call indicators, status bar

## Changes
- pkg/agent/events.go - Event types and listener interface for agent loop notifications
- pkg/tui/ - New package: bubbletea Model with viewport, textarea, lipgloss styles, event bridge
- cmd/picoclaw/cmd_agent.go - Replaced readline REPL with tea.NewProgram(), removed dead code
- One-shot mode (-m flag) is unchanged

## Test Plan
- [x] 11 TUI model unit tests pass
- [x] All 20 pkg/ test suites pass
- [x] Race detector passes (go test -race)
- [x] go vet clean
- [x] Binary builds and picoclaw agent --help works
- [ ] Manual smoke test: run picoclaw agent and verify TUI launches in alt-screen

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -43,6 +43,7 @@ pkg/
     openai_compat/  OpenAI-compatible HTTP provider
   session/          Session persistence (~/.picoclaw/workspace/sessions/*.json)
   tools/            Tool system — one file per tool (exec, web, edit, filesystem, etc.)
+  tui/              Bubbletea terminal UI (chat viewport, markdown, tool indicators)
 ```
 
 ## Code Style
--- a/README.md
+++ b/README.md
@@ -255,7 +255,8 @@ picoclaw onboard
 **4. Chat**
 
 ```bash
-picoclaw agent -m "What is 2+2?"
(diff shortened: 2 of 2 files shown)


Zhaoyikaiii:
Thank you for your work on this. At the moment, we are planning to introduce a separate bootstrap/launcher repository that will communicate with the current Pico project. We believe this separation will help us keep the core system clean while allowing more flexibility for orchestration and distributed features.

Would you be open to contributing this implementation in that new repository instead? We think it may be a better architectural fit there, and we would be happy to collaborate on shaping the direction together.

rbansal42:
Sounds good!

xiaket:
Thanks for your contribution! Closing now per above convo.

PR #604, opened 2026-02-21
fix: preserve reasoning_content for OpenAI-compatible reasoning models

### LINKED ISSUE #588: [Bug] Moonshot kimi-k2.5 reasoning_content missing in tool calls leads to API Error 400
## Quick Summary

When using the Moonshot AI model kimi-k2.5 (a reasoning model) via the picoclaw-gateway, the API returns a 400 Bad Request error during multi-turn conversations or tool calls.

Error Log
Plaintext```
  Status: 400
  Body: {"error":{"message":"thinking is enabled but reasoning_content is missing in assistant tool call message at index 4","type":"invalid_request_error"}}```
Environment
Model: moonshot/kimi-k2.5

PicoClaw Version: Latest (main branch)

winterfx:
Models like Moonshot kimi-k2.5 and DeepSeek-R1 return a reasoning_content field in assistant messages. When thinking is enabled, the API requires this field to be echoed back in subsequent requests. PicoClaw was silently dropping it, causing 400 errors on tool-call round-trips.

- Add ReasoningContent to Message and LLMResponse types
- Parse reasoning_content in openai_compat parseResponse()
- Carry reasoning_content through assistant tool-call messages
- Add unit test for reasoning_content parsing

Fixes #588

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

- **Reference URL:**
- **Reasoning:**reasoning_content is a standard field in OpenAI-compatible APIs for reasoning models. It was completely absent from the codebase — not parsed, not stored, not echoed back.

## 🧪 Test Environment
  - **Hardware:** Mac (Apple Silicon)
  - **OS:** macOS
  - **Model/Provider:** Moonshot kimi-k2.5
  - **Channels:** CLI

## 📸 Evidence (Optional)
<details>

image
after fixed

image

</details>

## ☑️ Checklist

### DIFF
--- a/pkg/agent/loop.go
+++ b/pkg/agent/loop.go
@@ -622,8 +622,9 @@ func (al *AgentLoop) runLLMIteration(
 
 		// Build assistant message with tool calls
 		assistantMsg := providers.Message{
-			Role:    "assistant",
-			Content: response.Content,
+			Role:             "assistant",
+			Content:          response.Content,
+			ReasoningContent: response.ReasoningContent,
 		}
 		for _, tc := range normalizedToolCalls {
 			argumentsJSON, _ := json.Marshal(tc.Arguments)
--- a/pkg/providers/openai_compat/provider.go
+++ b/pkg/providers/openai_compat/provider.go
@@ -148,8 +148,9 @@ func parseResponse(body []byte) (*LLMResponse, error) {
 	var apiResponse struct {
 		Choices []struct {
 			Message struct {
-				Content   string `json:"content"`
-				ToolCalls []struct {
+				Content          string `json:"content"`
+				ReasoningContent string `json:"reasoning_content"`
+				ToolCalls        []struct {
 					ID       string `json:"id"`
 					Type     string `json:"type"`
 					Function *struct {
@@ -221,10 +222,11 @@ func parseResponse(body []byte) (*LLMResponse, error) {
 	}
 
 	return &LLMResponse{
-		Content:      choice.Message.Content,
-		ToolCalls:    toolCalls,
-		FinishReason: choice.FinishReason,
-		Usage:        apiResponse.Usage,
+		Content:          choice.Message.Content,
+		ReasoningContent: choice.Message.ReasoningContent,
+		ToolCalls:        toolCalls,
+		FinishReason:     choice.FinishReason,
+		Usage:            apiResponse.Usage,
 	}, nil
 }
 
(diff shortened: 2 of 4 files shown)


alexhoshina (approved):
LGTM! Ready for @yinwm  to merge.

tinwritescode (approved):

nikolasdehor (approved):
Clean fix for a real problem. Reasoning models (DeepSeek-R1, kimi-k2.5, QwQ) return reasoning_content in assistant messages and require it echoed back on subsequent turns. Without this, tool-call round-trips fail with 400 errors because the API expects the field to be present.

What is correct:
- ReasoningContent field added to both Message and LLMResponse in protocoltypes/types.go with omitempty so it is invisible for non-reasoning models.
- Parsed from the API response in openai_compat/provider.go parseResponse().
- Carried through the assistant message construction in loop.go when tool calls are present -- this is the critical path, since the reasoning content needs to survive the assistant->tool->assistant round-trip.
- The test covers the important case: reasoning content + tool calls in the same response.

Edge cases to consider (non-blocking):
- Streaming responses: If the codebase has a streaming path (SSE), reasoning_content may arrive as a separate delta field. This PR handles the non-streaming path correctly. If streaming is added later, it will need the same treatment.
- Anthropic's thinking field: Anthropic uses a different field name (thinking blocks in the content array). This is correctly scoped to openai_compat only, so no conflict.
- Cost/token implications: reasoning_content can be very large (thousands of tokens of chain-of-thought). Echoing it back on every turn means context grows faster. This is working-as-designed per the API spec, but worth noting for users of reasoning models with limited context windows.

The test is well-structured -- it validates that reasoning content, regular content, and tool calls are all parsed correctly from a single response. Approving.

winterfx:
Hi @yinwm, just a gentle reminder about the PR when you have a moment 🙂
Would really appreciate your review. Thanks!

Orgmar:
@winterfx Good find on reasoning_content being silently dropped. Getting 400 errors on tool-call round-trips with models like kimi-k2.5 must have been frustrating to debug. Carrying the field through the full message chain is the right fix.

We're putting together the **PicoClaw Dev Group** on Discord for contributors to chat and collaborate. If you'd like to join, send an email to `support@sipeed.com` with the subject `[Join PicoClaw Dev Group] winterfx` and we'll get you the invite link.

Issue #572, opened 2026-02-21
[Feature] Add `/new` or `/clear` command to start a fresh session


rahul-hudle:
## Description

There is no way for users to start a fresh conversation without restarting the picoclaw process. Over time, conversations accumulate context and may hit token limits or become confused by stale context.

## Current Behavior

No session management commands exist. The only way to start fresh is to restart the process or manually delete session files.

## Proposed Behavior

Add `/new` (or `/clear`) as a slash command:

```
User: /new
PicoClaw: Starting a new conversation. Previous session saved as session_abc123.
```

Behavior:
1. Save current session to disk (if not already persisted)
2. Create a new empty session
3. Reload system prompts (IDENTITY.md, AGENT.md, SOUL.md, USER.md)
4. Confirm to user

Optionally add `/sessions` to list recent sessions and `/switch session <id>` to resume a previous one.

## Location

- `pkg/agent/loop.go` — add to `handleCommand`
- `pkg/session/manager.go` — session lifecycle methods

## Related

- #574 (`picoclaw sessions` CLI subcommand)

nikolasdehor:
This is a quality-of-life feature that would significantly improve the interactive experience. A few implementation thoughts:

**Session management in `pkg/agent/loop.go`**

The `/new` command should:
1. Flush the current session context (conversation history)
2. Preserve agent configuration (model, tools, skills)
3. Reload personality files (IDENTITY.md, AGENT.md, etc.)
4. Reset the token counter

```go
case "new", "clear":
    sessionID := a.session.ID()
    a.session.Save() // persist current session
    a.session.Reset() // clear conversation history
    a.reloadSystemPrompts()
    return fmt.Sprintf("New session started. Previous session: %s", sessionID), true
```

**Session listing (`/sessions`)**

This depends on how sessions are persisted. If `pkg/session/manager.go` already writes to disk, listing them is straightforward. The output should show:
- Session ID
- Start time
- Message count
- Last model used
- First user message (truncated) as a preview

**Scope consideration**

For the gateway (non-CLI) use case, `/new` per chat is important — each WhatsApp/Telegram/Discord conversation should be independently clearable. The session key already includes the channel + chat_id, so this should be natural.

Related: #574 (`picoclaw sessions` CLI subcommand) covers the CLI-side management. These two could share the same underlying session lifecycle methods.

AIWintermuteAI:
@Zepan 
[...]

PR #613, opened 2026-02-21
Feat: Add ‘/New’ and ‘/Clear’ Commands To Reset Context

jrswab:
## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue

Implements #572 

- **Reference URL:**
- **Reasoning:**

## 🧪 Test Environment
- **Hardware:** VPS with 2vpcu and 8GiB Ram
- **OS:** Ubuntu 24.04 
- **Model/Provider:** Kimi k2.5
- **Channels:** Telegram

## 📸 Evidence (Optional)
<details>

Screenshot from 2026-02-21 17-57-26

</details>

## ☑️ Checklist

### DIFF
--- a/pkg/agent/loop.go
+++ b/pkg/agent/loop.go
@@ -462,7 +462,64 @@ func (al *AgentLoop) startNewSessionForMessage(msg bus.InboundMessage) (string, 
 	overrideKey := buildSessionOverrideKey(msg.Channel, msg.ChatID, agent.ID)
 	al.sessionOverride.Set(overrideKey, newSessionKey)
 
+	// Clean up the old session file from disk to prevent unbounded accumulation.
+	if sessionKey != "" {
+		if err := agent.Sessions.Delete(sessionKey); err != nil {
+			logger.WarnCF("agent", "Failed to delete old session file", map[string]any{
+				"session_key": sessionKey,
+				"error":       err.Error(),
+			})
+		}
+	}
+
 	return formatNewSessionResponse(al.cfg, agent), nil
+}
+
+// clearSessionForMessage clears the current session history and summary in-place,
+// keeping the same session key. Unlike startNewSessionForMessage, this does not
+// create a new session key or set an override.
+func (al *AgentLoop) clearSessionForMessage(msg bus.InboundMessage) (string, error) {
+	if constants.IsInternalChannel(msg.Channel) {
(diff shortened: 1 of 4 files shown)


nikolasdehor (commented):
Good feature implementation with thorough tests. The session override store approach is clean — it avoids modifying the session manager and keeps the reset logic contained. A few observations:

**Session override store grows unboundedly.**
Every `/new` command adds an entry to `sessionOverride.overrides` that is never cleaned up. Over time, this map will grow proportionally to the number of unique `(channel, chatID, agentID)` tuples that have used `/new`. For a busy instance with many users, this could become significant. Consider:
- Replacing the old entry when `/new` is used again for the same tuple (this already happens via `Set`, so it is bounded per-tuple -- this is fine)
- Actually, re-reading the code, each tuple only has one entry at a time since `Set` overwrites. So the bound is the number of unique chats, which is reasonable. Disregard the growth concern.

**`/new` and `/clear` do the same thing.**
Both commands call `startNewSessionForMessage` identically. The issue (#572) and README suggest `/clear` clears the current session while `/new` starts a fresh one. Currently there is no semantic difference. Consider making `/clear` reset the current session in-place (clear history, keep session key) rather than creating a new session key. This matters if other parts of the system reference the session key (e.g., session persistence, analytics).

**Unbounded session accumulation on disk.**
Each `/new` call creates a new session key via `fmt.Sprintf("agent:%s:new:%d", agent.ID, time.Now().UnixNano())`. The old session is saved to disk via `agent.Sessions.Save(sessionKey)` but never deleted. Over many `/new` invocations, session files will accumulate on disk. This should be documented or a cleanup mechanism added.

**`agent.Sessions.Save` error is returned but not communicated to the user.**
In `startNewSessionForMessage`, if `Save` fails, the error propagates to `handleCommand` which formats it as `"Error: %v"`. This is fine for debugging but exposes internal errors to end users. Consider a friendlier message.

**Test coverage is solid.** The cross-chat isolation test and session persistence test are particularly good — they verify the two most important properties of this feature.

**Minor: `.opencode/` in `.gitignore`**
This is a valid addition but unrelated to the feature. Could be a separate trivial PR, but no objection to including it.

Overall this is a clean implementation. The main suggestion is to differentiate `/clear` from `/new` semantically, and to consider session file cleanup.

PR #622, opened 2026-02-22
feat: HTTP channel.

JaviLib:
📝 Description
This PR introduces the new HTTP channel feature for picoclaw, implemented with server-side markdown rendering using Goldmark, allowing the assistant to be accessed through a web interface by default and right after the installation.

🗣️ Type of Change
 ✨ New feature (non-breaking change which adds functionality)

🤖 AI Code Generation
 🛠️ Mostly AI-generated (AI draft, Human verified/modified)

🧪 Test Environment
Hardware: Mac mini M2 Pro
OS: macOS
Model/Provider: stepfun/step-3.5-flash (via picoclaw)
Channel: using the new HTTP (web interface) for testing purposes live.

☑️ Checklist
 My code/docs follow the style of this project.
 I have performed a self-review of my own changes.
 I have updated the documentation accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,12 +268,51 @@ Talk to your picoclaw through Telegram, Discord, DingTalk, LINE, or WeCom
 
 | Channel      | Setup                              |
 | ------------ | ---------------------------------- |
+| **HTTP**     | Automatic (enabled by default)    |
 | **Telegram** | Easy (just a token)                |
 | **Discord**  | Easy (bot token + intents)         |
 | **QQ**       | Easy (AppID + AppSecret)           |
 | **DingTalk** | Medium (app credentials)           |
 | **LINE**     | Medium (credentials + webhook URL) |
 | **WeCom**    | Medium (CorpID + webhook setup)    |
+
+<details>
+<summary><b>HTTP</b> (Default)</summary>
+
+The HTTP channel is enabled by default and runs on `http://127.0.0.1:8070`. No configuration needed!
+
+**Access the web interface:**
+
+1. Start the gateway:
+```bash
+picoclaw gateway
+```
+
+2. Open your browser and navigate to `http://127.0.0.1:8070`
+
+**Configuration:**
+
+```json
+{
+  "channels": {
+    "http": {
+      "enabled": true,
+      "host": "127.0.0.1",
+      "port": 8070
+    }
+  }
+}
+```
+
+**Features:**
+
+- Real-time chat interface
+- Markdown support with proper rendering
+- Responsive design for mobile and desktop
+
+The HTTP channel is perfect for testing, local development, or running PicoClaw on a headless server with a web UI.
+
+</details>
 
 <details>
 <summary><b>Telegram</b> (Recommended)</summary>


xiaket:
Heya @JaviLib , thanks for your contribution!

I have a concern over the memory usage brought by this feature, do you mind showing us some proof that this is not going to have huge negative impact on the RSS? Thanks!

JaviLib:
Hi, thank you for your consideration. I think it will not have any significant over memory usage, and in any case, it can be disabled at any moment.
I asked my AI to make some benchmarks, I reviewed them, and they convinced me. I include the files in case you want to run them.

[benchmarks_http_20260222.zip](https://github.com/user-attachments/files/25466763/benchmarks_http_20260222.zip)

This is the conclusion it takes:

## 📊 Benchmark Results

Attached are the performance comparison results between HTTP and Telegram channels:

Key findings:
- ⏱️ HTTPChannelMemory: ~10× slower than Telegram (expected due to Goldmark rendering)
- 💾 HTTP uses ~100× more memory (HTML rendering overhead)
- ✅ Acceptable trade-off: overhead paid for rich web formatting
- 🎯 HTTP channel works correctly in real-world tests

Regarding resource usage:
The benchmarks confirm that the HTTP channel has significant overhead, but Picoclaw remains suitable for low-resource environments. According to the project's README, Picoclaw is designed to run on $10 boards with <10MB RAM. The Telegram channel (the default for constrained devices) maintains minimal overhead, while the HTTP channel provides enhanced web formatting when resources are available.

HTTP is optional:
- HTTP channel is enabled by default for convenience
- Users on very limited resources (where HTTP rendering is difficult or impossible) can disable HTTP and use lighter channels like Telegram
- This flexibility ensures Picoclaw stays true to its lightweight philosophy across all hardware profiles

ZIP contents:
- 23 benchmark files (go test -bench)
- Detailed metrics: ns/op, bytes/op, allocs/op
- No personal data - only technical metrics

The results confirm that Markdown rendering overhead is significant but within expected range for the gained web functionality. The core lightweight architecture remains intact for resource-constrained deployments. 🚀

xiaket:
Thanks! I've reviewed the results in zip. But I don't think I can find any data/result for RSS? I'm expecting something like the table in https://github.com/sipeed/picoclaw/pull/598

JaviLib:
Thanks!. I've measured RSS. These are the results:

| Configuration | RSS |
|---------------|-----|
| Official v0.1.2 (no HTTP) | 25.1 MB |
| Modified (current GitHub branch + HTTP disabled) | 31.7 MB |
| Modified (HTTP enabled) | 38.3 MB |

Platform: darwin arm64 (other platforms may have less requirements).

Key findings:
- Official build already exceeds 20 MB limit (25.1 MB)
- Modified build adds ~6.6 MB overhead even when HTTP is disabled (likely linked dependencies)
- HTTP channel adds ~13.2 MB when activated

Question: Is the requirement that the *total* RSS must be <20 MB, or that the *increment* from the HTTP channel must be limited?

If the former, the core itself may be optimized to meet the limit.

Potential optimizations if those extra MB matter:
- Optimize/trim embedded HTML & CSS
- Replace the Markdown rendering library (goldmark) with a lighter alternative, or even remove it completely.
- Implement lazy loading for unused data/assets (if viable)
- Strip symbols & use -ldflags="-s -w" in build

Ready to proceed based on your clarification.

xiaket:
Thanks @JaviLib !

I think the team is more leaning towards not to have new features that dramatically increase memory footprint at the moment. Our selling point at the moment is low memory cost so it can run on dirty cheap hardware.

I'm totally with you in that these extra memory usage is not a lot. IT IS NOT, however, for a $10 hardware, it means whether it can run or not.

Later on, this project may decide to loose up the limit a bit, or have a stream of feature that will never get shipped to low-end devices, but for now, I don't think we should proceed.

Thank you very much for opening this PR! Do appreciate your patience with me on this!

JaviLib:
Ok, no problem. One question, how much would it be needed to be reduced for you to accept this PR. I can try to make the effort.

xiaket:
@JaviLib I would say I can look away if that increase is of the order of KB, not MB.

JaviLib:
Ok, I will try. I'll keep you informed. Thanks.

nikolasdehor (changes requested):
Interesting feature — a built-in web UI is useful for quick testing. However, there are several security and correctness issues that need to be addressed before merging.

**Critical: XSS vulnerability via `div.innerHTML = processedContent`**

In the JavaScript, user messages are rendered with:
```js
processedContent = content.replace(/\n/g, '<br>');
div.innerHTML = processedContent;
```
User input is inserted directly into the DOM via `innerHTML` without sanitization. An attacker (or even a user testing edge cases) can inject arbitrary HTML/JS. For user messages, use `textContent` instead of `innerHTML`. For assistant messages (which contain server-rendered HTML), you should sanitize on the server side or use a library like DOMPurify on the client.

**Critical: WebSocket origin check accepts all origins**
```go
var upgrader = websocket.Upgrader{
    CheckOrigin: func(r *http.Request) bool {
        return true
    },
}
```
This allows any website to open a WebSocket to the picoclaw instance. If a user has picoclaw running locally and visits a malicious page, that page can send messages through the bot. At minimum, restrict to same-origin or make the allowed origins configurable.

**No auth on any route**
The `/api/chat` endpoint and WebSocket have no authentication whatsoever. The `IsAllowed` check is never called in any handler. The allowlist config field exists but is dead code. Combined with the default `0.0.0.0` fallback in `Start()` (vs the documented `127.0.0.1` default), this means the channel could be open to the network without auth.

**Default host inconsistency:**
- `defaults.go` sets `Host: "127.0.0.1"` (good)
- But `Start()` falls back to `"0.0.0.0"` if host is empty (bad — this fallback should be `"127.0.0.1"` to match)

**`/api/chat` is fire-and-forget with no response delivery:**
The REST endpoint publishes the message to the bus and immediately returns `{"status":"ok"}`. But there is no way for the caller to receive the LLM response — that would only be delivered via WebSocket. The REST endpoint should either be removed or documented as "trigger only" with a note that responses require a WebSocket connection.

**Concurrent WebSocket writes are not safe:**
`session.Conn.WriteJSON` in `Send()` can be called from multiple goroutines (the bus dispatcher). The gorilla/websocket docs explicitly state: "Connections support one concurrent reader and one concurrent writer." You need a write mutex per connection, or use a buffered channel to serialize writes.

**`min()` redefinition:**
Go 1.21+ has a builtin `min()`. The custom `min` function at the bottom of `http.go` will shadow it and cause confusion (and is only used in a commented-out debug line anyway). Remove it.

**Hardcoded Spanish strings in the template:**
The welcome message, button text, and status messages are in Spanish ("Escribe tu mensaje...", "Enviar", "Conectado"). This should either be English (matching the rest of the project) or configurable.

**No tests:**
This is a 761-line addition with zero test coverage. At minimum, unit tests for `renderMarkdown`, the REST API handler, and the WebSocket lifecycle should be included.

**Missing newline at end of files:**
Both `http.go` and `index.html` are missing a trailing newline.

The feature is worthwhile but needs the security issues addressed before merge.

PR #623, opened 2026-02-22
fix: should use fmt.Printf instead of fmt.Print(fmt.Sprintf(...))

yofabr:
This PR replaces `fmt.Print(fmt.Sprintf(...))` with `fmt.Printf` for cleaner and more idiomatic Go code.

Using `fmt.Printf` avoids unnecessary string formatting and improves readability.

No functional changes were introduced.


xiaket (approved):
LGTM

Orgmar:
@yofabr Clean refactor! Replacing fmt.Print(fmt.Sprintf(...)) with fmt.Printf is the idiomatic Go way and avoids the unnecessary intermediate string allocation. Small changes like this add up across a codebase.

We're putting together the **PicoClaw Dev Group** on Discord for contributors. If you'd like in, shoot an email to `support@sipeed.com` with the subject `[Join PicoClaw Dev Group] yofabr` and we'll send you the invite link!

PR #638, opened 2026-02-22
flexibility of the `WebSearchTool` by allowing the `count` argument to accept both numeric and string values

Xeven777:
This pull request to `pkg/tools/web.go` enhances the flexibility of the `WebSearchTool` by allowing the `count` argument to accept both numeric and string values. This makes the tool more robust when handling input types.

Improvements to argument handling:

* Updated the `Execute` method in `WebSearchTool` to support `count` as either a `float64` or a `string`, converting string values to integers and validating them within the allowed range.
* Added the `strconv` package import to enable string-to-integer conversion.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -73,7 +73,6 @@ nfpms:
     package_name: picoclaw
     file_name_template: >-
       {{ .PackageName }}_
-      {{- .Version }}_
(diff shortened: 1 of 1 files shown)


Xeven777:
@nikolasdehor New PR here with clean diff: https://github.com/sipeed/picoclaw/pull/741

PR #676, opened 2026-02-23
feat: add WATI WhatsApp Business API channel

kenyeung128:
## Summary

- Add a new `wati` channel for direct WATI WhatsApp Business API integration (HTTP webhook inbound + REST API outbound)
- New `WATIChannel` in `pkg/channels/wati.go` following the LINE channel pattern
- WATI v3 API: outbound via `POST /api/ext/v3/conversations/messages/text` with Bearer token auth
- Webhook handler supports GET (hub.challenge verification) and POST (incoming messages) with multi-shape payload normalization
- `WATIConfig` with full env var support (`PICOCLAW_CHANNELS_WATI_*`), default port 18794

## Files Changed

| File | Change |
|------|--------|
| `pkg/channels/wati.go` | New WATI channel implementation |
| `pkg/config/config.go` | Add `WATIConfig` type and field in `ChannelsConfig` |
| `pkg/config/defaults.go` | Add WATI default config values |
| `pkg/channels/manager.go` | Register WATI channel in init (gated on `Enabled + APIToken`) |
| `config/config.example.json` | Add WATI example configuration |

## Test plan

- [x] `make build` compiles successfully
- [ ] Set `PICOCLAW_CHANNELS_WATI_ENABLED=true` and `PICOCLAW_CHANNELS_WATI_API_TOKEN=<token>`, verify webhook server starts on port 18794
- [ ] GET `/webhook/wati?hub.challenge=test` returns `test`
- [ ] POST webhook with WATI payload triggers message handling
- [ ] Outbound messages POST to WATI v3 API correctly

### DIFF
--- a/config/config.example.json
+++ b/config/config.example.json
@@ -139,6 +139,16 @@
       "webhook_path": "/webhook/wecom-app",
       "allow_from": [],
       "reply_timeout": 5
+    },
+    "wati": {
+      "_comment": "WATI WhatsApp Business API - https://www.wati.io/",
+      "enabled": false,
+      "api_token": "YOUR_WATI_API_TOKEN",
+      "api_base_url": "https://live-mt-server.wati.io",
+      "webhook_host": "0.0.0.0",
+      "webhook_port": 18794,
+      "webhook_path": "/webhook/wati",
+      "allow_from": []
     }
   },
   "providers": {
--- a/pkg/channels/manager.go
+++ b/pkg/channels/manager.go
@@ -202,6 +202,19 @@ func (m *Manager) initChannels() error {
 		}
 	}
 
+	if m.config.Channels.WATI.Enabled && m.config.Channels.WATI.APIToken != "" {
+		logger.DebugC("channels", "Attempting to initialize WATI channel")
+		wati, err := NewWATIChannel(m.config.Channels.WATI, m.bus)
+		if err != nil {
+			logger.ErrorCF("channels", "Failed to initialize WATI channel", map[string]any{
+				"error": err.Error(),
+			})
+		} else {
+			m.channels["wati"] = wati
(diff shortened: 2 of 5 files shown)


nikolasdehor (changes requested):
Functional implementation but has security and correctness issues that should be addressed before merge:

**Security:**
1. **No webhook signature verification.** The WATI webhook endpoint accepts any POST request without verifying the request actually came from WATI. An attacker who discovers the webhook URL can inject fake messages. WATI provides webhook verification tokens - this should be validated. At minimum, check the `allow_from` IP restriction (which is configured but I do not see enforced in the webhook handler).

2. **No request body size limit.** `io.ReadAll(r.Body)` will read an unbounded body into memory. A malicious request with a multi-GB body would OOM the process. Add `http.MaxBytesReader` or `io.LimitReader`.

3. **API token in Bearer header.** This is correct for WATI v3, but make sure the token is never logged. The current code does not log it, which is good.

**Correctness:**
4. **`HandleMessage` signature mismatch.** The call `c.HandleMessage(senderID, chatID, text, nil, metadata)` does not match the `BaseChannel.HandleMessage` signature which typically takes `(ctx, peer, messageID, senderID, chatID, content, media, metadata, sender)`. The PR may not compile against the current base. Please verify.

5. **New HTTP client per request.** In `callAPI`, a new `http.Client` is created for every outbound message. This defeats connection pooling and can cause socket exhaustion under load. Create the client once in `NewWATIChannel` and reuse it.

6. **`fromMe` detection is fragile.** The `getBoolField` helper only checks `bool` type assertions. If WATI sends `"fromMe": "true"` (string), it would be missed and self-messages would be processed, potentially causing loops.

**Missing:**
- No webhook secret/token verification
- No rate limiting on inbound webhook
- No test file
- `HandleMessage` call needs the context parameter (the channel context `c.ctx`)

PR #681, opened 2026-02-23
fix: distinguish network timeouts from context window errors

### LINKED ISSUE #683: [BUG] Network timeouts misclassified as context window errors
## Bug Description

HTTP timeouts from LLM providers (e.g. z.ai) are incorrectly classified as context window errors. This causes:

1. Useless history compression triggered on network timeouts
2. Misleading message to users: "Context window exceeded. Compressing history and retrying..."
3. No actual retry of the failed request

## Steps to Reproduce

1. Configure picoclaw with z.ai (or any provider with slow/unstable network)
2. Send a message that triggers an LLM call
3. Wait for HTTP timeout (2 minutes)
4. Observe the error in logs:
   ```
   error=failed to send request: context deadline exceeded 
   (Client.Timeout exceeded while awaiting headers)
   ```

## Expected Behavior

- Timeout errors should be recognized as network timeouts
- The request should be retried with backoff
- No false "Context window exceeded" message

## Actual Behavior

- `context deadline exceeded` matches the broad `"context"` substring check in the retry loop
- `Client.Timeout` matches the broad `"token"` substring check
- Error is classified as a context window error
- History compression runs (pointless for a network issue)
- User sees "Context window exceeded" (misleading)
- After compression + retry fails, the error is not retried further

## Root Cause

In `pkg/agent/loop.go`, the context window error detection uses overly broad substring matching:

```go
isContextError := strings.Contains(errMsg, "token") ||
    strings.Contains(errMsg, "context") ||
    strings.Contains(errMsg, "invalidparameter") ||
    strings.Contains(errMsg, "length")
```

- `"context"` matches both `context_length_exceeded` and `context deadline exceeded`
- `"token"` matches both token limit errors and `Client.Timeout`

## Fix

PR #681

dimensi:
Fixes #683

## Summary

- HTTP timeouts (`context deadline exceeded`, `Client.Timeout exceeded`) were incorrectly classified as context window errors due to overly broad substring matching (`"context"`, `"token"`, `"length"`)
- This caused useless history compression and a misleading "Context window exceeded" message to users when the real problem was a network timeout (e.g. slow z.ai API response)
- Replaced broad checks with explicit timeout detection first, then specific context window error patterns (`context_length_exceeded`, `token limit`, `too many tokens`, `invalidparameter`, etc.)
- Timeout errors were not retried at all — now they retry up to 2 times with exponential backoff (5s, 10s)

## Test plan

- [x] Existing `TestAgentLoop_ContextExhaustionRetry` passes — real context window errors still trigger compression
- [x] Full test suite passes (`go test ./...`)
- [ ] Manual test: simulate z.ai timeout and verify no false "Context window exceeded" message
- [ ] Manual test: verify timeout retries with backoff in logs

### DIFF
--- a/pkg/agent/loop.go
+++ b/pkg/agent/loop.go
@@ -9,6 +9,7 @@ package agent
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


nikolasdehor (approved):
Good fix for a real problem. Network timeouts being misclassified as context window errors would cause unnecessary history compression and confusing user messages.

The approach is correct: check for timeout first, then check for context window errors with `!isTimeoutError` guard.

Minor observations:
1. `strings.Contains(errMsg, "max_tokens")` could false-positive on API responses that echo back the `max_tokens` parameter in error messages unrelated to context limits (e.g., "max_tokens must be positive"). This is an edge case but worth noting.

2. The backoff `time.Duration(retry+1) * 5 * time.Second` gives 5s/10s, which is reasonable but blocks the goroutine. For high-concurrency scenarios, consider using a timer with context cancellation. Not a blocker for this PR.

3. Note that PR #699 (decompose AgentLoop) extracts the same error detection into `IsContextWindowError()` in `error_classifier.go`. These two PRs will conflict. Whichever lands second should incorporate the timeout guard from this PR.

LGTM.

yinwm:
Thanks for the pr

Orgmar:
@dimensi Good catch on the timeout vs context window misclassification. Broad substring matching on "context" and "token" silently treating network timeouts as context exhaustion is a frustrating bug to track down. The retry with exponential backoff for actual timeouts is a nice addition too.

We have a **PicoClaw Dev Group** on Discord where contributors share ideas and collaborate. If you're interested, send an email to `support@sipeed.com` with the subject `[Join PicoClaw Dev Group] dimensi` and we'll send you the invite link.

dimensi:
Thanks, but I realized I don't know how to use openclaw, picoclaw, and similar bots in real life, so I won't for now)

PR #682, opened 2026-02-23
fix: add generate dependency to test and vet Makefile targets

Esubaalew:
## Summary

- `make test` and `make vet` fail on a fresh clone because `cmd/picoclaw/workspace` (needed by `//go:embed`) is gitignored and only created by `go generate`
- The `build` target already depends on `generate`, but `test` and `vet` did not

## Test plan

- `rm -rf cmd/picoclaw/workspace && make test` — passes (was FAIL before)
- `rm -rf cmd/picoclaw/workspace && make vet` — passes (was FAIL before)
- `make build` — still works as before

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -129,11 +129,11 @@ clean:
 	@echo "Clean complete"
 
 ## vet: Run go vet for static analysis
-vet:
+vet: generate
 	@$(GO) vet ./...
 
 ## test: Test Go code
-test:
+test: generate
 	@$(GO) test ./...
 
 ## fmt: Format Go code


Esubaalew:
@PixelTux

nikolasdehor (approved):
Minimal and correct fix. The `go:embed` directive requires the workspace directory to exist at compile time, and `go test` / `go vet` trigger compilation. Adding `generate` as a prerequisite for `test` and `vet` targets mirrors what `build` already does. LGTM.

PR #686, opened 2026-02-23
feat: add Deepgram as alternative STT provider

dimensi:
## Summary
- Extract `Transcriber` interface in `pkg/voice/transcriber.go` so channels are decoupled from a specific STT backend
- Add `DeepgramTranscriber` (`pkg/voice/deepgram.go`) — uses Deepgram REST API with nova-2 model, smart formatting, and language detection
- Add `Deepgram` provider config to `ProvidersConfig` (supports `providers.deepgram.api_key`, `model_list` with `deepgram/` prefix, and env var)
- Update Telegram, Discord, Slack channels to accept `voice.Transcriber` interface instead of `*voice.GroqTranscriber`
- Add Deepgram as fallback in `cmd_gateway.go`: Groq is tried first, then Deepgram

## Test plan
- [x] `go build ./...` compiles
- [x] `go test ./...` passes
- [ ] Deploy and send voice message in Telegram, verify transcription in logs

### DIFF
--- a/cmd/picoclaw/cmd_gateway.go
+++ b/cmd/picoclaw/cmd_gateway.go
@@ -121,7 +121,7 @@ func gatewayCmd() {
 	// Inject channel manager into agent loop for command handling
 	agentLoop.SetChannelManager(channelManager)
 
-	var transcriber *voice.GroqTranscriber
+	var transcriber voice.Transcriber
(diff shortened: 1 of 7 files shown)


dimensi:
ц

PR #696, opened 2026-02-23
fix: model parser - handle two / in model

### LINKED ISSUE #694: How to set config for openrouter api?
My current config structure

```
{
  "agents": {
    "defaults": {
      "workspace": "~/.picoclaw/workspace",
      "restrict_to_workspace": true,
      "model": "nemotron",
      "max_tokens": 256000,
      "temperature": 0.7,
      "max_tool_iterations": 20
    }
  },
  "model_list": [
    {
      "model_name": "nemotron",
      "model": "nvidia/nemotron-3-nano-30b-a3b:free",
      "api_key": "sk-or-v1-uuuuu",
      "api_base": "https://openrouter.ai/api/v1/chat/completions"
    }
  ],
  "channels": {
    "slack": {
      "enabled": false,
      "bot_token": "xapp-1-uuuu-uuuu-uuu",
      "app_token": "xapp-1-uuuu-uuuu-uuu",
      "allow_from": [
      "*"
      ]
    }
  },
  "tools": {
    "web": {
      "duckduckgo": {
        "enabled": true,
        "max_results": 5
      },
      "tavily": {
        "enabled": false,
        "api_key": "tvly-uuuu",
        "max_results": 5
      }
    },
    "cron": {
      "exec_timeout_minutes": 5
    },
    "exec": {
      "enable_deny_patterns": false,
      "custom_deny_patterns": []
    },
    "skills": {
      "registries": {
        "clawhub": {
          "enabled": true,
          "base_url": "https://clawhub.ai",
          "search_path": "/api/v1/search",
          "skills_path": "/api/v1/skills",
          "download_path": "/api/v1/download"
        }
      }
    }
  },
  "heartbeat": {
    "enabled": true,
    "interval": 30
  },
  "devices": {
    "enabled": false,
    "monitor_usb": true
  },
  "gateway": {
    "host": "0.0.0.0",
    "port": 18790
  }
}
```

jackkav:
## 📝 Description

In config.json "model": "openrouter/a/b:free" was being parsed correctly in factory_provider and incorrectly in model_ref

  Fix provider/model parsing so OpenRouter namespaced model IDs (for example stepfun/step-3.5-flash:free) are not misinterpreted
  as provider/model.

  ## Problem

  When using OpenRouter models with a slash in the model ID, logs could show errors like step-3.5-flash:free is not a valid model
  because parsing treated stepfun as the provider instead of part of the model path.

  ## Root Cause

  ParseModelRef interpreted any string containing / as provider/model, regardless of whether the prefix was actually a known
  provider.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue

Fixes #694

## Notes
- I added tests to both areas to cover the same case.
- Provider detection should maybe live in one place. Out of scope for this bug fix.
- I don't think stacking the provider before the model is a good idea, why not separate it, its just config used to lookup api url and api format.

## ☑️ Checklist

### DIFF
--- a/pkg/providers/model_ref.go
+++ b/pkg/providers/model_ref.go
@@ -19,12 +19,12 @@ func ParseModelRef(raw string, defaultProvider string) *ModelRef {
 
 	if idx := strings.Index(raw, "/"); idx > 0 {
 		prefix := strings.TrimSpace(raw[:idx])
+		model := strings.TrimSpace(raw[idx+1:])
+		if model == "" {
+			return nil
+		}
 		if isKnownProviderPrefix(prefix) {
 			provider := NormalizeProvider(prefix)
-			model := strings.TrimSpace(raw[idx+1:])
-			if model == "" {
-				return nil
-			}
 			return &ModelRef{Provider: provider, Model: model}
 		}
 	}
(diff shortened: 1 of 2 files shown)


muava12:
waiting this fix soon :)

nikolasdehor (commented):
Good bug fix with solid test coverage. The approach of checking against known provider prefixes before splitting is correct. A few concerns:

**Maintenance burden of the hardcoded list:**
`knownProviderPrefixes` is a manually maintained map that must stay in sync with `NormalizeProvider()` and the factory. When a new provider is added, developers must remember to update this map too. Consider deriving it programmatically from the existing provider normalization table, or at minimum add a comment pointing to the other locations that need updating.

**Subtle behavioral change for unknown providers:**
Before this PR, `ParseModelRef("custom/my-model", "openai")` would parse as `{Provider: "custom", Model: "my-model"}`. After this PR, it parses as `{Provider: "openai", Model: "custom/my-model"}` because "custom" is not in the known list. This is the intended fix for OpenRouter, but it also changes behavior for any user who has a custom/self-hosted provider with a prefix not in the list. Worth documenting in release notes.

**Test for NormalizeProvider alignment:**
The `isKnownProviderPrefix` function calls `NormalizeProvider(prefix)` before the map lookup, which is good. But the map values themselves should be the *normalized* forms. For example, `"github-copilot"` and `"github_copilot"` both normalize to the same canonical form - are both entries needed? If `NormalizeProvider` handles the normalization, only the canonical form needs to be in the map.

**gitignore addition:**
The `.gocache` and `.gomodcache` entries in `.gitignore` are fine but unrelated to the bug fix. Consider splitting into a separate commit for clean history.

Overall this is a good fix. The test cases are comprehensive and cover the edge cases well.

jackkav:
I'd rather this be addressed in the schema

PR #714, opened 2026-02-24
skills: install/reinstall CLI and refactor into skillsCmd

seanly:
- Add ParseInstallSpec, InstallFromGitHubEx, fetchTree, fetchDefaultBranch, validateSubpath; support repo@branch and optional subpath
- Add reinstall subcommand (force overwrite); install errors with hint when skill already exists
- Production install uses GitHub Trees API for full directory; baseURL mode for tests uses single SKILL.md
- Keep InstallFromGitHub(spec) as wrapper for backward compatibility
- Add installer_test.go: ParseInstallSpec, reinstall overwrite, already-exists
- Refactor skills CLI from main.go into skillsCmd() in cmd_skills.go
- Add docs/skills-cli.md

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

- **Reference URL:**
- **Reasoning:**

## 🧪 Test Environment
- **Hardware:** Mac M1 
- **OS:** Mac OS
- **Model/Provider:** Gemini 
- **Channels:** agent cli

## 📸 Evidence (Optional)
<details>

</details>

## ☑️ Checklist


nikolasdehor (commented):
Good feature addition -- the extended install spec (`owner/repo@branch`, subpath, archive support, reinstall) addresses real usability gaps. A few concerns:

**1. SSRF in archive install from URL.**
`InstallFromArchive` accepts a URL and fetches it with `http.Get`. There is no validation that the URL points to a legitimate archive host. An attacker who controls a skill spec could point the installer at an internal network endpoint (e.g. `http://169.254.169.254/latest/meta-data/` on cloud VMs). Consider allowlisting the URL scheme (https only) and potentially the domain.

**2. Path traversal in tar extraction (`pkg/utils/tar.go`).**
Without seeing the full implementation, tar archives can contain entries with `../` in their paths. The `ExtractTar` utility must validate that extracted paths stay within the destination directory. This is the same "Zip Slip" vulnerability class mentioned in Go security advisories.

**3. `fetchDefaultBranch` and `fetchTree` silently swallow errors.**
`fetchDefaultBranch` returns `"main"` and `nil` error on any failure (network, 404, rate limit). This means a typo in the repo name (`sipeed/picocaw` instead of `sipeed/picoclaw`) silently tries to install from `main` branch of a non-existent repo, and the user gets a confusing "failed to fetch tree" error instead of "repo not found." Consider at least logging a warning when the API call fails.

**4. No rate limiting awareness for GitHub API.**
Both `fetchDefaultBranch` and `fetchTree` hit the GitHub API without authentication. Unauthenticated requests are limited to 60/hour. A user running `install` a few times during skill development could hit this quickly. Consider supporting `GITHUB_TOKEN` from the environment for authenticated requests.

**5. `InstallFromGitHubEx` downloads files sequentially.**
Each blob is fetched individually via raw.githubusercontent.com. For a skill with many files, this could be slow. The archive-based install path is better for large skills. Worth documenting in the help text that archive install is faster for repos with many files.

**6. The `isArchiveInstallArg` heuristic could false-positive.**
A file named `./my-skill.zip` that does not exist would return `false` (correct), but a GitHub repo URL like `https://github.com/owner/repo.zip` would match the archive pattern (has `.zip` suffix and https prefix). This would try to download the GitHub URL as a raw archive instead of using the Trees API. Consider checking the URL path more carefully or prioritizing the GitHub install path when the URL matches `github.com`.

**7. `reinstall` deletes the entire skill directory before downloading.**
If the download fails after deletion, the user loses their existing skill installation. Consider downloading to a temp directory first, then swapping (like atomic file operations).

The test coverage looks reasonable for the parsing logic. The `fetchTree`/`fetchDefaultBranch` functions are not tested (hard to test without mocking the GitHub API), but the `baseURL` injection is a good pattern for testability.

Issue #704, opened 2026-02-24
Race condition in session history causes "tool_call_ids did not have response messages" (HTTP 400)


ffonseca:
## Bug Description

  When the agent processes a message that triggers multiple tool_calls, the session can be saved in a corrupted
  state where the `assistant` message with `tool_calls` exists but the corresponding `tool` result messages are
  missing. On the next incoming message, the LLM API returns a 400 error.

  ## Error

  Error processing message: LLM call failed after retries: API request failed:
    Status: 400
    Body: {
      "error": {
        "message": "An assistant message with 'tool_calls' must be followed by tool messages responding to each
  'tool_call_id'. The following tool_call_ids did not have response messages: call_AijviHCBvxqBudag94AHVlQO",
        "type": "invalid_request_error",
        "param": "messages.[12].role",
        "code": null
      }
    }

  ## Root Cause (suspected)

  In `pkg/agent/loop.go`, after the agent loop completes, `Save()` is called and then `maybeSummarize()` launches a
  background goroutine:

  ```go
  agent.Sessions.AddMessage(opts.SessionKey, "assistant", finalContent)
  agent.Sessions.Save(opts.SessionKey) // persists to disk

  go func() {
      al.summarizeSession(agent, sessionKey) // calls TruncateHistory(4) + Save()
  }()

  The summarization goroutine calls TruncateHistory(4) + Save() concurrently with new incoming messages being
  appended to the session. This race condition can leave the persisted history in a state where an assistant message
   with tool_calls exists without the corresponding tool result messages.

  Steps to Reproduce

  1. Configure a gateway with a Telegram channel
  2. Send a message that causes the agent to invoke 2 or more tool_calls in a single LLM response
  3. Immediately send a follow-up message before the previous interaction fully persists
  4. (Or: trigger the summarization threshold while a message with tool_calls is being processed)

  Impact

  - Next message from the user receives no response (error is logged but user sees nothing)
  - Workaround: restart the container (resets in-memory session state)
  - Manual fix without restart: reset the session file
  docker exec picoclaw-gateway sh -c 'echo "{\"key\":\"agent:main:main\",\"messages\":[]}" >
  /home/picoclaw/.picoclaw/workspace/sessions/agent_main_main.json'

  Suggested Fix

  On session load, validate that every assistant message with tool_calls is followed by the corresponding tool
  result messages. Strip any orphaned tool_calls messages before sending history to the LLM API.

[...]

PR #719, opened 2026-02-24
feat(memory): add SQLite-backed persistent session store

is-Xiaoen:
## 📝 Description

Add a new `pkg/memory/` package that implements a SQLite-backed session store using `modernc.org/sqlite` (pure Go, zero CGo, cross-compile friendly).

**Problem:** The current `pkg/session/manager.go` uses an in-memory map with full JSON serialization on every `Save()` call. This leads to:
- Race conditions when `summarizeSession()` runs concurrently with the main loop (#704)
- I/O bottleneck from full-file rewrites on every save (painful on embedded devices)
- No session lifecycle management (no TTL, no cleanup — `sessions/` grows unbounded)

**Solution:** A new `Store` interface whose methods are each an atomic SQLite transaction — no separate `Save()` needed. This eliminates the storage-level write conflicts entirely.

This PR is **purely additive** — it does not modify any existing code. Integration with `pkg/agent/loop.go` will be a follow-up PR.

### Key design decisions

- **modernc.org/sqlite**: pure Go driver, zero CGo dependency, ideal for cross-compilation to embedded targets
- **Single connection** (`SetMaxOpenConns(1)`): serializes all writes at the DB level, preventing corrupt interleaving
- **WAL journal mode**: readers never block writers, good concurrency on read-heavy workloads
- **ToolCalls as JSON column**: always read/written together with their message, no need for a separate table
- **Tuned for embedded**: 512KB cache, `synchronous=NORMAL`, 5s busy timeout

### Migration

`MigrateFromJSON()` imports existing `sessions/*.json` files into SQLite:
- Reads session key from JSON content (not filename), correctly handling `sanitizeFilename()` colon → underscore mapping
- Renames migrated files to `.json.migrated` as backup (non-destructive)
- Idempotent: safe to re-run after partial failures via `INSERT OR IGNORE`

### Files

```
pkg/memory/
├── store.go            // Store interface (8 methods, maps 1:1 to current SessionManager)
├── sqlite.go           // SQLiteStore implementation
├── sqlite_test.go      // 14 unit + 2 concurrency + 3 benchmark tests
├── migration.go        // JSON → SQLite migration
└── migration_test.go   // 7 migration tests
```

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue

Ref #704 — eliminates the storage-level race condition (concurrent `Save()` calls overwriting each other). Full application-level fix (summarize goroutine vs main loop) to follow in integration PR.

- **Reference URL:** 26M2W3 community meeting notes (Agent track: "introduce SQLite")
- **Reasoning:** The current JSON file approach fundamentally cannot support concurrent writes safely. SQLite with WAL provides ACID transactions with minimal overhead, and `modernc.org/sqlite` avoids CGo for embedded cross-compilation.

## 🧪 Test Environment
- **Hardware:** Intel i7-12650H (development), targeting Raspberry Pi / MaixCAM
- **OS:** Windows 11 (development), tests are platform-independent
- **Model/Provider:** N/A (storage layer, no LLM interaction)
- **Channels:** N/A

## 📸 Evidence (Optional)
<details>
<summary>Click to view test results and benchmarks</summary>

```
=== RUN   TestOpen_CreatesDatabase          --- PASS
=== RUN   TestOpen_ExistingDatabase         --- PASS
=== RUN   TestAddMessage_BasicRoundtrip     --- PASS
=== RUN   TestAddMessage_AutoCreatesSession --- PASS
=== RUN   TestAddFullMessage_WithToolCalls  --- PASS
=== RUN   TestAddFullMessage_ToolCallID     --- PASS
=== RUN   TestGetHistory_EmptySession       --- PASS
=== RUN   TestGetHistory_Ordering           --- PASS
=== RUN   TestSetSummary_GetSummary         --- PASS
=== RUN   TestTruncateHistory_KeepLast      --- PASS
=== RUN   TestTruncateHistory_KeepZero      --- PASS
=== RUN   TestSetHistory_ReplacesAll        --- PASS
=== RUN   TestConcurrent_AddAndRead         --- PASS
=== RUN   TestConcurrent_SummarizeRace      --- PASS
=== RUN   TestMigrateFromJSON_Basic         --- PASS
=== RUN   TestMigrateFromJSON_WithToolCalls --- PASS
=== RUN   TestMigrateFromJSON_MultipleFiles --- PASS
=== RUN   TestMigrateFromJSON_InvalidJSON   --- PASS
=== RUN   TestMigrateFromJSON_RenamesFiles  --- PASS
=== RUN   TestMigrateFromJSON_Idempotent    --- PASS
=== RUN   TestMigrateFromJSON_ColonInKey    --- PASS
ok      github.com/sipeed/picoclaw/pkg/memory   0.990s

BenchmarkAddMessage-16          12854     91923 ns/op    2880 B/op     81 allocs/op
BenchmarkGetHistory_100-16      12516     98754 ns/op   43640 B/op   1226 allocs/op
BenchmarkGetHistory_1000-16      1375    852323 ns/op  390874 B/op  12029 allocs/op
```

`go vet ./pkg/memory/...` — clean, no issues.

</details>

## ☑️ Checklist

### DIFF
--- a/pkg/memory/migration.go
+++ b/pkg/memory/migration.go
@@ -1,0 +1,137 @@
+package memory
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/sipeed/picoclaw/pkg/providers"
+)
+
+// jsonSession mirrors the JSON structure used by pkg/session for deserialization.
+type jsonSession struct {
+	Key      string              `json:"key"`
+	Messages []providers.Message `json:"messages"`
+	Summary  string              `json:"summary,omitempty"`
+	Created  time.Time           `json:"created"`
+	Updated  time.Time           `json:"updated"`
+}
+
+// MigrateFromJSON reads JSON session files from sessionsDir and imports them
+// into the given SQLiteStore. Successfully migrated files are renamed to
+// .json.migrated (not deleted) as a backup.
+//
+// Returns the number of sessions successfully migrated.
+// Files that are already .migrated or fail to parse are skipped (partial failures
+// do not stop the migration of other files).
+func MigrateFromJSON(ctx context.Context, sessionsDir string, store *SQLiteStore) (int, error) {
+	entries, err := os.ReadDir(sessionsDir)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return 0, nil
+		}
+		return 0, fmt.Errorf("memory: read sessions dir: %w", err)
+	}
+
+	migrated := 0
+	for _, entry := range entries {
+		if entry.IsDir() {
+			continue
+		}
+		name := entry.Name()
+		if !strings.HasSuffix(name, ".json") {
+			continue
+		}
+		// Skip already-migrated files.
+		if strings.HasSuffix(name, ".migrated") {
+			continue
+		}
+
+		filePath := filepath.Join(sessionsDir, name)
+		data, err := os.ReadFile(filePath)
+		if err != nil {
+			continue // skip unreadable files
+		}
+
+		var sess jsonSession
+		if err := json.Unmarshal(data, &sess); err != nil {
+			continue // skip invalid JSON
+		}
+
+		if sess.Key == "" {
+			continue // skip sessions with no key
+		}
+
+		if err := importSession(ctx, store, &sess); err != nil {
+			continue // skip on import error
+		}
+
+		// Rename to .json.migrated as backup.
+		_ = os.Rename(filePath, filePath+".migrated")
+		migrated++
+	}
+
+	return migrated, nil
+}
+
+// importSession imports a single JSON session into the SQLite store.
+func importSession(ctx context.Context, store *SQLiteStore, sess *jsonSession) error {
+	tx, err := store.db.BeginTx(ctx, nil)
+	if err != nil {
+		return err
+	}
+	defer tx.Rollback()
+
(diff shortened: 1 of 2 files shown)


is-Xiaoen:
Thanks for the thorough review! The dependency concern is totally fair — 7 transitive deps for session storage is a real cost for a "pico" tool.

One technical nuance I want to flag: **JSONL's append-only advantage breaks down for `TruncateHistory`**, which runs every ~20 messages during summarization. Truncating in JSONL means read-filter-rewrite the entire file — the exact same full-serialization pattern we're trying to move away from. `SetHistory` (used by `forceCompression`) has the same issue. Claude Code's `history.jsonl` works because it's a pure append-only audit log; PicoClaw's sessions need truncation and replacement on a regular basis, which are fundamentally different access patterns.

That said, the `Store` interface is deliberately backend-agnostic. Adding a JSONL-backed implementation behind the same interface would be straightforward. The real value of this PR is arguably the interface design (atomic operations, no separate `Save()`) rather than the specific backend.

The 26M2W3 meeting notes mentioned "introduce SQLite" for the Agent track, which is why I went this direction. But I understand that may be a premature jump for the current single-process setup.

Happy to implement a JSONL backend alongside or instead of SQLite — what direction would you prefer?

is-Xiaoen:
Superseded by #732 (JSONL approach per @yinwm's feedback). Closing this one to keep things clean.

PR #724, opened 2026-02-24
fix: prevent DefaultConfig template values from leaking into user model_list entries

### LINKED ISSUE #721: [BUG] DefaultConfig ModelList template values leak into user model_list entries via JSON decoder slice reuse
## Quick Summary

`LoadConfig` calls `DefaultConfig()` which pre-populates `ModelList` with ~18 template entries (each with provider-specific `APIBase` values). Go's `encoding/json` decoder **reuses existing slice elements** when unmarshaling into a non-nil slice — it only writes fields that are present in the JSON, leaving absent fields at whatever value was already in memory.

This means if a user's `config.json` contains a `model_list` entry **without** `api_base`, the decoder inherits the `APIBase` from whichever `DefaultConfig` template occupied that index position, silently overriding the expected behavior.

**Concrete example:** A user with this config:
\`\`\`json
{
  "model_list": [
    {
      "model_name": "minimax-m2.5-free",
      "model": "opencode/minimax-m2.5-free",
      "api_key": "sk-xxx"
    }
  ]
}
\`\`\`

Gets `APIBase: "https://open.bigmodel.cn/api/paas/v4"` (Zhipu's URL) injected into their entry — because `DefaultConfig().ModelList[0]` is the Zhipu `glm-4.7` template, and the JSON element at index 0 reuses that memory without resetting `APIBase`.

The request then hits the wrong provider endpoint, returning a 404.

##  Environment & Tools
- **PicoClaw Version:** v0.1.2-204-g8774526
- **Go Version:** go1.25.7
- **AI Model & Provider:** Any provider when `api_base` is omitted from `model_list` entry
- **Operating System:** macOS / Linux (affects all platforms)
- **Channels:** N/A (affects core config loading)

## 📸 Steps to Reproduce
1. Ensure `DefaultConfig().ModelList` is non-empty (it always is, with ~18 template entries)
2. Write a `config.json` with a `model_list` entry that omits `api_base` (relying on the provider's built-in default)
3. Run `picoclaw agent -m "Hello"`

## ❌ Actual Behavior

The `ModelConfig.APIBase` is silently set to the `APIBase` of whichever `DefaultConfig` template entry occupies the same slice index. For example, the first entry always inherits `"https://open.bigmodel.cn/api/paas/v4"` (Zhipu), regardless of the configured provider. The request hits the wrong endpoint and fails with a 404 (or worse, succeeds silently against the wrong provider).

Debug evidence from `LoadConfig`:
\`\`\`
0000: []config.ModelConfig{config.ModelConfig{
  ModelName:"minimax-m2.5-free",
  Model:"opencode/minimax-m2.5-free",
  APIBase:"https://open.bigmodel.cn/api/paas/v4",  // ← NOT in config.json!
  APIKey:"sk-xxx",
  ...
}}
\`\`\`

## ✅ Expected Behavior

`ModelConfig.APIBase` should be `""` (empty) when not specified in `config.json`, allowing the provider factory to apply the correct default via `getDefaultAPIBase(protocol)`.

## 💬 Additional Context

**Root cause:** Go's `encoding/json` decoder behavior for slices — from the [docs](https://pkg.go.dev/encoding/json#Unmarshal):
> *"To unmarshal JSON into a slice, Unmarshal resets the slice length to zero and then appends each element to the slice."*

"Resets slice length to zero" means `len = 0` but the **backing array is preserved**. When appending, existing backing-array memory is reused. Fields absent from the JSON are never written, so they retain the previous in-memory value from `DefaultConfig`.

**Fix:** Reset `cfg.ModelList = nil` before calling `json.Unmarshal` in `LoadConfig`. This forces a fresh allocation for all elements, eliminating the template value leak.

This bug affects **any user who omits `api_base` from a `model_list` entry** — which is the recommended and documented usage pattern. The impact is silent mis-routing to wrong provider endpoints.

mqyang56:
## Summary

Fixes #721

When `DefaultConfig()` pre-populates `ModelList` with ~18 template entries, Go's JSON decoder reuses the existing slice backing-array elements rather than zero-initializing them. This means fields absent from the user's JSON (e.g. `api_base`) silently inherit the template value at the same index position — for example, `ModelList[0]` is the Zhipu `glm-4.7` entry with `APIBase: "https://open.bigmodel.cn/api/paas/v4"`, which leaks into any user-defined model at index 0 that omits `api_base`.

**Fix**: pre-scan the JSON into a zero-value `Config` to count user-provided `model_list` entries. Only reset `cfg.ModelList = nil` when the user provides >0 entries, ensuring clean zero-value initialization during the real unmarshal. When the user provides 0 entries, the `DefaultConfig` built-in list is preserved as a fallback.

## Behavior change

| Scenario | Before | After |
|---|---|---|
| User `model_list` has 0 entries | DefaultConfig list used (correct) | DefaultConfig list used (correct) |
| User `model_list` has ≥1 entries, some fields omitted | Omitted fields leaked from DefaultConfig template | Omitted fields are zero-valued (correct) |

## Test plan

- [x] All existing `pkg/config` tests pass
- [x] Full project builds cleanly (`go build ./...`)
- Manual: set a single `model_list` entry with no `api_base` → confirm the model entry has empty `api_base` instead of the Zhipu template URL

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -499,6 +499,20 @@ func LoadConfig(path string) (*Config, error) {
 		return nil, err
 	}
 
+	// Pre-scan the JSON to check how many model_list entries the user provided.
+	// Go's JSON decoder reuses existing slice backing-array elements rather than
+	// zero-initializing them, so fields absent from the user's JSON (e.g. api_base)
+	// would silently inherit values from the DefaultConfig template at the same
+	// index position. We only reset cfg.ModelList when the user actually provides
+	// entries; when count is 0 we keep DefaultConfig's built-in list as fallback.
+	var tmp Config
+	if err := json.Unmarshal(data, &tmp); err != nil {
+		return nil, err
+	}
+	if len(tmp.ModelList) > 0 {
+		cfg.ModelList = nil
+	}
+
 	if err := json.Unmarshal(data, cfg); err != nil {
 		return nil, err
 	}


yinwm:
LGTM ✅

Orgmar:
@mqyang56 That DefaultConfig leak was a sneaky one. Slice reuse during JSON decode silently inheriting template values at the same index is the kind of bug that's easy to miss. Clean fix with the pre-scan approach.

We're building a **PicoClaw Dev Group** on Discord for contributors to connect and collaborate. If you're interested, drop an email to `support@sipeed.com` with the subject `[Join PicoClaw Dev Group] mqyang56` and we'll send the invite link your way.

PR #732, opened 2026-02-24
feat(memory): JSONL-backed session persistence

### LINKED ISSUE #711: [Feature] JSONL-backed session persistence with Store interface
## 🎯 The Goal / Use Case

The current session system (`pkg/session/manager.go`) stores everything in an in-memory map and serializes full JSON files to disk on every `Save()`. This works for simple cases, but breaks down in a few real scenarios:

1. **Race condition with summarization** — `maybeSummarize()` launches a background goroutine that calls `TruncateHistory()` + `Save()` while new messages are still being appended. This can corrupt persisted history and cause 400 errors on the next LLM call (see #704).
2. **Full serialization is expensive** — every `Save()` marshals the entire session to JSON and does a temp-file-rename dance. On embedded devices with limited I/O bandwidth, this adds latency that users can feel.
3. **No session lifecycle management** — there's no way to expire old sessions, no size limits, no cleanup. On a long-running gateway, `sessions/` just grows forever.
4. **No shared memory across agents** — the multi-agent work (#294, #423) will need some form of shared context pool. The current per-session JSON files can't support that.

I ran into the race condition issue (#704) myself while testing on a Pi, and when I dug into the session code I realized the problems go deeper than just adding a mutex.

## 💡 Proposed Solution

Replace the JSON file backend with SQLite, using a pure-Go driver (`modernc.org/sqlite` or `crawshaw.io/sqlite`) so we keep zero-CGo cross-compilation.

The idea is to keep `SessionManager`'s public API mostly the same so nothing upstream breaks, but swap the storage layer underneath.

Rough schema I'm thinking:

```sql
CREATE TABLE sessions (
    key TEXT PRIMARY KEY,
    summary TEXT DEFAULT '',
    created_at DATETIME,
    updated_at DATETIME
);

CREATE TABLE messages (
    id INTEGER PRIMARY KEY AUTOINCREMENT,
    session_key TEXT NOT NULL REFERENCES sessions(key),
    role TEXT NOT NULL,
    content TEXT,
    tool_calls TEXT,      -- JSON, nullable
    tool_call_id TEXT,    -- nullable
    created_at DATETIME DEFAULT CURRENT_TIMESTAMP
);

CREATE INDEX idx_messages_session ON messages(session_key, id);
```

Key benefits:
- `AddMessage` becomes an INSERT instead of full-file rewrite
- `TruncateHistory` + `Save` can run in a transaction — no more race condition
- Session expiry is just `DELETE FROM sessions WHERE updated_at < ?`
- Later, a `memory` table can hold long-term facts for multi-agent shared context

## 🛠 Potential Implementation

I'd break this into a few PRs:

**PR 1**: New `pkg/memory/` package with a `Store` interface + SQLite implementation + tests. Doesn't touch anything existing yet.

**PR 2**: Wire it into `AgentLoop` — replace `SessionManager.Save()` calls with the new store. Keep the JSON path as fallback (auto-detect: if `sessions/*.json` exists, migrate on first load).

**PR 3**: Add lifecycle management — TTL-based session expiry, configurable max message count per session, `/session stats` command or similar.

I want to keep the interface clean enough that someone could later swap in a different backend (Postgres for a server deployment, etc.) without touching agent code.

- [ ] This is a Core Feature

This was specifically mentioned in the 26M2W3 community meeting notes under the Agent board: "记忆系统：引入 SQLite". It also directly addresses #704 (race condition) and provides the foundation for #294 (multi-agent shared context).

## 🔄 Alternatives Considered

- **Just fix the race condition with a mutex**: Solves #704 but doesn't address the full-serialization cost or lifecycle management. Band-aid.
- **bbolt / badger (embedded KV stores)**: More overhead than SQLite for structured data. We need relational queries (messages by session, ordered by time) and SQLite is purpose-built for this.
- **Keep JSON but write incrementally (append-only log)**: Considered this. It handles the write performance, but makes truncation/expiry complicated and doesn't help with the shared memory use case.

## 💬 Additional Context

I've already had two PRs merged (#173, #186) and I'm familiar with the session/agent code paths. Happy to take this on if the direction looks right.

Would appreciate feedback from the agent board maintainers on:
1. Is `modernc.org/sqlite` acceptable, or is there a preference for another driver? (It adds ~3-4MB to the binary — need to check if that fits the 20M target from the meeting notes)
2. Should the migration from JSON be automatic, or a separate `picoclaw migrate` subcommand?
3. Any concerns about SQLite on the target embedded platforms?

is-Xiaoen:
## Summary

Implements `pkg/memory/` — a new session persistence layer using append-only JSONL files. This is a revised approach based on the feedback from #719, where @yinwm pointed out that JSONL fits this use case better than SQLite.

- **Zero new dependencies** — uses only stdlib + the project's existing `pkg/fileutil`
- **Append-only writes** — `AddMessage` is a single file append, no full-file rewrite
- **Logical truncation** — `TruncateHistory` updates a skip offset in `.meta.json` instead of rewriting the message file
- **Physical compaction** — `Compact` rewrites the JSONL file to reclaim disk space after repeated truncations
- **Agent-readable** — `.jsonl` files work directly with `read_file`, `tail`, `grep`, and agent skills
- **Crash-safe** — meta is always written before JSONL rewrites; file I/O uses `fileutil.WriteFileAtomic` (fsync + rename) for durability on flash storage; TruncateHistory always reconciles meta.Count against the actual file; corrupt lines are logged and skipped

### File layout

```
sessions/
├── telegram_123456.jsonl       # one message per line, append-only
├── telegram_123456.meta.json   # summary, skip offset, timestamps
```

### Store interface

The `Store` interface maps 1:1 to the current `SessionManager` API. Each method is an atomic operation — no separate `Save()` call. It returns structured data (messages + summary separately), which keeps the storage layer clean and lets the prompt builder handle provider-specific caching optimizations downstream.

```go
type Store interface {
    AddMessage(ctx, key, role, content) error
    AddFullMessage(ctx, key, msg) error
    GetHistory(ctx, key) ([]Message, error)
    GetSummary(ctx, key) (string, error)
    SetSummary(ctx, key, summary) error
    TruncateHistory(ctx, key, keepLast) error
    SetHistory(ctx, key, history) error
    Compact(ctx, key) error
    Close() error
}
```

### Crash safety design

The two-file design (`.jsonl` + `.meta.json`) introduces crash windows between the two writes. The ordering is carefully chosen to always degrade toward "more data visible" rather than data loss:

| Operation | Crash scenario | Effect | Recovery |
|---|---|---|---|
| `addMsg` | JSONL written, meta not updated | `meta.Count` stale by 1 | `TruncateHistory` always re-counts actual lines |
| `SetHistory` | Meta written (Skip=0), JSONL not rewritten | Old messages temporarily visible | Next `SetHistory` corrects |
| `Compact` | Meta written (Skip=0), JSONL not rewritten | Truncated messages reappear | Next `Compact`/`TruncateHistory` corrects |

All file rewrites (meta + JSONL compaction) use `fileutil.WriteFileAtomic` which fsyncs before rename — important for SD cards and flash storage on embedded devices.

### Concurrency

Session locking uses a fixed `[64]sync.Mutex` sharded array (FNV hash). Memory is O(1) regardless of total session count — important for a long-running daemon.

### Migration

`MigrateFromJSON()` reads legacy `sessions/*.json` files, writes them via `SetHistory` (atomic replace), and renames originals to `.json.migrated`. Idempotent — safe to retry after crash.

## Why JSONL over SQLite

Per @yinwm's review on #719:
- JSONL append is already atomic at the OS level — no transaction machinery needed
- Agent tools can directly read session files, which matters for an AI agent framework
- Zero external dependencies aligns with the "pico" philosophy
- Covers 90%+ of use cases; SQLite can be added later behind the same `Store` interface if complex queries are actually needed

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Test plan

33 tests total (all passing, 0 lint issues):

- **20 unit tests**: basic roundtrip, ordering, empty sessions, tool calls, tool call IDs, summary CRUD, truncation (keep N / keep 0 / keep more than exist / stale meta.Count), SetHistory replace + skip reset, colon-in-key filename mapping, crash recovery with partial lines, persistence across instances, session isolation
- **3 compaction tests**: compact removes skipped lines from disk, no-op when skip=0, append after compact works correctly
- **2 concurrency tests**: 10-goroutine concurrent writes; simulated #704 race (summarizer goroutine vs main loop)
- **3 benchmarks**: AddMessage throughput, GetHistory at 100 and 1000 messages
- **8 migration tests**: basic, tool calls, multiple files, invalid JSON skip, rename to .migrated, idempotent, colon-in-key, retry after crash (no duplicates)

```
$ go test ./pkg/memory/... -v -count=1
ok  	github.com/sipeed/picoclaw/pkg/memory	9.389s

$ golangci-lint run ./pkg/memory/...
0 issues.
```

## Scope

This PR only adds new files under `pkg/memory/` — no existing code is modified. Wiring into `AgentLoop` will be a separate PR.

Closes #711

### DIFF
--- a/pkg/memory/jsonl.go
+++ b/pkg/memory/jsonl.go
@@ -236,11 +236,19 @@ func (s *JSONLStore) addMsg(sessionKey string, msg providers.Message) error {
 		return fmt.Errorf("memory: open jsonl for append: %w", err)
 	}
 	_, writeErr := f.Write(line)
-	closeErr := f.Close()
 	if writeErr != nil {
+		f.Close()
 		return fmt.Errorf("memory: append message: %w", writeErr)
 	}
-	if closeErr != nil {
+	// Flush to physical storage before closing. This matches the
+	// durability guarantee of writeMeta and rewriteJSONL (which use
+	// WriteFileAtomic with fsync). Without Sync, a power loss could
+	// leave the append in the kernel page cache only — lost on reboot.
+	if syncErr := f.Sync(); syncErr != nil {
+		f.Close()
+		return fmt.Errorf("memory: sync jsonl: %w", syncErr)
+	}
+	if closeErr := f.Close(); closeErr != nil {
 		return fmt.Errorf("memory: close jsonl: %w", closeErr)
 	}
 


pkg/memory/jsonl.go
```diff
@@ -0,0 +1,439 @@
...
+//
+// Messages are never physically deleted from the JSONL file. Instead,
+// TruncateHistory records a "skip" offset in the metadata file and
+// GetHistory ignores lines before that offset. This keeps all writes
+// append-only, which is both fast and crash-safe.
+type JSONLStore struct {
+	dir string
+
+	mu    sync.Mutex
+	locks map[string]*sync.Mutex
```
Zhaoyikaiii: This is usually fine for small tools, but if picoclaw is handling tens of thousands of sessions, it's advisable to consider using an LRU cache to limit the number of locks, or adding a cleanup mechanism to the Close logic. A simpler approach is to directly use sync.Map's LoadOrStore.

pkg/memory/jsonl.go
```diff
@@ -0,0 +1,439 @@
...
+		return nil, fmt.Errorf("memory: open jsonl: %w", err)
+	}
+	defer f.Close()
+
+	var msgs []providers.Message
+	scanner := bufio.NewScanner(f)
+	// Allow up to 1 MB per line for messages with large content.
+	scanner.Buffer(make([]byte, 0, 64*1024), 1024*1024)
+
+	for scanner.Scan() {
```
Zhaoyikaiii, line 147: Currently, GetHistory performs an $O(N)$ scan of the entire JSONL file, which will degrade performance as session files grow. We can achieve $O(1)$ access by using byte offsets.1. Update Metadata SchemaAdd ActiveOffset to track the byte position of the first valid message.
```go
type sessionMeta struct {
    // ...
    ActiveOffset int64 `json:"active_offset"` 
}
```

2. Update TruncateHistoryCalculate the byte offset of the $N$-th message from the end during truncation.3. Optimize GetHistoryUse f.Seek to jump directly to the active conversation:Go// In GetHistory

```go
if meta.ActiveOffset > 0 {
    // Start scanning from here...
    if _, err := f.Seek(meta.ActiveOffset, io.SeekStart); err != nil {
        return nil, err
    }
}
```

Why: This prevents unnecessary CPU/Memory overhead from parsing discarded JSON lines, especially for long-lived sessions.

pkg/memory/jsonl.go
```diff
@@ -0,0 +1,439 @@
...
+	if err != nil {
+		_ = os.Remove(tmp)
+		return fmt.Errorf("memory: rename meta: %w", err)
+	}
+	return nil
+}
+
+// readMessages reads all valid JSON lines from a .jsonl file.
+// Malformed trailing lines (e.g. from a crash) are silently skipped.
+func readMessages(path string) ([]providers.Message, error) {
```
Zhaoyikaiii: or we could specify a offset as a param here

pkg/memory/jsonl.go
```diff
@@ -0,0 +1,439 @@
...
+
+	meta, err := s.readMeta(sessionKey)
+	if err != nil {
+		return err
+	}
+	if meta.Skip == 0 {
+		return nil
+	}
+
+	all, err := readMessages(s.jsonlPath(sessionKey))
```
Zhaoyikaiii: we could save last 20 messages.and using seek to skip the front messages.

pkg/memory/jsonl.go
is-Xiaoen: Good call — switched to `sync.Map` with `LoadOrStore`. Cleaner and removes the separate mutex entirely. Pushed in 5d73ee2.

pkg/memory/jsonl.go
is-Xiaoen: Makes sense. I went with a slightly different approach for now: added a `skip` parameter to `readMessages` so `GetHistory` and `Compact` both skip the first N lines without unmarshaling them. This avoids the CPU cost on truncated lines while keeping the implementation straightforward.

For the byte-offset `Seek` approach — I think that's a solid next step if sessions get really long. The trade-off right now is that `TruncateHistory` stays O(1) (just a metadata write), whereas computing byte offsets during truncation would make it O(N). Combined with `Compact` bounding the file size, the skip-scan approach should hold up well in practice. Happy to add `ActiveOffset` later if we see it becoming a bottleneck.

Pushed in 5d73ee2.

pkg/memory/jsonl.go
is-Xiaoen: Done — `readMessages` now takes a `skip int` parameter. Lines before the offset are scanned but not unmarshaled, saving the JSON parsing overhead. See 5d73ee2.

pkg/memory/jsonl.go
is-Xiaoen: Addressed — `Compact` now passes `meta.Skip` to `readMessages`, so the skipped front lines are scanned without unmarshaling. Same commit (5d73ee2).

is-Xiaoen:
**sync.Map → sharded lock array**: Replaced with a fixed `[64]sync.Mutex` pool, keys mapped via FNV hash. Memory is O(1) regardless of session count — no growth over the daemon's lifetime.

**Scanner buffer 1MB → 10MB**: Tool results (`read_file` on large files, web search dumps, etc.) can exceed 1MB easily. Bumped the cap to 10MB. The scanner still starts at 64KB and grows lazily, so normal messages don't pay for it.

is-Xiaoen:
这轮重点修了三个 crash safety 的问题：

**1. TruncateHistory 对账逻辑不够健壮**

之前只在 `meta.Count == 0` 时才去数文件行数。但 `addMsg` 写 JSONL 和写 meta 是两步操作，中间崩溃会导致 `meta.Count` 偏小（比如文件 101 行但 meta 说 100），非零的 stale count 直接被信任，`keepLast` 算出来的 skip 就会偏。

现在 TruncateHistory 每次都 countLines 对账，反正不是热路径，countLines 也只是扫行不 unmarshal。

**2. SetHistory / Compact 的写入顺序有 data loss 风险**

之前是先重写 JSONL 再更新 meta。如果在中间崩溃，meta 还保留着旧的 `Skip=90`，但 JSONL 文件只剩 10 行了——GetHistory 从第 91 行开始读，直接返回空。

调换了顺序：先写 meta（Skip=0），再重写 JSONL。中间崩溃的最坏情况是旧文件还在、meta 说 Skip=0，GetHistory 会多返回一些已截断的消息，但不会丢数据。下次操作自动修正。

**3. Migration 在崩溃重试时会重复写入**

之前迁移是 per-message 调 AddFullMessage，全部写完才 rename `.json` → `.json.migrated`。如果写了 50 条消息后崩溃（rename 还没执行），重试会把 100 条全部再写一遍，变成 150 条。

改成用 SetHistory 原子替换，重试直接覆盖而不是追加。

每个 fix 都有对应的测试用例验证。

is-Xiaoen:
@nikolasdehor Thanks for the thorough review — really helpful to have someone actually trace through the crash scenarios.

Responses inline:

**1. Lock shard collision** — Correct, no bug. Different sessions that collide on the same shard just serialize against each other unnecessarily. With 64 shards and typical session counts, this is a non-issue.

**2. meta.Count drift after crash in addMsg** — Good catch documenting this. As you noted, Count is only consumed by TruncateHistory (which re-counts), so the drift is harmless. But yeah, it's the kind of thing that could confuse someone reading the code without the crash analysis context.

**3. sanitizeKey lossy mapping** — Added a comment documenting this as intentional. The bidirectional alternative (URL-encoding etc.) would make file listings unreadable and complicate debugging, and key collisions between ":" and "_" variants are not a realistic concern for channel identifiers.

**4 & 5. No fsync in rewriteJSONL / addMsg** — Good point. Just pushed a commit that replaces the manual temp+rename in both `writeMeta` and `rewriteJSONL` with the project's `fileutil.WriteFileAtomic`, which does `Sync()` before rename + directory sync. This matters for SD cards and flash storage on the embedded devices picoclaw targets.

For `addMsg` appends, I intentionally left it without fsync — it's the hot path (called on every message), and the cost of fsync per append is significant. The worst case on power loss is losing the last appended message, which is acceptable for a session store. The crash-between-addMsg scenario (meta stale by 1) is already handled by TruncateHistory's reconciliation.

**6. Silent skip of corrupt lines** — Now logs a warning with the line number and filename: `memory: skipping corrupt line 2 in session.jsonl: unexpected end of JSON input`. Doesn't fail the entire read, but operators can see it happened.

Zhaoyikaiii:
LGTM

is-Xiaoen:
@Zhaoyikaiii Thanks! Could you submit a formal approval via the GitHub review UI when you get a chance? Looks like branch protection requires at least one approved review before merge. Also — is there anything else this needs to wait on, or is it good to go once approved?

alexhoshina:
PTAL @yinwm

is-Xiaoen:
@yinwm 感谢详细的 review！已修复，选择了 Option A — 在 `addMsg` 中添加 `f.Sync()`。

修改见 f9f726c：

```go
_, writeErr := f.Write(line)
if writeErr != nil {
    f.Close()
    return fmt.Errorf("memory: append message: %w", writeErr)
}
if syncErr := f.Sync(); syncErr != nil {
    f.Close()
    return fmt.Errorf("memory: sync jsonl: %w", syncErr)
}
if closeErr := f.Close(); closeErr != nil {
    return fmt.Errorf("memory: close jsonl: %w", closeErr)
}
```

现在三条写入路径的 fsync 行为完全一致：

| Method | fsync | Mechanism |
|--------|-------|-----------|
| `addMsg` | ✅ | `f.Sync()` before `f.Close()` |
| `writeMeta` | ✅ | `WriteFileAtomic` (内含 fsync) |
| `rewriteJSONL` | ✅ | `WriteFileAtomic` (内含 fsync) |

对于嵌入式设备（SD 卡存储），一致的 fsync 保证比性能更重要。`addMsg` 的 fsync 开销在可接受范围内 — 单次 append 的数据量通常只有几 KB，SD 卡的顺序写入延迟在 1-5ms 左右。

PR #734, opened 2026-02-24
fix(channels): fix memory hazards in channel abstraction layer

alexhoshina:
## 📝 Description
- Fix dispatcher busy-wait: remove select+default pattern that caused CPU spin after context cancellation; SubscribeOutbound handles ctx internally
- Add TTL janitor for typingStops/placeholders sync.Map entries to prevent unbounded accumulation when outbound paths fail
- Reduce queue buffers from 100 to 16 slots (~84% memory reduction)
- Optimize SplitMessage with index-based rune operations to reduce intermediate string/rune allocations
- Replace uuid.New() with atomic counter + random prefix for media scope IDs (eliminates per-call crypto/rand syscall)
- Lazy channel worker creation: defer goroutine+buffer allocation until channel.Start() succeeds

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

- **Reference URL:**
- **Reasoning:**

## 🧪 Test Environment
- **Hardware:** 
- **OS:** 
- **Model/Provider:** 
- **Channels:** 

## 📸 Evidence (Optional)
<details>

</details>

## ☑️ Checklist

### DIFF
--- a/pkg/channels/base.go
+++ b/pkg/channels/base.go
@@ -18,8 +18,8 @@ import (
 )
 
 var (
-	fastIDCounter uint64
-	fastIDPrefix  string
+	uniqueIDCounter uint64
+	uniqueIDPrefix  string
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


pkg/channels/base.go
```diff
@@ -14,6 +17,28 @@ import (
...
+	if _, err := rand.Read(b[:]); err != nil {
+		// fallback to time-based prefix
+		binary.BigEndian.PutUint64(b[:], uint64(time.Now().UnixNano()))
+	}
+	fastIDPrefix = hex.EncodeToString(b[:])
+}
+
+// fastID generates a unique ID using a random prefix and an atomic counter.
+// Much cheaper than uuid.New() which calls crypto/rand.Read on every invocation.
+func fastID() string {
```
xiaket: nit: I'm a little bit hesitant here as we should not judge a function. It should not have `fast` in its name. We should instead name it by what it does and how it works. For example, if we have used an alternative way to implement this id and it is more efficient, how should we call it?

Can I suggest we simply call it `uniqueID` instead?

xiaket: Also worth a comment that this id has nothing to do with security, and should never be used in contexts where unpredictability matters.

pkg/channels/split.go
xiaket, line 1: For changes in this file, we should probably have some tests.

PR #741, opened 2026-02-24
fix: enhance count argument handling in WebSearchTool to support string input

Xeven777:
This pull request updates the handling of the `count` argument in the `WebSearchTool.Execute` method to make it more robust. The main improvement is that the method can now accept both numeric and string values for the `count` parameter, converting strings to integers when possible.

**Improvements to argument handling:**

* Enhanced the parsing of the `count` argument in `WebSearchTool.Execute` to accept both `float64` and string types, converting string values to integers using `strconv.Atoi` and validating the range (1–10).
* Added the `strconv` package import in `pkg/tools/web.go` to support string-to-integer conversion.

### DIFF
--- a/pkg/tools/web.go
+++ b/pkg/tools/web.go
@@ -6,9 +6,11 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"math"
 	"net/http"
 	"net/url"
 	"regexp"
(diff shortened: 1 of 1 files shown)


Xeven777:
@xiaket review ?

PR #757, opened 2026-02-25
fix(cron): agent response not sent to channel after cron job

decker502:
## Description

Cron-triggered agent replies were silently discarded and never delivered to any channel (e.g. Telegram).

`ExecuteJob` calls `ProcessDirectWithChannel`, which bypasses the `AgentLoop` bus and invokes `processMessage` directly. The `Run()` loop's `PublishOutbound` — which is responsible for dispatching the response to the channel — never fires in this code path. The returned response was discarded with `_ = response`.

**Before:**
```go
_ = response // Will be sent by AgentLoop  ← never is
```

**After:**
```go
if response != "" {
    t.msgBus.PublishOutbound(bus.OutboundMessage{
        Channel: channel,
        ChatID:  chatID,
        Content: response,
    })
}
```

This is consistent with the `deliver=true` and command execution paths in the same function.

## Type of Change

## 🤖 AI Code Generation

- [x] 🤖 Fully AI-generated — AI (Claude) identified the root cause and wrote the fix; contributor reviewed and validated it

## Related Issue

N/A — identified from production logs showing agent response generated but not received on Telegram.

## Technical Context

Call chain: `CronTool.ExecuteJob` → `ProcessDirectWithChannel` → `processMessage` → `runAgentLoop` → returns `string`. The `Run()` loop that normally calls `PublishOutbound` is never involved when `ProcessDirectWithChannel` is used directly.

## 🧪 Test Environment
- **Hardware:** PC
- **OS:** Debian 13.02
- **Model/Provider:** Claude Sonnet 4.6
- **Channels:** Telegram

## Evidence

```
2026/02/25 10:10:32 [INFO] agent: Response: ... {agent_id=main, session_key=agent:main:main, iterations=1, final_length=369}
# → message never received on Telegram
```
## Checklist

- [x] I have read CONTRIBUTING.md
- [x] `make check` passes locally
- [x] My change is focused and does not bundle unrelated changes
- [x] I have disclosed AI involvement above

### DIFF
--- a/pkg/tools/cron.go
+++ b/pkg/tools/cron.go
@@ -327,7 +327,12 @@ func (t *CronTool) ExecuteJob(ctx context.Context, job *cron.CronJob) string {
 		return fmt.Sprintf("Error: %v", err)
 	}
 
-	// Response is automatically sent via MessageBus by AgentLoop
-	_ = response // Will be sent by AgentLoop
+	if response != "" {
+		t.msgBus.PublishOutbound(bus.OutboundMessage{
+			Channel: channel,
+			ChatID:  chatID,
+			Content: response,
+		})
+	}
 	return "ok"
 }


nikolasdehor (approved):
This is a well-analyzed bug fix with three focused improvements:

**1. Cron response delivery (the main fix):**
The root cause is clearly explained: `ProcessDirectWithChannel` bypasses the `AgentLoop` bus, so `PublishOutbound` never fires. The fix explicitly publishes the outbound message. This is correct and consistent with the `deliver=true` path mentioned in the PR description.

**2. Atomic cron store writes:**
The `saveStoreUnsafe` change to write-temp-then-rename is a good defensive fix. `os.Rename` is atomic on the same filesystem, preventing partial writes on crash. Minor note: on some filesystems and cross-device scenarios `os.Rename` can fail, but for `storePath + ".tmp"` this is always the same directory so it is fine.

**3. Telegram transport hardening:**
Adding TCP keepalive (30s) and explicit timeouts to the Telegram HTTP transport is a good improvement for long-poll stability. The refactoring to always use a custom transport (rather than only when proxy is configured) is cleaner.

One minor observation: the `response != ""` check in `ExecuteJob` means that if a cron job legitimately returns an empty response, it silently succeeds without any notification. That might be fine (no point sending an empty message), but worth documenting.

LGTM — clean, focused, well-documented fix.

Issue #762, opened 2026-02-25
exec tool safety guard false-positive: domain URL without scheme blocked as path outside working dir


westwind027:
## Summary
`exec` tool rejects valid network commands like `curl -s "wttr.in/Nanjing?format=3"` with `Command blocked by safety guard (path outside working dir)`.

## Logs
- Tool call: `exec({"command":"curl -s \"wttr.in/Nanjing?format=3\""})`
- Error: `Command blocked by safety guard (path outside working dir)`

## Reproduction
1. Configure `exec` tool with workspace restriction enabled.
2. Run command containing URL without scheme, e.g. `wttr.in/Nanjing?format=3`.
3. Command is blocked before execution.

## Expected
URL-like strings should not be treated as filesystem absolute paths.

## Actual
Guard detects `/Nanjing?...` segment as absolute local path and blocks command.

## Suspected root cause
Path regex runs after only stripping `scheme://...` URLs. Domain URLs without scheme are not stripped, so embedded `/...` is misclassified as local path.

westwind027:
Fixed and pushed.

- PR: https://github.com/sipeed/picoclaw/pull/763
- Commit: https://github.com/westwind027/picoclaw/commit/b704ccd

This updates `exec` safety guard to avoid treating domain URLs without scheme (e.g. `wttr.in/...`) as filesystem absolute paths, with regression tests.

Issue #761, opened 2026-02-25
QQ channel: GROUP_AT_MESSAGE_CREATE wrongly sent to C2C /v2/users API (11255 invalid request)


westwind027:
## Summary
QQ group @ messages (`GROUP_AT_MESSAGE_CREATE`) are being replied via C2C endpoint (`/v2/users/{id}/messages`) instead of group endpoint, causing API error `11255 invalid request`.

## Logs
- Event received: `GROUP_AT_MESSAGE_CREATE`
- Outbound request: `POST https://api.sgroup.qq.com/v2/users/E7AD29.../messages`
- Response: `{"message":"invalid request","code":11255,"err_code":11255}`

## Reproduction
1. Enable QQ channel and receive a group @ message.
2. Agent generates a normal text response.
3. Observe outbound API call and error logs.

## Expected
For group chat replies, use QQ group message endpoint (`PostGroupMessage` / `/v2/groups/{group_id}/messages`) with proper passive-reply fields.

## Actual
System posts to C2C user endpoint (`PostC2CMessage` / `/v2/users/{user_id}/messages`), request gets rejected.

## Suspected root cause
Send path only uses `ChatID` but does not preserve chat kind (group/direct), defaulting to C2C send API.

## Additional note
Emoji/content is not the root cause; plain text requests also fail with same `11255` when endpoint is mismatched.

westwind027:
Fixed and pushed.

- PR: https://github.com/sipeed/picoclaw/pull/763
- Commit: https://github.com/westwind027/picoclaw/commit/0eade09

This changes QQ outbound routing to use group endpoint for group chats and keeps passive-reply context (`msg_id`/`msg_seq`) per chat.

PR #763, opened 2026-02-25
fix: qq group reply routing and exec URL path-guard false positives

westwind027:
This PR includes fixes for #761 and #762.

- fix(qq): route group replies to group API and preserve reply context
- fix(exec): avoid URL false positives in workspace path guard

Commits:
- 0eade09
- b704ccd

### DIFF
--- a/pkg/tools/shell.go
+++ b/pkg/tools/shell.go
@@ -289,10 +289,29 @@ func (t *ExecTool) guardCommand(command, cwd string) string {
 			return ""
 		}
 
-		pathPattern := regexp.MustCompile(`[A-Za-z]:\\[^\\\"']+|/[^\s\"']+`)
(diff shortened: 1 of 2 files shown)


alexhoshina:
This should be two pull requests, not one, as it involves two different modules.
The channel system is currently being refactored. If convenient, you can develop based on the [refactor branch](https://github.com/sipeed/picoclaw/tree/refactor/channel-system). Thank you.

alexhoshina:
A PR should only contain modifications related to one module; we might close this PR

xiaket:
Thanks for your contribution! I will close this PR as we should not have two bug fixes within the same PR.

xiaket:
Please feel free to split this into two PRs and we will be happy to review them!

PR #790, opened 2026-02-25
fix: exclude prompt_cache_key for Gemini API requests

rordd:
## Problem

Gemini's OpenAI-compatible endpoint (`generativelanguage.googleapis.com`) rejects unknown fields in the request body. The `prompt_cache_key` field (added in #617) causes a 400 error:

```
Invalid JSON payload received. Unknown name "prompt_cache_key": Cannot find field.
```

## Solution

Only include `prompt_cache_key` in the request body for OpenAI-native endpoints. Skip it when the API base URL contains `generativelanguage.googleapis.com`.

## Testing

- Verified Gemini 2.5 Flash no longer returns 400 errors
- OpenAI endpoints continue to receive `prompt_cache_key` as before

### DIFF
--- a/pkg/providers/openai_compat/provider.go
+++ b/pkg/providers/openai_compat/provider.go
@@ -115,8 +115,12 @@ func (p *Provider) Chat(
 	// with the same key and reuse prefix KV cache across calls.
 	// The key is typically the agent ID — stable per agent, shared across requests.
 	// See: https://platform.openai.com/docs/guides/prompt-caching
+	// Prompt caching is only supported by OpenAI-native endpoints.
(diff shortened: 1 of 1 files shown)


Orgmar:
@rordd That Gemini compatibility fix is solid. The prompt_cache_key field silently breaking requests on generativelanguage.googleapis.com is the kind of edge case that's easy to overlook when adding OpenAI-specific features. Clean conditional approach.

We have a **PicoClaw Dev Group** on Discord where contributors share ideas and collaborate. If you're interested, send an email to `support@sipeed.com` with the subject `[Join PicoClaw Dev Group] rordd` and we'll get you the invite link.

PR #728, opened 2026-02-24
feat(media): integrate TTL cleanup into FileMediaStore

ex-takashima:
## Summary

- Integrate TTL-based background cleanup (L2 safety net) directly into `FileMediaStore`, so file deletion and in-memory ref removal happen atomically under the same mutex — no dangling references
- Add `MediaCleanupConfig` (`enabled`, `max_age_minutes`, `interval_minutes`) to `tools.media_cleanup` config section with sensible defaults (enabled, 30min max age, 5min interval)
- Wire up `NewFileMediaStoreWithCleanup` + `Start()`/`Stop()` in `cmd_gateway.go`

### Design decisions

- **L1/L2 hybrid**: `ReleaseAll(scope)` remains the primary (L1) cleanup path for scope-based immediate deletion. `CleanExpired()` acts as L2 safety net for leaked entries
- **Atomic cleanup**: Both file removal and in-memory ref/scope cleanup happen under a single `sync.RWMutex` lock, preventing races between cleanup and resolve
- **`refToScope` reverse map**: Enables `CleanExpired()` to find and clean scope mappings without scanning all scopes
- **`nowFunc` injection**: Allows deterministic time control in tests without `time.Sleep`

### Files changed

| File | Change |
|------|--------|
| `pkg/media/store.go` | `storedAt`, `refToScope`, `MediaCleanerConfig`, `CleanExpired()`, `Start()`/`Stop()` |
| `pkg/media/store_test.go` | 8 new tests (expiry, mixed ages, empty scopes, lifecycle, concurrency, consistency) |
| `pkg/config/config.go` | `MediaCleanupConfig` struct + `ToolsConfig.MediaCleanup` field |
| `pkg/config/defaults.go` | Default: `{Enabled: true, MaxAge: 30, Interval: 5}` |
| `cmd/picoclaw/cmd_gateway.go` | Wiring: `NewFileMediaStoreWithCleanup` + `Start()`/`Stop()` |

## Test plan

- [x] `go test ./pkg/media/ -v` — 15/15 PASS (7 existing + 8 new)
- [x] `go test ./pkg/config/ -v` — all PASS
- [ ] `go test ./pkg/media/ -race -v` — needs cgo-enabled environment (CI)
- [ ] `go build ./cmd/picoclaw/` — pre-existing `cmd_onboard.go` embed error on this branch

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #811, opened 2026-02-26
feat(line): add StartTyping and PlaceholderRecorder integration

ex-takashima:
## Summary

- Implement `TypingCapable` interface for LINE channel using the loading animation API (1:1 chats only)
- Integrate `PlaceholderRecorder.RecordTypingStop` in `processEvent` for Manager coordination
- Skip `RecordPlaceholder` — LINE has no message edit API, so placeholder messages would be uneditable
- Relax `callAPI` status check from `200` to `2xx` range (loading API may return `202`)

### LINE API constraints

| Feature | LINE Support | Implementation |
|---------|-------------|----------------|
| Typing indicator | Loading animation (1:1 only) | `StartTyping()` with 50s refresh |
| Message edit | Not supported | Not implemented |
| Message delete | Not supported (user-only unsend) | Not implemented |

### Changes

- `StartTyping()` — new method, periodic 50s refresh goroutine, context-cancel to stop
- `sendLoading()` — now accepts `context.Context` and returns `error`
- `processEvent()` — replaced direct `sendLoading` call with `StartTyping` + `PlaceholderRecorder` integration
- `callAPI()` — `200` check relaxed to `2xx`

Split from #728 per @alexhoshina's request to keep that PR focused on MediaStore.

## Test plan

- [ ] Verify loading animation appears in 1:1 LINE chats
- [ ] Verify no crash in group chats (typing skipped gracefully)
- [ ] Verify goroutine cleanup on context cancel
- [ ] CI passes (linter + tests)

### DIFF
--- a/pkg/channels/line/line.go
+++ b/pkg/channels/line/line.go
@@ -380,25 +380,23 @@ func (c *LINEChannel) processEvent(event lineEvent) {
 
 	// Thinking indicator (LINE loading animation is 1:1 only).
 	// For group/room chats, LINE provides no equivalent API.
+	// Only start if PlaceholderRecorder is available to avoid wasted API calls.
 	if !isGroup {
-		typingCtx, typingCancel := context.WithTimeout(c.ctx, 5*time.Minute)
-		stop, err := c.StartTyping(typingCtx, chatID)
-		if err == nil {
-			var stopOnce sync.Once
-			stopFn := func() {
-				stopOnce.Do(func() {
-					stop()
-					typingCancel()
-				})
-			}
-			if rec := c.GetPlaceholderRecorder(); rec != nil {
+		if rec := c.GetPlaceholderRecorder(); rec != nil {
+			typingCtx, typingCancel := context.WithTimeout(c.ctx, 5*time.Minute)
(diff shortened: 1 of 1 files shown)


pkg/channels/line/line.go
```diff
@@ -577,17 +598,50 @@ func (c *LINEChannel) sendPush(ctx context.Context, to, content, quoteToken stri
...
+
+	ticker := time.NewTicker(50 * time.Second)
+	go func() {
+		defer ticker.Stop()
+		for {
+			select {
+			case <-typingCtx.Done():
+				return
+			case <-ticker.C:
+				_ = c.sendLoading(typingCtx, chatID)
```
alexhoshina: The error here was discarded; there should at least be a log record

ex-takashima:
@alexhoshina Thanks for the review! Fixed:

- **`sendLoading` error logging** — ticker goroutine now logs refresh failures via `logger.DebugCF` instead of discarding
- **Skip typing without recorder** — `StartTyping` is now only called when `PlaceholderRecorder` is available, avoiding wasted API calls and unnecessary goroutine creation

Ready for re-review.

PR #833, opened 2026-02-26
fix: return fetched content to LLM in web_fetch tool

lucamartinetti:
## Summary

- `WebFetchTool.Execute` was setting `ForLLM` to a summary string (`"Fetched N bytes from URL (extractor: ..., truncated: ...)"`) instead of the actual extracted text
- This was introduced in ca781d4 when the Tool interface was refactored from `(string, error)` to `*ToolResult` — the old code returned the full content as a single string, but the split into `ForLLM`/`ForUser` accidentally hid the content from the LLM
- The model never saw the fetched page content and could not answer questions based on web fetches

## Fix

Return the extracted `text` in `ForLLM` (the field the LLM actually reads) instead of the metadata summary. `ForUser` continues to return the full JSON result.

## Test plan

- [x] Deployed in a sandboxed container and verified `web_fetch` now returns page content the LLM can use (tested with weather data queries)

### DIFF
--- a/pkg/tools/web.go
+++ b/pkg/tools/web.go
@@ -652,8 +652,14 @@ func (t *WebFetchTool) Execute(ctx context.Context, args map[string]any) *ToolRe
 	resultJSON, _ := json.MarshalIndent(result, "", "  ")
 
 	return &ToolResult{
-		ForLLM:  text,
-		ForUser: string(resultJSON),
+		ForLLM: string(resultJSON),
+		ForUser: fmt.Sprintf(
+			"Fetched %d bytes from %s (extractor: %s, truncated: %v)",
+			len(text),
+			urlStr,
+			extractor,
+			truncated,
+		),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


afjcjsbx:
great bug you found! but instead of replacing ForLLM with text I would do something like this:

```go
	return &ToolResult{
		ForLLM: string(resultJSON),
		ForUser: fmt.Sprintf(
			"Fetched %d bytes from %s (extractor: %s, truncated: %v)",
			len(text),
			urlStr,
			extractor,
			truncated,
		),
	}
```

there are also unit tests to fix, but nice PR!

lucamartinetti:
Thanks @afjcjsbx, good call! Applied your suggestion — ForLLM now gets the full JSON result and ForUser gets the summary. Also updated the tests to match.

afjcjsbx:
LGTM!

PixelTux (approved):
LGTM

afjcjsbx:
@lxowalle PTAL

afjcjsbx:
@xiaket PTAL

Orgmar:
@lucamartinetti Good catch on the web_fetch bug. Returning the summary metadata string instead of the actual fetched content to the LLM would make the tool basically useless for real web lookups. Subtle regression from the ToolResult refactor.

We have a **PicoClaw Dev Group** on Discord where contributors chat and collaborate. If you're interested, send an email to `support@sipeed.com` with the subject `[Join PicoClaw Dev Group] lucamartinetti` and we'll send you the invite link.

PR #867, opened 2026-02-27
fix: use Anthropic SDK provider for anthropic protocol with API key

echowxsy:
## Summary

- Fix `anthropic` protocol in `model_list` to use the real Anthropic SDK provider (`NewClaudeProviderWithBaseURL`) instead of the OpenAI-compatible HTTP provider
- The previous code sent OpenAI-format requests to Anthropic-compatible endpoints, causing 404 errors

## Problem

When using `model_list` with `anthropic/` protocol and an API key:

```json
{
  "model_name": "glm-4.7",
  "model": "anthropic/glm-4.7",
  "api_key": "your-key",
  "api_base": "https://api.z.ai/api/anthropic/v1"
}
```

`CreateProviderFromConfig` fell into the `case "anthropic"` branch (line 114) but created an `HTTPProvider` (OpenAI-compatible), which sent `/v1/chat/completions` format requests to an Anthropic Messages API endpoint → 404.

## Fix

Replace `NewHTTPProviderWithMaxTokensFieldAndRequestTimeout` with `NewClaudeProviderWithBaseURL` for the API key path, so it uses the official Anthropic SDK which sends proper `/v1/messages` format requests.

## Testing

Tested on a MaixCAM (RISC-V) device with Zhipu GLM coding plan endpoint (`https://api.z.ai/api/anthropic/v1`):

```
Before: 404 Not Found
After:  🦞 2 + 2 = 4 ✅
```

### DIFF
--- a/pkg/providers/factory_provider_test.go
+++ b/pkg/providers/factory_provider_test.go
@@ -152,6 +152,48 @@ func TestCreateProviderFromConfig_Anthropic(t *testing.T) {
 	if modelID != "claude-sonnet-4.6" {
 		t.Errorf("modelID = %q, want %q", modelID, "claude-sonnet-4.6")
 	}
+	// Should use ClaudeProvider (Anthropic SDK), not HTTPProvider
+	if _, ok := provider.(*ClaudeProvider); !ok {
+		t.Errorf("expected *ClaudeProvider, got %T", provider)
+	}
+}
+
+func TestCreateProviderFromConfig_AnthropicWithCustomAPIBase(t *testing.T) {
+	cfg := &config.ModelConfig{
+		ModelName: "test-anthropic-custom",
+		Model:     "anthropic/glm-4.7",
+		APIKey:    "test-key",
+		APIBase:   "https://api.z.ai/api/anthropic/v1",
+	}
+
+	provider, modelID, err := CreateProviderFromConfig(cfg)
+	if err != nil {
+		t.Fatalf("CreateProviderFromConfig() error = %v", err)
+	}
+	if provider == nil {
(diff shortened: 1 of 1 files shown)


echowxsy:
Thanks for the thorough review! Addressing each point:

**Complementary to PR #444**: Agreed — this PR fixes `CreateProviderFromConfig` (model_list path) while #444 fixes `CreateProvider` (single-provider path). Both should be merged for complete coverage.

**Dropped parameters (Proxy, MaxTokensField, RequestTimeout)**: This is intentional and consistent with the existing OAuth/token auth path (lines 115-121), which also uses `ClaudeProvider` without these parameters. The Anthropic SDK manages HTTP internally via `anthropic-sdk-go` and doesn't expose proxy/timeout at the constructor level. If proxy support for Anthropic protocol becomes needed, it would require passing `option.WithHTTPClient()` with a custom transport — that could be a separate enhancement PR.

**Empty apiBase**: Verified — `normalizeBaseURL("")` returns `defaultBaseURL` (`"https://api.anthropic.com"`), so empty `api_base` correctly falls back to the official Anthropic API endpoint.

**Tests**: Added in `3e1e2b7`:
- `TestCreateProviderFromConfig_Anthropic` — updated to verify `*ClaudeProvider` type (not `*HTTPProvider`)
- `TestCreateProviderFromConfig_AnthropicWithCustomAPIBase` — tests custom `api_base` (e.g., Zhipu coding plan endpoint)
- `TestCreateProviderFromConfig_AnthropicMissingAPIKey` — verifies proper error message

All provider tests pass: `go test -tags stdjson ./pkg/providers/` ✅

PR #730, opened 2026-02-24
fix: multi-tool-call results incorrectly dropped causing LLM API 400

winterfx:
## 📝 Description

When an assistant message contains multiple tool calls (e.g. ToolCalls: ["A", "B"]), the second and subsequent tool results were incorrectly dropped by sanitizeHistoryForProvider. The old code only checked the immediate predecessor, which for the second tool result was another tool  result — not the assistant message — causing it to be treated as orphaned and removed. This led to a mismatch between assistant tool calls and tool results, resulting in LLM API 400 errors.

The fix walks backwards over preceding tool messages to find the nearest assistant with ToolCalls. Added unit tests covering key edge cases.

## 🗣️ Type of Change
- [x] 🐞 Bug fix (non-breaking change which fixes an issue)
- [ ] ✨ New feature (non-breaking change which adds functionality)
- [ ] 📖 Documentation update
- [ ] ⚡ Code refactoring (no functional changes, no api changes)

## 🤖 AI Code Generation
- [ ] 🤖 Fully AI-generated (100% AI, 0% Human)
- [x] 🛠️ Mostly AI-generated (AI draft, Human verified/modified)
- [ ] 👨‍💻 Mostly Human-written (Human lead, AI assisted or none)

## 🔗 Related Issue

#704 
## 📚 Technical Context (Skip for Docs)
  - Reference URL: N/A
  - Reasoning: sanitizeHistoryForProvider used last := sanitized[len(sanitized)-1] to check if a tool result has a matching assistant. For multi-tool-call, tool results are consecutive([assistant(A,B), tool(A), tool(B)]), so tool(B)'s immediate predecessor is tool(A), not the assistant. The fix uses a backward loop that skips over tool messages to locate the originating assistant.

## 🧪 Test Environment
 - Hardware: Mac
 - OS: macOS 26
 - Model/Provider: OpenAI/Gemini/Moonshot AI
 - Channels: N/A

## 📸 Evidence (Optional)
<details>
<summary>Click to view Logs/Screenshots</summary>

</details>

## ☑️ Checklist
- [x] My code/docs follow the style of this project.
- [x] I have performed a self-review of my own changes.
- [ ] I have updated the documentation accordingly.

Issue #475, opened 2026-02-19
[BUG]: History compression orphans tool_call/tool_result pairs, causing Anthropic API errors


felipe9962:
# Bug: History compression orphans tool_call/tool_result pairs, causing Anthropic API errors

## Problem

When the context window fills up, both `forceCompression()` and `summarizeSession()` in `pkg/agent/loop.go` can split tool_call/tool_result pairs, leaving orphaned messages. Anthropic (and other providers) require that every `assistant` message with `tool_calls` has matching `tool` role responses, and vice versa. When pairs are broken, the API returns:

```
tool_use ids were provided that do not have a tool_use block
```

### Root Causes

**1. `forceCompression()` (line ~723)** cuts conversation at the midpoint without checking for tool_call/tool_result boundaries:

```go
mid := len(conversation) / 2
keptConversation := conversation[mid:]
```

If `mid` falls between an `assistant` message with `ToolCalls` and its corresponding `tool` result messages, the result messages become orphaned.

**2. `summarizeSession()` (line ~861)** keeps the last 4 messages via `TruncateHistory(key, 4)`, but those 4 messages may start with orphaned `tool` results:

```go
agent.Sessions.TruncateHistory(sessionKey, 4)
```

**3. Existing partial fix in `context.go` (line ~196, by Diegox-17)** only strips orphaned `tool` messages from the very start of history:

```go
for len(history) > 0 && (history[0].Role == "tool") {
    history = history[1:]
}
```

This doesn't handle orphaned `assistant` messages with `ToolCalls` at the end, or broken pairs in the middle after compression.

## Proposed Fix

Add a `sanitizeToolPairs()` helper that ensures tool_call ↔ tool_result integrity after any history modification. Apply it in both compression paths.

### Helper function

```go
// sanitizeToolPairs ensures every assistant message with ToolCalls has
// matching tool results, and every tool result has its preceding tool_call.
// Orphaned messages are removed to prevent provider API errors.
func sanitizeToolPairs(messages []providers.Message) []providers.Message {
    // Build set of tool_call IDs present in assistant messages
    toolCallIDs := make(map[string]bool)
    for _, m := range messages {
        if m.Role == "assistant" && len(m.ToolCalls) > 0 {
            for _, tc := range m.ToolCalls {
                toolCallIDs[tc.ID] = true
            }
        }
    }

    // Build set of tool_result IDs present
    toolResultIDs := make(map[string]bool)
    for _, m := range messages {
[...]

PR #871, opened 2026-02-27
fix: preserve tool call/response pairing in forceCompression

### LINKED ISSUE #870: [Bug] forceCompression breaks tool call/response pairing, causing 400 error
## Description

The `forceCompression` function in `pkg/agent/loop.go` (lines 771-820) aggressively drops the oldest 50% of messages when context limit is hit. However, it does not preserve the pairing between tool calls and tool responses, leading to orphaned `tool` role messages that cause API 400 errors.

## Error Message

```
Status: 400
Body:   {"error":{"code":"invalid_parameter_error","message":"<400> InternalError.Algo.InvalidParameter: messages with role \"tool\" must be a response to a preceeding message with \"tool_calls\".","param":null,"type":"invalid_request_error"},"request_id":"99d697db-271a-98c4-a6b4-41937d43e224"}, retry=0}
```

## Root Cause

In `forceCompression`:

```go
conversation := history[1 : len(history)-1]
mid := len(conversation) / 2
keptConversation := conversation[mid:]  // Simply cuts from the middle!
```

The code cuts messages from the middle without checking if:
1. A `tool` role message has a corresponding `assistant` message with `tool_calls` before it
2. The cut point breaks a tool call/response pair

### Example Scenario

History before compression:
```
[0] system: "You are..."
[1] user: "What's the weather?"
[2] assistant: [tool_calls: get_weather]
[3] tool: "Sunny, 25°C"
[4] assistant: "It's sunny..."
[5] user: "Thanks!"
```

After `forceCompression` (mid=2, keeps [3], [4]):
```
[0] system: "... [compression note]"
[1] tool: "Sunny, 25°C"      // ❌ Orphaned! No preceding assistant with tool_calls
[2] assistant: "It's sunny..."
[3] user: "Thanks!"
```

This results in a `tool` message without a preceding `assistant` message containing `tool_calls`, which violates API requirements.

## Affected Providers

- OpenAI-compatible APIs (including Zhipu/GLM)
- Anthropic
- Any provider that enforces strict tool call/response pairing

## Suggested Fix

Option 1: **Skip tool messages when finding cut point**
```go
// Find a safe cut point that doesn't break tool call/response pairs
func findSafeCutPoint(conversation []providers.Message) int {
    mid := len(conversation) / 2
    
    // Walk backwards from mid to find a user message (safe cut point)
    for i := mid; i >= 0; i-- {
        if conversation[i].Role == "user" {
            return i + 1 // Cut after user message
        }
    }
    
    // Walk forwards from mid to find a user message
    for i := mid; i < len(conversation); i++ {
        if conversation[i].Role == "user" {
            return i + 1
        }
    }
    
    return mid // Fallback (may still cause issues)
}
```

Option 2: **Drop tool call/response pairs together**
When dropping messages, ensure that if an `assistant` message with `tool_calls` is dropped, all corresponding `tool` responses are also dropped.

Option 3: **Remove orphaned tool messages after compression**
Add a sanitization step after compression to remove `tool` messages without preceding `assistant[tool_calls]`.

## Environment

- PicoClaw version: latest
- Provider: Zhipu GLM (also affects other OpenAI-compatible providers)

## Related Issues

- #704 Race condition in session history causes similar error
- #760 Codex provider fails with 400 for multiple tool calls

QuietyAwe:
## Problem

The `forceCompression` function was cutting messages from the middle without considering tool call/response pairs, causing orphaned `tool` role messages that trigger API 400 errors:

```
"messages with role 'tool' must be a response to a preceeding message with 'tool_calls'"
```

## Root Cause

When context limit is hit, `forceCompression` would simply cut at `mid = len(conversation) / 2`, which could break a tool call/response pair:

**Before compression:**
```
[0] user: "What's the weather?"
[1] assistant: [tool_calls: get_weather]
[2] tool: "Sunny, 25°C"
[3] assistant: "It's sunny..."
[4] user: "Thanks!"
```

**After compression (mid=2, keeps [2], [3], [4]):**
```
[0] system: "..."
[1] tool: "Sunny, 25°C"      // ❌ Orphaned! No preceding assistant with tool_calls
[2] assistant: "It's sunny..."
[3] user: "Thanks!"
```

## Solution

1. **`findSafeCutPoint()`** - Finds a safe cut point after a user message, ensuring tool call/response pairs are kept together. User messages are always safe to cut after because they don't have tool call dependencies.

2. **`removeOrphanedToolMessages()`** - Safety net that removes any orphaned tool messages at the start of the kept conversation (edge case handling).

## Testing

Added comprehensive unit tests:
- `TestFindSafeCutPoint` - Tests various scenarios for finding safe cut points
- `TestRemoveOrphanedToolMessages` - Tests orphaned message removal
- `TestForceCompressionPreservesToolPairs` - Integration test for tool pair preservation

All tests pass ✅

Fixes #870

### DIFF
--- a/pkg/agent/loop.go
+++ b/pkg/agent/loop.go
@@ -803,6 +803,9 @@ func (al *AgentLoop) forceCompression(agent *AgentInstance, sessionKey string) {
 	// Additional safety: remove orphaned tool messages at the start of kept conversation
 	keptConversation = removeOrphanedToolMessages(keptConversation)
 
+	// Additional safety: remove orphaned assistant messages with tool_calls at the end
+	keptConversation = removeOrphanedAssistantWithToolCalls(keptConversation)
+
 	newHistory := make([]providers.Message, 0, 1+len(keptConversation)+1)
 
 	// Append compression note to the original system prompt instead of adding a new system message
@@ -831,6 +834,7 @@ func (al *AgentLoop) forceCompression(agent *AgentInstance, sessionKey string) {
 
 // findSafeCutPoint finds a safe index to cut the conversation without breaking tool call/response pairs.
 // It starts from the mid-point and searches forward for a user message, which is always safe to cut after.
+// If no user message is found, it falls back to finding a safe cut after the last complete tool call/response sequence.
 func findSafeCutPoint(conversation []providers.Message, mid int) int {
 	// Search forward from mid to find a user message
 	for i := mid; i < len(conversation); i++ {
@@ -846,7 +850,42 @@ func findSafeCutPoint(conversation []providers.Message, mid int) int {
 		}
 	}
 
-	// No user message found (edge case), use mid but this may cause issues
+	// No user message found (edge case): find a safe cut after the last complete tool sequence
+	// A safe cut is after all tool results for a tool call, i.e., after a non-tool message
+	// that doesn't have tool_calls, or after the last tool result of a complete sequence.
+	for i := mid; i < len(conversation); i++ {
+		// Find a position after all consecutive tool messages
+		if conversation[i].Role != "tool" {
+			// Check if this is an assistant without tool_calls (safe cut point)
+			// or if we need to skip past any tool results
+			if conversation[i].Role == "assistant" && len(conversation[i].ToolCalls) == 0 {
+				return i + 1 // Cut after this assistant message
+			}
+			// If it's an assistant with tool_calls, we need to find the end of the tool results
+			if conversation[i].Role == "assistant" && len(conversation[i].ToolCalls) > 0 {
+				// Count how many tool results we expect
+				expectedResults := len(conversation[i].ToolCalls)
+				resultCount := 0
+				for j := i + 1; j < len(conversation) && resultCount < expectedResults; j++ {
+					if conversation[j].Role == "tool" {
+						resultCount++
+					}
+				}
+				// Cut after all tool results
+				if resultCount == expectedResults {
+					// Find the position after the last tool result
+					for j := i + expectedResults; j < len(conversation); j++ {
+						if conversation[j].Role != "tool" {
+							return j
+						}
+					}
+					return len(conversation)
+				}
+			}
+		}
+	}
+
+	// Ultimate fallback: use mid (may cause issues, but removeOrphanedToolMessages will help)
 	return mid
 }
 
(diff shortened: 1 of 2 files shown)


nikolasdehor (commented):
Good approach with findSafeCutPoint — finding a user message boundary is a clean prevention strategy. A few observations:

1. **Coverage gap**: This fix only covers `forceCompression()` but the same bug exists in `summarizeSession()` via `TruncateHistory(key, 4)`. PR #665 covers both paths.

2. **Fallback to mid**: `findSafeCutPoint` falls back to `mid` when no user message is found, which reintroduces the original bug. Edge case: conversations with very few user messages (e.g., one long user prompt followed by many tool calls).

3. **Orphaned assistant messages**: `removeOrphanedToolMessages` strips leading `tool` role messages but does not handle orphaned `assistant` messages with `ToolCalls` at the end of the kept conversation. These would cause the provider to expect tool results that were cut away.

4. **PR #730**: Already merged, addresses the same class of bug in `sanitizeHistoryForProvider()`. Consider whether this PR and #665 are still needed as additional layers of defense.

5. **Issue #870 vs #475**: Issue #870 appears to be a duplicate of #475 (filed 2026-02-19), which is the issue our PR #665 closes.

QuietyAwe:
Thanks for the thorough review @nikolasdehor! Great points. Let me address each:

**1. Coverage gap & PR #665:**
You're right that `summarizeSession()` has the same issue via `TruncateHistory()`. I'll review #665 to see if it covers both paths comprehensively. If so, this PR can serve as an additional defense layer focused specifically on `forceCompression()`.

**2. Fallback to `mid`:**
Agreed this is a theoretical risk. In practice, conversations almost always have user messages, so the fallback should rarely trigger. However, I can add an additional check to ensure the fallback cut point doesn't orphan tool pairs — perhaps by extending `findSafeCutPoint` to also consider cutting after the last complete tool call/response sequence.

**3. Orphaned assistant messages with ToolCalls:**
Good catch! If we cut away tool responses but keep an assistant message with `ToolCalls`, the provider will expect results that don't exist. I'll add `removeOrphanedAssistantWithToolCalls()` to handle this edge case.

**4. PR #730:**
Good to know #730 already addresses this in `sanitizeHistoryForProvider()`. These fixes can coexist as defense-in-depth — #730 sanitizes before provider calls, while this PR prevents bad cuts at the source.

**5. Issue #870 vs #475:**
Thanks for pointing this out. I'll reference #475 in the PR description and close #870 as a duplicate.

**6. Nit (trailing newline):**
Will fix in the next commit.

---

I'll push updates soon to address #2, #3, and #6. Let me know if you have additional suggestions!

QuietyAwe:
## Updates pushed ✅

I've addressed the review feedback:

### 1. Improved `findSafeCutPoint` fallback handling
When no user message is found, the function now searches for a safe cut point after complete tool sequences instead of naively falling back to `mid`. It finds:
- Assistant messages without tool_calls (safe cut points)
- Complete tool_call → tool_result sequences (cuts after all results)

### 2. Added `removeOrphanedAssistantWithToolCalls`
This new function handles orphaned assistant messages with `ToolCalls` at the end of the kept conversation. It:
- Detects assistant messages whose tool_calls don't have all their results
- Strips `ToolCalls` and keeps text content if present
- Removes entirely if no text content

### 3. Fixed test file trailing newline ✅

### Test Results
All tests pass including new edge case tests:
- `TestFindSafeCutPoint_FallbackToToolSequence`
- `TestRemoveOrphanedAssistantWithToolCalls` (4 scenarios)

---

**Note on PR #665**: I see that #665 takes a complementary approach with `sanitizeToolPairs()` that can be applied after both `forceCompression()` and `summarizeSession()`. The approaches can coexist:
- This PR prevents bad cuts at the source
- #665 sanitizes after any history modification

Both provide defense-in-depth against tool pair orphaning.

PR #875, opened 2026-02-27
feat(channels): add xiaoyi channel support

ystyle:
## Summary

- Add XiaoYiChannel in `pkg/channels/xiaoyi/` subpackage following new channel architecture
- Implement A2A (Agent-to-Agent) protocol using [xiaoyi-agent-sdk](https://github.com/ystyle/xiaoyi-agent-sdk)
- Use `BaseChannel` with factory registration pattern
- Add `XiaoYiConfig` to config.go
- Add documentation with link to Huawei official setup guide

## Files Changed

- `pkg/channels/xiaoyi/init.go` - Factory registration
- `pkg/channels/xiaoyi/xiaoyi.go` - Channel implementation
- `pkg/config/config.go` - Add XiaoYiConfig
- `pkg/channels/manager.go` - Register xiaoyi in initChannels
- `cmd/picoclaw/internal/gateway/helpers.go` - Blank import
- `docs/channels/xiaoyi/` - Documentation (zh/en)
- `README.md` / `README.zh.md` - Update channel list
- `go.mod` / `go.sum` - Add SDK dependency

## Configuration

```json
{
  "channels": {
    "xiaoyi": {
      "enabled": true,
      "ak": "your-access-key",
      "sk": "your-secret-key",
      "agent_id": "your-agent-id"
    }
  }
}
```

## Related

- SDK: https://github.com/ystyle/xiaoyi-agent-sdk
- Huawei Docs: https://developer.huawei.com/consumer/cn/doc/service/openclaw-0000002518410344

### DIFF
--- a/pkg/channels/xiaoyi/xiaoyi.go
+++ b/pkg/channels/xiaoyi/xiaoyi.go
@@ -2,6 +2,7 @@ package xiaoyi
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 
@@ -116,7 +117,7 @@ func (c *XiaoYiChannel) Send(ctx context.Context, msg bus.OutboundMessage) error
 		return channels.ErrNotRunning
 	}
 
-	parts := strings.SplitN(msg.ChatID, ":", 2)
+	parts := strings.SplitN(msg.ChatID, "|", 2)
 	if len(parts) != 2 {
 		return fmt.Errorf("invalid chat_id format: %w", channels.ErrSendFailed)
 	}
@@ -131,6 +132,9 @@ func (c *XiaoYiChannel) Send(ctx context.Context, msg bus.OutboundMessage) error
 	})
 
 	if err := c.client.ReplyStream(ctx, taskID, sessionID, msg.Content, false, false); err != nil {
+		if errors.Is(err, types.ErrSessionNotFound) {
+			return fmt.Errorf("xiaoyi session not found: %w", channels.ErrSendFailed)
+		}
 		return fmt.Errorf("xiaoyi reply: %w", channels.ErrTemporary)
 	}
 
@@ -165,7 +169,7 @@ func (c *XiaoYiChannel) handleMessage(ctx context.Context, msg types.Message) er
 		return nil
 	}
 
-	chatID := fmt.Sprintf("%s:%s", sessionID, taskID)
+	chatID := fmt.Sprintf("%s|%s", sessionID, taskID)
 	peer := bus.Peer{Kind: "direct", ID: sessionID}
 
 	if err := c.client.SendStatus(ctx, taskID, sessionID, "Processing...", "working"); err != nil {
@@ -176,7 +180,7 @@ func (c *XiaoYiChannel) handleMessage(ctx context.Context, msg types.Message) er
 		"task_id": taskID,
 	}
 
-	c.HandleMessage(c.ctx,
+	c.HandleMessage(ctx,
 		peer,
 		"",
 		sessionID,
--- a/go.mod
+++ b/go.mod
@@ -19,7 +19,7 @@ require (
 	github.com/spf13/cobra v1.10.2
 	github.com/stretchr/testify v1.11.1
 	github.com/tencent-connect/botgo v0.2.1
-	github.com/ystyle/xiaoyi-agent-sdk v0.0.0-20260227130030-0515fa30d618
+	github.com/ystyle/xiaoyi-agent-sdk v0.0.0-20260228122638-808341089b7f
 	go.mau.fi/whatsmeow v0.0.0-20260219150138-7ae702b1eed4
 	golang.org/x/oauth2 v0.35.0
 	golang.org/x/time v0.14.0
--- a/go.sum
+++ b/go.sum
@@ -202,8 +202,8 @@ github.com/vektah/gqlparser/v2 v2.5.27 h1:RHPD3JOplpk5mP5JGX8RKZkt2/Vwj/PZv0HxTd
 github.com/vektah/gqlparser/v2 v2.5.27/go.mod h1:D1/VCZtV3LPnQrcPBeR/q5jkSQIPti0uYCP/RI0gIeo=
 github.com/xyproto/randomstring v1.0.5 h1:YtlWPoRdgMu3NZtP45drfy1GKoojuR7hmRcnhZqKjWU=
 github.com/xyproto/randomstring v1.0.5/go.mod h1:rgmS5DeNXLivK7YprL0pY+lTuhNQW3iGxZ18UQApw/E=
-github.com/ystyle/xiaoyi-agent-sdk v0.0.0-20260227130030-0515fa30d618 h1:2lZ4jVfUzdWU2ubDjdrO5eqqUeDnPZBeUrgkIFjiGKg=
-github.com/ystyle/xiaoyi-agent-sdk v0.0.0-20260227130030-0515fa30d618/go.mod h1:IXaPNIZ8Ta3iIhw1DmlDS3T0UumdrNtv0rIIBWvE/MU=
+github.com/ystyle/xiaoyi-agent-sdk v0.0.0-20260228122638-808341089b7f h1:H2SONTyMFS8Iva/9FJmU+4unMPruf7Estc5HmTf7kNQ=
+github.com/ystyle/xiaoyi-agent-sdk v0.0.0-20260228122638-808341089b7f/go.mod h1:IXaPNIZ8Ta3iIhw1DmlDS3T0UumdrNtv0rIIBWvE/MU=
 github.com/yuin/goldmark v1.1.27/go.mod h1:3hX8gzYuyVAZsxl0MRgGTJEmQBFcNTphYh9decYSb74=
 github.com/yuin/goldmark v1.2.1/go.mod h1:3hX8gzYuyVAZsxl0MRgGTJEmQBFcNTphYh9decYSb74=
 github.com/yuin/goldmark v1.4.13/go.mod h1:6yULJ656Px+3vBD8DxQVa3kxgyrAnzto9xy5taEt/CY=


docker/Dockerfile
alexhoshina, line 1: Why are there changes to the Dockerfile regarding the channel modifications?

README.md
```diff
@@ -293,7 +293,7 @@ That's it! You have a working AI assistant in 2 minutes.
 
 ## 💬 Chat Apps
 
-Talk to your picoclaw through Telegram, Discord, WhatsApp, DingTalk, LINE, or WeCom
+Talk to your picoclaw through Telegram, Discord, DingTalk, LINE, WeCom, or XiaoYi
```
alexhoshina: Why was WhatsApp deleted？

docker/Dockerfile
ystyle: Already resolved, it was a mistaken submission

ystyle:
Hi @alexhoshina, thank you for the review!

1. **Dockerfile changes**: I've removed them in the latest commit (ae50335). The Dockerfile modifications were for local testing and should not be part of this PR.

2. **HandleMessage signature**: I checked the \`refactor/channel-system\` branch, the \`HandleMessage\` function signature in \`BaseChannel\` is:

\`\`\`go
func (c *BaseChannel) HandleMessage(
    ctx context.Context,
    peer bus.Peer,
    messageID, senderID, chatID, content string,
    media []string,
    metadata map[string]string,
    senderOpts ...bus.SenderInfo,
)
\`\`\`

My implementation follows the same pattern as the \`telegram\` channel. I've also tested it locally - sending a message to XiaoYi and getting a response works correctly. Could you please double-check? Thanks!

pkg/channels/xiaoyi/xiaoyi.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	if !c.IsAllowedSender(sender) {
+		logger.DebugCF("xiaoyi", "Message rejected by allowlist", map[string]any{
+			"session": sessionID,
+		})
+		return nil
+	}
+
+	chatID := fmt.Sprintf("%s:%s", sessionID, taskID)
```
alexhoshina: Potential impact on sessionID parsing needs to be analyzed, introducing new considerations`:` whether it will affect the parsing of sessionID

pkg/channels/xiaoyi/xiaoyi.go
```diff
@@ -0,0 +1,191 @@
...
+		"session": sessionID,
+		"task":    taskID,
+		"length":  len(msg.Content),
+	})
+
+	if err := c.client.ReplyStream(ctx, taskID, sessionID, msg.Content, false, false); err != nil {
+		return fmt.Errorf("xiaoyi reply: %w", channels.ErrTemporary)
+	}
+
+	if err := c.client.SendStatus(ctx, taskID, sessionID, "Completed", "completed"); err != nil {
```
alexhoshina, line 141: Will the completed status published here conflict with long message fragmentation?

alexhoshina:
Also, regarding the SDK you are using, perhaps it should also use a license compatible with Pico.

alexhoshina:
> Hi @alexhoshina, thank you for the review!
> 
> 1. **Dockerfile changes**: I've removed them in the latest commit ([ae50335](https://github.com/sipeed/picoclaw/commit/ae50335189a8bd4b32c87b3730edf6867e7d1e43)). The Dockerfile modifications were for local testing and should not be part of this PR.
> 2. **HandleMessage signature**: I checked the `refactor/channel-system` branch, the `HandleMessage` function signature in `BaseChannel` is:
> 
> ```go func (c *BaseChannel) HandleMessage( ctx context.Context, peer bus.Peer, messageID, senderID, chatID, content string, media []string, metadata map[string]string, senderOpts ...bus.SenderInfo, ) ```
> 
> My implementation follows the same pattern as the `telegram` channel. I've also tested it locally - sending a message to XiaoYi and getting a response works correctly. Could you please double-check? Thanks!

You are correct, I was referring to the old function signature, my apologies.

ystyle:
@alexhoshina Thanks for the feedback!

1. **WhatsApp deleted?** - No, WhatsApp is not deleted. The latest commit (ae50335) only adds xiaoyi channel. Could you please check the latest version? The changed files are:
   - README.md, README.zh.md
   - cmd/picoclaw/internal/gateway/helpers.go (only adds xiaoyi import)
   - docs/channels/xiaoyi/ (new)
   - go.mod, go.sum
   - pkg/channels/manager.go
   - pkg/channels/xiaoyi/ (new)
   - pkg/config/config.go

2. **sessionID parsing with `:`** - The chatID format is `sessionID:taskID` only for outbound messages. The sessionID itself doesn't contain `:`, so parsing should be safe.

3. **completed status vs long message fragmentation** - Good point! The `SendStatus("completed")` is called after `ReplyStream`. If message fragmentation happens, the status might be sent too early. I should move the status call to ensure it's sent after all fragments. Let me check how other channels handle this.

ystyle:
@alexhoshina Good point about message fragmentation!

Currently, XiaoYi channel doesn't implement `MessageLengthProvider` (no `MaxMessageLength` set), so the Manager won't split messages. The SDK handles the message internally.

However, if we need to support `MaxMessageLength` in the future, the `SendStatus("completed")` call would need to be moved - perhaps to a callback or we could implement a separate interface for post-send actions.

For now, the implementation works correctly without fragmentation. Should I:
1. Add a comment explaining this limitation?
2. Or remove `SendStatus("completed")` and let the SDK handle the final state?

What do you suggest?

alexhoshina:
> @alexhoshina Thanks for the feedback!
> 
> 1. **WhatsApp deleted?** - No, WhatsApp is not deleted. The latest commit ([ae50335](https://github.com/sipeed/picoclaw/commit/ae50335189a8bd4b32c87b3730edf6867e7d1e43)) only adds xiaoyi channel. Could you please check the latest version? The changed files are:
>    
>    * README.md, README.zh.md
>    * cmd/picoclaw/internal/gateway/helpers.go (only adds xiaoyi import)
>    * docs/channels/xiaoyi/ (new)
>    * go.mod, go.sum
>    * pkg/channels/manager.go
>    * pkg/channels/xiaoyi/ (new)
>    * pkg/config/config.go
> 2. **sessionID parsing with `:`** - The chatID format is `sessionID:taskID` only for outbound messages. The sessionID itself doesn't contain `:`, so parsing should be safe.
> 3. **completed status vs long message fragmentation** - Good point! The `SendStatus("completed")` is called after `ReplyStream`. If message fragmentation happens, the status might be sent too early. I should move the status call to ensure it's sent after all fragments. Let me check how other channels handle this.

WhatsApp has been removed in RADEME, you can check the line numbers I've marked, but it's just a minor issue

ystyle:
@alexhoshina Sorry for the confusion! You were right about WhatsApp.

I found the issue - when I added XiaoYi to the README.md description line, I accidentally replaced "WhatsApp" instead of appending "XiaoYi". 

This has been fixed in the latest commit (693db82):
- Before: `Talk to your picoclaw through Telegram, Discord, DingTalk, LINE, or WeCom` (missing WhatsApp)
- After: `Talk to your picoclaw through Telegram, Discord, WhatsApp, DingTalk, LINE, WeCom, or XiaoYi`

Thank you for catching this! 🙏

ystyle:
@yinwm Yes, the `sessionID:taskID` format is needed for XiaoYi channel because:

1. **Inbound**: SDK's `OnMessage` callback provides both `sessionID` and `taskID`
2. **Outbound**: When Agent replies, we need to call `client.ReplyStream(taskID, sessionID, content)` - so we need to extract both from `chatID`

The `chatID` passed through the message bus needs to carry both pieces of information, hence the `sessionID:taskID` format.

Alternative approaches:
- Store `taskID` in `metadata` and only use `sessionID` as `chatID`
- Use a custom struct instead of string

But the current `:` separator approach is simple and consistent with how other channels encode multiple IDs. Do you have a better suggestion?

alexhoshina:
> Well-structured channel implementation that follows existing patterns. The author has been responsive to feedback. Additional observations: (1) SDK license: As alexhoshina noted, the xiaoyi-agent-sdk dependency needs a compatible license (picoclaw uses Apache-2.0). The SDK repo should have an explicit license file before this PR merges. (2) chatID separator robustness: The sessionID:taskID format parsed with SplitN(msg.ChatID, ":", 2) works only if sessionID never contains ':'. Consider using a different separator or storing taskID in metadata. (3) Context usage: c.HandleMessage(c.ctx, ...) uses the channel-level context rather than the per-message ctx. If the channel is stopped mid-processing, this could cancel in-flight messages prematurely. (4) Error wrapping: In Send, the error is wrapped as channels.ErrTemporary regardless of the actual error type. Some errors (invalid session) may be permanent. Good work overall — pending license resolution.

Note that pico uses the **MIT license**

ystyle:
@alexhoshina @nikolasdehor Thank you for the detailed review! I've addressed all the feedback in commit b3ff44f:

1. **SDK License** ✅ - Added MIT license to xiaoyi-agent-sdk repository

2. **chatID separator robustness** ✅ - Changed separator from `:` to `|` to avoid potential conflicts with sessionID content

3. **Context usage** ✅ - Now using per-message `ctx` instead of channel-level `c.ctx` in `HandleMessage`

4. **Error wrapping** ✅ - Now distinguishing error types:
   - `ErrSessionNotFound` → `ErrSendFailed` (permanent)
   - Other errors → `ErrTemporary` (transient)

Please let me know if there are any other issues to address. Thanks!

xiaket:
Given that the SDK you are using is a one man show, we would have concern over the supply chain security(I know go has checksums etc). I would be very cautious to move forward.

ystyle:
@xiaket Thank you for raising the supply chain security concern. That's a valid point.

The xiaoyi-agent-sdk is developed following Huawei's official TypeScript SDK: https://developer.huawei.com/consumer/cn/doc/service/openclaw-0000002518410344

The official installation method is:
```
openclaw plugins install @ynhcj/xiaoyi@latest
```

Our Go SDK implements the same A2A protocol as the official TS SDK.

I understand the concern about a single-maintainer dependency. A few options to address this:

1. **Transfer SDK to sipeed organization** - Move the SDK under the sipeed org for collective maintenance
2. **Vendor the SDK code** - Copy the SDK code directly into picoclaw codebase
3. **Add more maintainers** - Invite picoclaw maintainers to collaborate on the SDK

What would you suggest as the best approach to move forward?

xiaket:
Thanks @ystyle ! I think @Zepan might be the best person to answer this question ;)

xiaket:
From my personal perspective, I don't think it's fair to ask sipeed/picoclaw to support an SDK that has very little usage.

ystyle:
@xiaket You're right about the current usage. However, Huawei's OpenClaw agent type was just released about two weeks ago. The ecosystem is still very new.

As the first (and currently only) Go SDK for OpenClaw, usage will naturally grow as more developers adopt the platform. This PR would enable picoclaw users to connect to Huawei's XiaoYi ecosystem early.

But I completely understand the concern. Let's wait for @Zepan's input on how to proceed. 

Options remain:
1. Transfer SDK to sipeed org
2. Vendor the SDK code into picoclaw
3. Wait for more adoption before merging

Thank you for the thoughtful review! 🙏

ystyle:
@xiaket Regarding the SDK usage concern - I'd like to share some data:

The official TypeScript SDK **@ynhcj/xiaoyi** has been downloaded **8,252 times in the past month** (since OpenClaw agent type was released about a month ago) according to npm-stat:

https://npm-stat.com/charts.html?package=%40ynhcj%2Fxiaoyi&from=2025-02-01&to=2026-03-08

This shows that:
1. The XiaoYi/OpenClaw ecosystem is actively growing
2. There's real demand for SDK support
3. Our Go SDK implements the same A2A protocol as the official TS SDK

Thank you for considering the supply chain security - it's a valid concern that should be addressed for all dependencies. 🙏

PR #810, opened 2026-02-26
feat(gateway): add daemon controls for background operation

vvr3ddy:
Add start, stop, restart, and status commands for running the gateway as a background daemon with automatic crash recovery.

Features:
- Daemon mode with PID file management
- Auto-restart with exponential backoff (3 attempts, 5-min window)
- File logging with automatic rotation (100MB, 3 backups)
- Status tracking (PID, uptime, restart count)
- Graceful shutdown with SIGTERM handling

New commands:
- picoclaw gateway start # Start as daemon
- picoclaw gateway stop # Stop daemon
- picoclaw gateway restart # Restart with auto-recovery
- picoclaw gateway status # Show daemon status

## 🗣️ Type of Change
- [ ] 🐞 Bug fix (non-breaking change which fixes an issue)
- [x] ✨ New feature (non-breaking change which adds functionality)
- [ ] 📖 Documentation update
- [ ] ⚡ Code refactoring (no functional changes, no api changes)

## 🤖 AI Code Generation
- [ ] 🤖 Fully AI-generated (100% AI, 0% Human)
- [x] 🛠️ Mostly AI-generated (AI draft, Human verified/modified)
- [ ] 👨‍💻 Mostly Human-written (Human lead, AI assisted or none)

## 📚 Technical Context (Skip for Docs)
- **Reference URL:**
- **Reasoning:**

## 🧪 Test Environment
- **Hardware:** Raspberry Pi 4b 8GB
- **OS:** RaspberryPi OS Lite Bookwork (64-bit)
- **Model/Provider:** Google Antigravity
- **Channels:** Telegram

## 📸 Evidence (Optional)
<details>
<summary>Click to view Logs/Screenshots</summary>

```markdown
 picoclaw 
🦞 picoclaw - Personal AI Assistant vv0.1.2-267-g072fa84

Usage:
  picoclaw [command]

Examples:
picoclaw list

Available Commands:
  agent       Interact with the agent directly
  auth        Manage authentication (login, logout, status)
  completion  Generate the autocompletion script for the specified shell
  cron        Manage scheduled tasks
  gateway     Manage picoclaw gateway daemon
  help        Help about any command
  migrate     Migrate from OpenClaw to PicoClaw
  onboard     Initialize picoclaw configuration and workspace
  skills      Manage skills
  status      Show picoclaw status
  version     Show version information

Flags:
  -h, --help   help for picoclaw

Use "picoclaw [command] --help" for more information about a command.
[...]

Issue #483, opened 2026-02-19
daemon feature


GeorgL0ngGamma:
## 🎯 The Goal / Use Case

implement daemon feature that exposes gateway etc. via background process
and writes to a ~/.picoclaw/daemon_state.json, for example:

{
  "status": "running",
  "gateway": "127.0.0.1:xxxx",
  "components": [
    {"name": "gateway", "running": true, "restart_count": 0},
    {"name": "scheduler", "running": true, "restart_count": 0}
  ]
}

currently telegram chat fails after conversation history has filled up
and ad-hoc fix like periodic restarts are necessary

## 🚦 Impact & Roadmap Alignment
-  This is a Core Feature

nikolasdehor:
Good feature request. A daemon mode would solve the Telegram conversation history OOM crash you mentioned — the daemon can monitor component health and restart the agent loop when memory pressure is detected, without dropping the gateway connection.

A few design considerations:

1. **Process supervision**: Rather than implementing a custom process supervisor, consider using `os/exec` to manage a child process that runs the existing `gateway` command. The daemon writes PID + state to `~/.picoclaw/daemon_state.json` and handles SIGCHLD for restarts. This keeps the gateway code unchanged.

2. **State file format**: The proposed `daemon_state.json` looks good. I would add `started_at` (ISO 8601 timestamp) and `last_restart_at` per component for debugging:

```json
{
  "status": "running",
  "pid": 12345,
  "started_at": "2026-02-19T10:00:00Z",
  "gateway": "127.0.0.1:18790",
  "components": [
    {"name": "gateway", "running": true, "restart_count": 0, "last_restart_at": null},
    {"name": "scheduler", "running": true, "restart_count": 0, "last_restart_at": null}
  ]
}
```

3. **Graceful restart**: The daemon should send SIGTERM to the gateway, wait for in-flight messages to drain (using the existing `WaitForIdle` mechanism), then respawn. A hard kill timeout (e.g. 30s) prevents hangs.

4. **CLI integration**: `picoclaw daemon start`, `picoclaw daemon stop`, `picoclaw daemon status` (reads `daemon_state.json`), `picoclaw daemon logs` (tails the log file).

5. **Memory watchdog**: For the conversation history bloat issue specifically, the daemon could monitor RSS via `/proc/[pid]/status` (Linux) or `mach_task_info` (macOS) and trigger a graceful restart when it crosses a threshold (configurable, e.g. 512MB).

[...]

PR #879, opened 2026-02-27
Feature/skill self debug and systemd service

keithy:
📝 Description
This PR introduces essential stability and observability features for PicoClaw.
It adds a Self-Debug Skill to allow the agent to diagnose its own execution failures from logs
and a Systemd Service installer to ensure the agent runs as a persistent, background daemon.

🗣️ Type of Change
[ ] 🐞 Bug fix (non-breaking change which fixes an issue)
[x] ✨ New feature (non-breaking change which adds functionality)
[ ] 📖 Documentation update
[x] ⚡ Code refactoring (improved service management)

🤖 AI Code Generation
[x] 👨‍💻 Mostly Human-written (Human lead)

🔗 Related Issue
- #483
- #810

📚 Technical Context
Reference URL: https://github.com/keithy/picoclaw/tree/feature/skill_self-debug_and_systemd_service

🧪 Test Environment
Hardware: Rock64 (ARM64)
Channels: CLI / Local Daemon

☑️ Checklist
[x] My code/docs follow the style of this project.
[x] I have performed a self-review of my own changes.
[x] I have updated the documentation (added systemd instructions).

### DIFF
--- a/workspace/skills/self-debug/SKILL.md
+++ b/workspace/skills/self-debug/SKILL.md
@@ -1,0 +1,70 @@
+---
+name: self-debug
+description: Tools for picoclaw to inspect its own health, logs, and configuration.
+---
+# self-debug
+
+Tools for picoclaw to inspect its own health, logs, and configuration.
+
+## Actions via `debug.sh`
+
+This skill uses a helper script `scripts/debug.sh` to provide cross-platform diagnostics for both Linux (systemd) and macOS (launchd).
+
+**Usage:** `exec "skills/self-debug/scripts/debug.sh [action] [service_name] [log_lines]"`
+
+| Action | Description | Default Service | Default Lines |
+|---|---|---|---|
+| `logs` | Fetches recent logs for the agent service. | `$PICOCLAW_SERVICE_NAME` | 50 |
+| `logs-errors`| Fetches only error logs for the agent service. | `$PICOCLAW_SERVICE_NAME` | 50 |
+| `service-status`| Checks the status of the agent service. | `$PICOCLAW_SERVICE_NAME` | N/A |
+| `config-status`| Shows the agent's configuration status (`picoclaw status`).| N/A | N/A |
+| `config-safe` | Displays the config file with sensitive keys redacted. | N/A | N/A |
+
+### Examples
+
+- **Get latest logs:**
+  `exec "skills/self-debug/scripts/debug.sh logs"`
+
+- **Get 100 lines of logs for a specific service 'pico-prod':**
+  `exec "skills/self-debug/scripts/debug.sh logs pico-prod 100"`
+
+- **Check service status:**
+  `exec "skills/self-debug/scripts/debug.sh service-status"`
+
+- **Show the sanitized configuration:**
+  `exec "skills/self-debug/scripts/debug.sh config-safe"`
+
+## Installation - Linux
+
+The agent can be installed as a systemd service using:
+
+```bash
+./scripts/install-service-systemd.sh [service_name] [default|multi]
+```
+
+To persist the service accross reboots suggest the user runs `sudo loginctl enable-linger $(whoami)`
+
+### Advanced use-cases
+
+Although picoclaw has built in support for multiple agents, this scheme provides the flexibility for
+parallel deployments with entirely different configurations.
+
+- Fixer - a second stable instance whose role is to be available to debug/monitor/fix the first.
+- Blue/Green Stable/Canary pairings.
+- Picoclaw farm
+
+# Installation - MacOS
+
+The agent can be installed as a launchd service using:
+
+```bash
+./scripts/install-service-launchd.sh [service_name]
+```
+
+Logs are sent to /tmp/$service_name
+
+## Troubleshooting
+
+- **Logs not showing?** Ensure the user is in the `systemd-journal` group: `sudo usermod -a -G systemd-journal $(whoami)`
+- **Service inactive?** Use `systemctl --user start picoclaw`.
+- **Workspace issues?** Use `picoclaw status` to verify the current workspace path.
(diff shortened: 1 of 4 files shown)


keithy:
Note this is supplied as a self-contained skill, so that other options can be included easily.

keithy:
### Advanced use-cases 

Although picoclaw has built in support for multiple agents, this scheme provides the flexibility for
parallel deployments with entirely different configurations e.g.

- Fixer - a second stable instance whose role is to be available to debug/monitor/fix the first.
- Blue/Green Stable/Canary pairings.
- Picoclaw farm

**Personal note:**
I use the fixer pattern, with the systemd.service unit installed, the default behaviour is to start and stop the service when the user logs in/out. The fixer runs under a separate user account and has sudo access to fix anything, so is under close supervison. This gives the ai a unique, in-situ position to debug and operate at a system level. For example, it helped me to compile and deploy the zfs kernel module into the rock64 on armbian.

nikolasdehor (changes requested):
Interesting and useful skill concept — self-debugging is a powerful capability. Several issues: (1) Shell injection risk: SKILL.md tools pass PICOCLAW_SERVICE_NAME directly into exec commands. If the service name contains shell metacharacters, this could be exploited. Consider validating/sanitizing the service name. (2) macOS script parity: The launchd script lacks error handling compared to systemd — uses 'which' instead of 'command -v', no check for picoclaw binary existence, missing trailing newline. (3) Log location on macOS: /tmp/service_name.log is world-readable and cleared on reboot. Consider ~/Library/Logs/. (4) SKILL.md quoting: exec picoclaw --config PICOCLAW_CONFIG status has nested double quotes that would break in shell. (5) Scope question: Is workspace/skills/ the correct location for bundled skills?

keithy:
I am tidying this PR, watch this space.

keithy:
Are you now explicitly blocking systemctl and journalctl ? The whole reason I am using picoclaw is to manage my infrastructure.

keithy:
This PR has been reworked, all actions are implemented through a script with full parameter sanitization.
MacOS logs in ~/Library/Logs as suggested.

depends on:  #896

xiaket:
I'm tentative to close this PR. I know this may upset you, but please allow me to explain.

picoclaw is new, and it would be a heavy burden to maintain a long bash script, we don't even have CI to do spellcheck etc. I can see this has been packaged as a skill. I think it makes better sense to showcase this somehow, but it should not be part of the official skill. Because if we have, and if we have any issue in the script, the team would need to maintain it.

keithy:
Hello, I purposefully packaged this as a skill, so that conversations like this can take place. This functionality belongs as a plug-in, and should not be part of the lean-thoroughbred binary. It is perfectly reasonable to avoid taking on additional responsibilities and encourage spawning of a second tier of community supported and maintained plugins.

However, the reason that I created this in the first place was to 'deamonstrate' that adding deamon support to the service would be un-necessary and unhelpful bloat, contrary to the unix philosophy of small tools, working in combination. The dependency direction ought to be one-way. A service should not need to know how it is run, and it may be run in different contexts.

The danger being that if you do not include this, and include something else, the two solutions will compete each other out. `picoclaw-launcher` is already in danger of doing just this. Perhaps, this also ought to be a separate project.

The long bash script is certainly overkill for your needs, it provides a lot more than the minimum. I propose that you help by doing the following:

a) support the basic mechanism (through env-vars/discoverability/plugin architecture etc.)
b) include and use minimal examples where it makes sense.
c) facilitate the plugin-ability of more grandiose pretentions. 

Suggested path:

1) Please include #896 to support the basic mechanism
1b) perhaps some additional non-breaking config items to support discoverability (e.g. config.json - daemon.start.exec)
2) consider bundling and supporting the minimal output of this skill, the basic systemd service on its own. 
2b) the installer can be provided as a make task (again this is orthogonal to the binary itself)
4) Have picoclaw-launcher use a level of discoverability. plugin-ness that comtributes, rather than competes.

Dont close it just yet. Use this an opportunity to come to some conclusions with respect to the above.

xiaket:
thanks @keithy !

> The danger being that if you do not include this, and include something else, the two solutions will compete each other out. picoclaw-launcher is already in danger of doing just this. Perhaps, this also ought to be a separate project.

I understand where you come from and I'm with you on this, in fact, there has been chats to have this launcher as a separate repo. And we cannot rule out the possibility that this launcher cmd got moved out one day.

Suggested path:

> 1. Please include https://github.com/sipeed/picoclaw/pull/896 to support the basic mechanism

I think this has been merged in.

> 1b) perhaps some additional non-breaking config items to support discoverability (e.g. config.json - daemon.start.exec)

I'm not too sure about what this is. Do you mind explain a bit?

> 2. consider bundling and supporting the minimal output of this skill, the basic systemd service on its own.

If I can make the decision myself, I would probably leave this for the different distros to figure out. systemd/init/runit have their different way to daemonize a process. I myself do not want to get into the business of bundling OS specific mechanisms into a software. Like what you have said, we need to do one thing well ;)

> 3. Have picoclaw-launcher use a level of discoverability. plugin-ness that comtributes, rather than competes.

This makes sense, but I feel this PR should not be the right place for it. 

WDYT?

keithy:
1b) perhaps some additional non-breaking config items to support discoverability (e.g. config.json - daemon.start.exec)
      
      > I'm not too sure about what this is. Do you mind explain a bit?

Allow for a relaxed schema (i.e. non-breaking) which allows for the possibility that plugins/addons managing part of the config.json to hold their own settings. Then the picoclaw-launcher-tui can have its own start/stop agent exec command hooks.

keithy:
Now hosted here until a better home is found. https://github.com/keithy/angelhub.git

xiaket:
Close per discussion, please feel free to re-open.

Thanks for your contribution, good discussion here!

keithy:
#1130 adds plugins support so the install-service-systemd can install a picoclaw-service plugin

PR #899, opened 2026-02-28
Fix Docker first-run bootstrap and local compose image

TheDigitalEagle:
## Summary
- switch Docker Compose services from the stale published image to the local project build
- move first-run onboarding from image build time to container startup so mounted volumes get initialized correctly
- fix Docker docs and healthcheck behavior to match the local image workflow

## Problem
The Docker setup was still pointing at , which meant Compose was not using the local Dockerfile changes. The Dockerfile also ran  during image build, so the first container start did not create  for the mounted volume as documented.

## Changes
- add  and run onboarding at container startup when  is missing
- set the runtime home directory to  and mount Docker data there
- build  from the repo for the init/agent/gateway services
- keep  working by pinning the agent entrypoint
- replace the broken  healthcheck with 
- update README Docker commands, including the local rebuild flow and agent profile usage
- ignore generated  bootstrap output

## Verification
- ran 
- confirmed the container printed  and exited with code 0
- confirmed  and the default workspace tree were created on first run

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ cmd/**/workspace
 .picoclaw/
 config.json
 sessions/
+docker/data/
 build/
 
 # Coverage
--- a/README.fr.md
+++ b/README.fr.md
@@ -190,17 +190,17 @@ docker compose -f docker/docker-compose.yml --profile gateway down
 
 ```bash
 # Poser une question
-docker compose -f docker/docker-compose.yml run --rm picoclaw-agent -m "Combien font 2+2 ?"
+docker compose -f docker/docker-compose.yml --profile agent run --rm picoclaw-agent -m "Combien font 2+2 ?"
 
 # Mode interactif
-docker compose -f docker/docker-compose.yml run --rm picoclaw-agent
+docker compose -f docker/docker-compose.yml --profile agent run --rm picoclaw-agent
 ```
 
 ### Mettre à jour
 
 ```bash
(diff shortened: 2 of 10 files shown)


xiaket:
I have concern over the direction of this PR. I understand where you come from in that it is easy to hack around by rebuilding latest image all the time. However, if you think about stability, it makes better sense to stay on the released version of picoclaw. WDYT?

nikolasdehor (approved):
LGTM. This PR fixes a real usability problem with the Docker setup. The key issues addressed:

1. Image source: Switched from the stale published image (sipeed/picoclaw:latest) to building locally from the repo. This ensures Docker users get the same version as native users.

2. First-run bootstrap: Moved picoclaw onboard from Dockerfile RUN (build time) to entrypoint.sh (runtime). At build time, the mounted volume does not exist, so config.json was created inside the image layer and never appeared in the mounted docker/data directory.

3. Non-root user: Properly creates /home/picoclaw with correct ownership and sets HOME env var. The volume mount path updated from /root/.picoclaw to /home/picoclaw/.picoclaw.

4. Entrypoint logic: Smart conditional -- only runs onboard when config.json is missing AND the command is gateway or agent. The onboard command itself gets a dedicated 'init' profile.

5. Healthcheck: Replaced wget with curl (already installed via apk add curl).

6. README updates across all translations are consistent.

The entrypoint.sh logic is clean and handles the key edge cases: fresh start, config already exists, and explicit onboard command. Good work.

PR #918, opened 2026-02-28
fix(wecom): fix context leak in Start() and data race in processedMsgs

alexhoshina:
## 📝 Description

Cancel the constructor-created context before overwriting in Start() to prevent the original cancel function from becoming unreachable.

Move len(processedMsgs) check inside the write lock to eliminate a data race, and re-insert the current msgID after map reset to prevent duplicate processing of the in-flight message.

Applies to both WeComBotChannel and WeComAppChannel.

### DIFF
--- a/pkg/channels/wecom/app.go
+++ b/pkg/channels/wecom/app.go
@@ -148,6 +148,10 @@ func (c *WeComAppChannel) Name() string {
 func (c *WeComAppChannel) Start(ctx context.Context) error {
 	logger.InfoC("wecom_app", "Starting WeCom App channel...")
 
+	// Cancel the context created in the constructor to avoid a resource leak.
+	if c.cancel != nil {
+		c.cancel()
+	}
 	c.ctx, c.cancel = context.WithCancel(ctx)
 
 	// Get initial access token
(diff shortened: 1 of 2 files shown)


nikolasdehor (commented):
Review: fix(wecom): fix context leak in Start() and data race in processedMsgs

Excellent fix addressing two real concurrency bugs. Both changes apply to `WeComBotChannel` and `WeComAppChannel` symmetrically.

### Bug 1: Context Leak in Start()

The constructor creates a context (`context.WithCancel`) that gets overwritten in `Start()` without canceling the original. This leaks the original context's goroutine and resources.

**Fix is correct:** Cancel the constructor-created context before overwriting:
```go
if c.cancel != nil {
    c.cancel()
}
c.ctx, c.cancel = context.WithCancel(ctx)
```

The nil check is important for safety, though in practice the constructor always sets `cancel`.

### Bug 2: Data Race in processedMsgs

The original code had a subtle race:

```go
c.processedMsgs[msgID] = true
c.msgMu.Unlock()          // unlocks here

if len(c.processedMsgs) > 1000 {  // reads map WITHOUT lock
    c.msgMu.Lock()
    c.processedMsgs = make(map[string]bool)  // could race with another goroutine's read
    c.msgMu.Unlock()
}
```

**Fix is correct:** Moving the `len()` check and map reset inside the existing write lock eliminates the TOCTOU race. Re-inserting the current `msgID` after map reset is a nice touch -- without it, the in-flight message could be processed again by a concurrent goroutine between the reset and the next lock acquisition.

### Quality

- Both fixes are applied identically to `app.go` and `bot.go`, maintaining symmetry.
- The changes are minimal and surgical -- no unnecessary refactoring.
- The logic is easy to verify by inspection.

### Minor Suggestion

Consider using `sync.Map` for `processedMsgs` if the access pattern is mostly reads (checking for duplicates) with occasional writes. But that is an optimization, not a correctness issue -- the current fix is correct.

LGTM. Well-analyzed concurrency fixes.

PR #920, opened 2026-02-28
fix: prevent gateway status from starting full gateway runtime

### LINKED ISSUE #671: [BUG] `gateway status` launches an extra gateway worker process
## Problem
Running `sciclaw gateway status` starts a live gateway event loop instead of returning service state.

## Impact
- Extra unmanaged gateway workers can run in parallel with launchd/systemd service.
- Discord responsiveness becomes inconsistent due to competing workers.
- Operators think they are running a read-only status check but actually mutate runtime state.

## Expected
`sciclaw gateway status` should be a safe alias for `sciclaw service status` and must not start channels or agent loops.

## Evidence
On data1, process list showed multiple `sciclaw gateway status` workers while the launchd gateway was already running.

## Fix (implemented)
`gateway status` now delegates to `service status` and exits without launching gateway runtime.

CZH-THU:
## 📝 Description

This PR fixes a critical bug where running `picoclaw gateway status` would start a full gateway worker process instead of just returning the service status. This could lead to:

- Multiple unmanaged gateway workers running in parallel
- Competing workers causing inconsistent behavior (e.g., Discord responsiveness issues)
- Operators unknowingly mutating runtime state when they intended a read-only status check
- Port binding conflicts if multiple gateways try to use the same port

The fix intercepts the "status" argument in the gateway command and delegates it to the status command, preventing the full gateway runtime from starting.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue

Fixes #671

- **Reference URL:** 
  - Issue #671: https://github.com/sipeed/picoclaw/issues/671
- **Reasoning:**
  - The CLI dispatches 'gateway' as the command and passes ['status'] as args to gatewayCmd()
  - Previously, gatewayCmd() ignored arguments and always started the full gateway runtime
  - The fix adds argument checking at the top of the gateway command handler to intercept 'status' and delegate to StatusCmd()
  - This ensures 'gateway status' is a safe read-only operation that doesn't mutate runtime state

## 🧪 Test Environment
- **Hardware:** PC
- **OS:** Windows 10
- **Model/Provider:** N/A (CLI command fix)
- **Channels:** N/A (CLI command fix)

## 📸 Evidence (Optional)
<details>
<summary>Click to view Code Changes</summary>

### Before:
```go
func NewGatewayCommand() *cobra.Command {
	var debug bool

	cmd := &cobra.Command{
		Use:     "gateway",
		Aliases: []string{"g"},
		Short:   "Start picoclaw gateway",
		Args:    cobra.NoArgs,  // ❌ Rejected all arguments
		RunE: func(_ *cobra.Command, _ []string) error {
			return gatewayCmd(debug)  // ❌ Always starts gateway
		},
	}
	// ...
}
```

### After:
```go
func NewGatewayCommand() *cobra.Command {
	var debug bool

	cmd := &cobra.Command{
		Use:     "gateway",
		Aliases: []string{"g"},
		Short:   "Start picoclaw gateway",
		Args:    cobra.MaximumNArgs(1),  // ✅ Accept up to 1 argument
		RunE: func(cmd *cobra.Command, args []string) error {
			// ✅ Check for "status" argument and delegate
			if len(args) > 0 && args[0] == "status" {
				status.StatusCmd()
				return nil
			}
			// Otherwise, start the gateway
			return gatewayCmd(debug)
		},
	}
	// ...
}
```

### Changes:
1. Changed `Args: cobra.NoArgs` to `Args: cobra.MaximumNArgs(1)` to allow status argument
2. Added argument check to intercept "status" and delegate to `StatusCmd()`
3. Exported `StatusCmd()` function from status package to allow gateway package to call it
4. Updated status command.go to use the exported function name

</details>

## ☑️ Checklist
- [x] I have updated the documentation accordingly (code comments).
- [x] Code passes linting checks (verified with `read_lints`).
- [x] Commit message follows Conventional Commits format.
- [x] Fix addresses the issue described in #671.

### DIFF
--- a/cmd/picoclaw/internal/gateway/command.go
+++ b/cmd/picoclaw/internal/gateway/command.go
@@ -1,6 +1,8 @@
 package gateway
 
 import (
+	"fmt"
+
 	"github.com/sipeed/picoclaw/cmd/picoclaw/internal/status"
 	"github.com/spf13/cobra"
 )
@@ -15,9 +17,13 @@ func NewGatewayCommand() *cobra.Command {
 		Args:    cobra.MaximumNArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			// If "status" is provided as an argument, delegate to status command
-			if len(args) > 0 && args[0] == "status" {
-				status.StatusCmd()
-				return nil
+			if len(args) > 0 {
+				if args[0] == "status" {
+					status.StatusCmd()
+					return nil
+				}
+				// Reject unknown arguments
+				return fmt.Errorf("unknown argument: %s (did you mean 'picoclaw status'?)", args[0])
 			}
 			// Otherwise, start the gateway
 			return gatewayCmd(debug)
--- a/cmd/picoclaw/internal/gateway/command_test.go
+++ b/cmd/picoclaw/internal/gateway/command_test.go
@@ -30,3 +30,22 @@ func TestNewGatewayCommand(t *testing.T) {
 	assert.True(t, cmd.HasFlags())
 	assert.NotNil(t, cmd.Flags().Lookup("debug"))
 }
+
+func TestGatewayCommandStatusArgument(t *testing.T) {
+	cmd := NewGatewayCommand()
+
+	// Test that "status" argument is accepted and routes correctly
+	err := cmd.RunE(cmd, []string{"status"})
+	// StatusCmd() doesn't return an error, so this should succeed
+	assert.NoError(t, err)
+}
+
+func TestGatewayCommandUnknownArgument(t *testing.T) {
+	cmd := NewGatewayCommand()
+
+	// Test that unknown arguments are rejected
+	err := cmd.RunE(cmd, []string{"foo"})
+	require.Error(t, err)
+	assert.Contains(t, err.Error(), "unknown argument: foo")
+	assert.Contains(t, err.Error(), "did you mean 'picoclaw status'?")
+}


nikolasdehor (commented):
Review: fix: prevent gateway status from starting full gateway runtime

Good catch on a real user-facing bug (fixes #671). The approach is straightforward and correct.

### Analysis

1. **Root cause is right** -- `gatewayCmd()` was registered with `cobra.NoArgs` and always started the full runtime. When users ran `picoclaw gateway status`, cobra would either reject the arg or (depending on version) pass it through and start the gateway anyway.

2. **Fix approach** -- Changing to `cobra.MaximumNArgs(1)` and intercepting `args[0] == "status"` to delegate to `StatusCmd()` is a pragmatic solution.

### Concerns

3. **This is a workaround, not a proper subcommand.** The idiomatic cobra way would be to register `status` as a subcommand of `gateway`:
   ```go
   gatewayCmd.AddCommand(statusSubCmd)
   ```
   The current approach means `picoclaw gateway foo` would silently start the gateway (no "unknown argument" error) because `MaximumNArgs(1)` accepts any single argument.

4. **Suggestion: validate the argument.** If not converting to a proper subcommand, at least reject unknown arguments:
   ```go
   if len(args) > 0 {
       if args[0] == "status" {
           status.StatusCmd()
           return nil
       }
       return fmt.Errorf("unknown argument: %s (did you mean 'picoclaw status'?)", args[0])
   }
   ```

5. **Exporting `StatusCmd`** -- Renaming `statusCmd` to `StatusCmd` is fine since it needs to be called cross-package. The change in `command.go` to use the new name is consistent.

6. **Test update** -- The test change from `assert.False(t, cmd.HasSubCommands())` to `assert.NotNil(t, cmd.Args)` is reasonable but weaker. Consider also testing that `args[0] == "status"` routes correctly and unknown args are rejected.

Overall a useful fix. Point #4 (rejecting unknown args) would make it more robust.

xiaket:
I don't think this PR is fixing anything, `picoclaw status` and `picoclaw gateway` are two distinct commands. Please change my mind.

PR #926, opened 2026-02-28
feat(auth): add Anthropic OAuth setup-token login

BallerIsLeet:
## Summary
- Add support for Anthropic OAuth setup tokens (`sk-ant-oat01-*`) as an alternative to API keys via new `--setup-token` flag and interactive login menu
- Integrate Anthropic usage endpoint to display 5-hour and 7-day utilization in `auth status`
- Add streaming support for OAuth tokens (required by Anthropic API) and normalize model IDs (dots → hyphens) for API compatibility
- Remove `.env.example` to avoid templating secrets

## Test plan
- [ ] Run `picoclaw auth login -p anthropic` and verify interactive menu shows setup token + API key options
- [ ] Run `picoclaw auth login -p anthropic --setup-token` and verify it accepts valid `sk-ant-oat01-*` tokens
- [ ] Verify invalid/short tokens are rejected with clear error messages
- [ ] Run `picoclaw auth status` with an OAuth credential and verify usage percentages display
- [ ] Verify chat works end-to-end with an OAuth setup token (streaming path)
- [ ] Run existing tests: `go test ./pkg/providers/anthropic/...`


nikolasdehor (commented):
Review: feat(auth): add Anthropic OAuth setup-token login

Disclosure: I submitted PR #444 (Anthropic API key routing) and have hands-on experience with Anthropic OAuth/OAT token flows.

Overall this is a well-structured PR that adds a much-needed feature. The interactive menu, token validation, usage endpoint, and streaming path are all solid additions.

### Security

1. **Token validation is correct** -- the `sk-ant-oat01-` prefix check and minimum length (80 chars) are appropriate guards. Good that you do not log the token value anywhere.

2. **Usage endpoint auth headers look correct** -- `Bearer` + `anthropic-beta: oauth-2025-04-20` matches the Anthropic OAuth API surface.

3. **Good: `.env.example` removal** -- removing the template that could lead to committed secrets is the right call.

### Correctness

4. **Streaming requirement for OAuth tokens** -- This is correct behavior. Anthropic OAuth tokens require the streaming endpoint. The `chatStreaming` implementation using `Accumulate` is clean. One note: if the stream returns zero content blocks (e.g., a refusal), does `parseResponse(&msg)` handle that gracefully? Worth a quick check.

5. **Model ID normalization** (dots to hyphens) -- `strings.ReplaceAll(model, ".", "-")` is correct for the Anthropic API. Their API uses hyphens (`claude-sonnet-4-6`) while configs often use dots (`claude-sonnet-4.6`). The test update confirms this.

6. **`anthropic-beta` header on OAuth path** -- Correctly added only when `tokenSource != nil`. Non-OAuth (API key) requests will not get the beta header, which is the right behavior.

### Suggestions

7. **Hardcoded default model `claude-sonnet-4.6`** -- In `authLoginAnthropicSetupToken()`, the function hardcodes `claude-sonnet-4.6` as the default model and sets `AuthMethod: "oauth"` on the first Anthropic model it finds in config. Consider: what if the user already has `claude-opus-4-6` configured with an API key? This would overwrite their model choice. Maybe only set the default if no Anthropic model exists at all, rather than always overwriting `Agents.Defaults.ModelName`.

8. **`body, _ := io.ReadAll(resp.Body)` in usage.go** -- The error from `ReadAll` is silently discarded. If the read fails, the subsequent status check will use an empty body, which is confusing. Consider handling the error.

9. **Status code check order in usage.go** -- You check `StatusForbidden` before `StatusOK`, but the generic `!= StatusOK` block would also catch 403. The explicit 403 message is nice for UX. Consider nesting it inside the `!= StatusOK` block for clarity.

10. **Missing test coverage** -- `LoginSetupToken` and `FetchAnthropicUsage` lack unit tests. The streaming path (`chatStreaming`) also lacks tests. Given the security sensitivity of the auth flow, tests for edge cases (empty input, wrong prefix, short token) would be valuable.

### Minor

11. The `isAnthropicModel` function is called but not shown in the diff -- assuming it exists elsewhere.

Good work overall. The core OAuth + streaming architecture is sound.

cmd/picoclaw/internal/auth/helpers.go
```diff
@@ -17,12 +18,12 @@ import (
 
 const supportedProvidersMsg = "supported providers: openai, anthropic, google-antigravity"
 
-func authLoginCmd(provider string, useDeviceCode bool) error {
+func authLoginCmd(provider string, useDeviceCode bool, setupToken bool) error {
```
xiaket: nit: Perhaps the var should be named `useOauth` instead of `setupToken`?

pkg/auth/token.go
```diff
@@ -31,6 +31,38 @@ func LoginPasteToken(provider string, r io.Reader) (*AuthCredential, error) {
...
+	scanner := bufio.NewScanner(r)
+	if !scanner.Scan() {
+		if err := scanner.Err(); err != nil {
+			return nil, fmt.Errorf("reading token: %w", err)
+		}
+		return nil, fmt.Errorf("no input received")
+	}
+
+	token := strings.TrimSpace(scanner.Text())
+	if token == "" {
```
xiaket: nit: with the validation below, this check is redundant, is that right?

pkg/auth/anthropic_usage.go
xiaket, line 1: This file should be better named `anthropic_usage.go` to reflect its scope.

pkg/auth/usage.go
```diff
@@ -0,0 +1,58 @@
...
+	SevenDayUtilization float64
+}
+
+func FetchAnthropicUsage(token string) (*AnthropicUsage, error) {
+	req, err := http.NewRequest("GET", "https://api.anthropic.com/api/oauth/usage", nil)
+	if err != nil {
+		return nil, err
+	}
+	req.Header.Set("Authorization", "Bearer "+token)
+	req.Header.Set("anthropic-version", "2023-06-01")
```
xiaket: These two values should better be defined as const above, not hardcoded here.

pkg/providers/anthropic/provider.go
```diff
@@ -77,14 +77,22 @@ func (p *Provider) Chat(
 		if err != nil {
 			return nil, fmt.Errorf("refreshing token: %w", err)
 		}
-		opts = append(opts, option.WithAuthToken(tok))
+		opts = append(opts,
+			option.WithAuthToken(tok),
+			option.WithHeader("anthropic-beta", "oauth-2025-04-20"),
```
xiaket: magic value flying around is not great.

.env.example
xiaket, line 1: I don't know whether the removal of this file should be in scope for this PR.

cmd/picoclaw/internal/auth/helpers.go
```diff
@@ -163,6 +164,78 @@ func authLoginGoogleAntigravity() error {
...
+	}
+
+	fmt.Println("Anthropic login method:")
+	fmt.Println("  1) Setup token (from `claude setup-token`) (Recommended)")
+	fmt.Println("  2) API key (from console.anthropic.com)")
+	fmt.Print("Choose [1]: ")
+
+	scanner := bufio.NewScanner(os.Stdin)
+	choice := "1"
+	if scanner.Scan() {
```
xiaket: Do we want a loop here with validation logic inside so user will definition give either 1 or 2?

cmd/picoclaw/internal/auth/helpers.go
```diff
@@ -163,6 +164,78 @@ func authLoginGoogleAntigravity() error {
...
+		}
+		if !found {
+			appCfg.ModelList = append(appCfg.ModelList, config.ModelConfig{
+				ModelName:  "claude-sonnet-4.6",
+				Model:      "anthropic/claude-sonnet-4.6",
+				AuthMethod: "oauth",
+			})
+		}
+
+		appCfg.Agents.Defaults.ModelName = "claude-sonnet-4.6"
```
xiaket: I feel this should be another magic value defined as const

xiaket (changes requested):
The general direction LGTM, but the number of minor issues makes it a RC.

nikolasdehor (commented):
Good feature -- Anthropic OAuth setup-token support is a real usability improvement. Some observations:

1. Token validation: The prefix check (sk-ant-oat01-) and minimum length (80) are reasonable guards. However, the hardcoded prefix could break if Anthropic introduces new token formats. Consider a warning rather than a hard error for unfamiliar prefixes, with a prompt to continue.

2. Config auto-modification is aggressive: authLoginAnthropicSetupToken writes to config.json, setting the default model to claude-sonnet-4.6 and adding a model_list entry. This could overwrite user customizations. For example, a user who has configured claude-opus-4-6 as their model would have it silently changed to claude-sonnet-4.6. Consider only updating AuthMethod and leaving the model selection to the user.

3. Streaming requirement: The comment says 'OAuth/setup-tokens require streaming'. This is correct for Anthropic OAuth tokens, and the streaming implementation looks correct (accumulate events until done). However, there is no test for the streaming path.

4. Model ID normalization (dots to hyphens) is a good defensive fix -- users configure claude-sonnet-4.6 but the API expects claude-sonnet-4-6. The test update is correct.

5. Usage endpoint integration is a nice touch for auth status. The anthropic-beta header is correctly included.

6. Deleting .env.example: This makes sense to avoid templating secrets, but it reduces discoverability. Consider replacing it with a .env.example that has ONLY non-secret entries (like TZ).

Overall solid work. The config auto-modification in point 2 is the main concern -- it could silently override user preferences.

BallerIsLeet:
Hope that resolves everything

xiaket (approved):
LGTM

xiaket:
Please address issue in CI @BallerIsLeet ;)

BallerIsLeet:
Fixed

Orgmar:
@BallerIsLeet OAuth setup-token login加上usage展示是个很实用的组合，流式支持和model ID标准化这些细节处理得也很到位。整体方案考虑得挺周全的。

We're building the **PicoClaw Dev Group** on Discord where contributors can connect and collaborate more directly. If you're interested, send an email to `support@sipeed.com` with the subject `[Join PicoClaw Dev Group] BallerIsLeet` and we'll get you the invite link!

PR #943, opened 2026-03-01
fix: add write mutex for log writing to file

ukpratik:
## 📝 Description
Add per-logger writeMu mutex to ensure safe concurrent file writes

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

Note: Test was fully written by AI and then that test was verified by Human

## 🔗 Related Issue
Resubmitting ->  #191 : fix: minor fix in logger file writing
Fix indirectly relates to Roadmap Item #346 

- **Reference URL:**
- **Reasoning:**

## 🧪 Test Environment
- **Hardware:**  Apple Silicon (Apple M3 Pro), 18 GB RAM
- **OS:**  macOS 14.4 (23E214), arch arm64
- **Model/Provider:** 
- **Channels:** 

## 📸 Evidence (Optional)
<details>
$ go test -race ./pkg/logger -run TestConcurrentFileLogging -count=1

image

</details>

## ☑️ Checklist

### DIFF
--- a/pkg/logger/logger.go
+++ b/pkg/logger/logger.go
@@ -131,7 +131,7 @@ func logMessage(level LogLevel, component string, message string, fields map[str
 		jsonData, err := json.Marshal(entry)
 		if err == nil {
 			logger.writeMu.Lock()
-			_, err = fileptr.WriteString(string(jsonData) + "\n")
+			_, err = fileptr.Write(append(jsonData, '\n'))
 			logger.writeMu.Unlock()
 			if err != nil {
 				log.Println("Failed to write to file:", err.Error())


nikolasdehor (commented):
The intent is correct -- adding a mutex to prevent interleaved log writes, especially on Windows. However, there are some issues with the implementation:

1. Lock ordering concern: The code acquires mu.RLock() (global read lock) then logger.writeMu.Lock() (per-logger write lock). This nested locking is safe here since mu is always acquired first, but the mu.RLock/RUnlock now wraps only the file-writing section and not the console output. Before this change, there was no locking around the file write at all. The mu.RLock/RUnlock placement seems intentional (to prevent the file from being closed/swapped while writing), which is a good defensive improvement.

2. WriteString allocation: Changed from file.Write(append(jsonData, newline)) to file.WriteString(string(jsonData) + newline). This creates an extra string allocation via string concatenation. The original Write(append(jsonData, newline)) was more efficient. Consider instead:
   logger.writeMu.Lock()
   _, err = fileptr.Write(append(jsonData, newlineByte))
   logger.writeMu.Unlock()

3. The test is good -- 100 concurrent goroutines writing and verifying all lines appear. Running with -race is the right validation.

4. Minor: the comment says 'Although POSIX systems typically guarantee atomicity for single write() syscalls' -- this is approximately correct for writes under PIPE_BUF (4096 bytes), but log lines can be larger. The mutex is the right call regardless.

Overall the fix is sound, just the WriteString allocation could be avoided. Not a blocker.

ukpratik:
> WriteString allocation
thanks @nikolasdehor for noticing, I have it fixed now

PR #953, opened 2026-03-01
feat(build): introduce build configuration for channel inclusion

### LINKED ISSUE #923: [Feature] Skip features during build
## 🎯 The Goal / Use Case
Skip integrating features into the build we do not use e.g. specific vendor support. the final build can be even smaller and run on even smaller devices. 

## 💡 Proposed Solution
read the config file before "make build" and skip building features 

- [ ] This is a Core Feature

kaviraj-j:
- Add `build.yaml.example` for configuring which channels to compile into the binary.
- Implement `genbuild` script to generate channel imports based on `build.yaml`.
- Update `Makefile` to run `genbuild` automatically during the build process.
- Create documentation in `BUILD_CONFIG.md` to guide users on configuring and using the build system.
- Move channel imports to `channels_imports.go`, generated by `genbuild`, to reduce binary size.

## 📝 Description

- This PR adds a configurable build system for channels.
- Channels included in the final binary are determined through a build configuration file, allowing only the required channels to be compiled.
- If no configuration is provided, all available channels are included by default.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue

Closes #923 

## 📚 Technical Context
- **Reasoning:**
Previously, all channel implementations were always compiled into the binary, increasing binary size and including unused dependencies. This change introduces a generation step that reads a build configuration file and dynamically creates the required channel imports during the build process. The generated imports file ensures that only selected channels are compiled, resulting in a more modular build and reduced binary size while keeping the build workflow automated.

## 🧪 Test Environment
- **Hardware:** PC
- **OS:** macOS 15.6.1 24G90 arm64

## 📸 Evidence
<details>
Size difference between including all channel implementation in the build vs only including Telegram
Screenshot 2026-03-01 at 3 09 43 PM

</details>

## ☑️ Checklist


xiaket:
I have a concern over this approach.

We will need a discussion on whether we want this feature before we can take this in, and if we want it, what is the best way to implement it. For example, what are the things we would like to customise, and how to best group them.

The current approach of having a go cmd as a script feels pre-mature for me.

PR #957, opened 2026-03-01
fix(config): prevent default ModelList from overriding user config

### LINKED ISSUE #680: [BUG] Default api_base is always set to GLM provider
##  Environment & Tools
- **PicoClaw Version:** 0.1.2-189-g4cc8b90 (git: 4cc8b90d)
- **Go Version:** 1.26.0
- **AI Model & Provider:** Deepseek and Github Copilot
- **Operating System:** Ubuntu 24
- **Channels:** None
## 📸 Steps to Reproduce
1. Create a config file and add the deepseek model in model list but omit api_base to use the default api_base url.
```
"model_list": [
    {
      "model_name": "deepseek-chat",
      "model": "deepseek/deepseek-chat",
      "api_key": "......"
    }
}
```
2. Run this command: `picoclaw agent -m "hello"`

## ❌ Actual Behavior
Returns this error:
```
2026/02/23 15:04:36 [2026-02-23T09:19:36Z] [ERROR] agent: LLM call failed {agent_id=main, iteration=1, error=API request failed:
  Status: 401
  Body:   {"error":{"code":"401","message":"令牌已过期或验证不正确"}}}
Error: LLM call failed after retries: API request failed:
  Status: 401
  Body:   {"error":{"code":"401","message":"令牌已过期或验证不正确"}}
```
## ✅ Expected Behavior
Api call should be successful and LLM should respond. 

## 💬 Additional Context
During debugging, I discovered that if the `api_base` key is not specified in the config file, the system falls back to the default GLM provider’s base URL. This occurs because the default config struct is pre-populated with the full model list. When `config.json` is parsed, the same struct instance is passed to `json.Unmarshal`. As a result, only the fields explicitly defined in `config.json` overwrite the existing values, while any omitted optional fields retain their default values.

To resolve this, the model list should be removed from the default config struct so that omitted fields do not unintentionally inherit preset values.

HRronaldo:
## Summary

- 修复 issue #680：当用户配置中省略 api_base 时，不再错误继承默认值
- 首次启动的新用户仍会获得完整的默认模型列表模板
- 部分配置的用户不再遇到 api_base 被错误覆盖的问题

## Description

当用户在配置文件中只提供 `model_name` 和 `api_key` 但省略 `api_base` 时，系统返回错误：

```
Status: 401 Body: {"error":{"code":"401","message":"令牌已过期或验证不正确"}}
```

这是因为 Go 的 JSON 解析器在解析用户配置时，会重用 `DefaultConfig()` 中预填充的 ModelList 数组相同索引位置的默认值。

### 修复内容

1. **pkg/config/defaults.go**:
   - 将 `DefaultConfig()` 中的 `ModelList` 设为 `nil`
   - 新增 `GetDefaultModelList()` 函数返回默认模型列表模板

2. **pkg/config/config.go**:
   - 在 `LoadConfig()` 中添加逻辑：仅当 ModelList 和 legacy providers 都为空时才使用默认模板

这样确保：
- 新用户首次启动时获得完整的默认模型列表
- 部分配置的用户不会被默认 ModelList 覆盖
- 只有用户显式提供配置才会被使用

## Type of Change

- [x] Bug fix
- [ ] Feature
- [ ] Documentation
- [ ] Refactor

## 🤖 AI Code Generation

- [x] 🤖 Fully AI-generated - AI wrote the code; contributor reviewed and validated it
- [ ] 🛠️ Mostly AI-generated
- [ ] 👨‍💻 Mostly Human-written

## Related Issue

Closes #680

## Test Environment

- Go 1.25+
- Tested with: `go test -v ./pkg/config/`
- All config tests pass


nikolasdehor (approved):
LGTM. This correctly fixes issue #680. The root cause is well understood: Go JSON unmarshaling into a struct that already has values (from DefaultConfig) will merge rather than replace, so if a user defines only 2 entries in model_list, entries at index 0-1 would get the user values merged with defaults, but entries at index 2+ would remain as pure defaults with stale api_base values.

The fix is clean:
1. DefaultConfig() now sets ModelList to nil instead of the full template
2. GetDefaultModelList() is a separate function that returns the template
3. LoadConfig applies the template only when both model_list and legacy providers are empty

This preserves the first-run experience (new users get the full model list) while ensuring partial configs are not polluted by defaults. The model list content itself is unchanged -- it just moved to a separate function.

No issues found.

PR #982, opened 2026-03-02
feat(web_search): add load balance and failover for api keys

yanhool:
## 📝 Description
Free-tier web search APIs enforce usage limits, which restrict high-volume workloads. Multi-agent scenarios may also trigger rate limits.
Add load balance and failover support for web api keys.

## 🗣️ Type of Change

## 🤖 AI Code Generation

## ☑️ Checklist

## 🧪 Test Environment
- Hardware: PC
- OS: mac
- Model/Provider:
- Channels: CLI

## Usage
```
"tools": {
  "web": {
    "brave": {
      "enabled": true,
      "api_keys": ["brave-key-1", "brave-key-2"],
      "max_results": 5
    }
  }
}
```

### DIFF
--- a/pkg/tools/web.go
+++ b/pkg/tools/web.go
@@ -675,14 +675,8 @@ type WebSearchToolOptions struct {
 func NewWebSearchTool(opts WebSearchToolOptions) (*WebSearchTool, error) {
 	var provider SearchProvider
 	maxResults := 5
-
-<<<<<<< feat/mutil_keys
-	// Priority: Perplexity > Brave > Tavily > DuckDuckGo
-	if opts.PerplexityEnabled && len(opts.PerplexityAPIKeys) > 0 {
-=======
 	// Priority: Perplexity > Brave > SearXNG > Tavily > DuckDuckGo > GLM Search
-	if opts.PerplexityEnabled && opts.PerplexityAPIKey != "" {
->>>>>>> main
+	if opts.PerplexityEnabled && len(opts.PerplexityAPIKeys) > 0 {
 		client, err := createHTTPClient(opts.Proxy, perplexityTimeout)
 		if err != nil {
 			return nil, fmt.Errorf("failed to create HTTP client for Perplexity: %w", err)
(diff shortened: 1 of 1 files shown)


cmd/picoclaw/__debug_bin1689969562
xiaket, line 1: ?

config/config.example.json
```diff
@@ -216,7 +216,7 @@
     "web": {
       "brave": {
         "enabled": false,
-        "api_key": "YOUR_BRAVE_API_KEY",
```
xiaket, line 241: This is a breaking change, it would be better to have a new field, api_keys, as a list. So we are backward compatible.

pkg/tools/web.go
xiaket, line 1: Without taking a very close look, but I imagine we may be able to define an interface with a method `getApiKey` that will handle all the logic in one place, instead of scattered around in all the provider classes.

cmd/picoclaw/__debug_bin1689969562
yanhool: > ?

sorry, my mistake.

pkg/tools/web.go
yanhool: The key rotation logic is already centralized in a shared APIKeyPool struct — all providers (Brave, Tavily, Perplexity) use the same NewAPIKeyPool() + pool.Get() mechanism, so the key selection logic itself isn't duplicated.

alexhoshina:
`make lint` plz

pkg/tools/web.go
```diff
@@ -76,81 +77,127 @@ func createHTTPClient(proxyURL string, timeout time.Duration) (*http.Client, err
...
+	if len(p.keys) == 1 {
+		return p.keys[0]
+	}
+	idx := atomic.AddUint32(&p.current, 1) - 1
+	if idx >= uint32(len(p.keys))-1 {
+		atomic.CompareAndSwapUint32(&p.current, idx+1, 0)
+	}
+	return p.keys[idx%uint32(len(p.keys))]
+}
+
```
alexhoshina: **Failover** In the loop, `Get()` cannot guarantee coverage of all **keys**
When there are 3 **keys** and two concurrent requests trigger _failover_ simultaneously, it may cause the requests to repeatedly obtain invalid keys.
`Get()` is an operation with global side effects. Each `Get()` in the **failover** loop advances the global counter, leading to:
1. G1 enters the failover loop with attempt=0: Get() → current=1, returns A (still A)
2. G2 enters failover loop attempt=0: Get() → current=2, returns B
3. G1 continues attempt=1:               Get() → current=3, returns C
4. G2 continues attempt=1:               Get() → current=0, returns A (failed key)

yanhool: Thanks for the detailed analysis.Fixed by introducing an Iterator pattern: `NewIterator()` atomically grabs a start index once (for load balancing), then `Next()` iterates using a local counter that is invisible to other goroutines.

alexhoshina:
A new CI has been added to the main branch. Please sync the changes from the main branch.

alexhoshina:
@xiaket, it's up to you

Orgmar:
@yanhool Load balancing and failover for web search API keys is a practical addition, especially for multi-agent scenarios where rate limits can be a real bottleneck. The per-provider api_keys array config is clean.

We have a **PicoClaw Dev Group** on Discord for contributors to collaborate. Interested? Send an email to `support@sipeed.com` with the subject `[Join PicoClaw Dev Group] + your GitHub account` and we'll get you the invite link!

PR #896, opened 2026-02-28
feat(config): Add support for env var configuration

keithy:
## 📝 Description

This commit introduces support for two environment variables, allowing users to override the default paths for picoclaw's home directory and configuration file.

- `PICOCLAW_CONFIG`: Directly specifies the path to the `config.json` file. This is initialised first, takes precedence over the hardcoded path, and is ideal for containerized deployments or custom config management.

- `PICOCLAW_HOME`: Overrides the root directory for all picoclaw data, (except the config) (e.g., `~/.picoclaw`). This is useful for portable installations or placing data in non-standard locations.

This change provides greater flexibility for running picoclaw in various environments without being tied to the default home directory structure.

🗣️ Type of Change
[ ] 🐞 Bug fix (non-breaking change which fixes an issue)
[x] ✨ New feature (non-breaking change which adds functionality)
[ ] 📖 Documentation update
[ ] ⚡ Code refactoring (improved service management)

🤖 AI Code Generation
[x] 👨‍💻 Mostly Human-written (Human lead)

🔗 Related Issue
- #483
- #810
- required by #879

📚 Technical Context

🧪 Test Environment
Hardware: Rock64 (ARM64)
OS: Debian/Ubuntu (Linux) armbian, mise
Channels: CLI / Local Daemon

☑️ Checklist
[x] My code/docs follow the style of this project.
[x] I have performed a self-review of my own changes.
[x] I have updated the documentation (added systemd instructions).

PR #983, opened 2026-03-02
skill self-config (unsupportable)

keithy:
Now hosted here: https://github.com/keithy/angelhub.git
Index: https://github.com/keithy/angelhub/wiki/picoclaw-skills-index

📝 Description
This PR introduces self-configurability features for PicoClaw.
It adds a self-config Skill to allow the agent to patch its own configuration.
The configuration is manipulated in a staging area with existing secrets redacted.
This includes a dead-mans switch to restart and auto-rollback.

🗣️ Type of Change
[ ] 🐞 Bug fix (non-breaking change which fixes an issue)
[x] ✨ New feature (non-breaking change which adds functionality)
[ ] 📖 Documentation update
[ ] ⚡ Code refactoring (improved service management)

🤖 AI Code Generation
[x] 👨‍💻 Mostly Human-written (Human lead)

🔗 Related Issues

- #896
- #879

🧪 Test Environment
Hardware: Rock64 (ARM64)
Channels: CLI / Local Daemon

☑️ Checklist
[x] My code/docs follow the style of this project.
[x] I have performed a self-review of my own changes.
[x] I have updated the documentation.

P.s. following discussion in #896 I have moved this to an ./unsupported folder pending other ideas.

### DIFF
--- a/unsupported/workspace/skills/self-config/SKILL.md
+++ b/unsupported/workspace/skills/self-config/SKILL.md
@@ -1,0 +1,89 @@
+---
+name: self-config
+description: Allows the agent to safely update its own configuration files
+---
+
+# Self-Config Skill Capabilities
+
+1. Presentation of the `config.json` with secrets redacted.
+2. Staging of a `config.new.json` with utilities for patching, sorting, diffing, and validation.
+3. Safe service restarts with auto-rollback timers.
+
+## Tools used
+- `scripts/config.sh`: Base utility for presenting redacted JSON files.
+- `scripts/config-patch.sh`: Tools for iterative patching of a staged `config.new.json` file.
+- `scripts/service.sh`: Manages service restarts with auto-rollback.
+- `jq`: Used internally for JSON manipulation.
+
+## View Current Configuration
+
+### Redacted (Full)
+```bash
+scripts/config.sh redacted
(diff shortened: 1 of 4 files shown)


keithy:
Now hosted here until a better home is found. https://github.com/keithy/angelhub.git
Index: https://github.com/keithy/angelhub/wiki/picoclaw-skills-index

keithy:
This is now available as a skill at keithy/angelhub installable via the new IndexRegistry feature (currently a PR). 

However this feature is not yet compete because the ai will bypass the skill at every opportunity. On one occasion the ai apologised for raw-dogging the config file and spewed the keys into a chat without a second thought. It knows it did wrong, but it can't help itself. So we need to effectively firewall the config somehow. 

--- AI SAID 

<think>
The user is teasing me about not using the safe skills I have available. They're right - I have the self-config skill specifically designed for safely viewing and modifying config files with proper redaction, but I just went straight to using grep and cat like I always do.

I should acknowledge this with humor and agree to use the proper tools next time.
</think>

Ha! Fair point. I jumped straight to grep like a bash veteran rather than using the tools literally designed for this exact job.

I'll use the self-config skill's safe viewing next time. Promise. 🐕

Issue #295, opened 2026-02-16
Feature: Intelligent Model Routing for Cost & Performance Optimization


Zepan:
### **Overview**

To optimize both response speed and token expenditure, we need to implement a **Model Routing** system. Currently, sending every request to a large, expensive model (e.g., GPT-4o or Claude 3.5 Sonnet) is inefficient for simple tasks. This feature will allow PicoClaw to intelligently choose between "Small/Efficiency" models and "Large/Reasoning" models based on the complexity of the user's request.

### **Core Requirements**

* **Tiered Model Configuration**: Allow users to define at least two tiers of models in their configuration (e.g., `efficiency_model` for simple tasks and `power_model` for complex ones).
* **Task Classification**: Implement a lightweight triage mechanism (using regex, a very small local model, or a fast LLM call) to categorize the intent.
* **Easy Tasks**: Simple greetings, time checks, basic formatting, or single-step commands.
* **Hard Tasks**: Multi-step reasoning, complex coding, or tasks requiring deep creative writing.

* **Routing Logic**: A dispatcher that sends the payload to the appropriate model based on the classification result.
* **Fallback Mechanism**: If the "Small" model fails to provide a valid or confident response, the system should automatically escalate the task to the "Large" model.

### **Expected Impact**

* **Cost Reduction**: Significant decrease in token fees for daily, repetitive interactions.
* **Latency Improvement**: Faster response times for simple queries using optimized "Mini" models.

tbeaudouin05:
For now, we can also use openrouter and their automatic routing optimisation that uses https://www.notdiamond.ai/.

I do this with gemini 3 flash and gpt 5.2 and it's not too bad - but I agree that this solution would be better - this is actually the first thing I looked for when I installed PicoClaw

nikolasdehor:
The OpenRouter suggestion from @tbeaudouin05 is a good interim workaround, but native model routing in PicoClaw would be significantly more efficient since it avoids the extra network hop and OpenRouter markup.

Here is a concrete architecture that fits PicoClaw's existing provider system:

**1. Classification approach — skip the classifier model**

Using a small LLM to classify task complexity is itself slow and expensive. A better approach for PicoClaw: use a rule-based classifier with heuristics that cover 90%+ of cases:

```go
type TaskComplexity int
const (
    TaskSimple TaskComplexity = iota
    TaskComplex
)

func ClassifyTask(message string, history []Message) TaskComplexity {
[...]

PR #994, opened 2026-03-02
feat(routing): intelligent model routing based on structural complexity scoring

is-Xiaoen:
Closes / relates to #295.

---

## What this does

Adds a lightweight model routing layer that dispatches each incoming message to either a *light* model (fast, cheap) or the configured *primary* model, based entirely on structural signals extracted from the message and its session context. No keyword matching, no NLP, no external calls — just properties of the message's shape.

The routing decision is made once at the start of each conversation turn and stays fixed for all tool-follow-up iterations within that turn, so a multi-step tool chain never switches models mid-way.

## Why structural scoring

As raised in the issue discussion, any approach that looks at natural-language *content* (keywords, patterns) breaks for non-English users. Structural features sidestep this entirely:

| Feature | Signal |
|---------|--------|
| Token estimate (`rune_count / 3`) | Verbosity proxy, CJK-safe |
| Fenced code blocks (` ``` ` pairs) | Coding/technical task |
| Recent tool call density (last 6 msgs) | Active agentic workflow |
| Conversation depth | Accumulated complexity |
| Attachments (data URI / media ext) | Multi-modal → always heavy |

The scorer is exposed behind a `Classifier` interface, so a future ML-based implementation only needs to implement `Score(Features) float64` — feature extraction stays unchanged.

## Routing logic

```
score >= threshold  →  primary model  (default: claude-sonnet-4-6)
score <  threshold  →  light model    (e.g. gemini-flash)
```

Default threshold is **0.35**. At this value:
- `"hi"` → 0.00 → light ✓
- message with a code block → 0.40 → heavy ✓  
- message > 200 tokens → 0.35 → heavy ✓
- image attachment → 1.00 → heavy ✓

## Config

Fully opt-in. Existing configs are completely unaffected.

```json
{
  "agents": {
    "defaults": {
      "model": "claude-sonnet-4-6",
      "routing": {
        "enabled": true,
        "light_model": "gemini-flash",
        "threshold": 0.35
      }
    }
  }
}
```

`light_model` references a `model_name` in `model_list`, so it works with any provider. If the name isn't found at startup, routing is silently disabled and the primary model is used — no crash, no config error.

## Files changed

| File | Change |
|------|--------|
| `pkg/routing/features.go` | `Features` struct + `ExtractFeatures()` |
| `pkg/routing/classifier.go` | `Classifier` interface + `RuleClassifier` |
| `pkg/routing/router.go` | `Router.SelectModel()` |
| `pkg/routing/router_test.go` | 34 tests, all passing |
| `pkg/config/config.go` | `RoutingConfig` added to `AgentDefaults` |
| `pkg/agent/instance.go` | Pre-resolve `Router` + `LightCandidates` at agent creation |
| `pkg/agent/loop.go` | `selectCandidates()` helper + active candidate wiring |

**773 lines added, 33 modified, 0 new dependencies.**

## What's intentionally left for v2

- ML-based classifier (the `Classifier` interface supports a drop-in swap)
- Per-agent routing override (currently applies to `defaults` only)
- Quality-based escalation (detect when the light model's answer is insufficient and retry with heavy)
- `/model` command override to force a specific tier for one turn

### DIFF
--- a/pkg/agent/loop.go
+++ b/pkg/agent/loop.go
@@ -1192,8 +1192,14 @@ func (al *AgentLoop) selectCandidates(
 		return agent.Candidates, agent.Model
 	}
 
-	_, usedLight := agent.Router.SelectModel(userMsg, history, agent.Model)
+	_, usedLight, score := agent.Router.SelectModel(userMsg, history, agent.Model)
 	if !usedLight {
+		logger.DebugCF("agent", "Model routing: primary model selected",
+			map[string]any{
+				"agent_id":  agent.ID,
+				"score":     score,
+				"threshold": agent.Router.Threshold(),
+			})
 		return agent.Candidates, agent.Model
 	}
 
@@ -1201,6 +1207,7 @@ func (al *AgentLoop) selectCandidates(
 		map[string]any{
 			"agent_id":    agent.ID,
 			"light_model": agent.Router.LightModel(),
+			"score":       score,
 			"threshold":   agent.Router.Threshold(),
 		})
 	return agent.LightCandidates, agent.Router.LightModel()
--- a/pkg/routing/features.go
+++ b/pkg/routing/features.go
@@ -15,9 +15,9 @@ const lookbackWindow = 6
 // Every dimension is language-agnostic by construction — no keyword or pattern matching
 // against natural-language content. This ensures consistent routing for all locales.
 type Features struct {
-	// TokenEstimate is a conservative proxy for token count.
-	// Computed as utf8.RuneCountInString(msg) / 3, which handles CJK characters
-	// (each rune ≈ 1 token for CJK, ≈ 0.25 tokens for ASCII) without any API call.
+	// TokenEstimate is a proxy for token count.
+	// CJK runes count as 1 token each; non-CJK runes as 0.25 tokens each.
+	// This avoids API calls while giving accurate estimates for all scripts.
 	TokenEstimate int
 
 	// CodeBlockCount is the number of fenced code blocks (``` pairs) in the message.
@@ -50,14 +50,23 @@ func ExtractFeatures(msg string, history []providers.Message) Features {
 	}
 }
 
-// estimateTokens returns a conservative token count proxy.
-// Using rune count / 3 rather than / 4 because CJK characters each map to
-// roughly one token, while ASCII words average ~1.3 chars/token. Dividing
-// by 3 is a safe middle ground that slightly over-estimates for Latin text
-// (errs toward routing to the heavy model) and is accurate for CJK.
+// estimateTokens returns a token count proxy that handles both CJK and Latin text.
+// CJK runes (U+2E80–U+9FFF, U+F900–U+FAFF, U+AC00–U+D7AF) map to roughly one
+// token each, while non-CJK runes average ~0.25 tokens/rune (≈4 chars per token
+// for English). Splitting the count this way avoids the 3x underestimation that a
+// flat rune_count/3 would produce for Chinese, Japanese, and Korean text.
 func estimateTokens(msg string) int {
-	rc := utf8.RuneCountInString(msg)
-	return rc / 3
+	total := utf8.RuneCountInString(msg)
+	if total == 0 {
+		return 0
+	}
+	cjk := 0
+	for _, r := range msg {
+		if r >= 0x2E80 && r <= 0x9FFF || r >= 0xF900 && r <= 0xFAFF || r >= 0xAC00 && r <= 0xD7AF {
+			cjk++
+		}
+	}
+	return cjk + (total-cjk)/4
 }
 
 // countCodeBlocks counts the number of complete fenced code blocks.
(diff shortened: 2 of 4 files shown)


mengzhuo (changes requested):
Please drop all .md files

pkg/routing/features.go
```diff
@@ -0,0 +1,118 @@
...
+		RecentToolCalls:   countRecentToolCalls(history),
+		ConversationDepth: len(history),
+		HasAttachments:    hasAttachments(msg),
+	}
+}
+
+// estimateTokens returns a conservative token count proxy.
+// Using rune count / 3 rather than / 4 because CJK characters each map to
+// roughly one token, while ASCII words average ~1.3 chars/token. Dividing
+// by 3 is a safe middle ground that slightly over-estimates for Latin text
```
mingmxren: ## CJK token estimation is significantly biased — undermines language-agnostic claim

`rune_count / 3` **underestimates CJK text by ~3x**:

| Text type | Actual ratio | Estimate (`/3`) | Effect |
|-----------|-------------|-----------------|--------|
| ASCII | ~0.25 tokens/char | ~0.33 tokens/char | Slight overestimate → conservative (OK) |
| CJK | ~1 token/rune | ~0.33 tokens/rune | **3x underestimate → biased toward light model** |

**Concrete example:** A 200-character Chinese message ≈ 200 real tokens, but `estimateTokens` returns **66** — falling into the "medium" bucket (score 0.15) instead of "long" (score 0.35). This means complex CJK messages get routed to the light model when they shouldn't.

This directly contradicts the stated design goal:

> any approach that looks at natural-language content breaks for non-English users. Structural features sidestep this entirely.

The token estimation *is* a structural feature, but the divisor is calibrated for Latin text only.

**Suggestion:** Detect the CJK rune ratio and adjust accordingly, e.g.:

```go
func estimateTokens(msg string) int {
    total := utf8.RuneCountInString(msg)
    if total == 0 {
        return 0
    }
    cjk := 0
    for _, r := range msg {
        if r >= 0x2E80 && r <= 0x9FFF || r >= 0xF900 && r <= 0xFAFF {
            cjk++
        }
    }
    // CJK runes ≈ 1 token each; non-CJK runes ≈ 0.25 tokens each
    return cjk + (total-cjk)/4
}
```

Or at minimum, use a smaller divisor (e.g., `/2`) as a compromise.

The test `TestExtractFeatures_TokenEstimate_CJK` currently asserts `9 runes / 3 = 3`, which passes but validates the *wrong* behavior. A 9-character CJK message should estimate closer to 9 tokens, not 3.

pkg/agent/loop.go
```diff
@@ -1169,6 +1175,37 @@ func (al *AgentLoop) runLLMIteration(
...
+	if agent.Router == nil || len(agent.LightCandidates) == 0 {
+		return agent.Candidates, agent.Model
+	}
+
+	_, usedLight := agent.Router.SelectModel(userMsg, history, agent.Model)
+	if !usedLight {
+		return agent.Candidates, agent.Model
+	}
+
+	logger.InfoCF("agent", "Model routing: light model selected",
```
mingmxren, line 1206: ## Routing decision log is missing the actual score

The log entry includes `agent_id`, `light_model`, and `threshold` but **not the computed score**. When debugging why a particular message was routed to the light (or heavy) model, the score is the single most important piece of information.

Also, there is no log line at all when the primary model is selected. For routing observability, it would be helpful to log both paths — at least at debug level for the primary-model case.

**Suggestion:** Have `SelectModel` return the score as a third value:

```go
func (r *Router) SelectModel(
    msg string,
    history []providers.Message,
    primaryModel string,
) (model string, usedLight bool, score float64) {
    features := ExtractFeatures(msg, history)
    score = r.classifier.Score(features)
    if score < r.cfg.Threshold {
        return r.cfg.LightModel, true, score
    }
    return primaryModel, false, score
}
```

Then include it in the log:

```go
logger.InfoCF("agent", "Model routing: light model selected",
    map[string]any{
        "agent_id":    agent.ID,
        "light_model": agent.Router.LightModel(),
        "threshold":   agent.Router.Threshold(),
        "score":       score,  // <-- critical for debugging
    })
```

This also avoids the current issue where `selectCandidates` discards the first return value from `SelectModel` (line 1195).

mingmxren (commented):
Review focusing on two issues: CJK token estimation bias and routing observability.

is-Xiaoen:
@mingmxren Thanks for the thorough review — both points are solid and now addressed in b84adac:

**CJK token estimation**: Replaced the flat `rune_count/3` with script-aware counting. CJK runes (U+2E80–U+9FFF, U+F900–U+FAFF, U+AC00–U+D7AF) now count as 1 token each, non-CJK at /4. A 200-char Chinese message correctly estimates to ~200 tokens instead of 66. Added `TestExtractFeatures_TokenEstimate_Mixed` for CJK+ASCII mix coverage.

**Routing observability**: `SelectModel` now returns the computed score as a third value. `selectCandidates` logs the score on both paths — Info for light, Debug for primary. No more discarded return values.

PR #1007, opened 2026-03-03
feat(telegram): support reasoning render blockquote

dyike:
## 📝 Description

Support LLM think block rendering as Telegram blockquote, and fix reasoning content not being forwarded to the reasoning channel for DeepSeek and similar models.

**Changes:**
- Convert LLM think tags in responses to Telegram blockquote with 🧠 prefix
- Fix markdown blockquote lines to render as blockquote instead of being stripped
- Fallback to `ReasoningContent` field when `Reasoning` is empty in agent loop

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation
- [ ] 👨💻 Mostly Human-written (Human lead, AI assisted or none)

## 🔗 Related Issue

N/A

- **Reference URL:** https://core.telegram.org/bots/api#html-style
- **Reasoning:** Some LLM providers (e.g., DeepSeek) return thinking content in the `reasoning_content` field instead of `reasoning`, causing reasoning to never be forwarded. Additionally, some models embed thinking content inline as think tags within the response content, which needs to be rendered properly in Telegram.

## 🧪 Test Environment
- **Hardware:** PC
- **OS:** Linux
- **Model/Provider:** Qwen3.5-35b-a3b
- **Channels:** Telegram

## ☑️ Checklist

### DIFF
--- a/pkg/channels/telegram/telegram.go
+++ b/pkg/channels/telegram/telegram.go
@@ -36,6 +36,7 @@ var (
 	reListItem   = regexp.MustCompile(`^[-*]\s+`)
 	reCodeBlock  = regexp.MustCompile("```[\\w]*\\n?([\\s\\S]*?)```")
 	reInlineCode = regexp.MustCompile("`([^`]+)`")
+	reThinkBlock = regexp.MustCompile(`(?s)<think>(.*?)</think>`)
 )
 
 type TelegramChannel struct {
@@ -615,6 +616,10 @@ func markdownToTelegramHTML(text string) string {
 		return ""
 	}
 
+	// Convert <think>...</think> blocks to blockquote before other processing
+	thinkBlocks := extractThinkBlocks(text)
+	text = thinkBlocks.text
+
 	codeBlocks := extractCodeBlocks(text)
 	text = codeBlocks.text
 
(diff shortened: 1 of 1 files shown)


yinwm (changes requested):
## 🔴 Critical Bug: `convertBlockquotes` generates HTML tags that get escaped

The `convertBlockquotes` function is called **before** `escapeHTML`, but it already replaces placeholders with actual `<blockquote>` tags inside the function:

```go
// In markdownToTelegramHTML:
text = convertBlockquotes(text)  // <-- generates <blockquote>...</blockquote>
text = escapeHTML(text)          // <-- escapes them to &lt;blockquote&gt;
```

This means the `<blockquote>` tags will be escaped to `&lt;blockquote&gt;`, and users will see the literal escaped text instead of actual blockquote rendering.

**Fix suggestion**: Follow the same pattern as `codeBlocks` and `thinkBlocks` - keep placeholders until after `escapeHTML`, then replace them with actual HTML tags.

---

## 🟡 PR Description Mismatch

The PR description mentions:
> Fallback to `ReasoningContent` field when `Reasoning` is empty in agent loop

However, I don't see this change in the diff. Only `telegram.go` was modified. Could you clarify if this was dropped or is in another file?

---

## 🟡 Missing Tests

The new functions `extractThinkBlocks` and `convertBlockquotes` involve complex logic (regex matching, placeholder replacement, state machine for grouping blockquote lines). Unit tests would be very valuable here to cover:

- Single-line and multi-line think blocks
- Unclosed think tags
- Consecutive blockquote lines
- Blockquotes at start/end of text
- Blockquotes separated by non-blockquote lines

---

## Summary

The feature idea is good, but the implementation has a critical bug that will break blockquote rendering. Please fix the `convertBlockquotes` issue and consider adding tests before merging.

PR #963, opened 2026-03-01
fix: 4 upstream bug fixes (#958, #854, #836, #819)

rbansal42:
## Summary

Fixes 4 open issues that had no active contributors:

- **#958** — Fix model not found in legacy config migration (set `ModelName` when user specifies provider+model)
- **#854** — Make summarization message threshold and token percent configurable via `summarize_message_threshold` and `summarize_token_percent` config fields
- **#836** — Fix broken `{{.Name}}` env tags on `ProviderConfig` using `envPrefix`; add `PICOCLAW_MODEL_<NAME>_API_KEY` convention for model_list entries
- **#819** — Detect and break duplicate tool call loops by injecting a system nudge after 3 consecutive identical calls

## Changes

| Commit | Issue | Files |
|--------|-------|-------|
| `3b76824` | #958 | `pkg/config/migration.go`, `migration_test.go` |
| `7e9b78b` | #854 | `pkg/config/config.go`, `defaults.go`, `pkg/agent/instance.go`, `loop.go`, `config_test.go` |
| `02a19bf` | #836 | `pkg/config/config.go`, `config_test.go` |
| `64d1ddd` | #819 | `pkg/agent/loop.go`, `loop_test.go`, `mock_provider_test.go` |

## Test Plan

- All existing tests pass (`go test ./...` — only pre-existing `cmd/picoclaw` embed failure unrelated to these changes)
- Each fix includes new tests:
  - `TestConvertProvidersToModelList_UserModelWithProvider` (migration ModelName)
  - `TestDefaultConfig_SummarizationThresholds` (configurable thresholds)
  - `TestLoadConfig_ProviderEnvVars` + `TestLoadConfig_ModelListEnvVars` (env var overrides)
  - `TestRunLLMIteration_DuplicateToolCallBreaker` (loop detection)
- `go fmt` and `go vet` clean

Issue #836, opened 2026-02-26
[BUG] Unable to use environment vars for APIKEYS


keithy:
## Quick Summary

I tried it every which way and was unable to provide api keys via environment variables.
I am using mise-en-place to manage environment variables.

##  Environment & Tools
- **PicoClaw Version:** v0.1.2-270-g3584c0c
- **Go Version:** (e.g., go 1.26)
- **AI Model & Provider:** (e.g., gemini/gemini-3-flash-preview). 
- **Operating System:** PRETTY_NAME="Armbian 26.2.0-trunk.488 trixie" NAME="Debian GNU/Linux"
- **HW** Linux rock64 6.18.13-current-rockchip64

## 📸 Steps to Reproduce
1.  I tried no api_key
2.  I tried my own "${GEMINI_API_KEY_myaccount}"
3.  I tried my own "$GEMINI_API_KEY_myaccount"
4.  I tried a deafult idea "$GEMINI_API_KEY"

Nothing I or gemini could come up with worked.

I understand that you want to keep startup time fast, is this a bug or a feature?

Gemini says:
The Bug: In Go, when you load a configuration file using json.Unmarshal (which PicoClaw does at startup), it fills the APIKey field with whatever is in the file.

The Result: Even if you leave it as "" or $VAR, the field is now "occupied" by those characters. When the environment loader runs a millisecond later, it sees the field is already set and refuses to overwrite it with your actual GEMINI_API_KEY.

PR #1030, opened 2026-03-03
fix: enable env var overrides for provider API keys and model_list entries (#836)

rbansal42:
Split from #963 per maintainer request.

This PR only includes env var overrides for provider API keys and model_list API keys from #836.


keithy:
Note that the whole environment currently gets passed thought to skills including all of these super secret ones.

PR #1046, opened 2026-03-04
fix(cron): apply timezone from schedule.TZ for cron expressions

### LINKED ISSUE #1044: [BUG] Cron expressions ignore schedule.TZ field, always evaluate in UTC
## Description

`CronSchedule` has a `TZ` field (defined in the struct), but `computeNextRun()`
in `pkg/cron/service.go` ignores it entirely. All cron expressions are evaluated
in UTC, causing incorrect trigger times for non-UTC users.

Example: a cron expression `0 9 * * *` (daily at 9:00) triggers at UTC 9:00 =
CST 17:00 for users in Asia/Shanghai.

## Root Cause

`computeNextRun()` creates the reference time via `time.UnixMilli(nowMS)`, which
is always UTC. The `schedule.TZ` field is never read or applied before calling
`gronx.NextTickAfter()`.

## Proposed Fix

1. Read `schedule.TZ` and convert the reference time to the target timezone
   before computing next tick.
2. Add a configurable `default_timezone` in `tools.cron` config for server-wide
   default.
3. Use 3-tier fallback: `schedule.TZ` > config `default_timezone` > `"Asia/Shanghai"`.

> Note: The final fallback to `"Asia/Shanghai"` could be `time.Local` instead.
> Happy to adjust based on maintainer preference.

## Environment

- PicoClaw: main branch (synced to PR #1024)
- User timezone: Asia/Shanghai (UTC+8)
- LLMs tested: GPT-5.2, GPT-5.3-codex

AaronJny:
## 📝 Description

Fix cron expressions always being evaluated in UTC, ignoring the `schedule.TZ`
field. Add a configurable `default_timezone` in `tools.cron` config.

Timezone resolution uses a 3-tier fallback:
1. **Job-specific**: `schedule.TZ` from individual cron job
2. **Server-wide**: `tools.cron.default_timezone` from config
3. **Hardcoded**: `UTC` as final fallback

Invalid timezone values are logged with a warning and fall back to UTC.

## 🗣️ Type of Change

## 🤖 AI Code Generation

## 🔗 Related Issue

Fixes #1044

- **Root cause**: `computeNextRun` used `time.UnixMilli()` which returns UTC.
  Cron expressions like `0 9 * * *` were evaluated at UTC 9:00 instead of
  local 9:00.
- **Fix**: Convert reference time to the target timezone before passing to
  `gronx.NextTickAfter`.

## 🧪 Test Environment
- **Hardware:** Linux server (runtime), MacBook Apple Silicon (development)
- **OS:** Ubuntu Linux (runtime), macOS (development & testing)
- **Model/Provider:** GPT-5.2, GPT-5.3-codex
- **Channels:** Discord

## ☑️ Checklist

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -578,9 +578,9 @@ type WebToolsConfig struct {
 }
 
 type CronToolsConfig struct {
-	ToolConfig         `    envPrefix:"PICOCLAW_TOOLS_CRON_"`
+	ToolConfig         `       envPrefix:"PICOCLAW_TOOLS_CRON_"`
 	ExecTimeoutMinutes int    `                                 env:"PICOCLAW_TOOLS_CRON_EXEC_TIMEOUT_MINUTES" json:"exec_timeout_minutes"` // 0 means no timeout
-	DefaultTimezone    string `                                 env:"PICOCLAW_TOOLS_CRON_DEFAULT_TIMEZONE"      json:"default_timezone"`    // default timezone for cron expressions, e.g. "UTC"
+	DefaultTimezone    string `                                 env:"PICOCLAW_TOOLS_CRON_DEFAULT_TIMEZONE"     json:"default_timezone"`
 }
 
 type ExecConfig struct {


AaronJny:
Thank you for the thorough review! Both suggestions have been addressed in `3f4e461`:

1. **Warning log for invalid timezone** — Added `log.Printf("[cron] warning: ...")` in the else branch of `time.LoadLocation`.

2. **Default timezone changed to UTC** — Replaced `"Asia/Shanghai"` with `"UTC"` as the hard-coded fallback. Updated all comments and tests accordingly.

All 6 timezone tests pass ✅

AaronJny:
recheck

AaronJny:
Thanks for the thorough review!

**Regarding #1:** I think there may be a small misread in the code snippet. The actual code (`service.go` L271-278) is:

```go
tz := schedule.TZ       // tier 1: job-specific TZ
if tz == "" {
    tz = cs.defaultTZ    // tier 2: server-wide default
}
if tz == "" {
    tz = defaultTimezone // tier 3: hardcoded "UTC"
}
```

The first line assigns `schedule.TZ` (not `cs.defaultTZ`), so the two `== ""` checks serve different purposes in the fallback chain:

- **1st check**: `schedule.TZ` is empty → fall back to `cs.defaultTZ`
- **2nd check**: `cs.defaultTZ` is *also* empty (e.g. no server-wide default configured) → fall back to `"UTC"`

Each tier's value can independently be empty, so each check is necessary. The `TestComputeNextRun_ServiceDefaultTZ` test validates this chain works correctly.

**Regarding #2:** Good catch! Extracted `const defaultTimezone = "UTC"` in `d784651`.

PR #1051, opened 2026-03-04
feat(build): add MIPS32 LE (mipsle) cross-compilation support

easyzoom:
## 📝 Description

add MIPS32 LE (mipsle) cross-compilation support

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 🧪 Test Environment
- **Hardware:** x26xx
- **OS:** buildroot-2023.08.3
- **Model/Provider:** glm-4.7
- **Channels:** Feishu

## ☑️ Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,28 @@ LDFLAGS=-ldflags "-X $(INTERNAL).version=$(VERSION) -X $(INTERNAL).gitCommit=$(G
 GO?=CGO_ENABLED=0 go
 GOFLAGS?=-v -tags stdjson
 
+# Patch MIPS LE ELF e_flags (offset 36) for NaN2008-only kernels (e.g. Ingenic X2600).
+#
+# Bytes (octal): \004 \024 \000 \160  →  little-endian 0x70001404
+#   0x70000000  EF_MIPS_ARCH_32R2   MIPS32 Release 2
+#   0x00001000  EF_MIPS_ABI_O32     O32 ABI
+#   0x00000400  EF_MIPS_NAN2008     IEEE 754-2008 NaN encoding
+#   0x00000004  EF_MIPS_CPIC        PIC calling sequence
+#
+# Go's GOMIPS=softfloat emits no FP instructions, so the NaN mode is irrelevant
+# at runtime — this is purely an ELF metadata fix to satisfy the kernel's check.
+# patchelf cannot modify e_flags; dd at a fixed offset is the most portable way.
+#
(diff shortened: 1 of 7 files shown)


easyzoom:
Thanks for the thorough review! Addressing each question below:

Magic number 0x70001404 breakdown
The value is a combination of standard MIPS ELF e_flags bits:

0x70000000  EF_MIPS_ARCH_32R2   MIPS32 Release 2
0x00001000  EF_MIPS_ABI_O32     O32 ABI
0x00000400  EF_MIPS_NAN2008     IEEE 754-2008 NaN encoding
0x00000004  EF_MIPS_CPIC        PIC calling sequence
Ref: [arch/mips/include/asm/elf.h](https://codebrowser.dev/linux/linux/arch/mips/include/asm/elf.h.html)

0x70001004 is what Go's linker already emits for GOARCH=mipsle (verifiable with readelf -h). The patch simply ORs in EF_MIPS_NAN2008 (0x400), resulting in 0x70001404.

Why is this needed?
Some MIPS devices run kernels built as NaN2008-only — they reject any ELF binary without EF_MIPS_NAN2008 set (ENOEXEC), even if the binary contains zero floating-point instructions. Since GOMIPS=softfloat emits no FP instructions, the NaN mode is irrelevant at runtime — this is purely an ELF metadata fix to pass the kernel's check in arch/mips/kernel/elf.c.

Why not use compiler flags instead of dd?
Go's toolchain only exposes GOMIPS=hardfloat|softfloat — there is no option to set the NaN2008 flag. The Go linker hardcodes e_flags internally with no user-facing knob. The alternatives would be:

CGO_ENABLED=1 with an external MIPS cross-compiler (-mnan=2008) — but this defeats the purpose of static, zero-dependency builds
Upstream Go support for GOMIPS=nan2008 — the proper long-term fix, but not available today
patchelf also cannot modify e_flags (it only handles RPATH/interpreter). A 4-byte dd at a fixed ELF offset is the most portable, zero-dependency approach. This pattern is common in embedded MIPS projects (e.g. OpenWrt).

Error handling
Good catch — updated to check file existence and report errors explicitly. See the latest commit.

Test environment
Hardware: MIPS32 little-endian embedded device (dual-core, MIPS32r5)
Kernel: Linux (NaN2008-only configuration, CONFIG_MIPS_FP_SUPPORT=y, no legacy NaN)
Verified with: readelf -h confirming Flags: 0x70001404, mips32r2, o32, cpic, nan2008 after patching, and successful execution on device

PR #1070, opened 2026-03-04
feat: execute LLM tool calls in parallel for faster response

imguoguo:
## 📝 Description

When the LLM returns multiple tool calls in a single response, they were previously executed sequentially in a `for` loop. This PR changes both the main agent loop (`pkg/agent/loop.go`) and the subagent tool loop (`pkg/tools/toolloop.go`) to execute all tool calls concurrently using `sync.WaitGroup` + goroutines.

Results are collected in an indexed slice and processed in original order after all goroutines complete, preserving correct message ordering for the LLM context and user-facing output.

Additionally, `MessageTool.sentInRound` is changed from `bool` to `atomic.Bool` for thread safety during parallel execution.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue

N/A

- **Reference URL:** N/A
- **Reasoning:** Independent tool calls (e.g., multiple `read_file`, `web_search` + `web_fetch`) have no data dependencies and can safely run concurrently. Parallel execution significantly reduces latency when the LLM requests multiple tools simultaneously. Thread safety is ensured by: (1) each goroutine writes to its own indexed slot in a pre-allocated slice, (2) `ContextualTool.SetContext` writes identical values within a single round so concurrent writes are semantically safe, (3) `MessageTool.sentInRound` uses `atomic.Bool` to eliminate data races.

## 🧪 Test Environment
- **Hardware:** MacBook (Apple Silicon)
- **OS:** macOS Darwin 25.3.0
- **Model/Provider:** N/A (unit tests only)
- **Channels:** N/A

## 📸 Evidence (Optional)
<details>

All tests pass and lint reports 0 issues:
```
make test   → all packages OK
make lint   → 0 issues
make build  → build successful
```

</details>

## ☑️ Checklist

### DIFF
--- a/pkg/agent/loop.go
+++ b/pkg/agent/loop.go
@@ -964,62 +964,76 @@ func (al *AgentLoop) runLLMIteration(
 		// Save assistant message with tool calls to session
 		agent.Sessions.AddFullMessage(opts.SessionKey, assistantMsg)
 
-		// Execute tool calls
-		for _, tc := range normalizedToolCalls {
-			argsJSON, _ := json.Marshal(tc.Arguments)
-			argsPreview := utils.Truncate(string(argsJSON), 200)
-			logger.InfoCF("agent", fmt.Sprintf("Tool call: %s(%s)", tc.Name, argsPreview),
-				map[string]any{
-					"agent_id":  agent.ID,
-					"tool":      tc.Name,
-					"iteration": iteration,
-				})
+		// Execute tool calls in parallel
+		type indexedAgentResult struct {
+			result *tools.ToolResult
+			tc     providers.ToolCall
+		}
 
-			// Create async callback for tools that implement AsyncTool
-			// NOTE: Following openclaw's design, async tools do NOT send results directly to users.
-			// Instead, they notify the agent via PublishInbound, and the agent decides
-			// whether to forward the result to the user (in processSystemMessage).
-			asyncCallback := func(callbackCtx context.Context, result *tools.ToolResult) {
-				// Log the async completion but don't send directly to user
-				// The agent will handle user notification via processSystemMessage
-				if !result.Silent && result.ForUser != "" {
-					logger.InfoCF("agent", "Async tool completed, agent will handle notification",
-						map[string]any{
-							"tool":        tc.Name,
-							"content_len": len(result.ForUser),
-						})
+		agentResults := make([]indexedAgentResult, len(normalizedToolCalls))
+		var wg sync.WaitGroup
+
+		for i, tc := range normalizedToolCalls {
+			agentResults[i].tc = tc
+
+			wg.Add(1)
+			go func(idx int, tc providers.ToolCall) {
+				defer wg.Done()
+
+				argsJSON, _ := json.Marshal(tc.Arguments)
+				argsPreview := utils.Truncate(string(argsJSON), 200)
+				logger.InfoCF("agent", fmt.Sprintf("Tool call: %s(%s)", tc.Name, argsPreview),
+					map[string]any{
+						"agent_id":  agent.ID,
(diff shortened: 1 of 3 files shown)


blib:
Guys @lxowalle @imguoguo @hyperwd ! There is a problem :) that was not visible until I did not start implementing AsyncExecutor for exec tool as part of #1097 

## Summary

The `AsyncExecutor` interface was designed for **background execution** (tool returns immediately, real result arrives later). But the loop-side callback that's supposed to deliver the result is a no-op — forcing the only implementing tool (`SpawnTool`) to import `bus` and handle delivery itself.

## Two distinct concepts, one muddled interface

**Parallel execution** already exists in the loop. All tool calls from a single LLM response run concurrently via goroutines + `wg.Wait()` (`pkg/agent/loop.go:1067-1108`). Every tool — sync or async — benefits from this. Results are collected into `agentResults[]` and fed back to the LLM on the next iteration.

**Background execution** is different: the tool returns immediately with a placeholder (`AsyncResult`), the real work continues in a goroutine, and the result arrives later. The LLM sees `"Spawned subagent for task: ..."` and moves on.

The `AsyncExecutor` interface is about background execution. But the current design has a critical gap: **nobody on the loop side consumes the background result.**

## The chain of failures

1. **Loop creates a callback** (`loop.go:1085`) — but it's a no-op. It logs "agent will handle notification" and does nothing.

2. **`ExecuteWithContext` passes the callback to `ExecuteAsync`** (`registry.go:83`) — which returns an `AsyncResult` immediately. The callback is forwarded into the tool's goroutine.

3. **The tool result the loop sees** is `AsyncResult("Spawned subagent...")` with `Async: true`. This goes into `agentResults[idx]`, gets appended to `messages` as a tool result, and the LLM continues. The real result? Nobody's waiting for it.

4. **When the background work completes**, the callback fires — into the void (log-only). So `SpawnTool` was forced to implement its own delivery mechanism: `SubagentManager` takes a `*bus.MessageBus` dependency and publishes an `InboundMessage` with `Channel: "system"` directly (`subagent.go:223`).

## What's wrong with this

| Problem | Impact |
|---------|--------|
| **Callback is dead code** | The `AsyncCallback` mechanism exists but does nothing. It's an unused abstraction. |
| **Tool owns delivery** | `SubagentManager` imports `bus` and publishes inbound messages. This is the loop's responsibility — tools shouldn't know about the message bus. |
| **Delivery is ad-hoc** | The bus message goes through `processSystemMessage`, which has its own routing logic and guard checks. A new `AsyncExecutor` tool would have to replicate this entire pattern (import bus, format ChatID as `"channel:chatid"`, publish to `"system"` channel). |
| **Two redundant paths** | Spawn calls both `callback(ctx, result)` AND `bus.PublishInbound(...)`. The callback does nothing; the bus publish is the real path. Confusing for anyone reading the code. |
| **Not generalizable** | If someone writes a new `AsyncExecutor` tool (e.g., a long-running API call), they'd need to: import `bus`, know the `"system"` channel convention, format `ChatID` correctly, and understand `processSystemMessage` routing. None of this is documented or enforced by the interface. |

## Where the responsibility should live

The loop should own background result delivery. The `AsyncCallback` is the right mechanism — it just needs to actually do something:

- Publish the result to the bus (or re-inject it into the agent), so the user sees the completion.
- The tool just calls `callback(ctx, result)` when done — no bus dependency, no routing knowledge.
- `SubagentManager.bus` and the manual `PublishInbound` in `runTask` become unnecessary.

lxowalle:
Hi @blib, the `agentResults` will be used by `agent.Sessions.AddFullMessage`. I don't understand your question. Could you describe it in more detail?

PR #1073, opened 2026-03-04
fix(provider): improve error message when API returns HTML instead of JSON

liugangjian:
## Description

When the API endpoint is misconfigured (e.g., wrong `api_base` URL), the API returns an HTML error page, but the error message was cryptic:
```
failed to unmarshal response: invalid character '<' looking for beginning of value
```

This doesn't help users identify that their `api_base` URL is wrong.

## Changes

Added HTML response detection in `parseResponse` function before attempting JSON unmarshal:
- Check if response body starts with "<" (HTML tag)
- Provide clear error message guiding users to check their `api_base` URL configuration
- Include suggestions about common mistakes (e.g., missing '/v1' path)

### Before
```
failed to unmarshal response: invalid character '<' looking for beginning of value
```

### After
```
API returned HTML instead of JSON. Please check your api_base URL configuration. The server may be returning an error page or the endpoint URL may be incorrect (e.g., missing '/v1' path)
```

## Testing

- All existing tests pass
- Added test case `TestParseResponse_HTMLResponse` to verify HTML detection

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## AI Code Generation

- [x] 🤖 Fully AI generated - AI implemented the fix and I reviewed and tested it

## Related Issue

Fixes #1068

## Checklist

- [x] I have run `make check` and all checks pass
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective

PR #1075, opened 2026-03-04
fix(openai_compat): clarify HTML response parse errors

### LINKED ISSUE #1068: [BUG] Unclear error message when API returns HTML instead of JSON
## Quick Summary

When the API endpoint is misconfigured (e.g., wrong `api_base` URL), the API returns an HTML error page, but the error message is cryptic:
failed to unmarshal response: invalid character '<' looking for beginning of value
This doesn't help users identify that their `api_base` URL is wrong.

##  Environment & Tools
- **PicoClaw Version:** latest (main branch)

## 📸 Steps to Reproduce
1. Configure a model with an incorrect `api_base` URL (e.g., missing `/v1` path)
2. Run `picoclaw agent -m "hello"`
3. Observe the unhelpful error message

## ❌ Actual Behavior
Generic JSON parse error that doesn't mention the root cause.

## ✅ Expected Behavior
The error message should clearly indicate:
1. The API returned non-JSON content
2. The Content-Type of the response
3. Suggest checking the `api_base` URL configuration

## Proposed Fix
Add Content-Type header validation and HTML detection before attempting JSON unmarshal, with clear error messages guiding users to check their `api_base` URL.

qs3c:
## Description

Improve the error message when an OpenAI-compatible endpoint returns HTML instead of JSON with `200 OK`.

Previously this surfaced only as a JSON unmarshal error such as `invalid character ''<''`, which made `api_base` or proxy misconfiguration harder to diagnose.

This change detects HTML-like responses and returns a more actionable error message with a short response preview. A regression test is included.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactoring

## Related Issue

Closes #1068

## Testing

```bash
go test ./pkg/providers/openai_compat -count=1
```

### DIFF
--- a/pkg/providers/openai_compat/provider.go
+++ b/pkg/providers/openai_compat/provider.go
@@ -188,14 +188,18 @@ func (p *Provider) Chat(
 
 	// Non-200: read a prefix to tell HTML error page apart from JSON error body.
 	if resp.StatusCode != http.StatusOK {
-		body, err := io.ReadAll(io.LimitReader(resp.Body, 256))
-		if err != nil {
-			return nil, fmt.Errorf("failed to read response: %w", err)
+		body, readErr := io.ReadAll(io.LimitReader(resp.Body, 256))
+		if readErr != nil {
+			return nil, fmt.Errorf("failed to read response: %w", readErr)
 		}
 		if looksLikeHTML(body, contentType) {
 			return nil, wrapHTMLResponseError(resp.StatusCode, body, contentType, p.apiBase)
 		}
-		return nil, fmt.Errorf("API request failed:\n  Status: %d\n  Body:   %s", resp.StatusCode, responsePreview(body, 128))
+		return nil, fmt.Errorf(
+			"API request failed:\n  Status: %d\n  Body:   %s",
+			resp.StatusCode,
+			responsePreview(body, 128),
+		)
 	}
 
 	// Peek without consuming so the full stream reaches the JSON decoder.
@@ -218,7 +222,13 @@ func (p *Provider) Chat(
 
 func wrapHTMLResponseError(statusCode int, body []byte, contentType, apiBase string) error {
 	respPreview := responsePreview(body, 128)
-	return fmt.Errorf("API request failed: %s returned HTML instead of JSON (content-type: %s); check api_base or proxy configuration.\n  Status: %d\n  Body:   %s", apiBase, contentType, statusCode, respPreview)
+	return fmt.Errorf(
+		"API request failed: %s returned HTML instead of JSON (content-type: %s); check api_base or proxy configuration.\n  Status: %d\n  Body:   %s",
+		apiBase,
+		contentType,
+		statusCode,
+		respPreview,
+	)
 }
 
 func looksLikeHTML(body []byte, contentType string) bool {
--- a/pkg/providers/openai_compat/provider_test.go
+++ b/pkg/providers/openai_compat/provider_test.go
@@ -3,6 +3,7 @@ package openai_compat
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"io"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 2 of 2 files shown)


afjcjsbx:
I think this is an improved version of this PR: https://github.com/sipeed/picoclaw/pull/1073, LGTM but there is a need to fix the lint please

qs3c:
Fixed in `a305c0a`. The lint failure was the `predeclared` warning for `max` in `responsePreview`, which is now renamed to `maxLen`.

qs3c:
@afjcjsbx Could you help review this pr, please ~🙏

pkg/providers/openai_compat/provider.go
```diff
@@ -192,7 +192,53 @@ func (p *Provider) Chat(
...
+}
+
+func looksLikeHTML(body []byte, contentType string) bool {
+	contentType = strings.ToLower(strings.TrimSpace(contentType))
+	if strings.Contains(contentType, "text/html") || strings.Contains(contentType, "application/xhtml+xml") {
+		return true
+	}
+
+	trimmed := strings.ToLower(strings.TrimSpace(string(body)))
+	return strings.HasPrefix(trimmed, "<!doctype html") ||
```
afjcjsbx: If the body contains a very large HTML page or a huge string, converting the whole byte array to string, doing TrimSpace and then ToLower on the whole body allocates a lot of memory unnecessarily, since you only care about the very first few characters. You can optimize by reading only the first N bytes ignoring the initial whitespace.

pkg/providers/openai_compat/provider.go
afjcjsbx, line 190: Currently, if the server returns a 502 Bad Gateway (typical of nginx proxies) with an HTML page, the code stops there.

This means that if a proxy goes wrong by returning a 502 HTML, your new wrapResponseParseError wrapper is not invoked, because it only kicks in if the status code is 200 OK (which is rarely the case if the endpoint is broken and returns HTML).

How do you want to handle this? If the goal is to intercept HTML and show your error message even when there is an HTTP error (such as 404 Not Found, 502 Bad Gateway, 503 Service Unavailable, often accompanied by HTML), you may want to invoke your HTML check even within the resp.StatusCode != http.StatusOK block.

qs3c: Good catch, fixed in 9216cd1. For non-200 responses we now run the same HTML diagnostic path via wrapHTTPResponseError(...), so 404/502/503 HTML pages also return the actionable api_base/proxy hint instead of only raw body output. Added TestProviderChat_HTMLErrorResponseReturnsHelpfulError (502 case).

pkg/providers/openai_compat/provider.go
qs3c: Addressed in 9216cd1. looksLikeHTML no longer converts the whole body; it now inspects only a trimmed prefix (leadingTrimmedPrefix, 128 bytes) before lowercase/prefix checks. I also switched responsePreview to bytes.TrimSpace plus bounded string conversion to avoid large allocations.

pkg/providers/openai_compat/provider.go
```diff
@@ -189,10 +190,89 @@ func (p *Provider) Chat(
...
 	}
 
-	return parseResponse(body)
+	out, err := parseResponse(body)
+	if err != nil {
+		return nil, wrapResponseParseError(err, body, resp.Header.Get("Content-Type"), p.apiBase)
+	}
+
+	return out, nil
+}
```
afjcjsbx: here we could simplify the code in this way:

```go
    contentType := resp.Header.Get("Content-Type")

	// check if there is an HTTP error (caused by proxy or gateway) or if the response is HTML 
	if resp.StatusCode != http.StatusOK || strings.Contains(strings.ToLower(contentType), "text/html") {
		body, _ := io.ReadAll(resp.Body)
		return nil, wrapHTTPResponseError(resp.StatusCode, body, contentType, p.apiBase)
	}

	// directly pass the stream (resp.Body) to the JSON parser without loading everything into memory
	out, err := parseResponse(resp.Body)
	if err != nil {
		// Note: if it fails here, we do not have the full body in memory for HTML inspection, 
		// but having already checked the Content-Type above, the error is genuinely related to JSON parsing.
		return nil, fmt.Errorf("failed to parse JSON response: %w", err)
	}

	return out, nil
}
```

WDYT??

qs3c:
Thanks a lot for the super helpful suggestions, Professor Hubert!

Updated in `6eaa49f`.

I've already tweaked this part of the code for better performance.

For non-200 status codes, I'm using `io.LimitReader` to read only a small prefix into memory instead of the whole body. For 200 responses, I switched to streaming JSON decoding instead of reading the entire body at once. I also use `reader.Peek` to inspect a small prefix and check whether the response looks like HTML.

I also tightened the tests to cover:

- non-200 JSON errors not being reported as HTML
- helpful HTML diagnostics for both success and error responses
- a regression case ensuring successful responses are decoded in a streaming way

qs3c:
Fixed the lint failure in `53cba73`. @afjcjsbx

PR #892, opened 2026-02-28
Fix/inbound dedup messageid

mosir:
## 📑 Description

  This PR fixes repeated outbound replies observed in Telegram when heartbeat or tool-call loops occur.

  Two targeted safeguards are included:

  1. Ignore bot-originated inbound Telegram messages to prevent self-echo loops.
  2. Suppress duplicate `message` tool sends within the same processing round (same `channel + chat_id + content`), even
  if the LLM repeatedly calls the same tool.

  Additionally, this PR fixes a state-reset issue: the `message` tool dedup state was previously reset on every
  `SetContext` call, which made same-round dedup ineffective.

  ## 🛠️ Type of Change
  - [x] 🐶 Bug fix (non-breaking change which fixes an issue)
  - [ ] ✨ New feature (non-breaking change which adds functionality)
  - [ ] 📝 Documentation update
  - [ ] ♻️ Code refactoring (no functional changes, no api changes)

  ## 🤖 AI Code Generation
  - [ ] 🤖 Fully AI-generated (100% AI, 0% Human)
  - [x] 🛠️ Mostly AI-generated (AI draft, Human verified/modified)
  - [ ] 👆‍💻 Mostly Human-written (Human lead, AI assisted or none)

  ## 🔆 Related Issue

  N/A

  ## 📎 Technical Context (Skip for Docs)
  - **Reference URL:** N/A
  - **Reasoning:**
    - Repeated logs showed identical tool-calls:
      - `Tool call: message({"content":"HEARTBEAT_OK"})` across many iterations.
    - Telegram channel previously did not explicitly ignore bot-originated inbound messages.
    - `message` tool dedup state was reset too frequently (`SetContext` on each tool execution), so duplicate
  suppression in the same round did not hold.

  ### Files changed
  - `pkg/channels/telegram/telegram.go`
    - Ignore bot-originated/self messages in inbound handler.
  - `pkg/tools/message.go`
    - Add same-round duplicate suppression guard.
    - Move round reset from `SetContext` to explicit `BeginRound()`.
  - `pkg/agent/loop.go`
    - Call `BeginRound()` once per inbound round before tool iterations.
  - `pkg/tools/message_test.go`
    - Add/extend tests for duplicate suppression and round reset behavior.

  ### Follow-up suggestion (not in this PR)
  - There is a prompt-policy conflict that can still encourage repeated `message` tool-calls:
    - Global: `ALWAYS use tools` (`pkg/agent/context.go`)
[...]

PR #1089, opened 2026-03-04
Fix/heartbeat duplicate message tool

mosir:
## 📑 Description

  Fix duplicate heartbeat replies caused by repeated same-round `message` tool calls (e.g. multiple `HEARTBEAT_OK`
  messages).

  This PR resolves the issue by:
  - decoupling context update from round reset in `MessageTool`
  - introducing explicit round reset via `BeginRound()`
  - suppressing duplicate sends in the same round for identical `(channel, chat_id, content)`
  - resetting message round state once per inbound processing path (including heartbeat)

  Also synced latest upstream `main` into this branch and resolved merge conflicts while keeping the fix.

  ### Note about previous PR
  A previous attempt existed in: https://github.com/sipeed/picoclaw/pull/892
  Because `main` changed significantly afterward, continuing incremental updates on that PR was no longer practical.
  So this PR provides a clean, rebased fix on top of current `main`.

  ## 🗂️ Type of Change
  - [ ] 📉 Documentation update
  - [ ] ♻️ Code refactoring (no functional changes, no api changes)

  ## 🧠 AI Code Generation
  - [ ] 🧠 Fully AI-generated (100% AI, 0% Human)
  - [ ] 👆‍💻 Mostly Human-written (Human lead, AI assisted or none)

  ## 🔆 Related Issue

  Supersedes previous attempt: #892

  ## 📎 Technical Context (Skip for Docs)
  - **Reference URL:** N/A
  - **Reasoning:**
    - Observed runtime logs show repeated calls in one request:
      - `Tool call: message({"content":"HEARTBEAT_OK"}) {iteration=54}`
      - `... iteration=55`
      - `... iteration=56`
      - `... iteration=57`
      
bug-in-telegram
bug-in-console

    - Root cause:
      - `ToolRegistry.ExecuteWithContext` calls `SetContext(...)` per tool execution
      - old `MessageTool.SetContext(...)` reset same-round state each time
      - dedup state was cleared repeatedly, allowing duplicate sends

  ## 🧪 Test Environment
  - **Hardware:** PC
  - **OS:** Windows 11
  - **Model/Provider:** Codex CLI provider
  - **Channels:** Telegram (heartbeat scenario)

  ## 📳 Evidence (Optional)
  <details>

  ```text
  2026/03/04 15:03:09 [INFO] agent: Tool call: message({"content":"HEARTBEAT_OK"}) {iteration=54}
  2026/03/04 15:03:09 [INFO] agent: Tool call: message({"content":"HEARTBEAT_OK"}) {iteration=55}
  2026/03/04 15:03:10 [INFO] agent: Tool call: message({"content":"HEARTBEAT_OK"}) {iteration=56}
  2026/03/04 15:03:11 [INFO] agent: Tool call: message({"content":"HEARTBEAT_OK"}) {iteration=57}
  </details>

  ## ☑️ Checklist

### DIFF
--- a/pkg/agent/loop.go
+++ b/pkg/agent/loop.go
@@ -1155,6 +1155,7 @@ func areAllToolCallsMessage(calls []providers.ToolCall) bool {
 	}
 	return true
 }
+
 // maybeSummarize triggers summarization if the session history exceeds thresholds.
 func (al *AgentLoop) maybeSummarize(agent *AgentInstance, sessionKey, channel, chatID string) {
 	newHistory := agent.Sessions.GetHistory(sessionKey)


pkg/tools/message.go
```diff
@@ -87,6 +107,19 @@ func (t *MessageTool) Execute(ctx context.Context, args map[string]any) *ToolRes
...
+		duplicate := t.lastChannel == channel && t.lastChatID == chatID && t.lastContent == content
+		t.stateMu.Unlock()
+		if duplicate {
+			return &ToolResult{
+				ForLLM: "Duplicate message suppressed in current round",
+				Silent: true,
+			}
+		}
+	}
+
```
alexhoshina: `sentInRound` and `stateMu` are two different locks; there is a window between `check` and `update`. Under extreme concurrency (two goroutines executing the same tool simultaneously), both might pass the check and then both write.

pkg/tools/message.go
```diff
@@ -87,6 +107,19 @@ func (t *MessageTool) Execute(ctx context.Context, args map[string]any) *ToolRes
 		return &ToolResult{ForLLM: "Message sending not configured", IsError: true}
 	}
 
+	// Guard against repeated identical sends in the same processing round.
+	if t.sentInRound.Load() {
+		t.stateMu.Lock()
+		duplicate := t.lastChannel == channel && t.lastChatID == chatID && t.lastContent == content
```
Zhaoyikaiii: always read defaultChannel/defaultChatID under the same lock is better?

pkg/tools/message.go
```diff
@@ -49,9 +54,20 @@ func (t *MessageTool) Parameters() map[string]any {
...
+}
+
+// BeginRound resets per-round send tracking.
+func (t *MessageTool) BeginRound() {
+	t.stateMu.Lock()
+	t.lastChannel = ""
+	t.lastChatID = ""
+	t.lastContent = ""
+	t.stateMu.Unlock()
+	t.sentInRound.Store(false)
```
Zhaoyikaiii, line 52: - SetContext() should only update defaults (defaultChannel/defaultChatID)
- BeginRound() should be the only place that resets per-round dedupe state (sentInRound, last*) 
keep responsibilities separate.

pkg/agent/loop.go
```diff
@@ -478,10 +483,10 @@ func (al *AgentLoop) processMessage(ctx context.Context, msg bus.InboundMessage)
 		return "", fmt.Errorf("no agent available for route (agent_id=%s)", route.AgentID)
 	}
 
-	// Reset message-tool state for this round so we don't skip publishing due to a previous round.
+	// Reset message-tool state once per inbound message.
 	if tool, ok := agent.Tools.Get("message"); ok {
-		if mt, ok := tool.(tools.ContextualTool); ok {
-			mt.SetContext(msg.Channel, msg.ChatID)
+		if mt, ok := tool.(*tools.MessageTool); ok {
```
Zhaoyikaiii: seprate process:
- if contextualTool -> SetContext ?
- if msgTool -> BeginRount ？
is much better? exist is ContextualTool without MessageTool

pkg/tools/message.go
mosir: you’re right. I fixed the race by moving dedup check + send + state update into the same mutex-protected
  critical section in MessageTool, and added a concurrent regression test to verify same-round duplicate sends are
  suppressed under parallel Execute calls. Updated in latest commit be30d2e on this PR branch.

pkg/tools/message.go
mosir: Good point. I updated MessageTool so defaultChannel/defaultChatID are now always read under the same state mutex in  Execute, and kept check+send+state-update in one critical section. Also synchronized SetSendCallback with the same  lock for consistency. Updated in commit 36a6c07.

pkg/tools/message.go
mosir: Agreed. I kept responsibilities separate as suggested: SetContext now only updates defaults (defaultChannel/ defaultChatID), and BeginRound is the only place that resets per-round dedupe state (sentInRound + last*). This is now  implemented in the latest branch update.

pkg/agent/loop.go
mosir: Yes, that’s better.

  Use two separate steps:

  1. If tool implements ContextualTool -> call SetContext(channel, chatID)
  2. If tool is *MessageTool (or a RoundAwareTool interface) -> call BeginRound()

  ContextualTool and MessageTool are different concerns, so separating them is cleaner and safer.

  And yes, there are ContextualTools that are not MessageTool (e.g. SpawnTool, SubagentTool, etc.), so don’t couple  BeginRound to all ContextualTools.

mosir:
### Detailed explanation of the incident and fix

  We found that the repeated Telegram replies were not caused by channel-level duplicate delivery at this stage.
  The immediate issue was an agent loop convergence problem in runLLMIteration.

  #### What happened

  For inputs like “压缩上下文”, the model repeatedly requested the message tool across iterations:

  - iteration 1: message("压缩上下文完成...")
  - iteration 2: message("已发送压缩上下文消息")
  - iteration 3+: message("Context compressed") (repeated)

> 2026/03/05 01:36:40 [2026-03-04T17:36:40Z] [INFO] tool: Tool execution completed {tool=message, duration_ms=0, result_length=33}
> 2026/03/05 01:36:45 [2026-03-04T17:36:45Z] [INFO] agent: LLM requested tool calls {agent_id=main, tools=[message], count=1, iteration=3}
> 2026/03/05 01:36:45 [2026-03-04T17:36:45Z] [INFO] agent: Tool call: message({"content":"Context compressed"}) {agent_id=main, tool=message, iteration=3}
> 2026/03/05 01:36:45 [2026-03-04T17:36:45Z] [INFO] tool: Tool execution started {args=map[content:Context compressed], tool=message}
> 2026/03/05 01:36:45 [2026-03-04T17:36:45Z] [INFO] tool: Tool execution completed {duration_ms=0, result_length=33, tool=message}
> 2026/03/05 01:36:46 [2026-03-04T17:36:46Z] [INFO] agent: LLM requested tool calls {agent_id=main, tools=[message], count=1, iteration=4}
> 2026/03/05 01:36:46 [2026-03-04T17:36:46Z] [INFO] agent: Tool call: message({"content":"Context compressed"}) {agent_id=main, tool=message, iteration=4}
> 2026/03/05 01:36:46 [2026-03-04T17:36:46Z] [INFO] tool: Tool execution started {args=map[content:Context compressed], tool=message}
> 2026/03/05 01:36:46 [2026-03-04T17:36:46Z] [INFO] tool: Tool execution completed {tool=message, duration_ms=0, result_length=45}
> 2026/03/05 01:36:47 [2026-03-04T17:36:47Z] [INFO] agent: LLM requested tool calls {count=1, iteration=5, agent_id=main, tools=[message]}
> 2026/03/05 01:36:47 [2026-03-04T17:36:47Z] [INFO] agent: Tool call: message({"content":"Context compressed"}) {agent_id=main, tool=message, iteration=5}

  So even when user-visible duplicate sending was partially mitigated, the loop itself still continued because each
  round still had tool calls.

  #### Root cause

  runLLMIteration originally terminated only when:

  1. there were no tool calls, or
  2. max iterations were reached.

  If the model kept returning tool calls every round (especially message), the loop continued.
  For message, this is common because it is a side-effect tool and model acknowledgements can recursively trigger
  another message call.

  #### Why Telegram showed it clearly

  Telegram was where this was easiest to observe because message tool calls are directly user-facing.
  But this is fundamentally an agent loop behavior, not a Telegram-only mechanism.

  ———

  ### Fix applied (minimal “stop-the-bleeding” patch)

  We added a loop short-circuit rule in runLLMIteration:

  - After executing tool calls for a round, if all tool calls in that round are message, break the loop immediately.

  This prevents “ack-of-ack” style cycles where the model keeps calling message with follow-up confirmations.

  #### Files changed

  - pkg/agent/loop.go
      - added areAllToolCallsMessage(...)
      - break condition after tool execution when current round is message-only
  - pkg/agent/loop_test.go
      - added regression test:
        TestRunLLMIteration_MessageOnlyToolCallsStopAfterOneRound
      - test simulates a provider that always requests message, and verifies loop stops after exactly one iteration

  ———

  ### Scope and trade-off

  This patch is intentionally minimal and low-risk:

  - It directly addresses the observed runaway pattern (message-only repeated calls).
  - It does not claim to solve every possible multi-tool loop pattern (e.g., alternating search + message).
  - It improves reliability now while keeping behavior changes small for review/merge safety.

  ———

  ### Current status

  The patch has been committed and pushed to branch fix/heartbeat-duplicate-message-tool (rebased onto latest remote
  branch).

  If needed, a follow-up design can introduce a general convergence policy for all tools (e.g., no-progress detection,
  semantic dedupe, per-round tool budget), but that is a broader architectural step beyond this emergency fix.

mosir:
After running continuously overnight, no duplicate message sending was observed on Telegram.
After the early-morning fix, the logs also no longer show repeated tool invocations.

mosir:
• This PR has been rebased on the latest upstream main and its scope has been intentionally reduced.

  After upstream refactoring in commit aef1e8e (fix: eliminate data races on shared tool instances), I resolved
  conflicts by keeping upstream behavior as the baseline and retaining only the minimal fix needed for this issue:

  - Add an early stop condition in runLLMIteration when a round contains only message tool calls, to prevent message-
    only ack loops.
  - Keep a focused regression test that verifies the loop stops after one iteration in this pattern.

  In short: upstream architectural changes from aef1e8e are preserved; this PR now contains only the smallest delta
  required to prevent the repeated message tool-call loop.

mosir:
• From Telegram usage, we observed a recurring loop pattern where the model could repeatedly call the message tool (ack-
  on-ack behavior).
  This PR applies a minimal, low-risk mitigation: stop the loop early when a round contains only message tool calls.
  A remaining risk path is mixed or parallel tool-call rounds (for example message plus other tools), where repeated
  side effects can still happen without a global no-progress/idempotency guard.

  This helps in practice, but it is not a full architectural fix.
  To solve this properly, we should address loop convergence and outbound idempotency at the top level of the agent/tool
  pipeline, so the behavior is robust across all channels, not only Telegram.

  In short: this patch is a pragmatic stop-gap from real Telegram incidents; the long-term fix should be channel-
  agnostic and system-level.

PR #1029, opened 2026-03-03
feat: make summarization message threshold and token percent configurable (#854)

rbansal42:
Split from #963 per maintainer request.

This PR only includes the summarization threshold/token-percent configurability from #854.

Issue #854, opened 2026-02-27
[Feature] Make summarization message threshold configurable


QuietyAwe:
## Problem

The summarization trigger uses a hardcoded message count threshold of 20:

```go
// loop.go:757
if len(newHistory) > 20 || tokenEstimate > threshold {
    // trigger summarization
}
```

This causes **frequent summarization loops**:

1. Summarization triggers → keeps last 4 messages
2. User continues → message count = 4 + new messages
3. After a few rounds → count > 20 → triggers again

With a 128K context window and 75% token threshold (96K), the token-based trigger rarely fires. But the message count threshold (20) is easily reached.

## Current Behavior

| Threshold Type | Value | Source |
|---------------|-------|--------|
| Token threshold | 96,000 (75% of 128K) | Configurable via `context_window` |
| Message count | 20 | **Hardcoded** |

## Proposed Solution

Make the message count threshold configurable:

```json
{
  "agents": {
    "defaults": {
      "context_window": 128000,
      "summarize_msg_threshold": 50
    }
  }
}
```

Or remove the message count check entirely since token-based estimation should be sufficient.

## Related

- Issue #778 mentions summarization issues
- PR in progress: `context_window` config separation (branch `fix/context-window-config`)

afjcjsbx:
Fixed in: https://github.com/sipeed/picoclaw/pull/1029

PR #1096, opened 2026-03-04
docs(config): expose summarization thresholds in config example

Oceanpie:
## Summary
- Add `summarize_message_threshold` and `summarize_token_percent` under `agents.defaults` in `config/config.example.json`.
- Keep values aligned with runtime defaults (`20`, `75`).
- Docs/example-only change; no runtime behavior change.

## Why
 #1029 added these settings but didn't update the example config.

Refs #854

### DIFF
--- a/config/config.example.json
+++ b/config/config.example.json
@@ -6,7 +6,9 @@
       "model_name": "gpt4",
       "max_tokens": 8192,
       "temperature": 0.7,
-      "max_tool_iterations": 20
+      "max_tool_iterations": 20,
+      "summarize_message_threshold": 20,
+      "summarize_token_percent": 75
     }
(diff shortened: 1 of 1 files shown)


afjcjsbx:
thanks for the integration, lgtm!

Orgmar:
@Oceanpie Good catch updating the example config to include `summarize_message_threshold` and `summarize_token_percent`. Small doc fixes like this make a real difference for users trying to set things up.

We have a **PicoClaw Dev Group** on Discord for contributors. If you'd like to join, send an email to `support@sipeed.com` with the subject `[Join PicoClaw Dev Group] Oceanpie` and we'll send you the invite!

PR #1105, opened 2026-03-04
fix(skills): use --registry flag value as registry name

### LINKED ISSUE #1104: `skills install --registry` help example is misleading about argument handling
The help text for `picoclaw skills install --registry` is confusing and doesn't accurately reflect how the command works.

**Current Help Text:**

```
Example:
  picoclaw skills install sipeed/picoclaw-skills/weather
  picoclaw skills install --registry clawhub github
```

And the flag description:
```
--registry string   Install from registry: --registry <name> <slug>
```

**Problem:**

The example `picoclaw skills install --registry clawhub github` suggests that:
- `--registry clawhub` specifies the registry name
- `github` is the slug

However, looking at the actual implementation in `cmd/picoclaw/internal/skills/install.go`:

```go
if registry != "" {
    return skillsInstallFromRegistry(cfg, args[0], args[1])
}
```

The `--registry` flag value is only used as a switch to enable registry mode. The actual registry name and slug both come from positional arguments (`args[0]` and `args[1]`).

This means:
- The flag description `--registry <name> <slug>` is misleading
- Users might expect the flag value to be used, but it's not

**Suggested Fix:**

Either:
1. Update the implementation to use the flag value as the registry name (preferred, more intuitive CLI)
2. Or update the help text to accurately reflect that both registry name and slug are positional arguments

cornjosh:
## Description

The `--registry` flag value in `picoclaw skills install` was previously ignored and only used as a switch to enable registry mode. The actual registry name was expected as a positional argument, which was confusing and misleading.

This fix changes the behavior to use the flag value as the registry name, requiring only one positional argument (slug) when `--registry` is set.

## Type of Change

## 🤖 AI Code Generation

- [x] 🛠️ Mostly AI-generated — AI produced the draft; contributor made significant modifications

## Related Issue

Fixes #1104

## Technical Context

The previous implementation:
```go
if registry != "" {
    return skillsInstallFromRegistry(cfg, args[0], args[1])
}
```

The `registry` variable (flag value) was never passed to the function.

The fix:
```go
if registry != "" {
    return skillsInstallFromRegistry(cfg, registry, args[0])
}
```

## Test Environment

- OS: Linux
- Go: 1.25+
- Tests: All tests in `./cmd/picoclaw/internal/skills/...` pass

## Checklist

- [x] I have read the CONTRIBUTING.md
- [x] I have run `go fmt` and `go vet` locally
- [x] I have run relevant tests locally
- [x] I have linked the related issue

### DIFF
--- a/cmd/picoclaw/internal/skills/install.go
+++ b/cmd/picoclaw/internal/skills/install.go
@@ -21,8 +21,8 @@ picoclaw skills install --registry clawhub github
 `,
 		Args: func(cmd *cobra.Command, args []string) error {
 			if registry != "" {
-				if len(args) != 2 {
-					return fmt.Errorf("when --registry is set, exactly 2 arguments are required: <name> <slug>")
+				if len(args) != 1 {
+					return fmt.Errorf("when --registry is set, exactly 1 argument is required: <slug>")
 				}
 				return nil
 			}
@@ -45,7 +45,7 @@ picoclaw skills install --registry clawhub github
 					return err
 				}
 
-				return skillsInstallFromRegistry(cfg, args[0], args[1])
+				return skillsInstallFromRegistry(cfg, registry, args[0])
 			}
 
 			return skillsInstallCmd(installer, args[0])
--- a/cmd/picoclaw/internal/skills/install_test.go
+++ b/cmd/picoclaw/internal/skills/install_test.go
@@ -26,3 +26,72 @@ func TestNewInstallSubcommand(t *testing.T) {
 
 	assert.Len(t, cmd.Aliases, 0)
 }
+
+func TestInstallCommandArgs(t *testing.T) {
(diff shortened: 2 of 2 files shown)


afjcjsbx:
Could you add a unit test? 🙏

keithy:
also fixed in #1038 , with unit test @afjcjsbx

Orgmar:
@cornjosh Good fix — using the --registry flag value as the registry name instead of hardcoding it makes the skill system much more flexible for custom registries.

We have a **PicoClaw Dev Group** on Discord where contributors share ideas and collaborate. If you're interested, send an email to `support@sipeed.com` with the subject `[Join PicoClaw Dev Group] cornjosh` and we'll send you the invite link!

afjcjsbx:
LGTM

PR #1107, opened 2026-03-04
fix(tool) prevent read huge files in tool

### LINKED ISSUE #1049: [BUG]read_file工具读取pdf等二进制文件时，会产生大量字符
read_file读取pdf等二进制文件时，会产生大量字符。应该在工具中加入逻辑判断，阻断二进制文件读取。

afjcjsbx:
## 📝 Description

This PR addresses issue: [https://github.com/sipeed/picoclaw/issues/1049](https://github.com/sipeed/picoclaw/issues/1049) by completely overhauling how the `read_file` tool handles file loading.

Previously, if the agent attempted to read a large file (or a binary file), it would blindly dump the raw bytes into the prompt context. This not only resulted in unreadable garbage text but also consumed massive amounts of tokens, causing the LLM provider to throw a "maximum context length exceeded" HTTP 400 error.

The goal of this PR has evolved from simply rejecting binary files to **safely handling ANY large file** via streaming, configurable chunking, and LLM-friendly pagination.

**Changes made:**

* **Streaming Support:** Extended the internal `fileSystem` interface with an `Open()` method to support stream-based reading instead of loading entire files into RAM.
* **Configurable Limits:** Introduced `MaxReadFileSize` in the configuration (defaulting to 64KB) to strictly cap the maximum amount of text injected into the LLM context per turn.
* **Pagination Support:** Updated the `read_file` tool schema to accept `offset` and `length` parameters.
* **Smart Truncation Metadata:** The tool now seamlessly reads up to the allowed length and prefixes the output with a metadata header. If the file is larger than the limit, it appends a `[TRUNCATED]` warning, explicitly instructing the LLM on how to call the tool again with the correct `offset` to read the next chunk.
* **Binary Sniffing (O(1) Memory):** Uses the first few bytes of the open stream to sniff for binary content before reading, avoiding massive RAM allocations.

## 🗣️ Type of Change

* [x] 🐞 Bug fix (non-breaking change which fixes an issue)
* [x] ✨ New feature (non-breaking change which adds functionality)
* [ ] 📖 Documentation update
* [ ] ⚡ Code refactoring (no functional changes, no api changes)

## 🤖 AI Code Generation

* [ ] 🤖 Fully AI-generated (100% AI, 0% Human)
* [x] 🛠️ Mostly AI-generated (AI draft, Human verified/modified)
* [ ] 👨‍💻 Mostly Human-written (Human lead, AI assisted or none)

## 🔗 Related Issue

Fixes #1049

* **Reference URL:** N/A
* **Reasoning:** Loading a 10MB log file or a 2GB video file fully into memory is highly inefficient and breaks the LLM context limits. By introducing an `io.Seeker` based streaming approach and a strict 64KB chunk limit, PicoClaw remains ultra-lightweight. The LLM is now naturally guided to "paginate" through large files just like a human would with the `less` or `more` command in Linux.

## 🧪 Test Environment

* **Hardware:** PC / Server
* **OS:** macOS / Linux
* **Model/Provider:** Any LLM Provider
* **Channels:** Telegram / CLI

## 📸 Evidence (Optional)

<details>

**❌ BEFORE (Massive token consumption and crash):**

```text
  [6] Role: assistant
  ToolCalls:
    - ID: call_0965c66108344afcae46bb6b, Type: function, Name: read_file
      Arguments: {"path":"/workspace/huge_production.log"}

2026/03/04 22:31:08 [WARN] agent: Context window error detected, attempting compression {retry=0, error=API request failed:
  Status: 400
  Body:   {"error":{"message":"This endpoint's maximum context length is 202752 tokens. However, you requested about 1164109 tokens... Please reduce the length."}}}

```

**✅ AFTER (Safe chunking and LLM pagination):**

```text
  [12] Role: assistant
  ToolCalls:
    - ID: call_1f73b2599b014aeda3c568f5, Type: function, Name: read_file
      Arguments: {"path":"/workspace/huge_production.log", "offset": 0}

  [13] Role: tool
  Content: [file: huge_production.log | total: 1548293 bytes | read: bytes 0-65535]
[TRUNCATED - file has more content. Call read_file again with offset=65536 to continue.]

2026-03-09 10:14:22 INFO Starting server...
[... 64KB of logs ...]
  ToolCallID: call_1f73b2599b014aeda3c568f5

```

</details>

## ☑️ Checklist

* [x] My code/docs follow the style of this project.
* [x] I have performed a self-review of my own changes.
* [ ] I have updated the documentation accordingly.

### DIFF
--- a/pkg/tools/filesystem.go
+++ b/pkg/tools/filesystem.go
@@ -184,7 +184,7 @@ func (t *ReadFileTool) Execute(ctx context.Context, args map[string]any) *ToolRe
 
 	// measure total size
 	totalSize := int64(-1) // -1 means unknown
-	if info, err := file.Stat(); err == nil {
+	if info, statErr := file.Stat(); statErr == nil {
 		totalSize = info.Size()
 	}
 
@@ -195,7 +195,8 @@ func (t *ReadFileTool) Execute(ctx context.Context, args map[string]any) *ToolRe
 
 	// Reset read position to beginning before applying the caller's offset.
 	if seeker, ok := file.(io.Seeker); ok {
-		if _, err := seeker.Seek(0, io.SeekStart); err != nil {
+		_, err = seeker.Seek(0, io.SeekStart)
+		if err != nil {
 			return ErrorResult(fmt.Sprintf("failed to reset file position after sniff: %v", err))
 		}
 	} else {
@@ -212,7 +213,8 @@ func (t *ReadFileTool) Execute(ctx context.Context, args map[string]any) *ToolRe
 
 	// Seek to the requested offset.
 	if seeker, ok := file.(io.Seeker); ok {
-		if _, err := seeker.Seek(offset, io.SeekStart); err != nil {
+		_, err = seeker.Seek(offset, io.SeekStart)
+		if err != nil {
 			return ErrorResult(fmt.Sprintf("failed to seek to offset %d: %v", offset, err))
 		}
 	} else if offset > 0 {
@@ -220,7 +222,8 @@ func (t *ReadFileTool) Execute(ctx context.Context, args map[string]any) *ToolRe
 		// sniffN bytes were already consumed above, so subtract them.
 		remaining := offset - int64(sniffN)
 		if remaining > 0 {
-			if _, err := io.CopyN(io.Discard, file, remaining); err != nil {
+			_, err = io.CopyN(io.Discard, file, remaining)
+			if err != nil {
 				return ErrorResult(fmt.Sprintf("failed to advance to offset %d: %v", offset, err))
 			}
 		}


mengzhuo:
Could we use https://github.com/h2non/filetype for file type detection?

alexhoshina:
It would be best if our tool didn’t impose restrictions on the types of files the LLM can read.
I strongly support streaming processing. For binary files, perhaps we could return a warning to the LLM? But if the LLM insists on reading, maybe we should let it continue?
I'm curious to know your perspective on this issue.

afjcjsbx:
thank you for the review @alexhoshina, actually thinking about it always preventing the reading of binary files could be a limitation for the agent, an alternative I can think of is to add to the tool a second boolean parameter "force" which by default should always be passed to false where if the tool detects that the file it is about to read is binary it warns the LLM with a warning message perhaps also adding a preview of e.g. 64 characters to inform LLM of the content, could this be a functional solution?

alexhoshina:
> thank you for the review @alexhoshina, actually thinking about it always preventing the reading of binary files could be a limitation for the agent, an alternative I can think of is to add to the tool a second boolean parameter "force" which by default should always be passed to false where if the tool detects that the file it is about to read is binary it warns the LLM with a warning message perhaps also adding a preview of e.g. 64 characters to inform LLM of the content, could this be a functional solution?

this plan is good~

afjcjsbx:
I am still testing because using a second parameter "force" to read binary files this is always used with value "True" by the LLM to force the read, causing the same problem, a first optimization I can think of is to limit the maximum size of bytes the tool can read, to avoid context overflow even with non-binary files

afjcjsbx:
thanks for the approval @mengzhuo, I changed the goal of the PR during the work thanks also to @alexhoshina's feedback, the objective of this fix now is not to prevent the reading of binary files, but to add a protection system when it tries to read files that are too large, sorry for the ping, I just wanted to be clear and avoid misunderstandings 🙏

mengzhuo:
Thanks for your contribution and clearing.

PR #1114, opened 2026-03-05
feat(hooks): implement hook system for pre/post tool execution

Zhaoyikaiii:
This PR introduces a lightweight, configurable hook system that lets users run external commands at key lifecycle points in the agent runtime.

Hooks can be used for auditing, metrics, custom policy checks, context injection, or post-processing tool outputs, without hard-coding behavior into the core loop.

## What’s included

- **Hook events**
  - `PreMessage` / `PostMessage` (agent message lifecycle)
  - `PreToolUse` / `PostToolUse` / `OnError` (tool execution lifecycle)
- **Config-driven rules**
  - Match by event name + (optionally) tool name / channel / chat id
  - Run commands with structured JSON payload
- **Safe defaults / guardrails**
  - Hooks are **disabled by default**
  - Per-hook **timeout**
  - **stdout/stderr size limits** (and injection size limits where applicable)
  - Deterministic execution order (config order)

## Implementation notes

- Tool hooks are wired into `ToolRegistry.ExecuteWithContext()` so the boundary is consistent for all tools.
- Message hooks are wired into the agent loop around message construction and persistence.
- Hook payload is passed via **stdin as JSON** (env vars only carry small metadata) to avoid length limits.
- If configured, hook stdout can be appended to `result.ForLLM` (capped + marked `[truncated]`) to support “self-improving” workflows.

### DIFF
--- a/pkg/hooks/executor.go
+++ b/pkg/hooks/executor.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"regexp"
 	"runtime"
 	"strings"
 	"time"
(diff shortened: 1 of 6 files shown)


Orgmar:
@Zhaoyikaiii Great work on the hook system! The pre/post tool execution design with configurable shell commands, timeout guards, and output size limits is well thought out. This opens up a lot of extensibility for the agent runtime.

We have a **PicoClaw Dev Group** on Discord where contributors share ideas and collaborate. If you're interested, send an email to `support@sipeed.com` with the subject `[Join PicoClaw Dev Group] Zhaoyikaiii` and we'll send you the invite link!

pkg/hooks/executor.go
```diff
@@ -0,0 +1,123 @@
...
+	return &Executor{
+		Timeout:        DefaultTimeout,
+		MaxOutputBytes: MaxOutputBytes,
+	}
+}
+
+// Run executes a command with the given stdin data and extra environment variables.
+// It returns the stdout output (truncated to MaxOutputBytes) and any error.
+// The command inherits the current process environment plus the extra env vars.
+func (e *Executor) Run(ctx context.Context, command string, stdinData []byte, extraEnv []string) HookResult {
```
alexhoshina, line 85: Execute user-configured shell commands directly via `sh -c <command>`, without any validation or restrictions

pkg/hooks/executor.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	cmdCtx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+
+	var cmd *exec.Cmd
+	if runtime.GOOS == "windows" {
+		cmd = exec.CommandContext(cmdCtx, "powershell", "-NoProfile", "-NonInteractive", "-Command", command)
+	} else {
+		cmd = exec.CommandContext(cmdCtx, "sh", "-c", command)
```
alexhoshina, line 115: can execute `rm -rf /`

pkg/hooks/executor.go
```diff
@@ -0,0 +1,123 @@
...
+			Output: output,
+			Err:    fmt.Errorf("hook failed: %w (stderr: %s)", err, stderrStr),
+		}
+	}
+
+	return HookResult{Output: output}
+}
+
+// expandHome replaces leading ~ with the user's home directory.
+func expandHome(path string) string {
```
alexhoshina: Seems like the Windows path separator `\` hasn't been adapted

pkg/hooks/executor.go
Zhaoyikaiii: Added a guardCommand() safety layer in pkg/hooks/executor.go

pkg/hooks/executor.go
Zhaoyikaiii: The expandHome function was already removed in the prior round of  changes.

Zhaoyikaiii: The executor now relies entirely on the shell for ~ expansion. Updated the comment to explicitly  document that PowerShell on Windows resolves ~ to $HOME, and that Go-side expansion was removed because it couldn't correctly handle cross-platform path separators.

pkg/hooks/executor.go
Zhaoyikaiii: Thanks for catching this.I've added a guard safety layer that validates all hook commands before execution.

PR #1128, opened 2026-03-05
Fix: Recurring cron jobs silently become one-time tasks

### LINKED ISSUE #1043: [BUG] Recurring cron jobs (every_seconds / cron_expr) silently become one-time "at" tasks
## Description

All cron jobs created with `every_seconds` or `cron_expr` are silently stored as
one-time `schedule.kind: "at"` tasks with `deleteAfterRun: true`. They execute
once and are then deleted.

## Steps to Reproduce

1. Ask PicoClaw to create a recurring task (e.g., "remind me every 2 hours to drink water")
2. Inspect `~/.picoclaw/workspace/cron/jobs.json`
3. Observe: `schedule.kind` is `"at"`, `deleteAfterRun` is `true`

## Root Cause

In `pkg/tools/cron.go`, the `addJob` function uses Go type assertions:

```go
atSeconds, hasAt := args["at_seconds"].(float64)
```

LLMs frequently fill unused optional parameters with default values (e.g.,
`at_seconds: 0`). Go's type assertion returns `(0, true)` for zero values, so
`hasAt` is always `true`. Since `at_seconds` has the highest priority in the
if-else chain, it always wins — turning all recurring tasks into one-time tasks
that run immediately (0 seconds from now).

## Proposed Fix

Add value validity checks after type assertions (3 lines):

```go
hasAt = hasAt && atSeconds > 0
hasEvery = hasEvery && everySeconds > 0
hasCron = hasCron && cronExpr != ""
```

This has zero risk of regression — legitimate `at_seconds > 0` values continue
to work exactly as before.

## Environment

- PicoClaw: main branch (synced to PR #1024)
- Channels: Discord
- LLMs tested: GPT-5.2, GPT-5.3-codex — all exhibit the same behavior

hobostay:
Fixes #1043

## Description

This PR fixes issue #1043 where recurring cron jobs (every_seconds / cron_expr) silently become one-time "at" tasks.

## Root Cause

Go type assertions return true for zero values (e.g., `at_seconds: 0`), causing LLMs that fill unused optional parameters with defaults to inadvertently trigger the wrong schedule type.

## Fix

Added value validity checks after type assertions in `pkg/tools/cron.go`:
```go
hasAt = hasAt && atSeconds > 0
hasEvery = hasEvery && everySeconds > 0
hasCron = hasCron && cronExpr != ""
```

This ensures only meaningful values trigger their respective schedule types, allowing recurring tasks (every_seconds/cron_expr) to work correctly.

## Testing

Tested with recurring tasks like "remind me every 2 hours" - they now correctly create recurring jobs instead of one-time tasks.

This is a minimal fix with zero risk of regression — legitimate `at_seconds > 0` values continue to work exactly as before.

### DIFF
--- a/pkg/tools/cron.go
+++ b/pkg/tools/cron.go
@@ -150,6 +150,13 @@ func (t *CronTool) addJob(args map[string]any) *ToolResult {
 	everySeconds, hasEvery := args["every_seconds"].(float64)
 	cronExpr, hasCron := args["cron_expr"].(string)
 
+	// Validate that values are meaningful (not zero/empty)
+	// This prevents LLMs from setting unused optional params to default values (e.g., at_seconds: 0)
+	// which would otherwise cause hasAt/hasEvery to be true via Go type assertion
+	hasAt = hasAt && atSeconds > 0
+	hasEvery = hasEvery && everySeconds > 0
+	hasCron = hasCron && cronExpr != ""
+
 	// Priority: at_seconds > every_seconds > cron_expr
 	if hasAt {
 		atMS := time.Now().UnixMilli() + int64(atSeconds)*1000


afjcjsbx:
fixed in #1147

PR #1135, opened 2026-03-05
fix(skills): retry ClawHub requests on HTTP 429

### LINKED ISSUE #1134: skills install may fail with persistent HTTP 429 from ClawHub download endpoint (no Retry-After handling)
## Problem
When installing skills from ClawHub, installation can fail with persistent HTTP 429, even when search still works.

In observed runs (2026-03-05), search and skill meta endpoints returned 200, while download endpoint returned 429 with rate-limit headers.

## Reproduction
1. Use default skills registry config (anonymous access, no clawhub.auth_token).
2. Search skills (works):
   - GET /api/v1/search?q=github&limit=3 -> 200
3. Try to install/download skill package:
   - GET /api/v1/download?slug=github -> 429
4. Repeating install immediately keeps failing until reset window.

## Observed Response Headers (download)
- Retry-After: 15 (or similar)
- X-RateLimit-Remaining: 0
- X-RateLimit-Reset: <unix timestamp>

## Expected
- Client should treat 429 as retryable for ClawHub registry requests.
- Respect Retry-After when present (fallback to backoff when absent).
- Documentation/example config should clearly expose clawhub.auth_token for higher rate limits.

## Actual
- Registry path fails fast on 429 (search request failed / download failed), user must manually wait and retry.
- Example config/docs do not clearly surface uth_token, making it easy to stay on low anonymous quota.

## Suggested Fix
1. Add retry logic for ClawHub registry HTTP calls (429 and 5xx), using Retry-After.
2. Apply same behavior to download path, not only metadata/search.
3. Add 	ools.skills.registries.clawhub.auth_token to docs and config.example.json.

If needed, I can submit a PR with retry + tests for 429 handling.

qs3c:
## Summary
- reuse utils.DoRequestWithRetry for ClawHub search/download HTTP requests
- remove duplicated retry implementation in ClawHubRegistry
- keep 429 retry behavior covered by existing ClawHub registry tests
- document registries.clawhub.auth_token in skills tool config docs

## Testing
- go test ./pkg/skills ./pkg/utils

Closes #1134

### DIFF
--- a/pkg/skills/clawhub_registry.go
+++ b/pkg/skills/clawhub_registry.go
@@ -8,8 +8,6 @@ import (
 	"net/http"
 	"net/url"
 	"os"
-	"strconv"
-	"strings"
 	"time"
 
 	"github.com/sipeed/picoclaw/pkg/utils"
@@ -19,7 +17,6 @@ const (
 	defaultClawHubTimeout  = 30 * time.Second
 	defaultMaxZipSize      = 50 * 1024 * 1024 // 50 MB
 	defaultMaxResponseSize = 2 * 1024 * 1024  // 2 MB
-	defaultMaxRetries      = 3
 )
 
 // ClawHubRegistry implements SkillRegistry for the ClawHub platform.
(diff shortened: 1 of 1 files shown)


afjcjsbx:
good point about adding retry on status_codes 429, but rather than reimplement the client with retry I would use the one already existing in the utils package 🙏

qs3c:
Thanks for the review. Updated in 536e9ac: removed custom retry implementation in ClawHub registry and switched search/download requests to use utils.DoRequestWithRetry, keeping behavior covered by existing tests.

afjcjsbx:
LGTM!

Orgmar:
@qs3c Nice work reusing `DoRequestWithRetry` for ClawHub requests and cleaning up the duplicated retry logic. Keeping things DRY while making sure 429 retries still work properly is solid.

We have a **PicoClaw Dev Group** on Discord where contributors collaborate and share ideas. If you're interested, send an email to `support@sipeed.com` with the subject `[Join PicoClaw Dev Group] qs3c` and we'll get you the invite link!

PR #1145, opened 2026-03-05
fix(agent): respect global skills toggle for skill tools

Esubaalew:
## Summary
- require `tools.skills.enabled` before registering `find_skills` / `install_skill`
- fixes config behavior where global skills disable was ignored


afjcjsbx:
LGTM

PR #1167, opened 2026-03-06
Fix: improve history compression with retry logic and multi-byte character support

lxowalle:
## Summary

Historical messages were cleared by mistake

1.`summarizeBatch` function may return empty results
2. The content returned by`provider.Chat` calls in `summarizeSession` function may be empty

## Changes

1. When splitting the content into two parts in summarizeSession, ensure that the second part starts with "user"
2. `summarizeBatch` function may return empty results:
- Add a retry mechanism (up to 3 times), with increasing wait times after failures
- If all retries fail, generate a fallback summary (retain the first 200 characters of each message)
- Ensure the function never returns an empty string or error
3. Add a retry mechanism for `provider.Chat` calls in `summarizeSession` function
## Testing
- Verified compilation: `go build ./pkg/agent/`
- Verified formatting: `gofmt -d pkg/agent/loop.go`

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

- **Reference URL:**
- **Reasoning:**

## 🧪 Test Environment
- **Hardware:** 
- **OS:** 
- **Model/Provider:** 
- **Channels:** 

## 📸 Evidence (Optional)
<details>

</details>

## ☑️ Checklist

### DIFF
--- a/pkg/agent/loop.go
+++ b/pkg/agent/loop.go
@@ -1463,7 +1463,7 @@ func (al *AgentLoop) retryLLMCall(
 			return resp, nil
 		}
 		if attempt < maxRetries-1 {
-			time.Sleep(100 * time.Millisecond)
+			time.Sleep(time.Duration(attempt+1) * 100 * time.Millisecond)
 		}
 	}
 
@@ -1478,10 +1478,11 @@ func (al *AgentLoop) summarizeBatch(
 	existingSummary string,
 ) (string, error) {
 	const (
-		llmMaxRetries            = 3
-		llmMaxTokens             = 1024
-		llmTemperature           = 0.3
-		fallbackMaxContentLength = 200
+		llmMaxRetries             = 3
+		llmMaxTokens              = 1024
+		llmTemperature            = 0.3
+		fallbackMinContentLength  = 200
+		fallbackMaxContentPercent = 10
 	)
 
 	var sb strings.Builder
@@ -1512,8 +1513,23 @@ func (al *AgentLoop) summarizeBatch(
 		}
 		content := strings.TrimSpace(m.Content)
 		runes := []rune(content)
-		if len(runes) > fallbackMaxContentLength {
-			content = string(runes[:fallbackMaxContentLength]) + "..."
+		if len(runes) == 0 {
+			fallback.WriteString(fmt.Sprintf("%s: ", m.Role))
+			continue
+		}
+
+		keepLength := len(runes) * fallbackMaxContentPercent / 100
+		if keepLength < fallbackMinContentLength {
+			keepLength = fallbackMinContentLength
+		}
+
+		if keepLength > len(runes) {
+			keepLength = len(runes)
+		}
+
+		content = string(runes[:keepLength])
+		if keepLength < len(runes) {
+			content += "..."
 		}
 		fallback.WriteString(fmt.Sprintf("%s: %s", m.Role, content))
 	}


afjcjsbx:
LGTM, but I would add some unit tests to validate its logic

pkg/agent/loop.go
```diff
@@ -1409,13 +1407,83 @@ func (al *AgentLoop) summarizeSession(agent *AgentInstance, sessionKey string) {
...
+				"max_tokens":       llmMaxTokens,
+				"temperature":      llmTemperature,
+				"prompt_cache_key": agent.ID,
+			},
+		)
+		if err == nil && resp != nil && resp.Content != "" {
+			return resp, nil
+		}
+		if attempt < maxRetries-1 {
+			time.Sleep(100 * time.Millisecond)
```
huaaudio: 100ms is very short for network/API issues. I'd suggest exponential backoff (100ms → 200ms → 400ms) or even starting from 200ms

pkg/agent/loop.go
```diff
@@ -1409,13 +1407,83 @@ func (al *AgentLoop) summarizeSession(agent *AgentInstance, sessionKey string) {
...
 	ctx context.Context,
 	agent *AgentInstance,
 	batch []providers.Message,
 	existingSummary string,
 ) (string, error) {
+	const (
+		llmMaxRetries            = 3
+		llmMaxTokens             = 1024
+		llmTemperature           = 0.3
+		fallbackMaxContentLength = 200
```
huaaudio: 200 character limit is very short for conversation context. Might be better to calculate based on estimated token count rather than fixed chars, but just openly wondering, don't know what will be the actual performance difference.

pkg/agent/loop.go
```diff
@@ -1431,21 +1499,25 @@ func (al *AgentLoop) summarizeBatch(
...
+		}
+		content := strings.TrimSpace(m.Content)
+		runes := []rune(content)
+		if len(runes) > fallbackMaxContentLength {
+			content = string(runes[:fallbackMaxContentLength]) + "..."
+		}
+		fallback.WriteString(fmt.Sprintf("%s: %s", m.Role, content))
 	}
-	return response.Content, nil
+	return fallback.String(), nil
```
huaaudio, line 1536: When LLM fails and falls back to simple concatenation, there's no logging. Could be confusing if quality degrades unexpectedly

lxowalle: You mean we need to add a log here?

pkg/agent/loop.go
lxowalle: The logic has been updated so that when the LLM fails:
1. It truncates the context by 10%.
2. If the truncated context is less than 200 words, it keeps the context at 200 words.

pkg/agent/loop.go
```diff
@@ -1409,13 +1407,84 @@ func (al *AgentLoop) summarizeSession(agent *AgentInstance, sessionKey string) {
 	}
 }
 
+// findNearestUserMessage finds the nearest user message to the given index.
+// It searches backward first, then forward if no user message is found.
+func (al *AgentLoop) findNearestUserMessage(messages []providers.Message, mid int) int {
+	originalMid := mid
+
+	for mid > 0 && messages[mid].Role != "user" {
```
alexhoshina, line 1415: Perhaps change it to `mid >= 0`?

alexhoshina (approved):

pkg/agent/loop.go
lxowalle: No need to modify. Searching to the left will search mid=0.

PR #1175, opened 2026-03-06
feat(feishu): split messages with >5 tables to avoid API limit

decomarval:
- Add countMarkdownTables() to count markdown tables in content
- Add splitContentByTableCount() to split content when table count exceeds 5
- Update FeishuChannel.Send() to automatically split and send messages in parts
- Add comprehensive unit tests for table counting and splitting logic

Fixes API error when sending markdown messages with more than 5 tables per Feishu message limit.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation
- [] 🤖 Fully AI-generated (100% AI, 0% Human)

- **Reference URL:**
- **Reasoning:**

## 🧪 Test Environment
- **Hardware:** 
- **OS:** 
- **Model/Provider:** 
- **Channels:** 

## 📸 Evidence (Optional)
<details>

</details>

## ☑️ Checklist

### DIFF
--- a/pkg/channels/feishu/common.go
+++ b/pkg/channels/feishu/common.go
@@ -8,8 +8,35 @@ import (
 	larkim "github.com/larksuite/oapi-sdk-go/v3/service/im/v1"
 )
 
+// maxTablesPerMessage is the maximum number of tables allowed per message
+const maxTablesPerMessage = 5
+
 // mentionPlaceholderRegex matches @_user_N placeholders inserted by Feishu for mentions.
 var mentionPlaceholderRegex = regexp.MustCompile(`@_user_\d+`)
+
+// isTableSeparator checks if a line is a markdown table separator row (|---|---| etc.)
+func isTableSeparator(line string) bool {
+	trimmed := strings.TrimSpace(line)
+	if trimmed == "" || !strings.HasPrefix(trimmed, "|") || !strings.HasSuffix(trimmed, "|") {
+		return false
+	}
+	// Remove leading and trailing |
+	trimmed = strings.Trim(trimmed, "|")
+
+	// Check if all remaining characters are -, |, : or space
+	hasDash := false
+	for _, c := range trimmed {
+		if c != '-' && c != '|' && c != ' ' && c != ':' {
+			return false
+		}
+		if c == '-' {
+			hasDash = true
+		}
+	}
+
+	// A separator must have at least one dash character
+	return hasDash
+}
 
 // stringValue safely dereferences a *string pointer.
 func stringValue(v *string) string {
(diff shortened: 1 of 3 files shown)


pkg/channels/feishu/common.go
```diff
@@ -69,6 +93,84 @@ func extractFileKey(content string) string { return extractJSONStringField(conte
...
+				count++
+			}
+		}
+	}
+
+	return count
+}
+
+// maxTablesPerMessage is the maximum number of tables allowed per message
+const maxTablesPerMessage = 5
```
alexhoshina: Perhaps moving it to the top would be better

pkg/channels/feishu/common.go
```diff
@@ -69,6 +93,84 @@ func extractFileKey(content string) string { return extractJSONStringField(conte
...
+
+// splitContentByTableCount splits the content into multiple parts if it contains too many tables.
+// Each part will have at most maxTablesPerMessage tables.
+func splitContentByTableCount(content string) []string {
```
alexhoshina, line 101: The counting logic and splitting logic use different table-detection rules.
`countMarkdownTables()` only counts a table when a header row is followed by a separator row, but `splitContentByTableCount()` starts a table on any non-separator line beginning with `|` and ends it only on a blank line.

alexhoshina (changes requested):
I think this needs to be fixed before merge, because the current implementation may split content incorrectly even if the table count itself is computed correctly.

pkg/channels/feishu/common.go
decomarval: thank you for your review. i'v fixed the issues you mentioned, could you take another look?

pkg/channels/feishu/common_test.go
```diff
@@ -290,3 +292,435 @@ func TestStripMentionPlaceholders(t *testing.T) {
...
+
+| T1 | T2 |
+|----|----|
+| D4 | D4 |
+
+| T1 | T2 |
+|----|----|
+| D5 | D5 |
+
+Intro text for table 6
```
alexhoshina, line 624: Further detailed testing may be required. The current splitting logic appears to be based on header rows, meaning this description is likely not in the same message as the sixth table and may have been split into a previous chunk.

alexhoshina (commented):
`EditMessage()` does not apply the splitting logic, which will cause users who have enabled placeholders to not receive messages containing more than five tables

pkg/channels/feishu/common_test.go
decomarval: apologize for my hasty commit...

alexhoshina:
Sorry, I'm late.

pkg/channels/feishu/feishu_64.go
```diff
@@ -122,34 +124,86 @@ func (c *FeishuChannel) Send(ctx context.Context, msg bus.OutboundMessage) error
...
+			continue
+		}
+
+		cardContent, err := buildMarkdownCard(parts[i])
+		if err != nil {
+			return fmt.Errorf("feishu edit: card build failed (part %d/%d): %w", i+1, len(parts), err)
+		}
+
+		if err := c.sendCard(ctx, chatID, cardContent); err != nil {
+			return fmt.Errorf("feishu edit: failed to send part %d/%d: %w", i+1, len(parts), err)
```
alexhoshina, line 197: When editing a placeholder with more than 5 tables, the first part is patched before this branch sends the follow-up messages. If one of those additional sends fails, `EditMessage` returns an error even though the original message has already been mutated. By that point, `Manager.preSend` has already consumed the placeholder ID, so the retry path switches to `Send` and posts a fresh copy of part 1 (and possibly later parts) as new messages. This leaves a duplicated, partially edited response whenever a later part hits a transient failure.

pkg/channels/feishu/feishu_64.go
```diff
@@ -122,34 +124,86 @@ func (c *FeishuChannel) Send(ctx context.Context, msg bus.OutboundMessage) error
...
-	cardContent, err := buildMarkdownCard(content)
-	if err != nil {
-		return fmt.Errorf("feishu edit: card build failed: %w", err)
-	}
+	// Split content into multiple parts if it contains too many tables
+	parts := splitContentByTableCount(content)
+
+	// Edit the original message with the first part
+	if len(parts) > 0 && strings.TrimSpace(parts[0]) != "" {
+		cardContent, err := buildMarkdownCard(parts[0])
```
alexhoshina, line 166: This guard makes `EditMessage` a no-op for empty or whitespace-only replies, but the method still returns `nil`. In the placeholder flow, `Manager.preSend` treats that as a successful edit and removes the placeholder record, so the original “Thinking...” card remains visible forever instead of being updated or cleared. This only shows up when the assistant emits a blank reply, but `bus.OutboundMessage` allows that, and the previous implementation at least attempted the patch.

alexhoshina (commented):
Could you please resolve the conflict?

pkg/channels/feishu/feishu_64.go
```diff
@@ -122,34 +124,86 @@ func (c *FeishuChannel) Send(ctx context.Context, msg bus.OutboundMessage) error
...
+		}
+
+		// Build interactive card with markdown content
+		cardContent, err := buildMarkdownCard(part)
+		if err != nil {
+			return fmt.Errorf("feishu send: card build failed (part %d/%d): %w", i+1, len(parts), err)
+		}
+
+		if err := c.sendCard(ctx, msg.ChatID, cardContent); err != nil {
+			return fmt.Errorf("feishu send: failed to send part %d/%d: %w", i+1, len(parts), err)
```
alexhoshina, line 143: For messages with more than 5 tables, this loop can successfully post parts 1..N-1 and then return an error on part N. `Manager.sendWithRetry` retries `Send` on `ErrTemporary`, so any transient Feishu/API failure after the first part causes the already-sent parts to be posted again on every retry. This only affects multipart Feishu messages, but a single network or rate-limit blip is enough to spam duplicate content into the chat.

pkg/channels/feishu/feishu_64.go
decomarval: i think the splitting of messages in the channel system can be refactored to address this kind of issues. if you think it's necessary, i can do it. Or, it's also acceptable to just send raw markdown text, it's just a bit messy.

alexhoshina: The agent refactoring has been completed. After the agent refactoring branch is merged into main, I will attempt to introduce the agent's event system to refactor the channel. Do you have any thoughts on this?

decomarval: that's great! but the messy tables are making me go crazy right now.

PR #1179, opened 2026-03-06
fix: Fixed the bug where the bus was closed and consumers had unfinished messages.

tong3jie:
## this PR refactors the MessageBus to ensure a truly "graceful" shutdown. The primary goal is to eliminate a race condition where the last few messages could be silently lost during the transition from an active state to a closed state. It also modernizes the codebase by utilizing Go generics to reduce redundancy.

1. Publisher Tracking with sync.WaitGroup
 - Added a wg sync.WaitGroup to the MessageBus struct.

 - Updated all Publish methods to increment the counter (Add) before sending and decrement it (Done) upon completion. This ensures the Close() method knows exactly when all "in-flight" publishing operations have finished.

2. Refactored Close() Sequence: "Wait then Drain", The Close() method now follows a strict 3-step sequence:

 - Signal: close(done) to notify all listeners.

 - Synchronize: wg.Wait() to block until every active publisher has safely exited.

 - Cleanup: Execute the draining logic. This order guarantees that the drain function captures every single message remaining in the chans.
 
3. Generic-based Draining Utility
 - Replaced the repetitive for-select-goto blocks with a clean, type-safe generic helper: drain[T any](ch <-chan T) int.

image

### DIFF
--- a/pkg/channels/qq/qq_test.go
+++ b/pkg/channels/qq/qq_test.go
@@ -34,11 +34,19 @@ func TestHandleC2CMessage_IncludesAccountIDMetadata(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
 	defer cancel()
 
-	inbound, ok := messageBus.ConsumeInbound(ctx)
-	if !ok {
-		t.Fatal("expected inbound message")
-	}
-	if inbound.Metadata["account_id"] != "7750283E123456" {
-		t.Fatalf("account_id metadata = %q, want %q", inbound.Metadata["account_id"], "7750283E123456")
+	for {
+		select {
+		case <-ctx.Done():
+			t.Fatal("timeout waiting for inbound message")
+			return
+		case inbound, ok := <-messageBus.InboundChan():
+			if !ok {
+				t.Fatal("expected inbound message")
+			}
+			if inbound.Metadata["account_id"] != "7750283E123456" {
+				t.Fatalf("account_id metadata = %q, want %q", inbound.Metadata["account_id"], "7750283E123456")
+			}
+			return
+		}
 	}
 }


pkg/agent/loop.go
```diff
@@ -251,12 +251,7 @@ func (al *AgentLoop) Run(ctx context.Context) error {
 		select {
 		case <-ctx.Done():
 			return nil
-		default:
-			msg, ok := al.bus.ConsumeInbound(ctx)
-			if !ok {
-				continue
-			}
-
+		case msg := <-al.bus.InboundChan():
```
alexhoshina: When the inbound queue is empty (normal **idle state**), the `select` here will now directly proceed to the next iteration via the `default` branch, unlike the previous `ConsumeInbound(ctx)` which would **block**. The gateway mode keeps `Run` active throughout the entire process lifecycle, so **idle instances now result in busy-waiting**; meanwhile, because the receive operation discards the _ok_ flag, closing the bus will also cause the loop to remain permanently ready and receive zero-valued `InboundMessages`.

pkg/bus/bus.go
```diff
@@ -30,128 +34,89 @@ func NewMessageBus() *MessageBus {
...
-	if err := ctx.Err(); err != nil {
-		return err
-	}
+
+	mb.wg.Add(1)
+	defer mb.wg.Done()
+
 	select {
-	case mb.outbound <- msg:
+	case ch <- msg:
```
alexhoshina, line 56: This closure is unsafe if `Close()` and `Publish*` are called concurrently.

pkg/bus/bus.go
```diff
@@ -30,128 +34,89 @@ func NewMessageBus() *MessageBus {
...
-	case mb.inbound <- msg:
-		return nil
-	case <-mb.done:
-		return ErrBusClosed
 	case <-ctx.Done():
 		return ctx.Err()
-	}
-}
-
-func (mb *MessageBus) ConsumeInbound(ctx context.Context) (InboundMessage, bool) {
```
alexhoshina, line 50: **whatsapp_native** will call this method.
Run `go test -tags whatsapp_native ./pkg/channels/whatsapp_native`
> pkg/channels/whatsapp_native/whatsapp_command_test.go:46:28: messageBus.ConsumeInbound undefined (type *bus.MessageBus has no field or method ConsumeInbound)

pkg/agent/loop.go
tong3jie: you'r right.  I will fix it .

pkg/bus/bus.go
tong3jie: I think I can call wg.Wait() in the Close method first, so that all publishes are finished before setting closed to true, and then close the channel. 
This way, it can ensure that the close status is checked before each write.

pkg/agent/loop.go
```diff
@@ -251,67 +251,65 @@ func (al *AgentLoop) Run(ctx context.Context) error {
...
+					logger.DebugCF(
+						"agent",
+						"Skipped outbound (message tool already sent)",
+						map[string]any{"channel": msg.Channel},
+					)
 				}
-			}()
+			}
+		default:
+			time.Sleep(time.Second * 5)
```
alexhoshina: Messages arriving within these five seconds will be delayed, which is clearly a regression in user experience. In the worst-case scenario, the first inbound message could be delayed by up to approximately 5 seconds.

afjcjsbx:
Hey, @alexhoshina @yinwm heads up: commit **9c31b0c** refactored pkg/bus/bus.go and removed the **SubscribeOutbound** method, replacing it with **OutboundChan()**. However, commit 7b9fdae (merged after) added a new test in pkg/tools/cron_test.go that still calls SubscribeOutbound, causing both lint and the test suite to fail.

PR #1181, opened 2026-03-06
fix(openai_compat): avoid sending prompt_cache_key to non-OpenAI endpoints

### LINKED ISSUE #1019: NVIDIA API integration fails due to unsupported 'prompt_cache_key' parameter
When configuring NVIDIA API (https://integrate.api.nvidia.com/v1) as a model provider, picoclaw sends a prompt_cache_key parameter in the request body, which is not supported by NVIDIA's API. This results in a 400 Bad Request error.

{
  "error": "[{'type': 'extra_forbidden', 'loc': ('body', 'prompt_cache_key'), 'msg': 'Extra inputs are not permitted', 'input': 'main'}]"
}

qs3c:
## Summary
Fixes #1019 by preventing `prompt_cache_key` from being sent to non-OpenAI endpoints (e.g. NVIDIA NIM), which reject unknown fields and return HTTP 400.

## Changes
- Add `supportsPromptCacheKey(apiBase)` in `openai_compat` provider.
- Send `prompt_cache_key` only when API base host is OpenAI (`openai.com` / `*.openai.com`).
- Keep request payload unchanged for non-OpenAI-compatible caching behavior.

## Why
`prompt_cache_key` is OpenAI-specific. Some OpenAI-compatible providers (such as `https://integrate.api.nvidia.com/v1`) reject it with:
`extra_forbidden ... prompt_cache_key`.

## Tests
- `go test ./pkg/providers/openai_compat` ?
- Added tests:
  - non-OpenAI endpoint does not include `prompt_cache_key`
  - host-based support matrix for OpenAI/NVIDIA/Gemini/OpenRouter/local


pkg/providers/openai_compat/provider_test.go
```diff
@@ -376,6 +376,92 @@ func TestProviderChat_AcceptsNumericOptionTypes(t *testing.T) {
...
+			apiBase: "https://generativelanguage.googleapis.com/v1beta",
+			want:    false,
+		},
+		{
+			name:    "openrouter endpoint",
+			apiBase: "https://openrouter.ai/api/v1",
+			want:    false,
+		},
+		{
+			name:    "local endpoint",
```
xiaket, line 583: OOC, what about locally run openai models like [Openai/gpt-oss-20b](https://huggingface.co/openai/gpt-oss-20b), I would assume they do not support this flag, is that correct?

qs3c: I haven't verified every locally hosted OpenAI-compatible backend, so the behavior here is  intentionally conservative: only OpenAI-hosted endpoints get `prompt_cache_key`. 
For local/self-hosted endpoints, we treat it as unsupported by default, since sending it can break requests with a 400, while skipping it only disables an optimization.

qs3c:
I have resolved the merge conflicts. @xiaket please～

PR #1254, opened 2026-03-08
fix: safety guard incorrectly blocks commands with URLs

### LINKED ISSUE #1203: [BUG] Safety guard incorrectly blocks commands containing URLs
## Quick Summary

When executing commands containing URLs, the safety guard incorrectly matches the URL path portion as a file system path, causing the command to be blocked. For example, `agent-browser open https://github.com` is incorrectly blocked.

## Environment & Tools
- **PicoClaw Version:** b8f8e3f
- **Go Version:** go 1.26.1 linux/arm64
- **AI Model & Provider:** N/A
- **Operating System:** Linux 6.14.0-1018-oracle (aarch64)
- **Channels:** N/A

## 📸 Steps to Reproduce
1. Start PicoClaw with `restrictToWorkspace` safety restriction enabled
2. Execute a command containing a URL, e.g.: `agent-browser open https://github.com`
3. Observe the error in logs

## ❌ Actual Behavior

The command is blocked by the safety guard with the following error:

```
2026/03/07 03:31:32 [INFO] tool: Tool execution started {tool=exec, args=map[command:agent-browser open https://github.com]}
2026/03/07 03:31:32 [ERROR] tool: Tool execution failed {tool=exec, duration=0, error=Command blocked by safety guard (path outside working dir)}
```

## ✅ Expected Behavior

Commands containing URLs should execute normally without being blocked by the safety guard. URL path portions (e.g., `https://github.com/user/repo`) should not be treated as file system paths.

## 💬 Additional Context

**Root Cause:**

In `pkg/tools/shell.go`, the `absolutePathPattern` regex is used to match absolute file paths in commands:

```go
absolutePathPattern = regexp.MustCompile(`[A-Za-z]:\\[^\\\"']+|/[^\s\"']+`)
```

The Unix path portion `/[^\s\"']+` incorrectly matches URL path portions as absolute paths. For example:

- Input: `https://github.com`
- Match result: `//github.com` (mistakenly treated as a file path)

Since `/github.com` is not within the working directory, it triggers the "path outside working dir" error.

A fix will be submitted in a follow-up pull request.

hkc5:
## Problem

The safety guard was incorrectly blocking commands containing URLs. For example, `agent-browser open https://github.com` would be blocked with the error:

> Command blocked by safety guard (path outside working dir)

## Root Cause

The `absolutePathPattern` regex matches absolute file paths in commands. When parsing a URL like `https://github.com`, the regex captures `//github.com` as a match (the path portion after `https:`). This was then treated as an absolute file path and blocked because it's outside the working directory.

## Fix

Added a check to skip any path match that starts with `//`, as these are URL path components, not file system paths.

## Changes

- Modified `guardCommand` in `pkg/tools/shell.go` to skip URL path components
- Added test case `TestShellTool_URLsNotBlocked` to verify the fix

Fixes #1203

### DIFF
--- a/pkg/tools/shell.go
+++ b/pkg/tools/shell.go
@@ -336,35 +336,34 @@ func (t *ExecTool) guardCommand(command, cwd string) string {
 			return ""
 		}
 
-		matches := absolutePathPattern.FindAllString(cmd, -1)
+		// Web URL schemes whose path components (starting with //) should be exempt
+		// from workspace sandbox checks. file: is intentionally excluded so that
+		// file:// URIs are still validated against the workspace boundary.
+		webSchemes := []string{"http:", "https:", "ftp:", "ftps:", "sftp:", "ssh:", "git:"}
 
-		for _, raw := range matches {
+		matchIndices := absolutePathPattern.FindAllStringIndex(cmd, -1)
+
+		for _, loc := range matchIndices {
+			raw := cmd[loc[0]:loc[1]]
+
 			// Skip URL path components that look like they're from web URLs.
 			// When a URL like "https://github.com" is parsed, the regex captures
 			// "//github.com" as a match (the path portion after "https:").
-			// These double-slash prefixes indicate URL paths, not file system paths.
-			// However, we must NOT skip file:// URIs as they could escape the sandbox.
-			// Only skip if preceded by a web URL scheme (http:, https:, ftp:, etc.).
-			if strings.HasPrefix(raw, "//") {
-				// Check if this // path is preceded by a web URL scheme
-				// by looking for patterns like "http://", "https://", "ftp://" before the match
-				idx := strings.Index(cmd, raw)
-				if idx > 0 {
-					// Look for the scheme prefix (e.g., "https:") before the //
-					before := cmd[:idx]
-					// Check if it ends with a web URL scheme followed by colon
-					// Web schemes: http, https, ftp, ftps, sftp, ssh, git
-					webSchemes := []string{"http:", "https:", "ftp:", "ftps:", "sftp:", "ssh:", "git:"}
-					isWebURL := false
-					for _, scheme := range webSchemes {
-						if strings.HasSuffix(before, scheme) {
-							isWebURL = true
-							break
-						}
+			// Use the exact match position (loc[0]) so that duplicate //path substrings
+			// in the same command are each evaluated at their own position.
+			if strings.HasPrefix(raw, "//") && loc[0] > 0 {
+				before := cmd[:loc[0]]
+				isWebURL := false
+
+				for _, scheme := range webSchemes {
+					if strings.HasSuffix(before, scheme) {
+						isWebURL = true
+						break
 					}
-					if isWebURL {
-						continue
-					}
+				}
+
+				if isWebURL {
+					continue
 				}
 			}
 
--- a/pkg/tools/shell_test.go
+++ b/pkg/tools/shell_test.go
@@ -515,3 +515,29 @@ func TestShellTool_FileURISandboxing(t *testing.T) {
 		}
 	}
 }
+
+// TestShellTool_URLBypassPrevented verifies that a command cannot bypass the workspace
+// sandbox by smuggling a real path after a URL that contains the same //path substring.
+// e.g. "echo https://etc/passwd && cat //etc/passwd" must still be blocked.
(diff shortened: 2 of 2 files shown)


hkc5:
@imguoguo

hkc5:
Thanks for the review Nicolas, I guess we need another reviewer
tagging from the CONTRIBUTING.md
@yinwm @alexhoshina @lxowalle @Zhaoyikaiii

pkg/tools/shell.go
```diff
@@ -339,6 +339,14 @@ func (t *ExecTool) guardCommand(command, cwd string) string {
 		matches := absolutePathPattern.FindAllString(cmd, -1)
 
 		for _, raw := range matches {
+			// Skip URL path components that look like they're from URLs.
+			// When a URL like "https://github.com" is parsed, the regex captures
+			// "//github.com" as a match (the path portion after "https:").
+			// These double-slash prefixes indicate URL paths, not file system paths.
+			if strings.HasPrefix(raw, "//") {
```
alexhoshina: `file:` **URI** is also skipped, bypassing workspace sandbox checks. Attackers can use `file:Paths like `/etc/passwd` **escaping**

alexhoshina:
`make lint` plz

pkg/tools/shell.go
```diff
@@ -339,6 +339,35 @@ func (t *ExecTool) guardCommand(command, cwd string) string {
...
+					// Web schemes: http, https, ftp, ftps, sftp, ssh, git
+					webSchemes := []string{"http:", "https:", "ftp:", "ftps:", "sftp:", "ssh:", "git:"}
+					isWebURL := false
+					for _, scheme := range webSchemes {
+						if strings.HasSuffix(before, scheme) {
+							isWebURL = true
+							break
+						}
+					}
+					if isWebURL {
```
alexhoshina: This URL exemption can be bypassed because strings.Index(cmd, raw) always returns the first occurrence of a matched `//... `substring,not the occurrence currently being validated. If the command contains the same`//etc/passwd` text once inside an `https://...` URL and again as a real shell path later, both matches are treated as URL components and the second one is no longer checked against the wor kspace restriction. In that case commands like `echo https://etc/passwd && cat //etc/passwd` can read outside the working directory.

pkg/tools/shell.go
```diff
@@ -336,9 +336,37 @@ func (t *ExecTool) guardCommand(command, cwd string) string {
...
+			// When a URL like "https://github.com" is parsed, the regex captures
+			// "//github.com" as a match (the path portion after "https:").
+			// Use the exact match position (loc[0]) so that duplicate //path substrings
+			// in the same command are each evaluated at their own position.
+			if strings.HasPrefix(raw, "//") && loc[0] > 0 {
+				before := cmd[:loc[0]]
+				isWebURL := false
+
+				for _, scheme := range webSchemes {
+					if strings.HasSuffix(before, scheme) {
```
alexhoshina, line 359: nit: `HTTPS:example.com` will be blocked. This issue does not affect merging; it is merely a minor detail.

pkg/tools/shell_test.go
```diff
@@ -443,3 +443,101 @@ func TestShellTool_CustomAllowPatterns(t *testing.T) {
...
+	// The URL path components (e.g., "//github.com") should be recognized as URLs,
+	// not as file system paths.
+	commands := []string{
+		"agent-browser open https://github.com",
+		"curl https://api.example.com/data",
+		"wget http://example.com/file",
+		"browser open https://github.com/user/repo",
+		"fetch ftp://ftp.example.com/file.txt",
+		"git clone https://github.com/sipeed/picoclaw.git",
+	}
```
alexhoshina, line 466: nit: Real network requests may cause the exec tool to time out or slow down the CI.

hkc5:
Thanks @alexhoshina for the thorough review! Happy to follow up with separate PRs for any remaining nits, would appreciate a merge when you get a chance!

Orgmar:
@hkc5 Clean fix for the URL false positive in the safety guard. The regex matching //github.com from URLs as an absolute path is a subtle edge case. Simple check for the // prefix handles it well.

We have a **PicoClaw Dev Group** on Discord where contributors can connect and collaborate. If you're interested, send an email to `support@sipeed.com` with the subject `[Join PicoClaw Dev Group] + your GitHub account` and we'll send you the invite link!

PR #1283, opened 2026-03-09
fix(feishu): preserve mention identity and sender open_id in messages

Vast-Stars:
## Summary
- Replace `@_user_N` placeholders with `@Name(open_id:xxx)` instead of stripping them, so downstream tools can extract user IDs for API calls
- Bot's own mentions are still stripped for trigger detection
- Fix mention handling not applied in p2p (direct) chats
- Inject sender `open_id` into message content and metadata so the LLM knows who sent the message

## Test plan
- [x] Unit tests updated and passing (`go test ./pkg/channels/feishu/ -run TestStripMention`)
- [ ] Manual test: send message with @mentions in group chat, verify names and open_ids appear in content
- [ ] Manual test: send message with @mentions in direct chat, verify same behavior
- [ ] Verify bot mention is still stripped correctly in group trigger detection

### DIFF
--- a/pkg/channels/feishu/common_test.go
+++ b/pkg/channels/feishu/common_test.go
@@ -256,26 +256,26 @@ func TestStripMentionPlaceholders(t *testing.T) {
 			name:    "user mention preserved with name and id",
 			content: "@_user_1 hello",
 			mentions: []*larkim.MentionEvent{
-				{Key: strPtr("@_user_1"), Id: &larkim.UserId{OpenId: strPtr("ou_user456")}, Name: strPtr("张三")},
+				{Key: strPtr("@_user_1"), Id: &larkim.UserId{OpenId: strPtr("ou_user456")}, Name: strPtr("ZhangSan")},
 			},
-			want: "@张三(open_id:ou_user456) hello",
+			want: "@ZhangSan(open_id:ou_user456) hello",
 		},
 		{
 			name:    "mixed bot and user mentions",
-			content: "@_user_1 给我和 @_user_2 拉个群",
+			content: "@_user_1 create a group with @_user_2",
 			mentions: []*larkim.MentionEvent{
 				{Key: strPtr("@_user_1"), Id: &larkim.UserId{OpenId: strPtr(botOpenID)}, Name: strPtr("Bot")},
-				{Key: strPtr("@_user_2"), Id: &larkim.UserId{OpenId: strPtr("ou_user789")}, Name: strPtr("李四")},
+				{Key: strPtr("@_user_2"), Id: &larkim.UserId{OpenId: strPtr("ou_user789")}, Name: strPtr("LiSi")},
 			},
-			want: "给我和 @李四(open_id:ou_user789) 拉个群",
+			want: "create a group with @LiSi(open_id:ou_user789)",
 		},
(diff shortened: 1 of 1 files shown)


pkg/channels/feishu/feishu_64.go
```diff
@@ -394,6 +400,19 @@ func (c *FeishuChannel) handleMessageReceive(ctx context.Context, event *larkim.
 	if sender != nil && sender.TenantKey != nil {
 		metadata["tenant_key"] = *sender.TenantKey
 	}
+	if senderID != "" {
+		metadata["sender_open_id"] = senderID
```
alexhoshina, line 404: `extractFeishuSenderID` prioritizes returning **UserId**, followed by **OpenId**, and then **UnionId**. Therefore, **senderID** is mostly **user_id**, not **open_id**.
```go
func extractFeishuSenderID(sender *larkim.EventSender) string {
	if sender == nil || sender.SenderId == nil {
		return ""
	}

	if sender.SenderId.UserId != nil && *sender.SenderId.UserId != "" {
		return *sender.SenderId.UserId
	}
	if sender.SenderId.OpenId != nil && *sender.SenderId.OpenId != "" {
		return *sender.SenderId.OpenId
	}
	if sender.SenderId.UnionId != nil && *sender.SenderId.UnionId != "" {
		return *sender.SenderId.UnionId
	}

	return ""
}
```

pkg/channels/feishu/feishu_64.go
```diff
@@ -394,6 +400,19 @@ func (c *FeishuChannel) handleMessageReceive(ctx context.Context, event *larkim.
...
+	}
+	// Store the full open_id for API use (senderID may be user_id instead)
+	if sender != nil && sender.SenderId != nil && sender.SenderId.OpenId != nil && *sender.SenderId.OpenId != "" {
+		metadata["sender_open_id"] = *sender.SenderId.OpenId
```
alexhoshina, line 408: If the `OpenId` happens to be empty while the `UserId` is not, an ID of the wrong type will be stored in the metadata.

pkg/channels/feishu/feishu_64.go
```diff
@@ -417,6 +431,17 @@ func (c *FeishuChannel) handleMessageReceive(ctx context.Context, event *larkim.
...
 	}
 
+	// Prepend sender identity so the LLM knows who sent the message.
+	if sender != nil && sender.SenderId != nil {
+		openID := ""
+		if sender.SenderId.OpenId != nil {
+			openID = *sender.SenderId.OpenId
+		}
+		if openID != "" {
+			content = fmt.Sprintf("[sender: open_id=%s] %s", openID, content)
```
alexhoshina, line 441: Is it necessary to prefix all messages with `[sender: open_id=xxx]`?

pkg/channels/feishu/common_test.go
alexhoshina, line 1: Missing tests for **open_id only**

alexhoshina:
These two comments apply to this PR   
https://github.com/sipeed/picoclaw/pull/1318#discussion_r2922509126
https://github.com/sipeed/picoclaw/pull/1318#discussion_r2922512917

PR #1296, opened 2026-03-10
chore: remove duplicated timestamp from Go log output

yorkhellen:
## 📝 Description

This PR removes the duplicated timestamp in log output 
image

The go standard log package automatically add a timestamp to each log entry, while the project’s logging already includes its own timestamp field. This change disables the default Go log timestamp to avoid redundant time information in logs.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## ☑️ Checklist

### DIFF
--- a/pkg/logger/logger.go
+++ b/pkg/logger/logger.go
@@ -52,6 +52,8 @@ type LogEntry struct {
 func init() {
 	once.Do(func() {
 		logger = &Logger{}
+		// disable default log prefix date & timestamp
+		log.SetFlags(0)
 	})
 }
 


afjcjsbx:
for me it is ok, however I would like to hear someone else's opinion as well because adding **log.SetFlags(0)** surely solve the double timestamp problem in picoclaw logs however there is to consider that the standard log package of Go acts as a global singleton. Calling **log.SetFlags(0)** changes the default logger behavior for the whole application, including all imported third-party libraries.

If any library (e.g., a database driver, HTTP server, etc.) internally uses log.Println() or log.Printf() to print its own errors or warnings, those printouts will suddenly lose the date and time, making debugging those packages much more difficult.

putueddy:
Clean pick, trivial change (+2/-0). Removes the redundant `log.Ldate|log.Ltime` prefix since the structured logger already adds timestamps. Log output is noticeably cleaner now. Tested on Pi 4, v0.2.1.

Issue #1247, opened 2026-03-08
[BUG] OpenRouter provider keep removing "openrouter/" prefix when sending the POST request over the network


confusionhill:
##  Environment & Tools
- **PicoClaw Version:** picoclaw v0.2.0-284-g66e6fb6
- **Go Version:** go1.25.7
- **AI Model & Provider:**  openrouter-free & openrouter
- **Operating System:**  Linux/ Raspberry Pi OS
- **Channels:** NA

## 📸 Steps to Reproduce
1.  Follow instruction from https://github.com/sipeed/picoclaw/blob/81dfdf5f45d8f503189b9d7087c3e47e61b27c63/docs/troubleshooting.md?plain=1#L8
2. Run `picoclaw agent -m "hel"`
3. Will retrieve ` LLM call failed {agent_id=main, iteration=1, error=API request failed`

## ❌ Actual Behavior
network call fails because openrouter prefix is removed when sending`POST` request to openrouter

```bash
curl https://openrouter.ai/api/v1/chat/completions \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer $OPENROUTER_API_KEY" \
  -d '{
  "model": "qwen3-next-80b-a3b-instruct:free",
  "messages": [
      {
        "role": "user",
        "content": "What is the meaning of life?"
      }
    ]
  
}'
```

## ✅ Expected Behavior

It should include it provider just like on open router example

```bash
curl https://openrouter.ai/api/v1/chat/completions \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer $OPENROUTER_API_KEY" \
  -d '{
  "model": "qwen/qwen3-next-80b-a3b-instruct:free",
  "messages": [
      {
        "role": "user",
        "content": "What is the meaning of life?"
      }
    ]
  
}'
```

## 💬 Additional Context

i asumed this is a noticable bug by the dev, if not will send patch over the weekend

confusionhill:
if anyone encountering the same issue, try adding a random prefix

ex
```
{
      "model_name": "openrouter-free",
      "model": "[randomword]/openrouter/free",
     "api_key": "somekey",
      "api_base": "https://openrouter.ai/api/v1"
    }
```

instead of 
```
  "model": "openrouter/free"
```

agardnerIT:
The above advice doesn't work on `0.2.3`
```
{
      "model_name": "openrouter-free",
      "model": "fdsk/openrouter/free",
[...]

PR #1308, opened 2026-03-10
Fix(provider) Prevent openrouter Provider models from getting cut

confusionhill:
# 📝 Description
This PR fixes an issue where model name prefixes (e.g., `qwen/`) were being incorrectly truncated when using the **OpenRouter** provider. For OpenRouter, vendor prefixes are required to correctly route the request (e.g., `qwen/qwen3-next-80b-a3b-instruct:free`).

The fix updates `ExtractProtocol` to be aware of the `APIBase`. When the URL contains `openrouter.ai`, the full model identifier (including the vendor prefix) is preserved as the `modelID`, while still allowing standard prefix-stripping for native providers (like Groq or OpenAI) to maintain backward compatibility.

## 🚀 Type of Change
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Documentation update
- [ ] Refactor

## 🤖 AI Code Generation
- [ ] 🤖 Fully AI-generated
- [ ] 🛠️ Mostly AI-generated
- [x] 👨‍💻 Mostly Human-written (Developed with AI assistance but human-led implementation)

## 🔗 Related Issue
https://github.com/sipeed/picoclaw/issues/1247

## 🛠️ Technical Context
The core problem was that `ExtractProtocol` was stripping everything before the first slash unconditionally. This PR modifies `ExtractProtocol` to accept the `apiBase` URL. If the URL points to OpenRouter, the full model string (e.g., `qwen/model-id`) is returned.

#### Before (Broken - Prefix cut)
```json
{
  "model_name": "openrouter-free",
  "model": "stepfun/step-3.5-flash:free",
  "api_key": "",
  "api_base": "https://openrouter.ai/api/v1"
}
```

#### After (Previously required workaround)
```json
{
  "model_name": "openrouter-free",
  "model": "gemini/stepfun/step-3.5-flash:free",
  "api_key": "",
  "api_base": "https://openrouter.ai/api/v1"
}
```

Relevant changes:
- **`pkg/providers/factory_provider.go`**: Modified `ExtractProtocol` to accept `url` and `CreateProviderFromConfig` to pass `cfg.APIBase`.
- **`pkg/agent/instance.go`**: Updated the `ensureProtocol` helper to correctly pass the API base.
- **`pkg/providers/factory_provider_test.go`**: Added test cases for the new URL-aware protocol extraction.
- **`pkg/providers/openai_compat/provider_test.go`**: Added regression tests for model normalization with OpenRouter.

## 🧪 Test Environment
- **OS**: macOS
- **Go Version**: 1.25+
- **Provider**: OpenRouter
- **Model tested**: `qwen/qwen3-235b-a22b-2507`

## ✅ Checklist
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes
- [x] `make check` passes locally

### DIFF
--- a/pkg/agent/instance.go
+++ b/pkg/agent/instance.go
@@ -183,7 +183,7 @@ func NewAgentInstance(
 				if fullModel == raw {
 					return ensureProtocol(fullModel), true
 				}
-				_, modelID := providers.ExtractProtocol(fullModel)
+				_, modelID := providers.ExtractProtocol(fullModel, cfg.GetAPIBase())
 				if modelID == raw {
 					return ensureProtocol(fullModel), true
 				}
--- a/pkg/providers/factory_provider.go
+++ b/pkg/providers/factory_provider.go
@@ -38,15 +38,20 @@ func createCodexAuthProvider() (LLMProvider, error) {
 
 // ExtractProtocol extracts the protocol prefix and model identifier from a model string.
 // If no prefix is specified, it defaults to "openai".
+// If the model is from OpenRouter, it returns the model as is.
 // Examples:
-//   - "openai/gpt-4o" -> ("openai", "gpt-4o")
-//   - "anthropic/claude-sonnet-4.6" -> ("anthropic", "claude-sonnet-4.6")
-//   - "gpt-4o" -> ("openai", "gpt-4o")  // default protocol
-func ExtractProtocol(model string) (protocol, modelID string) {
+//   - "openai/gpt-4o", "https://openrouter.ai/api/v1" -> ("openai", "gpt-4o")
+//   - "anthropic/claude-sonnet-4.6", "https://api.anthropic.com/v1" -> ("anthropic", "claude-sonnet-4.6")
+//   - "gpt-4o", "https://api.openai.com/v1" -> ("openai", "gpt-4o")  // default protocol
+//   - "openrouter/gpt-4o", "https://openrouter.ai/api/v1" -> ("openrouter", "openrouter/gpt-4o")
+func ExtractProtocol(model string, url string) (protocol, modelID string) {
 	model = strings.TrimSpace(model)
 	protocol, modelID, found := strings.Cut(model, "/")
 	if !found {
 		return "openai", model
+	}
+	if strings.Contains(strings.ToLower(url), "openrouter.ai") {
+		return protocol, model
 	}
 	return protocol, modelID
 }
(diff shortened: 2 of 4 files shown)


confusionhill:
> This addresses the same root cause as PRs #1312 and #1303 (issue #1247) but from a different angle -- making `ExtractProtocol` URL-aware so it preserves the vendor prefix for OpenRouter models.
> 
> The approach has merit but introduces some concerns:
> 
> 1. **API change to `ExtractProtocol`** -- adding a `url` parameter changes the signature of a function that is called from multiple places. The PR correctly updates both call sites (`factory_provider.go` and `instance.go`), but any future callers must remember to pass the API base URL.
> 2. **OpenRouter-specific logic in a generic function** -- `ExtractProtocol` is supposed to be a generic protocol extractor, but now it has an `if strings.Contains(url, "openrouter.ai")" check. This creates a precedent where other providers with similar multi-segment model IDs (e.g., `nvidia/meta/llama-3.1-8b`) would need their own special cases. Consider whether a more general approach (e.g., a "preserve full model ID" flag in `ModelConfig`) would be better.
> 3. **Return value semantics change** -- for OpenRouter, the function now returns `(protocol, fullModel)` instead of `(protocol, modelID)`. This means the `modelID` return value contains the protocol prefix, which could confuse callers that expect `modelID` to be the prefix-stripped value.
> 4. **Good test coverage** -- the test additions in both `factory_provider_test.go` and `openai_compat/provider_test.go` are thorough and cover the key scenarios (OpenRouter preserves prefix, native providers strip prefix, unknown prefixes at OpenRouter).
> 5. **nvidia case already existed** -- the existing test shows `nvidia/meta/llama-3.1-8b -> ("nvidia", "meta/llama-3.1-8b")` which already preserves multi-segment model IDs for nvidia. This suggests the real issue is specific to how OpenRouter models are treated downstream, not in `ExtractProtocol` itself.
> 
> The fix works, but I think the cleaner long-term solution is the approach in PRs #1312/#1303 (returning resolved model from candidates) rather than making the protocol extractor URL-aware.

i agree upon your points, it was tightly coupled. my focus was on fixing openrouter it self, like a patch, a quick patch.

i would also endorse 
https://github.com/sipeed/picoclaw/pull/1303

as it has the cleanest approach between the 3

PR #1318, opened 2026-03-10
fix(feishu): invalidate cached token on auth error to enable retry recovery

Vast-Stars:
## Summary

- The Lark SDK v3's built-in token retry loop does not clear stale tokens from cache when the server returns error `99991663` (tenant_access_token invalid), causing all API calls to fail until the token naturally expires (~2 hours)
- Implement a custom `tokenCache` (implementing `larkcore.Cache`) with an `InvalidateAll()` method, injected via `lark.WithTokenCache()`
- On any API response with code `99991663`, invalidate the cache so the next application-level retry fetches a fresh token

## Changes

- Added `tokenCache` struct with `Get`/`Set`/`InvalidateAll` methods
- Wired custom cache into `lark.NewClient` via `WithTokenCache()`
- Added `invalidateTokenOnAuthError` helper called in all API methods: `sendCard`, `EditMessage`, `SendPlaceholder`, `ReactToMessage`, `fetchBotOpenID`, `sendImage`, `sendFile`, `downloadResource`

## Test plan

- [x] `make build` compiles successfully
- [x] `go test ./pkg/channels/feishu/` all pass
- [ ] Deploy and verify Feishu bot message sending recovers automatically after token expiration

### DIFF
--- a/pkg/channels/feishu/feishu_64.go
+++ b/pkg/channels/feishu/feishu_64.go
@@ -29,11 +29,17 @@ import (
 	"github.com/sipeed/picoclaw/pkg/utils"
 )
 
+// errCodeTenantTokenInvalid is the Feishu API error code for an expired/revoked
+// tenant_access_token. The Lark SDK's built-in retry does not clear its cache
+// on this error, so we do it ourselves.
+const errCodeTenantTokenInvalid = 99991663
+
 type FeishuChannel struct {
 	*channels.BaseChannel
-	config   config.FeishuConfig
-	client   *lark.Client
-	wsClient *larkws.Client
+	config     config.FeishuConfig
+	client     *lark.Client
+	wsClient   *larkws.Client
+	tokenCache *tokenCache // custom cache that supports invalidation
 
 	botOpenID atomic.Value // stores string; populated lazily for @mention detection
 
@@ -47,10 +53,12 @@ func NewFeishuChannel(cfg config.FeishuConfig, bus *bus.MessageBus) (*FeishuChan
 		channels.WithReasoningChannelID(cfg.ReasoningChannelID),
 	)
 
+	tc := newTokenCache()
 	ch := &FeishuChannel{
 		BaseChannel: base,
 		config:      cfg,
-		client:      lark.NewClient(cfg.AppID, cfg.AppSecret),
+		tokenCache:  tc,
+		client:      lark.NewClient(cfg.AppID, cfg.AppSecret, lark.WithTokenCache(tc)),
 	}
 	ch.SetOwner(ch)
 	return ch, nil
@@ -147,6 +155,7 @@ func (c *FeishuChannel) EditMessage(ctx context.Context, chatID, messageID, cont
 		return fmt.Errorf("feishu edit: %w", err)
 	}
 	if !resp.Success() {
+		c.invalidateTokenOnAuthError(resp.Code)
 		return fmt.Errorf("feishu edit api error (code=%d msg=%s)", resp.Code, resp.Msg)
 	}
 	return nil
@@ -186,6 +195,7 @@ func (c *FeishuChannel) SendPlaceholder(ctx context.Context, chatID string) (str
 		return "", fmt.Errorf("feishu placeholder send: %w", err)
 	}
 	if !resp.Success() {
+		c.invalidateTokenOnAuthError(resp.Code)
 		return "", fmt.Errorf("feishu placeholder api error (code=%d msg=%s)", resp.Code, resp.Msg)
 	}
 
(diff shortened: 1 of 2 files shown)


CLAUDE.md
alexhoshina, line 1: Should not be submitted

README.local.md
alexhoshina, line 1: Should not be submitted

CLAUDE.md
Vast-Stars: sorry

README.local.md
Vast-Stars: FIxed！

CLAUDE.md
Vast-Stars: done

pkg/channels/feishu/feishu_64.go
```diff
@@ -830,3 +868,55 @@ func extractFeishuSenderID(sender *larkim.EventSender) string {
...
+	}
+	return e.value, nil
+}
+
+// InvalidateAll removes all cached tokens, forcing fresh acquisition.
+func (c *tokenCache) InvalidateAll() {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	clear(c.store)
+}
```
alexhoshina: Perhaps create a separate file? It's up to you

pkg/channels/feishu/feishu_64.go
```diff
@@ -830,3 +868,55 @@ func extractFeishuSenderID(sender *larkim.EventSender) string {
...
+
+func (c *tokenCache) Get(_ context.Context, key string) (string, error) {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	e, ok := c.store[key]
+	if !ok || e.expireAt.Before(time.Now()) {
+		return "", nil
+	}
+	return e.value, nil
+}
```
alexhoshina: `Get()` returns only a null value when detecting an expired entry, but does not remove the entry from the map. During long-term operation, expired keys will continue to accumulate.

alexhoshina (changes requested):

pkg/channels/feishu/feishu_64.go
Vast-Stars: Good idea, I'll move it to a separate file.

pkg/channels/feishu/feishu_64.go
Vast-Stars: `RLock` change to `Lock`，use `delete(c.store, key)` to clean up when item expired

alexhoshina:
`make lint` plz

pkg/channels/feishu/feishu_64.go
```diff
@@ -417,6 +436,17 @@ func (c *FeishuChannel) handleMessageReceive(ctx context.Context, event *larkim.
...
 	}
 
+	// Prepend sender identity so the LLM knows who sent the message.
+	if sender != nil && sender.SenderId != nil {
+		openID := ""
+		if sender.SenderId.OpenId != nil {
+			openID = *sender.SenderId.OpenId
+		}
+		if openID != "" {
+			content = fmt.Sprintf("[sender: open_id=%s] %s", openID, content)
```
alexhoshina: This causes all messages to start with `[sender: open_id=xxxx]`, which prevents the command system from working, as the command system only recognizes messages that begin with `/` or `!`.
Original message: `/help`
Will be processed as: `[sender: open_id=ou_xxx] /help`
`handleCommand` cannot recognize it as a command, so the message will be **sent to the agent as ordinary chat text** and **not executed** as a command.

pkg/channels/feishu/feishu_64.go
```diff
@@ -394,6 +405,19 @@ func (c *FeishuChannel) handleMessageReceive(ctx context.Context, event *larkim.
...
+	}
+	// Store the full open_id for API use (senderID may be user_id instead)
+	if sender != nil && sender.SenderId != nil && sender.SenderId.OpenId != nil && *sender.SenderId.OpenId != "" {
+		metadata["sender_open_id"] = *sender.SenderId.OpenId
+	}
+
+	// Replace mention placeholders for all chat types: bot mentions are stripped, others become @Name(open_id:xxx)
+	if len(message.Mentions) > 0 {
+		knownBotID, _ := c.botOpenID.Load().(string)
+		content = stripMentionPlaceholders(content, message.Mentions, knownBotID)
```
alexhoshina: For example
Group message: `@Alice /help`
will be processed as: `@Alice(open_id:...) /help`
causing the command to be unrecognized.

Vast-Stars: Sorry, the old commits were mixed with other changes. Now the commit history is clean.

PR #1319, opened 2026-03-10
fix(agent): skip final reply only when message tool sent to same chat

Vast-Stars:
## Summary

- Fix: when the message tool sends to a different chat_id (e.g. a newly created Feishu group), the final reply to the original chat was incorrectly suppressed
- Track the target chat_id in `MessageTool` and only skip the outbound reply when it matches the inbound message's chat_id

## Changes

- `pkg/tools/message.go`: add `sentChatID` field, `SentToChatID()` method, reset in `ResetSentInRound`
- `pkg/agent/loop.go`: `alreadySent` condition now checks `SentToChatID() == msg.ChatID`

## Test Plan

- [x] `go test ./pkg/tools/` — passed
- [x] `go vet` — clean
- [x] Manual test: bot creates Feishu group and sends message to it, original private chat still receives the final reply

## bug behavior
拉群之后，能在新群里发消息，但是在原始对话里无法继续。
After bot create group and send messages in the new group, the bot can not send messages in the old chat, which still have a streaming stuck card
image

### DIFF
--- a/pkg/agent/loop.go
+++ b/pkg/agent/loop.go
@@ -346,7 +346,7 @@ func (al *AgentLoop) Run(ctx context.Context) error {
 					if defaultAgent != nil {
 						if tool, ok := defaultAgent.Tools.Get("message"); ok {
 							if mt, ok := tool.(*tools.MessageTool); ok {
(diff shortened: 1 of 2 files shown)


TempestShaw:
It would be better to keep both claude.md and README.local.md in the local environment only.

PR #1321, opened 2026-03-10
fix(providers): resolve correct provider per fallback candidate

Vast-Stars:
## Summary

- Fix: fallback chain reused the primary model's provider instance (API key + endpoint) for all candidates
- When the primary model was zhipu and fallbacks were moonshot/anthropic/minimax, every fallback request was sent with zhipu's API key, causing 401 across the board
- Add a lazy `resolveProvider` method that looks up each candidate's `ModelConfig` from `model_list` and creates the correct provider instance with the right API key and base URL

## Root Cause

The `run` closure in the fallback execution always called `agent.Provider.Chat()`, where `agent.Provider` was the single provider instance created at startup for the primary model. Although the fallback chain correctly identified candidates like `moonshot/kimi-k2.5`, the actual HTTP requests were sent through the primary (zhipu) provider with the wrong credentials.

## Changes

- `pkg/agent/loop.go`:
  - Add `providerCache sync.Map` field to `AgentLoop`
  - Add `resolveProvider(providerName, model, fallbackProvider)` method that lazily creates and caches providers per candidate from `model_list` config
  - Update fallback `run` closure to use `resolveProvider()` instead of `agent.Provider`

## Test Plan

- [x] `go test ./pkg/providers/` — passed
- [x] `make vet` — clean
- [ ] Manual test: intentionally break primary model key, verify fallback correctly switches to other providers

### DIFF
--- a/pkg/agent/loop.go
+++ b/pkg/agent/loop.go
@@ -48,6 +48,7 @@ type AgentLoop struct {
 	mediaStore     media.MediaStore
 	transcriber    voice.Transcriber
 	cmdRegistry    *commands.Registry
+	providerCache  sync.Map // map[string]providers.LLMProvider — lazily created per provider/model
 }
 
 // processOptions configures how a message is processed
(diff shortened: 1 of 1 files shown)


Vast-Stars:
Reopening from a clean branch based on main.

PR #1324, opened 2026-03-10
fix(providers): resolve correct provider per fallback candidate

Vast-Stars:
## Summary

- Fix: fallback chain reused the primary model's provider instance (API key + endpoint) for all candidates
- When the primary model was zhipu and fallbacks were moonshot/anthropic/minimax, every fallback request was sent with zhipu's API key, causing 401 across the board
- Add a lazy `resolveProvider` method that looks up each candidate's `ModelConfig` from `model_list` and creates the correct provider instance with the right API key and base URL

## Root Cause

The `run` closure in the fallback execution always called `agent.Provider.Chat()`, where `agent.Provider` was the single provider instance created at startup for the primary model. Although the fallback chain correctly identified candidates like `moonshot/kimi-k2.5`, the actual HTTP requests were sent through the primary (zhipu) provider with the wrong credentials.

## Changes

- `pkg/agent/loop.go`:
  - Add `providerCache sync.Map` field to `AgentLoop`
  - Add `resolveProvider(providerName, model, fallbackProvider)` method that lazily creates and caches providers per candidate from `model_list` config
  - Update fallback `run` closure to use `resolveProvider()` instead of `agent.Provider`

## Test Plan

- [x] `go test ./pkg/providers/` — passed
- [x] `make vet` — clean
- [ ] Manual test: intentionally break primary model key, verify fallback correctly switches to other providers

### DIFF
--- a/pkg/agent/loop.go
+++ b/pkg/agent/loop.go
@@ -48,6 +48,7 @@ type AgentLoop struct {
 	mediaStore     media.MediaStore
 	transcriber    voice.Transcriber
 	cmdRegistry    *commands.Registry
+	providerCache  sync.Map // map[string]providers.LLMProvider — lazily created per provider/model
 }
 
 // processOptions configures how a message is processed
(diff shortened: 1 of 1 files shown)


nikolasdehor (commented):
This addresses a real and important bug -- fallback candidates using the primary model's provider (wrong API key/endpoint). The fix is architecturally sound.

Some observations:

1. **sync.Map usage** -- using `sync.Map` for the provider cache is fine for this read-heavy workload. However, there is a subtle race in `resolveProvider`: two goroutines could both miss the `Load`, both create a provider, and both call `Store`. This is benign (last-write-wins, both values are equivalent) but you could use `LoadOrStore` to avoid creating duplicate provider instances.

2. **Case-insensitive comparison** -- `strings.EqualFold(entryModel, target)` is good for user-friendliness, but note that `providerName + "/" + model` (the `target`) is constructed from the candidate's fields, which may already be lowercased or normalized elsewhere. If the model list has "OpenRouter/GPT-4o" and the candidate has "openrouter/gpt-4o", this handles it correctly.

3. **Fallback to agent's default provider** -- when the model is not found in `model_list`, falling back to `fallbackProvider` (the agent's primary provider) and caching it is reasonable, but it means the same provider gets cached under a key it doesn't truly belong to. If the config is later reloaded with the correct model entry, the stale cache would persist until restart. Consider whether the cache should have an invalidation mechanism, or whether "not found" results should not be cached.

4. **Missing test coverage** -- the PR description mentions passing tests, but no new test was added for `resolveProvider` itself. A unit test verifying that different candidates get different provider instances would strengthen confidence.

Overall the approach is correct and fixes a real production issue. The concerns above are minor and could be addressed in a follow-up.

putueddy:
Tested this on a Raspberry Pi 4 running v0.2.1 with Kimi (moonshot) as primary. Before this fix, fallback candidates were all getting 401s because they inherited the primary provider's API key. After cherry-picking, fallbacks correctly resolve their own provider credentials. Clean pick, no conflicts. LGTM 👍

PR #1331, opened 2026-03-10
fix(config): correct off-by-one in round-robin model selection

### LINKED ISSUE #1153: [BUG]Duplicate model_name entries don't fall back correctly — always resolves to second entry, ignores first and third
## Quick Summary
When multiple model_list entries share the same model_name (as implied by the fix(config): allow duplicate model_name for load balancing commit), the first entry is skipped entirely and the second is always used — even when the first model works correctly in isolation. If the second entry fails, it errors out rather than falling through to the third.

##  Environment & Tools
- **PicoClaw Version:** (0.2.0)
- **Go Version:** (1.25.7)
- **AI Model & Provider:** (Google)
- **Operating System:** (Debian Stable)
- **Channels:** (Telegram)

## 📸 Steps to Reproduce
1. Add three model_list entries with the same model_name, 
e.g.:
{ "model_name": "google", "model": "gemini/gemini-3.1-flash-lite-preview", ... },
{ "model_name": "google", "model": "gemini/gemini-3-flash-preview", ... },
{ "model_name": "google", "model": "gemini/gemini-flash-lite-latest", ... }
2. Set agents.defaults.model_name to "google"
3. Send a message — it always hits the second entry
4. Make the second entry invalid — it errors instead of trying the third

## ❌ Actual Behavior
1. First entry is always skipped
2. Second entry is always used
3. No fallback to third entry on failure

## ✅ Expected Behavior
Either round-robin across all matching entries, or try them in order with fallback on failure. The first entry should not be silently skipped.

nikolasdehor:
Closes #1153

## Problem

When multiple `model_list` entries share the same `model_name` (for load
balancing), the round-robin selector in `GetModelConfig` always skips the
first entry on its initial call.

**Root cause:** `rrCounter.Add(1)` returns the *post-increment* value (1 on
the first call), so `1 % N` resolves to index 1 — the second entry.  The
first entry (index 0) is only reached after N calls instead of on the very
first call.

With 3 entries the observed pattern is: 2nd → 3rd → 1st → 2nd → ...
instead of: 1st → 2nd → 3rd → 1st → ...

## Fix

Subtract 1 from the counter before the modulo operation:

```go
idx := (rrCounter.Add(1) - 1) % uint64(len(matches))
```

This makes the sequence start at index 0 while keeping atomic increment
semantics for concurrent safety.

## Tests

- `TestGetModelConfig_RoundRobinSequence` — 3 entries, verifies all 3 are
  reached within one full cycle.
- `TestGetModelConfig_TwoEntries_BothReachable` — 2 entries, verifies both
  are hit within 2 calls.
- All existing config tests continue to pass.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -914,8 +914,11 @@ func (c *Config) GetModelConfig(modelName string) (*ModelConfig, error) {
 		return &matches[0], nil
 	}
 
-	// Multiple configs - use round-robin for load balancing
-	idx := rrCounter.Add(1) % uint64(len(matches))
+	// Multiple configs - use round-robin for load balancing.
+	// Subtract 1 because Add returns the new (post-increment) value,
+	// so without the subtraction the first entry (index 0) is skipped
+	// on the initial call.
+	idx := (rrCounter.Add(1) - 1) % uint64(len(matches))
 	return &matches[idx], nil
 }
 
--- a/pkg/config/model_config_test.go
+++ b/pkg/config/model_config_test.go
@@ -80,6 +80,55 @@ func TestGetModelConfig_RoundRobin(t *testing.T) {
 	}
 }
 
+func TestGetModelConfig_RoundRobinSequence(t *testing.T) {
+	cfg := &Config{
+		ModelList: []ModelConfig{
+			{ModelName: "seq-model", Model: "openai/model-a", APIKey: "key1"},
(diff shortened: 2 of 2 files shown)


putueddy:
Nice catch. The Add(1) - 1 fix is simple and correct — verified on our Pi 4 build. Tests pass including the new TestGetModelConfig_RoundRobinSequence. Would be good to merge this soon since it affects anyone with multiple model_list entries sharing the same model_name.

PR #1337, opened 2026-03-11
fix: resolve gateway binary path, pass --config flag, and clarify emp…

### LINKED ISSUE #1290: web backend Gateway config
# Gateway 自启动配置缺失与路径解析问题报告

## 问题描述

在使用 `picoclaw-launcher` (Web Backend) 控制面板通过调用 `/api/gateway/start` 接口启动 `picoclaw` 网关服务时，遇到以下两类报错：

1. **二进制文件路径未找到 `executable file not found in $PATH`**: 
   - HTTP Status 500
   - `Failed to auto-start gateway: failed to start gateway: exec: "picoclaw": executable file not found in $PATH`
2. **配置文件缺省导致启动奔溃**: 
   - 终端日志显示 gateway 进程启动后立即退出 (Exit Code 1)
   - `Error: error creating provider: model "" not found in model_list: model "" not found in model_list or providers`

此外，页面调用 `/api/gateway/status` 等待时会出现 Gateway 状态异常 (Status: "stopped", Start_Allowed: true)。

## 问题根本原因

1. **相对路径兼容性缺失**：在 `web/backend/api/gateway.go` 的 `findPicoclawBinary()` 中，仅扫描了启动器所在目录和系统环境变量的 `$PATH`。但是在本地 Makefile 测试环境中，`picoclaw` 编译于 `build/picoclaw` 路径下，而 `picoclaw-launcher` 编译与执行在 `web/backend/dist` 中，这导致跨目录启动时无法定位到网关程序二进制文件。
2. **配置隔离**：`picoclaw-launcher` 支持通过命令行参数接收配置文件路径 (例如: `make run CONFIG_PATH=/path/to/config.json`)。但在 `api/gateway.go` 的启动逻辑中，`exec.Command(execPath, "gateway")` 并没有将该配置路径转发给启动的网关。
   - `picoclaw gateway` 存在默认的配置读取逻辑 `~/.picoclaw/config.json`。
   - 当 `~/.picoclaw/config.json` 不存在或者其中并未有效配置模型提供商(Provider)时，Gateway 瞬间抛出 Panic 退出，即 Exit status 1。这就解释了为什么 Launcher 显示为 Stopped 但实际上后端有触发过 Start().

### 1. 完善二进制文件路径探测 `findPicoclawBinary()`
在搜索策略中添加相对当前执行目录和工作目录的前置路径检查，例如：
```go
// 匹配 web/backend 结构与全局 build 输出
for _, rel := range []string{"..", "../..", "../../.."} {
    candidate := filepath.Join(dir, rel, "build", binaryName)
    if info, err := os.Stat(candidate); err == nil && !info.IsDir() {
        return candidate
    }
}
```

### 2. 将配置路径显式传递给 Gateway 进程
修改 `startGatewayLocked()` 中 `exec.Command` 的参数列表：
```go
// 通过 --config 显示指定配置给网关
cmd := exec.Command(execPath, "gateway", "--config", h.configPath)
```

## 测试结果
修复后，通过面板控制网关启动，网关不再受限于未全局安装的限制，且将成功读入面板指定的 `config.json` 并进入正常 Running 状态，与状态 API 相挂钩同步无误。

snac21:
This PR addresses the 3 issues reported regarding the gateway startup process, incorporating the latest feedback:

1. **Binary path resolution:** `findPicoclawBinary()` now checks the `PICOCLAW_BINARY` environment variable first as an explicit override. As suggested, the fragile relative path probing (`../../build/picoclaw`) was removed. It now falls back to the same directory as the executable (which naturally covers the production build layout) before relying on `$PATH`.
2. **Config path forwarding:** By setting the `PICOCLAW_CONFIG` environment variable in the child process before `cmd.Start()`. This seamlessly integrates with the existing CLI config resolution logic.
3. **Crash on empty model:** Improved robustness in `pkg/providers/factory.go`. Instead of panicking with a confusing "no API key configured for model: " error when an empty model reaches the default switch case, it now checks if both the provider and model are empty and returns a clear `"no model configured: agents.defaults.model is empty"` error early. 

All tests have been updated and pass successfully.

Fixes #1290

### DIFF
--- a/web/backend/api/gateway.go
+++ b/web/backend/api/gateway.go
@@ -134,6 +134,12 @@ func (h *Handler) startGatewayLocked() (int, error) {
 	execPath := findPicoclawBinary()
 
 	cmd := exec.Command(execPath, "gateway")
+	// Forward the launcher's config path via the environment variable that
+	// GetConfigPath() already reads, so the gateway sub-process uses the same
+	// config file without requiring a --config flag on the gateway subcommand.
+	if h.configPath != "" {
+		cmd.Env = append(os.Environ(), "PICOCLAW_CONFIG="+h.configPath)
+	}
 
 	stdoutPipe, err := cmd.StdoutPipe()
 	if err != nil {
@@ -530,18 +536,32 @@ func (h *Handler) currentGatewayStatus() string {
 }
 
 // findPicoclawBinary locates the picoclaw executable.
-// Tries the same directory as the current executable first, then falls back to $PATH.
+// Search order:
+//  1. PICOCLAW_BINARY environment variable (explicit override)
+//  2. Same directory as the current executable
+//  3. Falls back to "picoclaw" and relies on $PATH
 func findPicoclawBinary() string {
-	if exe, err := os.Executable(); err == nil {
-		dir := filepath.Dir(exe)
-		candidate := filepath.Join(dir, "picoclaw")
-		if runtime.GOOS == "windows" {
-			candidate += ".exe"
+	binaryName := "picoclaw"
+	if runtime.GOOS == "windows" {
+		binaryName = "picoclaw.exe"
+	}
+
+	// 1. Explicit override via environment variable
+	if p := os.Getenv("PICOCLAW_BINARY"); p != "" {
+		if info, _ := os.Stat(p); info != nil && !info.IsDir() {
+			return p
 		}
+	}
+
+	// 2. Same directory as the launcher executable
+	if exe, err := os.Executable(); err == nil {
+		candidate := filepath.Join(filepath.Dir(exe), binaryName)
 		if info, err := os.Stat(candidate); err == nil && !info.IsDir() {
 			return candidate
 		}
 	}
+
+	// 3. Fall back to PATH lookup
 	return "picoclaw"
 }
 
--- a/pkg/providers/factory.go
+++ b/pkg/providers/factory.go
@@ -40,6 +40,10 @@ func resolveProviderSelection(cfg *config.Config) (providerSelection, error) {
 	providerName := strings.ToLower(cfg.Agents.Defaults.Provider)
 	lowerModel := strings.ToLower(model)
 
+	if providerName == "" && model == "" {
+		return providerSelection{}, fmt.Errorf("no model configured: agents.defaults.model is empty")
+	}
+
 	sel := providerSelection{
 		providerType: providerTypeHTTPCompat,
 		model:        model,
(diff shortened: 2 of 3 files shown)


web/backend/api/gateway.go
```diff
@@ -133,7 +133,11 @@ func (h *Handler) startGatewayLocked() (int, error) {
 	// Locate the picoclaw executable
 	execPath := findPicoclawBinary()
 
-	cmd := exec.Command(execPath, "gateway")
+	args := []string{"gateway"}
+	if h.configPath != "" {
+		args = append(args, "--config", h.configPath)
```
wj-xiao: `picoclaw gateway` does not support a `--config` flag today, so this change will make launcher-driven gateway startup fail immediately.
The subcommand only defines `--debug` and `--no-truncate`, and still uses `cobra.NoArgs`, so passing `--config <path>` causes the child process to exit with unknown flag: `--config`.

Since the launcher always has a non-empty config path, this affects both auto-start and manual start. If the goal is to make the gateway use the launcher’s config file, a safer approach would be to set `PICOCLAW_CONFIG` in the child process environment before `cmd.Start()`, because the CLI already resolves its config path from that environment variable.

web/backend/api/gateway.go
```diff
@@ -530,18 +534,50 @@ func (h *Handler) currentGatewayStatus() string {
...
-			candidate += ".exe"
-		}
+
+		// 2. Same directory as the launcher executable
+		candidate := filepath.Join(dir, binaryName)
 		if info, err := os.Stat(candidate); err == nil && !info.IsDir() {
 			return candidate
 		}
+
+		// 3. Relative paths — covers running from web/backend/dist/
```
wj-xiao: The relative-path fallback here looks unnecessary. In our actual build layout, both `picoclaw-launcher` and `picoclaw` are emitted into the same `build/` directory, so the previous “same directory as the current executable” check should already cover the production case.
For typical development flows, this `../../build /` `../build` heuristic is also fairly brittle because `os.Executable()` often points to a temporary Go build location rather than `web/backend/dist/`.

Given that this PR already adds `PICOCLAW_BINARY` as an explicit override, I think that plus same-directory lookup and the final $PATH fallback is the robust part; this extra relative-path probing seems speculative and may not be worth keeping unless there is a concrete packaging layout that depends on it.

web/backend/api/gateway.go
snac21: Updated it as requested.Could you please review it again?

Orgmar:
@snac21 Nice work tackling three issues in one PR. Using PICOCLAW_BINARY as an explicit override and removing the fragile relative path probing is much more robust. The clear error message for empty model config is also a good UX improvement.

We have a **PicoClaw Dev Group** on Discord for contributors. Want to join? Send an email to `support@sipeed.com` with the subject `[Join PicoClaw Dev Group] + your GitHub account` and we'll get you the invite link!

PR #1342, opened 2026-03-11
feat(provider): Enhance OpenAI Provider Robustness and Feature Parity

tong3jie:
## 📝 Description

This PR implements a significant refactor of the openai_compat provider to improve production readiness, memory safety, and compatibility across diverse LLM vendors.

1. Resilience: Automatic Retry & Backoff

- Integrated hashicorp/go-retryablehttp to replace the standard http.Client.

- Implemented Exponential Backoff to automatically handle transient failures such as 429 Too Many Requests and 5xx Server Errors.

- Added WithRetry Functional Option to allow granular control over MaxRetries, MinWait, and MaxWait timings.

2. Security & Defensive Programming

- Explicit Context Validation: Added ctx.Done() checks before processing large payloads to ensure canceled or timed-out requests release resources immediately.

- OOM Prevention: Introduced io.LimitReader for response bodies (capped at 10MB) to protect the service from memory exhaustion if a provider returns unexpected massive payloads.

- HTML Error Detection: Enhanced the looksLikeHTML logic to provide clearer error messages when API gateways or proxies return HTML error pages instead of valid JSON.

3. Protocol & Vendor Adaptability

- Dynamic Token Field Resolution: Refactored resolveMaxTokenField to automatically toggle between max_tokens and max_completion_tokens based on the model family (o1, gpt-5, GLM, etc.).

- Optimized Model Normalization: Implemented a prefix-based lookup table (vendorPrefixes) for normalizeModel, providing a more scalable way to strip provider prefixes without breaking routing.

4. Code Quality & Architectural Improvements

- Improved Functional Options pattern for better extensibility.

- Optimized HTTP Transport initialization with proper proxy support.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

- **Reference URL:**
- **Reasoning:**

## 🧪 Test Environment
- **Hardware:** x86
- **OS:** macos, linux

## 📸 Evidence (Optional)
<details>

</details>

## ☑️ Checklist


tong3jie:
I also think it's best to remove this hard-coded approach and adopt a configurable one to enhance flexibility.  

I will fix it right now.

yinwm (changes requested):
## Pre-Landing Review Summary

Thanks for this comprehensive enhancement! The retry mechanism and OOM protection are valuable additions. I found a few issues that should be addressed before merging.

---

### 🔴 Critical

#### 1. Missing "novita" vendor prefix

**File:** `pkg/providers/openai_compat/provider.go:30-44`

The original code's switch statement included `"novita"` in the vendor prefix list, but the new `vendorPrefixes` map is missing it:

```go
// Original code
case "litellm", "moonshot", ..., "minimax", "novita":

// New code
var vendorPrefixes = map[string]struct{}{
    ...
    "minimax":    {},
    // "novita" is missing!
}
```

This will cause `novita/xxx` model names to not be properly normalized, breaking routing for Novita providers.

**Fix:** Add `"novita": {},` to the `vendorPrefixes` map.

---

### 🟡 Informational

#### 2. Ineffective ctx.Done() check position

**File:** `pkg/providers/openai_compat/provider.go:219-226`

Checking `ctx.Done()` after `resp.Body.Close()` doesn't serve a meaningful purpose — the request has already been sent and the response received. If context cancellation support is needed, it should be checked before the request or during response body reading.

```go
defer resp.Body.Close()

select {
case <-ctx.Done():
    return nil, ctx.Err()  // Too late — resources already consumed
default:
}
```

**Suggestion:** Either remove this code block, or move it before the `Do()` call. Note that `retryablehttp` already supports context cancellation natively, so manual checking may be unnecessary.

#### 3. Redundant Timeout setting

**File:** `pkg/providers/openai_compat/provider.go:91-92`

The timeout is set twice on the same underlying object:

```go
retryClient.HTTPClient.Timeout = defaultRequestTimeout
// ...
p.httpClient.Timeout = defaultRequestTimeout  // Same object, redundant
```

Since `retryClient.StandardClient()` returns a wrapper around the same `HTTPClient`, setting the timeout twice is confusing.

**Suggestion:** Remove the `p.httpClient.Timeout` line and keep only one timeout assignment.

---

### 📝 Test Naming

#### 4. Test name/intent mismatch

**File:** `pkg/providers/openai_compat/provider_test.go:14-53`

`TestProviderChat_MaxTokensFieldNull` tests an empty string (`""`), not null. Also, the test uses a GLM model (`glm-4.7`) but expects `max_tokens` — this contradicts the PR description which states GLM should use `max_completion_tokens`.

**Suggestion:** 
- Rename to `TestProviderChat_EmptyMaxTokensFieldUsesDefault`
- Use a non-GLM model (e.g., `gpt-4o`) to test the default `max_tokens` behavior

---

**Recommendation:** Please fix the critical issue (#1) before merging. Issues #2-4 are code quality improvements that would be good to address together.

yinwm (changes requested):
## Pre-Landing Review — PR #1342

Overall direction is solid. The retry mechanism at the HTTP transport layer, OOM protection, config-driven `max_tokens_field`, and vendor prefix map are all good improvements. Found **2 blocking issues** that need fixing before merge, plus a few suggestions.

---

### 🔴 CRITICAL (must fix)

**C1: HTTP Transport initialization regression**

```go
// PR code (provider.go NewProvider)
transport := &http.Transport{}
```

This creates a bare `*http.Transport` without TLS config, HTTP/2 support, or keep-alive settings. The original `common.NewHTTPClient` uses `http.DefaultTransport.(*http.Transport).Clone()`, which preserves all system defaults. This can cause:
- TLS handshake failures for endpoints requiring specific TLS versions/ciphers
- Loss of HTTP/2 multiplexing (performance regression)
- Different connection pooling behavior

**Fix**: Use `http.DefaultTransport.(*http.Transport).Clone()` then set `Proxy` on the clone.

---

**C2: ChatStream should bypass retryablehttp**

```go
// Current code (provider.go ChatStream)
streamClient := &http.Client{Transport: p.httpClient.Transport}
```

Now that `p.httpClient.Transport` is `*retryablehttp.RoundTripper`, SSE streaming requests go through retry logic. This is problematic:
- Retrying a partially consumed stream causes server-side state inconsistency
- Retry wait times stack on top of streaming latency, degrading UX
- Some providers' streaming endpoints are not idempotent

**Fix**: ChatStream should use the underlying raw `*http.Transport` directly, or create a separate non-retry client for streaming.

---

### 🟡 Suggestions

**S1: 10MB hard limit may truncate legitimate responses**

`maxResponseSize = 10 * 1024 * 1024` is applied to all success-path responses. Models returning very long outputs (extensive tool calls, long-form generation) could exceed this. Consider raising the limit (e.g., 50MB) or only applying it when Content-Type is not JSON.

---

**S2: Retry logging disabled**

```go
retryClient.Logger = nil
```

Production environments benefit from retry event observability. With logging disabled, 429 retries and 5xx backoff are invisible. Consider using `log.Default()` or a custom logger.

---

**S3: `WithRetry(0, ...)` semantics are confusing**

`maxRetries=0` means "don't override" (keeps default 3), while `maxRetries=1` means "retry once". The zero-value-as-noop pattern is error-prone. Consider using `-1` for "don't override", or a pointer type `*int`.

---

**S4: `LinearJitterBackoff` vs `ExponentialBackoff`**

Linear jitter may not back off aggressively enough under sustained failures. The default exponential backoff is generally recommended for production retry policies.

---

### ✅ What's done well

- Config-driven `max_tokens_field` — clean, no more hardcoded model name matching
- `vendorPrefixes` map — more maintainable than switch/case
- OOM protection with `io.LimitReader`
- HTML error detection on 200 OK responses (covers misbehaving API gateways)
- 114 lines of new tests covering retry, default fields, and error handling

---

**Summary**: Fix C1 (Transport) and C2 (Stream retry bypass) and this is ready to go. The suggestions are non-blocking improvements.

PR #1356, opened 2026-03-11
fix: spawn tool now respects target agent's model config

### LINKED ISSUE #1322: [BUG] spawn ignores target agent model and always uses caller model
## Quick Summary
When `spawn` is called with `agent_id`, the spawned subagent currently runs with the caller agent's model instead of the target agent's configured model.

## Environment
- PicoClaw: current main (observed on Docker + Telegram)
- Config: multiple agents in `agents.list`, e.g. `main` -> `economico`, `analista` -> `premium`

## Steps to Reproduce
1. Configure two agents with different models in `agents.list`.
2. Allow spawn from `main` to `analista` using `subagents.allow_agents`.
3. Ask main agent to run `spawn` with `agent_id=analista`.
4. Check logs / behavior.

## Actual Behavior
Spawned task runs with the manager default model (from caller), not the target agent model.
In our case this caused errors like model alias mismatch in subagent path.

## Expected Behavior
When `agent_id` is provided, spawn should resolve and use that target agent's model (with fallback to caller/default only if unresolved).

## Notes
I have a fix prepared that adds a resolver from target `agent_id` -> target model and applies it in subagent tool loop execution.

treepudding:
## Summary
Fixes #1322

Previously, when `spawn` tool was called with `agent_id` parameter, the subagent would use the caller agent's model instead of the target agent's configured model.

## Changes
- Add `modelResolver` field to `SubagentManager`
- Add `SetModelResolver()` method to inject model resolution function
- In `runTask()`, use target agent's model when `agent_id` is specified
- Add test case for model resolver functionality

## Test Plan
- [x] All existing tests pass (`go test ./pkg/tools/... ./pkg/agent/...`)
- [x] New `TestSubagentManager_ModelResolver` test passes
- [x] Build succeeds (`go build ./...`)

### DIFF
--- a/pkg/agent/loop.go
+++ b/pkg/agent/loop.go
@@ -224,6 +224,13 @@ func registerSharedTools(
 			if cfg.Tools.IsToolEnabled("subagent") {
 				subagentManager := tools.NewSubagentManager(provider, agent.Model, agent.Workspace)
 				subagentManager.SetLLMOptions(agent.MaxTokens, agent.Temperature)
+				// Set model resolver so spawn can use target agent's model
+				subagentManager.SetModelResolver(func(targetAgentID string) string {
+					if targetAgent, ok := registry.GetAgent(targetAgentID); ok {
+						return targetAgent.Model
+					}
+					return ""
+				})
 				spawnTool := tools.NewSpawnTool(subagentManager)
 				currentAgentID := agentID
 				spawnTool.SetAllowlistChecker(func(targetAgentID string) bool {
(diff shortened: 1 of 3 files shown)


putueddy:
Cherry-picked cleanly. The fix ensures subagents spawned with `agent_id` use the target agent's configured model instead of inheriting the caller's. All tests pass including the new `TestSubagentManager_ModelResolver`. Tested on Pi 4, ARM64, v0.2.1.

treepudding:
@alexhoshina Hi,this PR is waiting for CI approval. Could you please approve the workflow run? Thanks!

PR #1385, opened 2026-03-11
fix(web): restore the last active chat session

### LINKED ISSUE #1373: [BUG]在web对话页面，每次点击对话项目都是默认显示新的对话。希望修改为默认显示上次的对话内容。
## Quick Summary

在web对话页面，每次点击对话项目都是默认显示新的对话。希望修改为默认显示上次的对话内容。

##  Environment & Tools
- **PicoClaw Version:** (v0.2.2)
- **AI Model & Provider:** (modelscope.cn)
- **Operating System:** (Android Termux)
- **Channels:** ( Feishu.)

Alix-007:
Fixes #1373.

## Summary
- persist the last active chat session id in local storage
- hydrate that session history when the chat page opens again
- keep manual session switching on the same loading path

## Testing
- pnpm exec tsc --noEmit
- pnpm exec eslint src/hooks/use-pico-chat.ts

### DIFF
--- a/web/frontend/src/hooks/use-pico-chat.ts
+++ b/web/frontend/src/hooks/use-pico-chat.ts
@@ -1,6 +1,12 @@
 import dayjs from "dayjs"
 import { useAtomValue } from "jotai"
-import { useCallback, useEffect, useRef, useState } from "react"
+import {
+  type SetStateAction,
+  useCallback,
+  useEffect,
+  useRef,
+  useState,
+} from "react"
 import { useTranslation } from "react-i18next"
 import { toast } from "sonner"
 
@@ -131,7 +137,27 @@ export function usePicoChat() {
   const isConnectingRef = useRef(false)
   const msgIdCounter = useRef(0)
   const activeSessionIdRef = useRef(activeSessionId)
-  const hydratedInitialSessionRef = useRef(false)
+  const messagesRevisionRef = useRef(0)
+
+  const setTrackedMessages = useCallback(
+    (nextState: SetStateAction<ChatMessage[]>) => {
+      setMessages((prev) => {
+        const next =
+          typeof nextState === "function"
+            ? (
+                nextState as (prevState: ChatMessage[]) => ChatMessage[]
+              )(prev)
+            : nextState
+
+        if (next !== prev) {
+          messagesRevisionRef.current += 1
+        }
+
+        return next
+      })
+    },
+    [],
+  )
 
   // Keep ref in sync
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


web/frontend/src/hooks/use-pico-chat.ts
```diff
@@ -109,18 +125,67 @@ export function usePicoChat() {
...
+
+    return detail.messages.map((m, i) => ({
+      id: `hist-${i}-${Date.now()}`,
+      role: m.role as "user" | "assistant",
+      content: m.content,
+      timestamp: fallbackTime,
+    }))
+  }, [])
+
+  useEffect(() => {
```
wj-xiao, line 180: The hydration effect is not safe under React StrictMode. In development, React runs the effect setup/cleanup/setup sequence on initial mount. The first pass sets the hydratedInitialSessionRef flag and starts the restore request, the cleanup marks it cancelled, and the second pass exits early because the flag is already set. That means the restored messages never get applied, so the feature appears broken locally.

web/frontend/src/hooks/use-pico-chat.ts
```diff
@@ -325,14 +390,7 @@ export function usePicoChat() {
...
       try {
-        const detail = await getSessionHistory(sessionId)
-        const fallbackTime = detail.updated
-        const historyMessages = detail.messages.map((m, i) => ({
-          id: `hist-${i}-${Date.now()}`,
-          role: m.role as "user" | "assistant",
-          content: m.content,
-          timestamp: fallbackTime,
-        }))
+        const historyMessages = await loadSessionMessages(sessionId)
```
wj-xiao, line 427: The initial restore request is not guarded against later state changes. While loadSessionMessages(storedSessionId) is in flight, the websocket can already reconnect and receive new messages, or the user can switch to another session. When the restore promise resolves, it unconditionally calls setMessages(historyMessages), which can overwrite newer live messages or leave messages out of sync with activeSessionId.

Alix-007: Adjusted in 16ce1a8. The initial restore path now captures a message revision before requesting history and skips applying the result if the user switched sessions or any newer messages landed while the request was in flight, so it no longer overwrites live state.

web/frontend/src/hooks/use-pico-chat.ts
Alix-007: Adjusted in 16ce1a8. I removed the one-shot hydration gate so the initial restore effect remains safe under React StrictMode, and the restore now only applies if the session and local message revision are still unchanged when the request resolves.

PR #1391, opened 2026-03-12
fix(config): start model round robin from the first match

### LINKED ISSUE #1153: [BUG]Duplicate model_name entries don't fall back correctly — always resolves to second entry, ignores first and third

Alix-007:
## Summary
- start duplicate `model_name` round-robin selection from the first matching entry instead of the second
- make the regression test verify the exact selection order across repeated calls
- preserve the existing load-balancing behavior after the first pick

## Testing
- `go test ./pkg/config -run ''TestGetModelConfig_(RoundRobin|Concurrent|Found|NotFound|EmptyList)''`

Fixes #1153

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -921,7 +921,7 @@ func (c *Config) GetModelConfig(modelName string) (*ModelConfig, error) {
 	}
 
 	// Multiple configs - use round-robin for load balancing
-	idx := rrCounter.Add(1) % uint64(len(matches))
+	idx := (rrCounter.Add(1) - 1) % uint64(len(matches))
 	return &matches[idx], nil
 }
 
--- a/pkg/config/model_config_test.go
+++ b/pkg/config/model_config_test.go
@@ -54,6 +54,8 @@ func TestGetModelConfig_EmptyList(t *testing.T) {
 }
 
 func TestGetModelConfig_RoundRobin(t *testing.T) {
+	rrCounter.Store(0)
+
 	cfg := &Config{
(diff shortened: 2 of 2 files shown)


Alix-007:
Superseded by #1702, which rebases the same round-robin fix onto current main and includes deterministic first-match regression coverage.

PR #1402, opened 2026-03-12
fix(cron): publish deliver=false agent replies

### LINKED ISSUE #1058: [BUG] Cron deliver=false tasks silently discard LLM response — messages never reach channel
## Description

When a cron job has `deliver=false` (the default for agent-processed tasks),
the LLM processes the job message successfully but the response is silently
discarded. The user never receives any output in their Discord channel.

## Root Cause

In `pkg/tools/cron.go`, `ExecuteJob` calls `ProcessDirectWithChannel` for
`deliver=false` tasks. The response is returned but immediately discarded:

```go
response, err := t.executor.ProcessDirectWithChannel(...)
return "ok"
```

The comment claims "Will be sent by AgentLoop", but this is incorrect:
- `ProcessDirectWithChannel` calls `processMessage` → `runAgentLoop`
- `runAgentLoop` uses `SendResponse: false`, so it does NOT publish the response
- The normal `Run()` inbound loop would publish after `processMessage` returns,
  but `ProcessDirectWithChannel` bypasses `Run()` entirely

Result: the LLM response is computed, returned to `ExecuteJob`, and then thrown away.

## Impact

All cron tasks using `deliver=false` (agent-processed tasks like "remind me
every 2 hours to check X") produce results that never reach the user. Only
`deliver=true` (direct message) and `command` (shell execution) paths work.

## Environment

- PicoClaw: main branch
- Channels: Discord

Alix-007:
Fixes #1058.

- publish agent responses for deliver=false cron jobs
- keep empty/whitespace responses silent
- add ExecuteJob regression tests

### DIFF
--- a/pkg/tools/cron.go
+++ b/pkg/tools/cron.go
@@ -341,7 +341,14 @@ func (t *CronTool) ExecuteJob(ctx context.Context, job *cron.CronJob) string {
 		return fmt.Sprintf("Error: %v", err)
 	}
 
-	// Response is automatically sent via MessageBus by AgentLoop
-	_ = response // Will be sent by AgentLoop
+	if strings.TrimSpace(response) != "" {
+		pubCtx, pubCancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer pubCancel()
+		t.msgBus.PublishOutbound(pubCtx, bus.OutboundMessage{
+			Channel: channel,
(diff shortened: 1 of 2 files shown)


Alix-007:
Superseded by #1699, which rebases the same fix onto current main and includes the same deliver=false cron reply regression coverage.

PR #1440, opened 2026-03-12
fix(irc): accept string auto-join channel config

Alix-007:
## Summary


Alix-007:
Follow-up update: I pushed a forward-only sync commit that preserves the original IRC auto-join fix while rebasing the semantics onto the current config parser. Local validation on this branch: `go test ./pkg/channels/irc -count=1` and `go test ./pkg/config -run TestFlexibleStringSliceUnmarshalJSON_String -count=1` both pass.

Alix-007:
Closing this to keep the queue focused on higher-value fixes.

I re-ranked the remaining open PRs by impact and maintenance cost, and this one falls into the low-value bucket for now:
- relatively niche IRC config compatibility fix
- branch is conflicting with `main`
- no maintainer review context to preserve

If this bug becomes urgent again, it is better to reopen it from a fresh branch on current `main` than to keep this stale PR in the active queue.

PR #1441, opened 2026-03-12
fix(skill): tighten weather location matching guidance

Alix-007:
## Summary


Orgmar:
@Alix-007 Thanks for tightening up the weather location matching logic, small refinements like this make the skill more reliable.

By the way, we've got a **PicoClaw Dev Group** on Discord for contributors to chat and collaborate. Interested? Just drop an email to `support@sipeed.com` with the subject `[Join PicoClaw Dev Group] + your GitHub account` and we'll send you the invite!

PR #1453, opened 2026-03-13
fix(provider): classify transport send failures for fallback

### LINKED ISSUE #1419: Fallback chain aborts on retriable OpenRouter transport resets
## Summary

`picoclaw agent` can fail against OpenRouter with transport errors like `connection reset by peer`, but the fallback chain treats them as unclassified and aborts before trying configured fallback models.

In the same environment, direct `curl` requests to OpenRouter succeed, which points to brittle transport error handling in the PicoClaw request path rather than a general connectivity failure.

## Reproduction

Observed on an Orange Pi Zero board running the `ops` profile against OpenRouter `stepfun/step-3.5-flash`.

Service logs show errors like:

```text
LLM call failed after retries: fallback: unclassified error from openrouter/stepfun/step-3.5-flash:
failed to send request: Post "https://openrouter.ai/api/v1/chat/completions":
read tcp ...: read: connection reset by peer
```

At the same time:

- `curl -I https://openrouter.ai/api/v1/models` succeeds over IPv4 and IPv6
- authenticated direct `curl` requests to `/chat/completions` succeed repeatedly
- `picoclaw agent -m ... --model openrouter-stepflash` intermittently fails with the transport reset above

## Problems

1. `ClassifyError()` does not treat common send-request transport failures as retriable, so they become `fallback: unclassified error ...` and stop the fallback chain.
2. `FallbackChain.Execute()` is invoked from the agent loop with `agent.Provider.Chat(...)` for every candidate, so cross-provider fallback candidates do not get their own provider/base URL instantiated from `model_list`.

## Expected

- Transport-level send failures like `connection reset by peer`, `connection refused`, `no route to host`, `unexpected EOF`, etc. should be classified as retriable.
- Fallback execution should resolve and instantiate the provider for each candidate model entry, not reuse only the active provider instance.

## Proposed fix

- Extend the error classifier to mark common transport-send failures as `FailoverTimeout` / retriable.
- Resolve a provider from the candidate model reference inside the fallback execution path so fallback candidates actually use their configured provider settings.

Alix-007:
## Summary
- classify common transport send/request failures as retriable timeout-style fallback errors
- cover connection reset, connection refused, no route to host, unexpected EOF, and TLS handshake timeout
- add focused classifier regression tests for these transport patterns

Closes #1419.

## Testing
- go test ./pkg/providers -run TestClassifyError_TransportSendPatterns -count=1

### DIFF
--- a/pkg/providers/error_classifier.go
+++ b/pkg/providers/error_classifier.go
@@ -48,6 +48,12 @@ var (
 		substr("timed out"),
 		substr("deadline exceeded"),
 		substr("context deadline exceeded"),
+		substr("failed to send request"),
+		substr("connection reset by peer"),
+		substr("connection refused"),
+		substr("no route to host"),
+		substr("unexpected eof"),
+		substr("tls handshake timeout"),
 	}
 
 	billingPatterns = []errorPattern{
--- a/pkg/providers/error_classifier_test.go
+++ b/pkg/providers/error_classifier_test.go
@@ -154,6 +154,28 @@ func TestClassifyError_TimeoutPatterns(t *testing.T) {
 	}
 }
 
+func TestClassifyError_TransportSendPatterns(t *testing.T) {
+	patterns := []string{
+		`failed to send request: Post "https://openrouter.ai/api/v1/chat/completions": read tcp 127.0.0.1:1->127.0.0.1:2: read: connection reset by peer`,
+		`failed to send request: dial tcp 127.0.0.1:443: connect: connection refused`,
+		`failed to send request: dial tcp 127.0.0.1:443: connect: no route to host`,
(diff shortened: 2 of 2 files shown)


Alix-007:
Closing this to avoid keeping another duplicate branch in the queue.

After re-checking issue #1419 and the current PR queue, this branch is not the best line to keep active:
- there are already overlapping open PRs on the same transport-classification topic, especially #1465 and #1430
- this PR only addresses the classifier half of #1419, while the issue description also calls out candidate-provider resolution behavior
- with multiple near-identical open PRs, keeping this one open adds queue noise instead of helping review throughput

If the maintainers later prefer this exact patch shape, it can be reopened, but for now I think reducing duplicate PRs is the better move.

PR #1455, opened 2026-03-13
fix(telegram): reconnect long polling after disconnect

### LINKED ISSUE #216: Telegram gateway can stop responding after long-poll updates channel closes
## Problem
In Telegram polling mode, when the updates channel closes unexpectedly, the bot logs a reconnect message but does not actually restart polling. After that point, Telegram appears connected but no longer receives new updates.

## Affected behavior
- Gateway starts normally
- Telegram channel shows connected
- After channel closure/network hiccup, bot stops responding to user messages until process restart

## Reproduction
1. Start gateway with Telegram enabled.
2. Trigger an updates-channel interruption (temporary network issue, Telegram polling interruption, etc.).
3. Observe log like: `Updates channel closed, reconnecting...`.
4. Send Telegram message to bot.
5. No inbound processing occurs.

## Root cause
When `updates` channel closes (`ok == false`), handler exits instead of creating a new long-poll stream.

## Expected behavior
- Initial startup should fail fast if polling cannot start.
- After successful startup, polling should auto-reconnect with retry/backoff when disconnected.

## Proposed fix
A fix is implemented in PR #215:
- add reconnect loop after runtime disconnect
- keep fail-fast behavior on first connect
- add privacy mode warning for group-chat expectations

PR: https://github.com/sipeed/picoclaw/pull/215

Alix-007:
## Summary
- restart Telegram long polling when the handler or updates stream stops unexpectedly
- keep the initial startup behavior fail-fast, but add a reconnect loop for runtime disconnects
- add a regression test covering handler restart after disconnect

Fixes #216

## Test plan
- go test ./pkg/channels/telegram -run TestRunPollingLoop_RestartsAfterHandlerStops -count=1

### DIFF
--- a/pkg/channels/telegram/telegram.go
+++ b/pkg/channels/telegram/telegram.go
@@ -58,7 +58,7 @@ type TelegramChannel struct {
 
 type telegramBotHandler interface {
 	Start() error
-	StopWithContext(context.Context) error
+	StopWithContext(ctx context.Context) error
 }
 
 func NewTelegramChannel(cfg *config.Config, bus *bus.MessageBus) (*TelegramChannel, error) {


mymmrac:
@Alix-007 I don't think this functionality is needed as long polling will stop only if context is closed, but in that case we don't need to restart it, so why this functionality is needed?

Alix-007:
The reconnect path here is only for runtime disconnects where the updates stream stops while the channel context is still alive. Issue #216 documents that case: the gateway stays marked connected, logs a reconnect message, and then stops receiving Telegram updates until process restart.

So this does not restart after normal shutdown/cancel; runPollingLoop exits immediately once ctx.Err() != nil. It only re-establishes polling when bh.Start() returns with a live context after the updates channel has closed unexpectedly.

I also pushed a small lint-only follow-up and re-ran:
- go test ./pkg/channels/telegram -run TestRunPollingLoop_RestartsAfterHandlerStops -count=1

Alix-007:
Closing this for now.

After re-checking the current implementation and the upstream `telego` behavior, I do not have enough repository-specific evidence to keep this PR open:
- picoclaw uses `github.com/mymmrac/telego v1.7.0`, not `go-telegram-bot-api`
- `telego`'s `UpdatesViaLongPolling` already retries `GetUpdates` errors internally by default, so ordinary polling/network hiccups do not appear to close the updates channel in the way this PR assumes
- the current regression test exercises a forced handler/channel-stop path, but it does not demonstrate a real `telego` failure mode seen in picoclaw itself

Given that, and given the PR is now conflicting with `main`, I think the right move is to close instead of continuing to maintain a fix without a solid reproducer.

If we get a telego-specific reproducer or logs showing the updates stream really exits while the channel context is still alive, I can reopen this as a narrower follow-up.

PR #1457, opened 2026-03-13
fix(agent): retry transient provider failures

### LINKED ISSUE #629: [BUG] Didn't retry if meet LLM call failed
##  Environment & Tools
- **PicoClaw Version:** v0.1.2
- **AI Model & Provider:** (OpenRouter)
- **Operating System:** (Ubuntu 22.04)
- **Channels:** (Discord)

## 📸 Steps to Reproduce
1. Run a long task
2. Server may return HTTP 500 sometimes
3. Task hang without retry, logs output LLM call failed 

## ❌ Actual Behavior
Task dropped

## ✅ Expected Behavior
Retry and continue

Alix-007:
## Summary
- retry transient single-provider LLM failures inside the agent loop instead of aborting immediately
- reuse provider error classification so HTTP 500/502/503 style failures are treated like temporary outages
- add a regression test proving the agent recovers after one transient provider failure

Fixes #629

## Test plan
- go test ./pkg/agent -run TestAgentLoop_ContextExhaustionRetry -count=1
- go test ./pkg/agent -run TestAgentLoop_TransientProviderRetry -count=1

### DIFF
--- a/pkg/agent/loop.go
+++ b/pkg/agent/loop.go
@@ -1152,15 +1152,17 @@ func (al *AgentLoop) runLLMIteration(
 			isTransientProviderError := classifiedErr != nil && classifiedErr.Reason == providers.FailoverTimeout
 
 			// Detect real context window / token limit errors, excluding network timeouts.
-			isContextError := !isTimeoutError && !isTransientProviderError && (strings.Contains(errMsg, "context_length_exceeded") ||
-				strings.Contains(errMsg, "context window") ||
-				strings.Contains(errMsg, "maximum context length") ||
-				strings.Contains(errMsg, "token limit") ||
-				strings.Contains(errMsg, "too many tokens") ||
-				strings.Contains(errMsg, "max_tokens") ||
-				strings.Contains(errMsg, "invalidparameter") ||
-				strings.Contains(errMsg, "prompt is too long") ||
-				strings.Contains(errMsg, "request too large"))
+			isContextError := !isTimeoutError &&
+				!isTransientProviderError &&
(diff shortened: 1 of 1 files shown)


Alix-007:
Pushed an updated branch for this PR. It keeps the transient single-provider retry fix, resolves the current pkg/agent/loop drift against main, and keeps the targeted regression in place.

Local verification:
- go test ./pkg/agent -run TestAgentLoop_ -count=1

Alix-007:
Pushed a follow-up lint-only fix on top of the retry update to satisfy golines in pkg/agent/loop.go and reran go test ./pkg/agent -run TestAgentLoop_ -count=1 locally.

Alix-007:
Refreshed this branch onto the current `main` and force-updated the PR head.

During the refresh I explicitly dropped an old conflict-resolution commit that had accumulated unrelated registry/provider reload changes, so the PR is back to the intended narrow scope: transient retry handling for single-provider calls plus the focused regression test.

Local verification after the refresh:
- `go test ./pkg/agent -run 'TestAgentLoop_' -count=1`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.10.1 run ./pkg/agent/...`

GitHub PR checks are green again.

Alix-007:
Closing this to reduce queue noise. The branch is conflicting again, the issue lane already has another open carrier (#866), and this PR has aged without maintainer review despite several newer focused fixes being merged first. If we revisit this bug, it should come back as a fresh clean-base follow-up rather than continue on this stale branch.

PR #1484, opened 2026-03-13
fix(line): add request body size limit to webhook handler

### LINKED ISSUE #1407: [BUG] LINE webhook handler reads unlimited bodies (DoS risk)
## Quick Summary

A new unauthenticated DoS vector was introduced with the LINE channel: unbounded request body reads allow remote memory exhaustion on the webhook server.

The LINE webhook handler uses io.ReadAll on the request body without applying a MaxBytesReader or any explicit size limit before verifying the signature or parsing JSON. Because the LINE webhook server is bound to a configurable host/port (default 0.0.0.0 in the new config) and must be reachable from the internet, an unauthenticated attacker can send oversized POSTs to the webhook endpoint and force the process to allocate large amounts of memory, leading to denial-of-service or crashes.

## Evidence
Webhook handler reads the entire request body via io.ReadAll without any size limit or MaxBytesReader, enabling remote memory exhaustion before signature verification.
[code](https://github.com/alexhoshina/picoclaw/blob/f294a71bc5771bf896fd13c47e651692d4925a5c/pkg/channels/line.go#L176-L207)

suboss87:
## Summary

- Add `io.LimitReader` with 4 MB cap to LINE webhook handler to prevent unauthenticated DoS via unbounded request body reads
- Reject oversized payloads with `413 Request Entity Too Large` before signature verification or JSON parsing
- Follows the same pattern already used by the WeCom channel (`pkg/channels/wecom/aibot.go`)

## Problem

The LINE webhook handler uses `io.ReadAll(r.Body)` without any size limit. Since the webhook endpoint must be internet-reachable, an attacker can send arbitrarily large POST bodies to exhaust memory and crash the process.

## Fix

```go
const maxBodySize = 4 << 20 // 4 MB
body, err := io.ReadAll(io.LimitReader(r.Body, maxBodySize+1))
// ... error handling ...
if len(body) > maxBodySize {
    http.Error(w, "Request body too large", http.StatusRequestEntityTooLarge)
    return
}
```

The `maxBodySize+1` trick (read one byte more than the limit) allows distinguishing between a body that fits exactly vs. one that exceeds the limit, without silently truncating valid payloads.

## Test plan

- [ ] Verify normal LINE webhook events (< 4 MB) are processed correctly
- [ ] Verify oversized POST bodies return 413 status
- [ ] Verify `go vet` and `make lint` pass

Fixes #1407

### DIFF
--- a/pkg/channels/line/line.go
+++ b/pkg/channels/line/line.go
@@ -166,12 +166,18 @@ func (c *LINEChannel) webhookHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	body, err := io.ReadAll(r.Body)
+	// Limit request body to 4 MB to prevent memory exhaustion (DoS).
+	const maxBodySize = 4 << 20 // 4 MB
+	body, err := io.ReadAll(io.LimitReader(r.Body, maxBodySize+1))
 	if err != nil {
 		logger.ErrorCF("line", "Failed to read request body", map[string]any{
 			"error": err.Error(),
 		})
 		http.Error(w, "Bad request", http.StatusBadRequest)
+		return
+	}
+	if len(body) > maxBodySize {
+		http.Error(w, "Request body too large", http.StatusRequestEntityTooLarge)
 		return
 	}
 


suboss87:
Closing — the LINE webhook body size limit was already fixed upstream (maxWebhookBodySize = 1 MiB at package level). My change is now redundant.

PR #1510, opened 2026-03-13
Improve white-space and general rendering of CommonMark in Matrix

dim:
## 📝 Description

This is a follow up to my previous PR #1370. Some Matrix clients, most notably FluffyChat, struggle with the amount of white-space added during the HTML conversion. This patch improves the output considerably. 

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue

N/A

- **Reference URL:**
- **Reasoning:** Matrix clients apply default browser paragraph margins to <p> elements, which stack with the margins of surrounding block elements (lists, code blocks, headings) and produce excessive vertical spacing.

## 🧪 Test Environment
- **Hardware:** Raspberry Pi 5
- **OS:** Debian 13
- **Model/Provider:** Gemini
- **Channels:** Matrix

## 📸 Evidence (Optional)
<details>

image

</details>

## ☑️ Checklist

### DIFF
--- a/pkg/channels/matrix/matrix.go
+++ b/pkg/channels/matrix/matrix.go
@@ -273,8 +273,9 @@ func (c *MatrixChannel) Stop(ctx context.Context) error {
 }
 
 func markdownToHTML(md string) string {
-	p := parser.NewWithExtensions(parser.CommonExtensions | parser.AutoHeadingIDs)
-	renderer := mdhtml.NewRenderer(mdhtml.RendererOptions{Flags: mdhtml.CommonFlags})
+	extensions := (parser.CommonExtensions | parser.NoEmptyLineBeforeBlock) &^ parser.DefinitionLists
+	p := parser.NewWithExtensions(extensions)
+	renderer := mdhtml.NewRenderer(mdhtml.RendererOptions{Flags: mdhtml.UseXHTML})
 	return strings.TrimSpace(string(markdown.ToHTML([]byte(md), p, renderer)))
 }
 
--- a/pkg/channels/matrix/matrix_test.go
+++ b/pkg/channels/matrix/matrix_test.go
@@ -340,23 +340,96 @@ func TestMatrixOutboundContent(t *testing.T) {
 }
 
 func TestMarkdownToHTML(t *testing.T) {
-	tests := []struct {
+	cases := []struct {
 		name     string
-		input    string
-		contains string
+		md       string
+		rendered string
 	}{
-		{"bold", "**hello**", "<strong>hello</strong>"},
-		{"italic", "_world_", "<em>world</em>"},
-		{"header", "### Title", "<h3"},
-		{"code block", "```\nfoo()\n```", "<code>"},
-		{"inline code", "`x`", "<code>x</code>"},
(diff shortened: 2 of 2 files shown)


horsley:
Thanks for working on this. I understand the motivation, but I’m not convinced this change belongs in picoclaw in its current form.

My main concern is scope and ownership. picoclaw’s responsibility here should be to generate valid, spec-compliant Matrix HTML, not to encode rendering workarounds for specific Matrix clients unless we have strong evidence that the current output is actually invalid or broadly incompatible.

From the PR description, the problem seems to be that *some* Matrix clients — especially FluffyChat — do not render the current output well. That sounds more like a client-side rendering/CSS issue, or possibly something to address upstream in the Markdown renderer, rather than something picoclaw should permanently bake into its default Matrix output for all users.

This PR does not just tweak presentation in a minimal way. It introduces AST-level post-processing to suppress paragraph tags and changes parser behavior (`DefinitionLists`), which means picoclaw is no longer just using the Markdown library’s output — it is now applying channel-specific structural rewrites. That is a much larger responsibility boundary, and I think the bar for merging that kind of change should be fairly high.

Before I’d be comfortable merging this, I think we need at least:

1. **A stronger spec-based justification**  
   Please show that the current output is actually undesirable from a Matrix-spec perspective, not just visually suboptimal in one client.

2. **Cross-client validation**  
   This should be tested in multiple mainstream Matrix clients, not only FluffyChat. Otherwise we risk fixing one client’s rendering quirk by changing output for everyone else.

3. **Semantic safety**  
   The transformation should preserve Markdown/CommonMark structure. Right now it appears to change semantics in valid cases such as multi-paragraph list items, where paragraph boundaries can be collapsed.

4. **A clearer justification for why this belongs in picoclaw**  
   If this is really a client-specific compatibility workaround, I would prefer a narrower or opt-in approach rather than changing the default output globally.

So my current position is: I’m not opposed to Matrix-specific adaptation in principle, but I don’t think we have enough evidence yet that this is the right layer to solve it, nor enough proof that the change is safe across clients.

**References**
- CommonMark 0.31.2: https://spec.commonmark.org/0.31.2/
- Matrix Client-Server API (`org.matrix.custom.html` guidance): https://spec.matrix.org/v1.13/client-server-api/

dim:
@horsley I appreciate the concern:

1. I updated the test to demonstrate the output for a longer document. The new version behaves just like Element (the reference Matrix client) while the previous version was adding additionaly, unnecessary tags.
2. I validated the output in Element Desktop, Element Android and FluffyChat. There are still rendering issues with Element Android and FluffyChat but those are bugs in their apps, not something I am trying to work around here at all. I will flag this with the relevant GH projects separately.
3. I will add a separate test for this
4. I hope the above explains it. I am aiming for a simple, safe and minimal CommonMark to HTML conversion in picoclaw, nothing more.

dim:
OK, I reduced the code to the bare minimum and added good tests. No hacks, not "workarounds", all plain and simple. PTAL and let me know if that's acceptable.

horsley (approved):
LGTM

PR #1536, opened 2026-03-14
Fix: allow picoclaw media tempdir

alexhoshina:
## 📝 Description

This PR fixes the non-image attachment flow under the default workspace sandbox.

  Previously, inbound documents/audio/video were often downloaded into `/tmp/picoclaw_media/...`, but tools such as `read_file`,
  `list_dir`, restricted `exec`, and `send_file` were still limited to the agent workspace. As a result, the model could receive a
  local file path that it still could not access.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

 ## What Changed
 
- Centralized the shared PicoClaw media temp directory path into `pkg/media/tempdir.go`
  - Updated media download/write paths to reuse the centralized temp dir helper
    - `pkg/utils/media.go`
    - `pkg/channels/feishu/feishu_64.go`
    - `pkg/channels/matrix/matrix.go`
  - Extended agent read allow-path construction to always include the PicoClaw media temp directory
  - Passed that allowlist through to tools that need to read local files
    - `read_file`
    - `list_dir`
    - `exec` (including `working_dir` validation)
    - `send_file`
  - Added path validation support for explicitly allowed read paths while preserving existing workspace restriction behavior for
  normal paths
  - Added regression tests covering:
    - agent tools reading/listing/executing within the media temp dir
    - `send_file` accepting a whitelisted temp media path
    - matrix temp dir helper behavior
  - Removed the now-unused `validatePath` wrapper after the allowlist-aware refactor

### DIFF
--- a/pkg/tools/filesystem.go
+++ b/pkg/tools/filesystem.go
@@ -82,22 +82,19 @@ func isAllowedPath(path string, patterns []*regexp.Regexp) bool {
 	}
 
 	cleaned := filepath.Clean(path)
+	if !filepath.IsAbs(cleaned) {
+		return false
+	}
 	if !matchesAllowedPath(cleaned, patterns) {
 		return false
 	}
 
-	resolved, err := filepath.EvalSymlinks(cleaned)
-	if err == nil {
-		return matchesAllowedPath(resolved, patterns)
-	}
-	if os.IsNotExist(err) {
-		parentResolved, parentErr := resolveExistingAncestor(filepath.Dir(cleaned))
-		if parentErr == nil {
-			return matchesAllowedPath(parentResolved, patterns)
-		}
+	resolved, err := resolvePathAgainstExistingAncestor(cleaned)
+	if err != nil {
+		return false
 	}
 
-	return false
+	return matchesAllowedPath(resolved, patterns)
 }
 
 func matchesAllowedPath(path string, patterns []*regexp.Regexp) bool {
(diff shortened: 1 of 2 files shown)


horsley:
Thanks for the fix — this is a solid direction overall.

One security concern: `whitelistFs.matches` currently uses regex-only matching (`matchesAllowedPath`) and does not use the symlink-resolving logic in `isAllowedPath`.

This can allow symlink escape from an allowed dir. Example: if `/tmp/picoclaw_media/link_out -> /etc`, then `read_file` on `/tmp/picoclaw_media/link_out/passwd` matches the allow regex and can read `/etc/passwd`.

Could we switch `whitelistFs.matches` to `isAllowedPath(path, patterns)` (or equivalent canonical-path validation) so whitelist access is symlink-safe?

alexhoshina:
> Thanks for the fix — this is a solid direction overall.
> 
> One security concern: `whitelistFs.matches` currently uses regex-only matching (`matchesAllowedPath`) and does not use the symlink-resolving logic in `isAllowedPath`.
> 
> This can allow symlink escape from an allowed dir. Example: if `/tmp/picoclaw_media/link_out -> /etc`, then `read_file` on `/tmp/picoclaw_media/link_out/passwd` matches the allow regex and can read `/etc/passwd`.
> 
> Could we switch `whitelistFs.matches` to `isAllowedPath(path, patterns)` (or equivalent canonical-path validation) so whitelist access is symlink-safe?

Thank you for your review, done.

PR #1610, opened 2026-03-15
fix(web): require CIDR allowlist for public mode

### LINKED ISSUE #1529: Security: refuse public web mode when no allowed_cidrs are configured
## Summary
`picoclaw-web -public` currently listens on `0.0.0.0` even when `allowed_cidrs` is empty. Because the IP allowlist middleware treats an empty CIDR list as “no restriction”, the public web console is exposed without any authentication boundary.

## Why this matters
The web console exposes powerful management endpoints, including config editing and session inspection. Public mode should fail closed unless the operator explicitly constrains network access.

### Manual
1. Start the web backend with `picoclaw-web -public`.
2. Leave `launcher-config.json` with an empty `allowed_cidrs` list.
3. From another host on the same network, access endpoints such as:
   - `GET /api/config`
   - `GET /api/sessions`
   - `GET /api/pico/token`
4. Observe that the requests succeed.

### Suggested regression test
- Add a startup validation test that sets `public=true` and `allowed_cidrs=[]`.
- Expected result: startup fails with a clear error, unless an explicit “insecure public mode” escape hatch is provided.

## Proposed implementation
- Refuse to start in public mode when `allowed_cidrs` is empty.
- If maintainers want an override, make it very explicit (for example `--insecure-public`) rather than the default behavior.
- Update docs/help text to explain the requirement.

## Acceptance criteria
- Public mode fails closed without an allowlist.
- Non-public localhost mode keeps the current behavior.
- Docs/help text explain how to expose the console safely.
- A regression test covers the startup validation path.

yzxlr:
## Summary
- fail closed when PicoClaw Web is started in public/LAN mode without any allowed_cidrs configured
- reject saving launcher configs that enable public mode without an explicit CIDR allowlist
- update UI/help text and add regression coverage for the new validation

Closes #1529

## Testing
- go test ./web/backend/...

### DIFF
--- a/web/backend/api/gateway_host_test.go
+++ b/web/backend/api/gateway_host_test.go
@@ -31,8 +31,9 @@ func TestBuildWsURLUsesRequestHostWhenLauncherPublicSaved(t *testing.T) {
 	configPath := filepath.Join(t.TempDir(), "config.json")
 	launcherPath := launcherconfig.PathForAppConfig(configPath)
 	if err := launcherconfig.Save(launcherPath, launcherconfig.Config{
-		Port:   18800,
-		Public: true,
+		Port:         18800,
+		Public:       true,
+		AllowedCIDRs: []string{"192.168.1.0/24"},
 	}); err != nil {
 		t.Fatalf("launcherconfig.Save() error = %v", err)
 	}
--- a/README.md
+++ b/README.md
@@ -207,6 +207,7 @@ Open http://localhost:18800 in your browser. The launcher manages the gateway pr
 
 > [!WARNING]
 > The web console does not yet support authentication. Avoid exposing it to the public internet.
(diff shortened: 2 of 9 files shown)


web/backend/api/gateway_host_test.go
```diff
@@ -31,8 +31,9 @@ func TestBuildWsURLUsesRequestHostWhenLauncherPublicSaved(t *testing.T) {
 	configPath := filepath.Join(t.TempDir(), "config.json")
 	launcherPath := launcherconfig.PathForAppConfig(configPath)
 	if err := launcherconfig.Save(launcherPath, launcherconfig.Config{
-		Port:   18800,
-		Public: true,
+		Port:         18800,
+		Public:       true,
+		AllowedCIDRs: []string{"192.168.1.0/24"},
```
sky5454, line 36: Not Universal, Class C IP address is not only 192.168.1.x,  should test lists that contain multiple elements

PR #1630, opened 2026-03-16
fix(provider): normalize anthropic model ids for API key path

### LINKED ISSUE #1624: fix(provider/anthropic): model ID dots not normalized to dashes in API-key and anthropic-messages paths
## Description

When configuring `anthropic/claude-sonnet-4.6` as the default model with an API key, the Anthropic API returns a 404 because the model name is sent with dots instead of dashes.

## Environment

- **Platform**: macOS (Darwin 25.3.0)
- **Config**: `model: "anthropic/claude-sonnet-4.6"` with API key auth

## Steps to Reproduce

1. Configure `config.json`:
```json
{
  "agents": {
    "defaults": {
      "model_name": "claude-sonnet-4.6",
      "model": "anthropic/claude-sonnet-4.6"
    }
  }
}
```
2. Run picoclaw with any prompt.

## Expected Behavior

The model ID `claude-sonnet-4.6` is normalized to `claude-sonnet-4-6` before being sent to the Anthropic API.

## Actual Behavior

```
ERR agent loop.go:1161 (agent.(*AgentLoop).runLLMIteration) > LLM call failed
error="API request failed:\n  Status: 404\n  Body: {\"error\":{\"code\":\"not_found_error\",
\"message\":\"model: claude-sonnet-4.6 was not found. Did you mean claude-sonnet-4-6?\",
\"type\":\"i..."
```

## Root Cause

The dot-to-dash normalization (`strings.ReplaceAll(model, ".", "-")`) only exists in the SDK-based Anthropic provider (`pkg/providers/anthropic/provider.go:214`), which is used for OAuth/token auth. Two other code paths lack it:

| Code path | Used when | Normalization |
|---|---|---|
| `pkg/providers/anthropic/provider.go` | OAuth/token auth | Yes |
| `pkg/providers/factory_provider.go` (anthropic case) | API key auth | **Missing** |
| `pkg/providers/anthropic_messages/provider.go` | anthropic-messages protocol | **Missing** |

## Relation to #269

Issue #269 reported a similar 404 symptom but with a different root cause (wrong provider type via the old config format). This bug affects the newer protocol-prefix config format where the provider selection is correct but the model name isn't normalized.

## Proposed Fix

Add `strings.ReplaceAll(modelID, ".", "-")` in:
1. `CreateProviderFromConfig` for the `"anthropic"` case
2. `anthropic_messages.buildRequestBody`

Alix-007:
## Summary
- normalize Anthropic model IDs on the OpenAI-compatible API-key path
- convert dotted Claude aliases like `claude-sonnet-4.6` to the dash form Anthropic expects
- add regression coverage for both normalization and the outbound request body

## Testing
- `go test ./pkg/providers/openai_compat -run TestNormalizeModel_UsesAPIBase -count=1`
- `go test ./pkg/providers/openai_compat -run TestProviderChat_NormalizesAnthropicModelForAnthropicAPIBase -count=1`
- `go test ./pkg/providers -run TestCreateProviderFromConfig_Anthropic -count=1`

Closes #1624.

### DIFF
--- a/pkg/providers/openai_compat/provider_test.go
+++ b/pkg/providers/openai_compat/provider_test.go
@@ -511,13 +511,15 @@ func TestProviderChat_AcceptsNumericOptionTypes(t *testing.T) {
 }
 
 func TestNormalizeModel_UsesAPIBase(t *testing.T) {
+	const anthropicAPIBase = "https://api.anthropic.com/v1"
+
 	if got := normalizeModel("deepseek/deepseek-chat", "https://api.deepseek.com/v1"); got != "deepseek-chat" {
 		t.Fatalf("normalizeModel(deepseek) = %q, want %q", got, "deepseek-chat")
 	}
-	if got := normalizeModel("claude-sonnet-4.6", "https://api.anthropic.com/v1"); got != "claude-sonnet-4-6" {
+	if got := normalizeModel("claude-sonnet-4.6", anthropicAPIBase); got != "claude-sonnet-4-6" {
 		t.Fatalf("normalizeModel(anthropic plain) = %q, want %q", got, "claude-sonnet-4-6")
 	}
-	if got := normalizeModel("anthropic/claude-sonnet-4.6", "https://api.anthropic.com/v1"); got != "claude-sonnet-4-6" {
+	if got := normalizeModel("anthropic/claude-sonnet-4.6", anthropicAPIBase); got != "claude-sonnet-4-6" {
 		t.Fatalf("normalizeModel(anthropic prefixed) = %q, want %q", got, "claude-sonnet-4-6")
 	}
 	if got := normalizeModel("openrouter/auto", "https://openrouter.ai/api/v1"); got != "openrouter/auto" {


Alix-007:
Addressed the current red CI item on this PR. This update only reshapes the new Anthropic normalization coverage so golines/gofmt stay clean; the provider logic and assertions are unchanged. Local verification: go test ./pkg/providers/openai_compat -run TestNormalizeModel_UsesAPIBase -count=1 and go test ./pkg/providers/openai_compat -run TestProviderChat_NormalizesAnthropicModelForAnthropicAPIBase -count=1.

pavelanni:
Hi @Alix-007, thanks for looking into this! Just a heads-up — I opened the issue (#1624) and already have a fix in #1626 that addresses the same bug.

The two PRs take different approaches:

- **#1626** normalizes the model ID in `factory_provider.go` (the `"anthropic"` case) and in `anthropic_messages/provider.go`, where the protocol is already known.
- **#1630** adds normalization inside `openai_compat/provider.go` by detecting `api.anthropic.com` in the API base URL.

A couple of concerns with the URL-sniffing approach:
1. It won't cover custom proxies or alternative Anthropic-compatible endpoints that don't contain `api.anthropic.com` in the URL.
2. It doesn't fix the `anthropic-messages` provider path, which has the same issue.

Happy to discuss if the maintainers prefer a different direction.

Alix-007:
Closing this in favor of #1626. That branch comes from the issue author, covers both the anthropic API-key path and the anthropic-messages path, and avoids URL sniffing on api_base. I’d rather reduce reviewer load here than keep two overlapping fixes open.

Issue #1634, opened 2026-03-16
bug: per-agent model config selects model string but not provider — all agents share a single provider instance


securityguy:
## Description

When multiple agents are configured in `agents.list`, each with a different `model.primary`, the model string is correctly routed per-agent but the **provider instance is shared across all agents** and is determined solely by `agents.defaults.model_name`. This means the per-agent `model.primary` setting only changes the model string passed to `Chat()` — it has no effect on which provider binary is invoked.

## Root Cause

In `cmd/picoclaw/internal/gateway/helpers.go`, a single provider is created from the default config:

```go
provider, modelID, err := providers.CreateProvider(cfg)  // uses agents.defaults.model_name
agentLoop := agent.NewAgentLoop(cfg, msgBus, provider)
```

`NewAgentLoop` → `NewAgentRegistry` passes this **same provider instance** to every agent. The fallback chain in `loop.go` calls:

```go
func(ctx context.Context, provider, model string) (*providers.LLMResponse, error) {
    return agent.Provider.Chat(ctx, messages, providerToolDefs, model, llmOpts)
}
```

The `provider` parameter (the protocol string from `FallbackCandidate.Provider`) is passed in but **ignored** — `agent.Provider` is always the single shared instance.

## Impact

Consider this config:

```json
"agents": {
    "defaults": { "model_name": "claude-cli" },
    "list": [
        { "id": "alice", "model": { "primary": "gemini-cli" } },
        { "id": "bob",   "model": { "primary": "codex-cli"  } }
    ]
}
```

- The shared provider is `ClaudeCliProvider` (from defaults)
- Alice's `"primary": "gemini-cli"` causes `ClaudeCliProvider.Chat(model="gemini-cli")` to be called, which passes `--model gemini-cli` to the `claude` binary — rejected with `ModelNotFoundError`
- Bob's `"primary": "codex-cli"` causes `ClaudeCliProvider.Chat(model="codex-cli")` — same failure

This makes it impossible to assign different CLI providers to different agents in a single picoclaw instance. HTTP-based providers (OpenAI, Anthropic, etc.) are less affected because the model string is passed through the API request, but cross-provider fallback chains (e.g. `claude-cli` falling back to `gemini-cli`) also do not work correctly.

## Expected Behaviour

[...]

PR #1637, opened 2026-03-16
fix(agent): dispatch per-candidate provider in fallback chain

securityguy:
## Summary

- All agents previously shared a single `LLMProvider` instance created from `agents.defaults.model_name`
- Per-agent `model` config only changed the model string passed to `Chat()` — it never changed which provider was invoked
- Cross-provider fallback chains (e.g. `gemini-cli` falling back to `claude-cli`) failed silently, and different CLI providers could not be assigned to different agents

## Solution

Introduces `ProviderDispatcher` which lazily creates and caches provider instances keyed by `"protocol/modelID"`:

- Thread-safe (sync.RWMutex), double-checked locking pattern
- Providers are stateless so caching is safe across agents with no data leakage
- `Flush(cfg)` method clears the cache on config reload
- The fallback chain's `run` closure resolves the correct provider via the dispatcher, falling back to `agent.Provider` for backward compatibility

References #1634 (keeping issue open for tracking)

## Test plan

- [ ] `go test ./pkg/providers/... -run TestProviderDispatcher -race` — 5 tests, all pass including race detector
- [ ] `go test ./pkg/agent/...` — existing agent loop tests pass with updated `NewAgentLoop` signature
- [ ] `go build ./...` — compiles cleanly

### DIFF
--- a/cmd/picoclaw/internal/agent/helpers.go
+++ b/cmd/picoclaw/internal/agent/helpers.go
@@ -47,9 +47,10 @@ func agentCmd(message, sessionKey, model string, debug bool) error {
 		cfg.Agents.Defaults.ModelName = modelID
 	}
 
+	dispatcher := providers.NewProviderDispatcher(cfg)
 	msgBus := bus.NewMessageBus()
 	defer msgBus.Close()
-	agentLoop := agent.NewAgentLoop(cfg, msgBus, provider)
+	agentLoop := agent.NewAgentLoop(cfg, msgBus, provider, dispatcher)
 	defer agentLoop.Close()
 
 	// Print agent startup info (only for interactive mode)
--- a/cmd/picoclaw/internal/gateway/helpers.go
+++ b/cmd/picoclaw/internal/gateway/helpers.go
@@ -81,8 +81,9 @@ func gatewayCmd(debug bool) error {
 		cfg.Agents.Defaults.ModelName = modelID
 	}
 
+	dispatcher := providers.NewProviderDispatcher(cfg)
 	msgBus := bus.NewMessageBus()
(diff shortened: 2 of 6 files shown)


yinwm (commented):
Thanks for the thorough implementation — the tests and thread-safety considerations are appreciated. However, after comparing this with #2143, I believe #2143 is the better fit for this project. Here's my reasoning:

**Why #2143 over this PR:**

1. **YAGNI** — #2143 solves the same bug in 43 lines vs 286 lines. The `ProviderDispatcher` with double-check locking, lazy creation, and `Flush()` is over-engineered for a map that's written once at init and read-only thereafter. No locks are needed.

2. **Correct lifecycle** — `CandidateProviders` lives on `AgentInstance` (per-agent config), while `ProviderDispatcher` lives on `AgentLoop` (shared across agents). Provider configuration follows the agent, not the loop. Multiple agents with different fallback chains need independent provider maps.

3. **Lower API surface impact** — #2143 doesn't change `NewAgentLoop`'s signature or touch 6 call sites. This PR changes the public API and requires updating every caller + test.

4. **No global state** — `ProviderDispatcher` is a global cache that needs explicit `Flush()` on config reload. #2143's map is created during `NewAgentInstance` and naturally discarded when the agent is recreated on reload.

**What I'd like to salvage from this PR if possible:**
- The unit tests are well-structured — it would be great if @corevibe555 could add similar tests to #2143
- The `Flush()` concept is worth documenting as a pattern for when lazy initialization is actually needed in the future

I'm going to close this PR in favor of #2143. Thank you for the contribution and the careful implementation — the quality of work here is clearly strong.

PR #1642, opened 2026-03-16
新增asr和tts功能

kkroid:
独立子模块 `cmd/picoclaw-voice`，实现完整端到端语音对话链路

- 协议层参考 xiaozhi 私有 WebSocket 协议设计，未绑定具体硬件。
- 流式ASR: 豆包 / FunASR 本地离线版本
- 流式TTS: 豆包 / Fish Speech 本地离线版本


lxowalle:
Nice, we've also recently discussed integrating ASR+TTS functionality internally. I think there are still some TODO to improve:
1. Need to finalize the configuration file to allow users to customize their own ASR and TTS models. For example, setting which TTS model to use, how to configure the KEY, how to enable it, etc.
2. Need to determine how to handle non-streaming scenarios, such as when an audio file is sent through the channel
3. Integrate ASR+TTS into the gateway

kkroid:
> Nice, we've also recently discussed integrating ASR+TTS functionality internally. I think there are still some TODO to improve:
> 
> 1. Need to finalize the configuration file to allow users to customize their own ASR and TTS models. For example, setting which TTS model to use, how to configure the KEY, how to enable it, etc.
> 2. Need to determine how to handle non-streaming scenarios, such as when an audio file is sent through the channel
> 3. Integrate ASR+TTS into the gateway

Thanks for your feedback
1. Config — Currently using env vars + .env file. Happy to align with whatever config format the team prefers (e.g. extending config.json/config.yaml).
2. Non-streaming — Will look into supporting audio file input via channels as a follow-up.
3. Gateway integration — Currently this PR keeps it as a standalone module. I agree that integrating them into the main gateway is a good idea.

opcache:
666 ，支持，做了我想做的，牛啊

lxowalle:
https://github.com/sipeed/picoclaw/issues/1648 has a more detailed proposal.

Issue #1652, opened 2026-03-16
[BUG] GLM Coding Plan is not support PicoClaw


xpader:
GLM Coding Plan is not support PicoClaw,  config glm coding will consuming balance or get error soon:
```
  Status: 429
  Body:   {"error":{"code":"1113","message":"余额不足或无可用资源包,请充值。"}}
```

My Config:
```
...
  "agents": {
    "defaults": {
      "workspace": "/home/***/.picoclaw/workspace",
      "restrict_to_workspace": true,
      "allow_read_outside_workspace": false,
      "provider": "",
      "model": "glm-4.7",
      "max_tokens": 32768,
      "max_tool_iterations": 50,
      "summarize_message_threshold": 20,
      "summarize_token_percent": 75
    }
  },
...
"model_list": [
    {
      "model_name": "glm-4.7",
      "model": "zhipu/glm-4.7",
      "api_base": "https://open.bigmodel.cn/api/paas/v4",
      "api_key": "***"
    },
...
```

xpader:
Only supported these tools and OpenClaw:

<img width="1877" height="445" alt="Image" src="https://github.com/user-attachments/assets/0a114609-f5bc-41cb-839c-5d3132d4fab1" />

hyperwd:
Hi @xpader,

Thanks for reporting this issue!

## Analysis

The problem is likely related to the API endpoint you're using. Your current configuration uses the **OpenAI-compatible endpoint**:

```json
"api_base": "https://open.bigmodel.cn/api/paas/v4"
```

This endpoint may consume your **balance quota** instead of the **Coding Plan** quota, which could explain the "insufficient balance" error (429) you're encountering.

## Solution

To use the **GLM Coding Plan**, you likely need to use Zhipu AI's **Anthropic Messages API endpoint** with the Anthropic Messages protocol format.

**Good news**: I've just submitted PR #1663 to fix an issue with GLM's nil input handling for the anthropic-messages provider. Once this PR is merged and released, you can configure it like this:

```json
{
  "model_list": [
    {
      "model_name": "glm-4.7",
      "model": "anthropic-messages/glm-4.7",
      "api_base": "https://open.bigmodel.cn/api/anthropic",
      "api_key": "your_api_key_here"
    }
  ]
}
```

**Key changes:**
- Change `"model"` from `"zhipu/glm-4.7"` to `"anthropic-messages/glm-4.7"`
[...]

PR #1663, opened 2026-03-17
fix(providers): handle nil input in GLM series tool_use blocks

hyperwd:
## Problem

Zhipu AI's GLM series models may return tool_use blocks with null input field, which causes their API to reject subsequent requests with error:
`ClaudeContentBlockToolResult object has no attribute id`

## Solution

- Add defensive nil check for tool call Arguments field
- Replace nil input with empty object `{}` to comply with Anthropic spec
- Prevent API errors while maintaining backward compatibility with other providers

This is a follow-up fix to PR #1284.

### DIFF
--- a/pkg/providers/anthropic_messages/provider.go
+++ b/pkg/providers/anthropic_messages/provider.go
@@ -221,11 +221,17 @@ func buildRequestBody(
 
 			// Add tool_use blocks
 			for _, tc := range msg.ToolCalls {
+				// Handle nil Arguments (GLM-4 may return null input)
+				input := tc.Arguments
+				if input == nil {
+					input = map[string]any{}
+				}
+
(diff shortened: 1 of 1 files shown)


hyperwd:
## Related Issue

This PR also fixes #1652 - it resolves the nil input issue that was preventing users from using the **GLM Coding Plan** via the `anthropic-messages` provider.

Once this PR is merged, users can configure GLM to use the Coding Plan quota with:

```json
{
  "model_name": "glm-4.7",
  "model": "anthropic-messages/glm-4.7",
  "api_base": "https://open.bigmodel.cn/api/anthropic",
  "api_key": "your_api_key_here"
}
```

**Closes #1652**

afjcjsbx:
LGTM!

PR #1666, opened 2026-03-17
feat(mcp): Refactor MCP Manager: Process Lifecycle, Resilience, and Concurrency Safety

tong3jie:
## 📝 Description

This PR addresses several critical areas in the mcp.Manager implementation, specifically focusing on resource management, process lifecycle control for Stdio servers, and concurrency safety. The refactor decouples transport logic, optimizes environment variable handling, and mitigates potential memory leaks and zombie processes.

### Key Changes
1. Process Lifecycle & Resource Safety
- Zombie Process Mitigation: Added SysProcAttr configuration to exec.Command for stdio transports. This ensures that on Unix-like systems, child processes receive a termination signal if the parent process exits unexpectedly.

- Graceful Shutdown Enhancement: Optimized the Close() method using atomic.Bool and sync.WaitGroup. The manager now ensures all in-flight tool calls are completed before closing sessions, with a built-in timeout to prevent hanging.

2. Architectural Refactoring & Abstraction
- Transport Factory Pattern: Extracted transport initialization from the monolithic ConnectServer method into private newStdioTransport and newSSETransport helpers. This separation of concerns (SoC) improves code readability and simplifies unit testing.

- Header Handling: Updated headerTransport to use Header.Add instead of Set (or vice versa based on requirement) to ensure critical default headers are not inadvertently overwritten.

3. Performance & Concurrency Optimization
- Optimized Environment Merging: Introduced godotenv utility that pre-allocates map capacity based on os.Environ(). This reduces memory allocations and redundant system calls during server startup.

- Concurrency Fixes: Identified and addressed a potential Data Race in GetServers. Added recommendations for deep-copying Tools slices to prevent external mutations from affecting the internal state of the manager.

4. Resilience Foundations
- Config Persistence: The ServerConnection now persists the config.MCPServerConfig, enabling the future implementation of automated reconnection strategies when a session is dropped.

5. Asynchronous Resilience & Health Monitoring

- Active Health Checks: Implemented a background monitor that performs periodic heartbeats using ListTools.

- Status State Machine: Introduced Online/Offline states to allow Fast-Fail mechanism in CallTool, preventing request accumulation during server downtime.

- Exponential Backoff Reconnection: Servers that fail health checks are moved to an asynchronous reconnection queue with a 1s to 16s exponential backoff strategy, ensuring self-healing without blocking the main execution thread.
## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 🧪 Test Environment
- **Hardware:** x86
- **OS:** macos, linux

## 📸 Evidence (Optional)
<details>

</details>

## ☑️ Checklist

### DIFF
--- a/pkg/mcp/manager.go
+++ b/pkg/mcp/manager.go
@@ -1,22 +1,37 @@
 package mcp
 
 import (
-	"bufio"
+	"cmp"
 	"context"
 	"errors"
 	"fmt"
+	"maps"
 	"net/http"
-	"os"
-	"os/exec"
 	"path/filepath"
-	"strings"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/modelcontextprotocol/go-sdk/mcp"
 
 	"github.com/sipeed/picoclaw/pkg/config"
 	"github.com/sipeed/picoclaw/pkg/logger"
+)
+
+const (
+	clientName    = "picoclaw"
+	clientVersion = "1.0.0"
+
+	logModule = "mcp"
+	maxFails  = 3
+)
+
+var (
+	ErrManagerClosed = errors.New("MCP manager is closed")
+
+	ErrInvalidServerConfig = errors.New("either URL or command must be provided")
+
+	ErrStdioCommandRequired = errors.New("command is required for stdio transport")
 )
 
 // headerTransport is an http.RoundTripper that adds custom headers to requests
@@ -26,91 +41,21 @@ type headerTransport struct {
 }
 
 func (t *headerTransport) RoundTrip(req *http.Request) (*http.Response, error) {
-	// Clone the request to avoid modifying the original
 	req = req.Clone(req.Context())
-
-	// Add custom headers
 	for key, value := range t.headers {
-		req.Header.Set(key, value)
+		req.Header.Add(key, value)
 	}
+	base := cmp.Or(t.base, http.DefaultTransport)
 
-	// Use the base transport
-	base := t.base
-	if base == nil {
-		base = http.DefaultTransport
-	}
 	return base.RoundTrip(req)
-}
-
-// loadEnvFile loads environment variables from a file in .env format
-// Each line should be in the format: KEY=value
-// Lines starting with # are comments
-// Empty lines are ignored
-func loadEnvFile(path string) (map[string]string, error) {
-	file, err := os.Open(path)
-	if err != nil {
-		return nil, fmt.Errorf("failed to open env file: %w", err)
-	}
-	defer file.Close()
-
-	envVars := make(map[string]string)
-	scanner := bufio.NewScanner(file)
-	lineNum := 0
-
-	for scanner.Scan() {
-		lineNum++
-		line := strings.TrimSpace(scanner.Text())
-
-		// Skip empty lines and comments
-		if line == "" || strings.HasPrefix(line, "#") {
-			continue
-		}
-
-		// Parse KEY=value
-		parts := strings.SplitN(line, "=", 2)
-		if len(parts) != 2 {
-			return nil, fmt.Errorf("invalid format at line %d: %s", lineNum, line)
-		}
-
-		key := strings.TrimSpace(parts[0])
-		value := strings.TrimSpace(parts[1])
-
-		if key == "" {
-			return nil, fmt.Errorf("invalid format at line %d: empty key", lineNum)
-		}
-
-		// Remove surrounding quotes if present
-		if len(value) >= 2 {
-			if (value[0] == '"' && value[len(value)-1] == '"') ||
-				(value[0] == '\'' && value[len(value)-1] == '\'') {
-				value = value[1 : len(value)-1]
-			}
-		}
-
-		envVars[key] = value
-	}
-
-	if err := scanner.Err(); err != nil {
-		return nil, fmt.Errorf("error reading env file: %w", err)
-	}
-
-	return envVars, nil
-}
-
-// ServerConnection represents a connection to an MCP server
-type ServerConnection struct {
-	Name    string
-	Client  *mcp.Client
-	Session *mcp.ClientSession
-	Tools   []*mcp.Tool
 }
 
 // Manager manages multiple MCP server connections
 type Manager struct {
 	servers map[string]*ServerConnection
 	mu      sync.RWMutex
-	closed  atomic.Bool    // changed from bool to atomic.Bool to avoid TOCTOU race
-	wg      sync.WaitGroup // tracks in-flight CallTool calls
+	closed  atomic.Bool
+	wg      sync.WaitGroup
 }
 
 // NewManager creates a new MCP manager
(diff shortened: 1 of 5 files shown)


pkg/mcp/manager.go
```diff
@@ -133,64 +78,48 @@ func (m *Manager) LoadFromMCPConfig(
...
-							"server":   name,
-							"env_file": serverCfg.EnvFile,
-							"error":    err.Error(),
-						})
-					errs <- err
-					return
-				}
-				serverCfg.EnvFile = filepath.Join(workspace, serverCfg.EnvFile)
-			}
+			serverCfg.EnvFile = filepath.Join(workspace, serverCfg.EnvFile)
```
alexhoshina, line 116: `filepath.Join(workspace, serverCfg.EnvFile)` does more than just rewrite relative paths for env files. For common stdio configurations where env_file is not set, `filepath.Join("/ws", "")` becomes "**/ws**", causing `godotenv.Read` to attempt to parse the entire workspace directory as an env file, which ultimately prevents the service from starting at all.

pkg/mcp/manager.go
```diff
@@ -207,210 +136,83 @@ func (m *Manager) LoadFromMCPConfig(
...
+		}
 
-	logger.InfoCF("mcp", "MCP server initialization complete",
-		map[string]any{
+		logger.WarnCF(logModule, "Initialized with partial failures", map[string]any{
+			"failed":    len(allErrors),
 			"connected": connectedCount,
 			"total":     enabledCount,
 		})
+		return fmt.Errorf("partial MCP initialization failure: %w", err)
```
alexhoshina: Existing MCP connections should be preserved.
Currently, if any configured **MCP server** goes down, even if another one has successfully connected, it will cause all MCP tools to become completely unavailable.

pkg/mcp/manager.go
```diff
@@ -530,3 +323,100 @@ func (m *Manager) GetAllTools() map[string][]*mcp.Tool {
...
+			if attempt < len(backoff) {
+				wait = backoff[attempt]
+			}
+
+			time.Sleep(wait)
+			attempt++
+
+			logger.DebugCF("mcp", "Reconnection attempt", map[string]any{"server": name, "attempt": attempt})
+
+			err := m.ConnectServer(context.Background(), name, conn.Config)
```
alexhoshina, line 418: Stop reconnecting after the **manager** has been closed

alexhoshina (changes requested):

pkg/mcp/manager.go
tong3jie: don't think the scenario you mentioned will happen.

godotenv.Read will read file but not dir . if the path was a dir , it return error.

if the path was be "/ws",  t will be regarded as a file. 

if the file dose not exist ,that will return an  error, otherwise, it shall return empty map.

corrections are welcome .

@alexhoshina

pkg/mcp/manager.go
tong3jie: yes,  I agree with your point. 

I will fix it .

pkg/mcp/manager.go
tong3jie: thanks, you'r right.

I'll fix it.

yinwm (changes requested):
## Code Review — PR #1666

Great architectural direction. The transport factory pattern, `ServerConnection` extraction, health monitoring, and auto-reconnection are all solid improvements. However, I found **4 critical bugs** that need fixing before merge.

---

### 🔴 CRITICAL (must fix)

**C1: `filepath.Join(workspace, "")` crashes on empty EnvFile**

Already flagged by @alexhoshina but worth emphasizing severity. The original code had `if serverCfg.EnvFile != ""` guard before joining. The PR removed it, so when `EnvFile` is empty (the most common case for stdio servers):

```
filepath.Join("/workspace", "") → "/workspace"
godotenv.Read("/workspace") → attempts to read a *directory* as a file → crash
```

This breaks service startup for all stdio servers without an explicit `env_file` configured.

**Fix**: Restore the empty-value guard before calling `filepath.Join`.

---

**C2: `createTransport` uses wrong context and dummy server name**

```go
func (conn *ServerConnection) createTransport(cfg config.MCPServerConfig) (mcp.Transport, error) {
    // ...
    return conn.newSSETransport(context.Background(), "temp", cfg)
    return conn.newStdioTransport(context.Background(), "temp", cfg)
}
```

Two problems:
1. `context.Background()` instead of the caller's `ctx` — stdio child processes created via `exec.CommandContext` won't be cancelled when the original context is cancelled
2. `"temp"` as the server name — all log messages from transport creation show "temp" instead of the real server name, making debugging impossible

**Fix**: Pass `ctx` and `name` through to `createTransport` from `newServerConnection`.

---

**C3: `validateConfig` called inside the loop — O(N²) validation**

```go
for name, serverCfg := range mcpCfg.Servers {
    if !serverCfg.Enabled { continue }
    if err := m.validateConfig(mcpCfg); err != nil {  // validates ALL servers on every iteration
        return ...
    }
}
```

`validateConfig` iterates all servers in the config. Calling it inside the `for range` loop means N enabled servers → N² validation passes.

**Fix**: Call `validateConfig` once before the loop.

---

**C4: `handleServerOffline` infinite reconnection with no upper bound**

```go
go func() {
    for {
        time.Sleep(wait)
        err := m.ConnectServer(context.Background(), name, conn.Config)
        if err == nil { return }
        // ❌ no max retry limit — runs forever if server is permanently unavailable
    }
}()
```

Problems:
- No maximum retry count — if a server is permanently down, this goroutine runs forever
- Uses `context.Background()` — even after `Manager.Close()`, the goroutine continues during `time.Sleep` windows
- On reconnection success, calls `m.startMonitor(name)` which spawns yet another goroutine, potentially stacking monitors

**Fix**: Add a max retry limit (e.g., 10 attempts). Use a cancellable context tied to the manager's lifecycle. Cancel all reconnection goroutines in `Close()`.

---

### 🟡 WARNING (should fix)

**W1: `SysProcAttr{Setpgid: true}` won't compile on Windows**

`syscall.SysProcAttr` fields differ between Unix and Windows. `Setpgid` doesn't exist on Windows. Add a build tag (`//go:build !windows`) or move to a `_unix.go` file.

---

**W2: `GetServers` returns shallow copy — external code can mutate internal state**

`maps.Copy` copies the map entries (pointer values), so callers get the same `*ServerConnection` pointers. External code can directly modify `conn.Tools`, `conn.Session`, etc.

Consider returning an interface or documenting that callers must not mutate.

---

**W3: `Header.Add` vs `Header.Set` semantics change**

The PR changed from `req.Header.Set(key, value)` to `req.Header.Add(key, value)`. If the base transport or middleware already sets the same header, `Add` appends a duplicate instead of replacing. Verify this is intentional.

---

### ✅ What's done well

- **Transport factory** (`newStdioTransport` / `newSSETransport`) — clean SoC
- **ServerConnection extraction** to `server.go` — much better file organization
- **godotenv** replacing hand-rolled parser — eliminates maintenance burden
- **Health monitoring + auto-reconnect** — essential for production MCP usage
- **`Setpgid: true`** — prevents zombie stdio processes (Unix)
- **`maps.Copy`** — idiomatic Go 1.21+ patterns

---

**Summary**: Fix C1–C4 and this is a strong merge candidate. The architecture is right; these are implementation bugs that need addressing.

PR #1680, opened 2026-03-17
fix(agent): recover direct answer after tool limit

### LINKED ISSUE #1641: [BUG]picclaw not work after a few days with max_tool_iterations erorr
##  Environment & Tools
- **PicoClaw Version:** 
picoclaw 0.2.2 (git: 4a8a2e9)
  Build: 2026-03-11T12:25:50Z
  Go: 1.25.7
- **Go Version:** go version go1.25.7 (Red Hat 1.25.7-1.0.1.module+an8.10.0+11387+fcfff780) linux/amd64
- **AI Model & Provider:** OpenAI compatiable
- **Operating System:** Aliyun Anolis OS 8.8 x86_64
- **Channels:** Feishu

## 📸 Steps to Reproduce
1. Chat with picoclaw for a few days
2. Occur the following errors:

```
I've completed processing but have no response to give. Increase max_tool_iterations in config.json.
```

and can not chat any more

## ❌ Actual Behavior

print error and stop use

## ✅ Expected Behavior

Auto handle this, not stop use

## 💬 Additional Context

I wonder to know is max_tool_iterations can set very big.can you auto remove some tools when reach the limit?

Image

Alix-007:
## Summary
- attempt one final no-tool LLM pass when the agent exhausts `max_tool_iterations`
- keep the synthetic fallback prompt out of session history so later turns stay clean
- add regression coverage for the recovery path and the existing context-retry path

## Problem
When a model keeps asking for tools until `max_tool_iterations` is reached, PicoClaw currently falls straight back to:

`I've completed processing but have no response to give. Increase \`max_tool_iterations\` in config.json.`

That leaves weaker tool-calling models stuck, and long-running sessions can become practically unusable after the first bad tool loop.

## Solution
After the normal tool loop hits its limit, the agent now makes one extra LLM call with the same accumulated conversation and tool results, but with the tool list hidden and a short instruction to answer directly from the available context.

If that direct-answer recovery still fails or still asks for tools, PicoClaw keeps the existing default response. This makes the fix conservative while removing the dead-end for models that already have enough context to answer.

## Testing
- `go test ./pkg/agent -run 'TestAgentLoop_(ToolLimitFallsBackToDirectAnswer|ContextExhaustionRetry)$'`
- `go test ./pkg/agent`

Closes #1641

### DIFF
--- a/pkg/agent/loop.go
+++ b/pkg/agent/loop.go
@@ -1372,7 +1372,77 @@ func (al *AgentLoop) runLLMIteration(
 		})
 	}
 
+	if finalContent == "" && iteration >= agent.MaxIterations && agent.MaxIterations > 0 {
+		logger.WarnCF("agent", "Tool iteration limit reached, attempting direct answer without tools",
+			map[string]any{
+				"agent_id":       agent.ID,
+				"iteration":      iteration,
+				"max_iterations": agent.MaxIterations,
+			})
+
+		llmOpts := map[string]any{
+			"max_tokens":       agent.MaxTokens,
+			"temperature":      agent.Temperature,
+			"prompt_cache_key": agent.ID,
+		}
+		if agent.ThinkingLevel != ThinkingOff {
+			if tc, ok := agent.Provider.(providers.ThinkingCapable); ok && tc.SupportsThinking() {
+				llmOpts["thinking_level"] = string(agent.ThinkingLevel)
+			}
+		}
+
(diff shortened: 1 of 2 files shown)


Alix-007:
Superseded by #1686.

This PR head branch was based on a divergent fork history, which made it conflict against upstream `main` even though the change itself was fine. I recreated the same fix on a clean branch based on the current upstream `main` so it can merge normally.

PR #1690, opened 2026-03-17
docs(exec): document build tool guard limitation

### LINKED ISSUE #529: [Task] Security Guard Bypass via Build Tools - Documentation Needed
# Security Guard Bypass via Build Tools - Documentation Needed

## Issue Description

The picoclaw safety guard has a **significant architectural bypass**: it only inspects the direct command being executed, not child processes spawned by that command.

### Current Behavior

- ✅ `make build` → compiles binaries (allowed)
- ✅ `make run` → executes compiled binaries (allowed)
- ❌ `./malicious_binary` → direct execution (blocked)

**The safety guard does NOT block execution of compiled binaries when invoked by development tools** (make, go run, cargo, npm, etc.). It only blocks direct binary execution by picoclaw.

### Proof of Concept

```makefile
# Malicious Makefile that bypasses safety guard
exploit: exploit.go
	go build -o exploit exploit.go
	./exploit  # ← Executes without being blocked
```

An attacker with ability to write files to the workspace could:
1. Write malicious Go/C/Rust code
2. Create a Makefile that compiles and executes it
3. Instruct picoclaw to run `make run`
4. The malicious code executes with picoclaw's privileges

## Request: Document This as a Known Architectural Limitation

Rather than fixing this (which would require fundamental architectural changes), **I request documenting this limitation clearly** so users understand the security boundaries.

### Proposed Documentation Addition

Add a new section to `docs/security.md` (or create `docs/security-limitations.md`) with the following content:

> ## ⚠️ Known Architectural Limitation
>
> ### Indirect Execution via Build Tools
>
> The picoclaw safety guard **can only inspect the command being executed directly**, not child processes spawned by that command.
>
> **Consequence:** Any tool that can execute binaries (such as `make`, `go run`, `cargo`, `npm run`, build scripts, etc.) can completely bypass the safety guard.
>
> **Example bypass:**
>
> ```makefile
> # This Makefile compiles and executes arbitrary code
> exploit: exploit.go
>     go build -o exploit exploit.go
>     ./exploit  # ← Executes without being blocked
> ```
>
> **Security Implications:**
> - The safety guard does **NOT** protect against malicious compiled code executed through Makefiles, build scripts, or development tools.
> - A malicious actor with the ability to create files in the workspace could:
>   1. Write malicious code in Go/C/Rust/etc.
>   2. Create a Makefile that compiles and executes it
>   3. Invoke `make run` from picoclaw → code executes without blocking
>
> **Recommendations for Users:**
> 1. **Never execute unreviewed code**: Treat any Makefile or build script as executable code. Review it before execution.
> 2. **Use sandboxing**: For high-risk environments, run picoclaw inside a container or VM with restrictive policies.
> 3. **Use `approve` mode**: Enable IM approval mode for build/compile tasks, though this may be cumbersome for legitimate workflows.
> 4. **Understand guard limits**: The safety guard is designed as a basic defense against direct dangerous commands (`rm -rf /`, `curl | sh`, etc.), **not** as a complete security system.
>
> **Note for Developers:**
> This limitation is inherent to the current architecture. True isolation would require:
> - Running all commands in a sandbox that restricts child process creation
> - OR recursively inspecting every process's command line (expensive and complex)
> - OR using SELinux/AppArmor/BPF to filter exec() at kernel level

### Why Document Instead of Fix?

Fixing this properly requires one of:
1. **Sandboxing all commands** (e.g., using containers, seccomp, or similar) - major architectural change
2. **Recursive process inspection** - complex, fragile, and may not catch all cases
3. **Kernel-level filtering** - platform-specific and requires elevated privileges

These are out of scope for picoclaw's design philosophy of being lightweight and simple.

Therefore, the appropriate response is to **document this as a known limitation** so users understand what the safety guard does and does not protect against.

## Additional Context

This was discovered during analysis of picoclaw's security policy engine (see PR #421). The current implementation only filters the top-level command string before execution, without considering what that command might spawn.

The bypass works because:
- `make`, `go`, `cargo`, etc. are whitelisted as "safe" developer tools
- The safety guard doesn't intercept or inspect child processes
- Once a developer tool is running, it can execute arbitrary binaries

## Impact

- **Severity**: Medium to High (depending on threat model)
- **Affected users**: Anyone who trusts the safety guard to prevent arbitrary code execution
- **Real-world risk**: Low for typical personal use, but could be exploited in scenarios where:
  - Multiple users share a picoclaw instance
  - Untrusted code or Makefiles are present in the workspace
  - Picoclaw is used in automated/CI contexts with untrusted inputs

## Labels

- `security`
- `documentation`
- `wontfix` (or `design-limitation`)
- `needs-discussion`

Alix-007:
## Summary
- document that the exec safety guard only inspects the top-level command PicoClaw launches
- clarify that child processes spawned by build tools like `make`, `go run`, `cargo`, and `npm run` are outside that guard boundary
- add concrete mitigation guidance in both the README security section and the exec tool reference

## Testing
- git diff --check

Closes #529

### DIFF
--- a/README.md
+++ b/README.md
@@ -861,6 +861,21 @@ Even with `restrict_to_workspace: false`, the `exec` tool blocks these dangerous
 * `shutdown`, `reboot`, `poweroff` — System shutdown
 * Fork bomb `:(){ :|:& };:`
 
+#### Known Limitation: Child Processes From Build Tools
+
+The exec safety guard only inspects the command line PicoClaw launches directly. It does not recursively inspect child
+processes spawned by allowed developer tools such as `make`, `go run`, `cargo`, `npm run`, or custom build scripts.
+
+That means a top-level command can still compile or launch other binaries after it passes the initial guard check. In
+practice, treat build scripts, Makefiles, package scripts, and generated binaries as executable code that needs the same
+level of review as a direct shell command.
+
+For higher-risk environments:
+
+* Review build scripts before execution.
+* Prefer approval/manual review for compile-and-run workflows.
+* Run PicoClaw inside a container or VM if you need stronger isolation than the built-in guard provides.
+
 #### Error Examples
 
 ```
--- a/docs/tools_configuration.md
+++ b/docs/tools_configuration.md
@@ -84,6 +84,22 @@ By default, PicoClaw blocks the following dangerous commands:
 - Git: `git push`, `git force`
 - Other: `eval`, `source *.sh`
 
+### Known Architectural Limitation
+
+The exec guard only validates the top-level command sent to PicoClaw. It does **not** recursively inspect child
+processes spawned by build tools or scripts after that command starts running.
+
+Examples of workflows that can bypass the direct command guard once the initial command is allowed:
+
+- `make run`
+- `go run ./cmd/...`
+- `cargo run`
+- `npm run build`
+
+This means the guard is useful for blocking obviously dangerous direct commands, but it is **not** a full sandbox for
+unreviewed build pipelines. If your threat model includes untrusted code in the workspace, use stronger isolation such
+as containers, VMs, or an approval flow around build-and-run commands.
+
 ### Configuration Example
 
 ```json


afjcjsbx:
Thanks for the documentation improvement! LGTM!

PR #1692, opened 2026-03-17
fix(config): inherit provider defaults for matching model_list entries

### LINKED ISSUE #1635: [BUG] model_list does not inherit api_key/api_base from providers (even with matching protocol prefix)
## Quick Summary

When defining `providers` in config (e.g. litellm) and referencing matching protocol prefixes in `model_list`, ModelConfig does not inherit api_key/api_base unless they are manually specified in every model_list entry. This causes inconvenient errors and is non-DRY.

## Environment & Tools
- **Go Version:** N/A (running official binary release)
- **Operating System:** As per official Dockerfile
- **Channels:** Discord, Telegram

## 📸 Steps to Reproduce
1. Add litellm provider to config:
    ```json
    "providers": {
      "litellm": { "api_key": "123", "api_base": "http://host:4000/v1" }
    }
    ```
2. Add model_list entries referencing litellm prefix but without duplicated keys:
    ```json
    "model_list": [
      { "model_name": "kimi1", "model": "litellm/kimi-1" },
      { "model_name": "kimi2", "model": "litellm/kimi-2" }
    ]
    ```
3. Start up the system.

## ❌ Actual Behavior

Both models fail with:
```
Error creating provider: api_key or api_base is required for HTTP-based protocol "litellm"
```

## ✅ Expected Behavior

If a model_list entry has a protocol prefix matching a defined provider (e.g. litellm/), the ModelConfig should inherit api_key, api_base, proxy, etc., so the user does not need to repeat them.

### Current Architecture

- The current config loader only invokes auto-migration from providers to model_list if model_list is completely empty.
- If any entries exist in model_list, the providers block is silently ignored (see pkg/config/config.go `LoadConfig` & `migration.go`).
- ModelConfig entries require all credentials to be explicitly set, even for identical protocol prefixes.
- There is no check at runtime or config load to merge/inherit providers.

### How to Add Without Breaking Current Behavior

- When loading/parsing model_list, for each entry missing api_key/api_base, check if its protocol prefix matches a configured provider.
- If found, automatically inherit these fields from the matching provider.
- This does not break current config formats, because explicit specification would always override inherited values.
- It is fully backward compatible, and enables DRY config for users with many model variants under one provider.
- If desired, provide an optional flag for strict mode or warnings if inheritance is not possible.

### Refactor/Roadmap

Is this addressed by any planned refactor or milestone? If not, could you consider it for a future release?

Alix-007:
## Summary
- inherit missing model_list fields from the matching legacy provider config during LoadConfig
- preserve explicit model_list overrides so user-provided values still win
- cover litellm, bare openai models, and copilot alias inheritance with regression tests

Fixes #1635

## Testing
- go test ./pkg/config -run 'TestLoadConfig_(ModelListInheritsProviderFields|ModelListKeepsExplicitOverrides|ModelListInheritsOpenAIProviderForBareModel|ModelListInheritsCopilotProviderAlias)$'
- go test ./pkg/config

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -555,9 +555,51 @@ func TestLoadConfig_WebToolsProxy(t *testing.T) {
 	}
 }
 
-func TestLoadConfig_ModelListInheritsProviderFields(t *testing.T) {
+func loadModelConfigFromJSON(t *testing.T, configJSON, modelName string) *ModelConfig {
+	t.Helper()
+
 	tmpDir := t.TempDir()
 	configPath := filepath.Join(tmpDir, "config.json")
+	if err := os.WriteFile(configPath, []byte(configJSON), 0o600); err != nil {
+		t.Fatalf("os.WriteFile() error: %v", err)
+	}
+
+	cfg, err := LoadConfig(configPath)
+	if err != nil {
+		t.Fatalf("LoadConfig() error: %v", err)
+	}
+
+	modelCfg, err := cfg.GetModelConfig(modelName)
+	if err != nil {
+		t.Fatalf("GetModelConfig() error: %v", err)
+	}
+
+	return modelCfg
+}
+
+func assertModelConfigProviderInheritance(
+	t *testing.T,
+	modelCfg *ModelConfig,
+	wantAPIKey, wantAPIBase, wantProxy string,
+	wantRequestTimeout int,
+) {
+	t.Helper()
+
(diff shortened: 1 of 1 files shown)


Alix-007:
Closing this as superseded by #1786, which already merged the linked #1635 fix. Keeping both open would only duplicate the same config inheritance lane. Thanks for the reviewable branch and tests here.

PR #1697, opened 2026-03-17
feat(provider): add per-model cooldown scope option

### LINKED ISSUE #1612: [Feature] Add option for per-model cooldown for specific providers
## 🎯 The Goal / Use Case

Some providers (such as LiteLLM router, Ollama or other multi-model backends) expose multiple models under a single provider interface. The current cooldown mechanism operates at the **provider level**, which can cause unnecessary downtime if only one model is problematic. However, for other providers, the provider-level cooldown is still preferred for simplicity.

## 💡 Proposed Solution

- Add a configuration option or flag that enables **per-model cooldown** instead of provider-level cooldown for designated providers.
- When this option is set, only the affected (provider, model) pair should enter cooldown upon error, while the rest models of that provider remain active.
- Default behaviour should remain provider-level cooldown unless this option is enabled.

- Extend provider or config to accept a per-model-cooldown flag or option (e.g. `provider.cooldown_strategy: "per-model"` or similar).
- Update `CooldownTracker` or similar code to check this flag and key cooldowns accordingly.
- Ensure fallback chain logic respects this flag per provider.

- [ ] This is a Core Feature

## 🔄 Alternatives Considered

- Keeping only provider-level cooldown (as today), at cost of unnecessary downtime.
- Only enabling per-model cooldown globally (losing flexibility for simple providers).

## 💬 Additional Context

- Useful especially for provider acting as router (e.g., LiteLLM, Ollama) to reduce user-facing downtime.
- This does NOT change the default - only when the flag/option is set for a provider.
- Changing cooldown granularity as a configurable option keeps backwards compatibility and offers greatest flexibility.

Alix-007:
Fixes #1612

## Summary
- add `model_list[*].cooldown_strategy` to opt specific backends into per-model cooldown tracking
- keep provider-wide cooldowns as the default behavior
- attach resolved cooldown keys to fallback candidates during agent initialization
- add regression tests for config validation, candidate resolution, and sibling-model fallback behavior

## Validation
- `go test ./pkg/providers -run 'TestFallback_(CooldownSkip|PerModelCooldownDoesNotSkipSiblingModel|DefaultProviderCooldownSkipsSiblingModel)$'`\n- `go test ./pkg/agent -run 'TestNewAgentInstance_(ResolveCandidatesFromModelListAlias|ResolvePerModelCooldownKeys)$'`\n- `go test ./pkg/config -run 'Test(ModelConfig_Validate|Config_ValidateModelList)$'`\n- `go test ./pkg/providers`\n- `go test ./pkg/agent`\n- `go test ./pkg/config`\n\n## Notes\n- `go test ./...` still fails on upstream baseline in `cmd/picoclaw/internal/onboard` and `pkg/tools`, unrelated to this change.

### DIFF
--- a/pkg/agent/instance.go
+++ b/pkg/agent/instance.go
@@ -185,6 +185,10 @@ func NewAgentInstance(
 				logger.WarnCF("agent", "Routing light model config invalid; routing disabled",
 					map[string]any{"light_model": rc.LightModel, "agent_id": agentID, "error": err.Error()})
 			} else {
+				// Reuse the resolved candidate's canonical provider/model form so
+				// routing init stays aligned with model-list resolution behavior.
+				lightModelCfg.Model = providers.ModelKey(resolved[0].Provider, resolved[0].Model)
+
 				lp, _, err := providers.CreateProviderFromConfig(lightModelCfg)
 				if err != nil {
 					logger.WarnCF("agent", "Routing light model provider init failed; routing disabled",
--- a/pkg/agent/instance.go
+++ b/pkg/agent/instance.go
@@ -167,8 +167,9 @@ func NewAgentInstance(
 		summarizeTokenPercent = 75
 	}
 
-	// Resolve fallback candidates
+	// Resolve fallback candidates.
 	candidates := resolveModelCandidates(cfg, defaults.Provider, model, fallbacks)
+	candidates = applyCooldownKeys(cfg, candidates)
 
 	// Model routing setup: pre-resolve light model candidates at creation time
 	// to avoid repeated model_list lookups on every incoming message.
@@ -192,7 +193,7 @@ func NewAgentInstance(
 						LightModel: rc.LightModel,
 						Threshold:  rc.Threshold,
 					})
-					lightCandidates = resolved
+					lightCandidates = applyCooldownKeys(cfg, resolved)
 					lightProvider = lp
 				}
 			}
@@ -225,6 +226,64 @@ func NewAgentInstance(
 		Router:                    router,
 		LightCandidates:           lightCandidates,
 		LightProvider:             lightProvider,
+	}
+}
+
+func applyCooldownKeys(cfg *config.Config, candidates []providers.FallbackCandidate) []providers.FallbackCandidate {
+	if len(candidates) == 0 {
+		return candidates
+	}
+
+	resolved := make([]providers.FallbackCandidate, len(candidates))
+	copy(resolved, candidates)
+
+	for i := range resolved {
+		resolved[i].CooldownKey = resolveCooldownKey(cfg, resolved[i])
+	}
+
+	return resolved
+}
+
+func resolveCooldownKey(cfg *config.Config, candidate providers.FallbackCandidate) string {
+	if candidate.Provider == "" {
+		return ""
+	}
+
+	if cfg != nil && candidateUsesPerModelCooldown(cfg, candidate) {
+		return providers.ModelKey(candidate.Provider, candidate.Model)
+	}
+
+	return candidate.Provider
+}
+
+func candidateUsesPerModelCooldown(cfg *config.Config, candidate providers.FallbackCandidate) bool {
+	if cfg == nil {
+		return false
+	}
+
+	candidateKey := providers.ModelKey(candidate.Provider, candidate.Model)
+	for i := range cfg.ModelList {
+		ref := providers.ParseModelRef(cfg.ModelList[i].Model, "openai")
+		if ref == nil {
+			continue
+		}
+		if providers.ModelKey(ref.Provider, ref.Model) != candidateKey {
+			continue
+		}
+		if normalizeCooldownStrategy(cfg.ModelList[i].CooldownStrategy) == "model" {
+			return true
+		}
+	}
+
+	return false
+}
+
+func normalizeCooldownStrategy(strategy string) string {
+	switch strings.ReplaceAll(strings.ToLower(strings.TrimSpace(strategy)), "_", "-") {
+	case "model", "per-model":
+		return "model"
+	default:
+		return "provider"
 	}
 }
 
--- a/pkg/agent/instance.go
+++ b/pkg/agent/instance.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/sipeed/picoclaw/pkg/config"
@@ -239,7 +240,7 @@ func applyCooldownKeys(cfg *config.Config, candidates []providers.FallbackCandid
 	copy(resolved, candidates)
 
 	for i := range resolved {
-		resolved[i].CooldownKey = resolveCooldownKey(strategyLookup, resolved[i])
+		resolved[i].CooldownKey = resolveCooldownKey(strategyLookup, resolved[i], resolved)
 	}
 
 	return resolved
@@ -267,7 +268,11 @@ func buildCooldownStrategyLookup(cfg *config.Config) map[string]string {
 	return strategyLookup
 }
 
-func resolveCooldownKey(strategyLookup map[string]string, candidate providers.FallbackCandidate) string {
+func resolveCooldownKey(
+	strategyLookup map[string]string,
+	candidate providers.FallbackCandidate,
+	candidates []providers.FallbackCandidate,
+) string {
 	if strings.TrimSpace(candidate.Provider) == "" {
 		if strings.TrimSpace(candidate.Model) == "" {
 			return ""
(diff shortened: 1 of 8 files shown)


yinwm (changes requested):
## PR #1697 Code Review: Per-Model Cooldown Scope

Thanks for this feature! The design is solid and the implementation is clean. I have a few concerns that should be addressed before merging.

---

### 🔴 Critical Issues

#### 1. Light Model Routing - Cooldown Key May Not Resolve Correctly

**Location**: `pkg/agent/instance.go:204-212`

```go
lightModelCfg := providers.ModelConfig{Primary: rc.LightModel}
resolved := providers.ResolveCandidatesWithLookup(lightModelCfg, defaults.Provider, resolveFromModelList)
if len(resolved) > 0 {
    router = routing.New(...)
    lightCandidates = applyCooldownKeys(cfg, resolved)  // ← Potential issue
}
```

The `applyCooldownKeys` function calls `candidateUsesPerModelCooldown`, which iterates through `cfg.ModelList` to find a matching `CooldownStrategy`. However, there's a potential mismatch:

- `rc.LightModel` might be an alias (e.g., `"light-model"`) or a direct model string
- `resolveFromModelList` resolves aliases to actual model strings
- The matching logic in `candidateUsesPerModelCooldown` uses `ParseModelRef` on `ModelList[i].Model`

If the format returned by `resolveFromModelList` differs slightly from what's stored in `ModelList[i].Model` (case sensitivity, extra spaces, provider aliases), the match will fail.

**Recommendation**: Add a test case to verify light_model cooldown key resolution works correctly.

---

#### 2. O(n×m) Complexity in `applyCooldownKeys`

**Location**: `pkg/agent/instance.go:268-287`

```go
func applyCooldownKeys(cfg *config.Config, candidates []providers.FallbackCandidate) {
    for i := range resolved {
        resolved[i].CooldownKey = resolveCooldownKey(cfg, resolved[i])  // For each candidate
    }
}

func candidateUsesPerModelCooldown(...) bool {
    for i := range cfg.ModelList {  // Iterates entire ModelList
        ref := providers.ParseModelRef(cfg.ModelList[i].Model, "openai")  // Re-parses every time
        // ...
    }
}
```

**Problem**:
- N candidates × M ModelList entries = O(n×m)
- Each iteration calls `ParseModelRef`, re-parsing the same strings

**Impact**: With 20 models and 5 fallback candidates, this performs 100 string parsing operations. While this happens at agent creation (not runtime), it's still unnecessary overhead.

**Recommendation**: Pre-build a lookup map:

```go
func applyCooldownKeys(cfg *config.Config, candidates []providers.FallbackCandidate) []providers.FallbackCandidate {
    if len(candidates) == 0 || cfg == nil {
        return candidates
    }

    // Pre-build lookup map to avoid repeated parsing
    strategyMap := make(map[string]string)  // modelKey -> normalized strategy
    for i := range cfg.ModelList {
        ref := providers.ParseModelRef(cfg.ModelList[i].Model, "openai")
        if ref == nil {
            continue
        }
        key := providers.ModelKey(ref.Provider, ref.Model)
        strategyMap[key] = normalizeCooldownStrategy(cfg.ModelList[i].CooldownStrategy)
    }

    // Now O(n) instead of O(n×m)
    for i := range resolved {
        resolved[i].CooldownKey = resolveCooldownKeyWithMap(resolved[i], strategyMap)
    }
    return resolved
}
```

---

### 🟡 Medium Issues

#### 3. Duplicate Normalize Logic

Two places have nearly identical normalize logic:

- `pkg/agent/instance.go:295-301` - `normalizeCooldownStrategy()` returns string
- `pkg/config/config.go:632-638` - `isValidCooldownStrategy()` returns bool

**Recommendation**: Export a single `NormalizeCooldownStrategy` function from `config` package and reuse it.

---

#### 4. Empty Provider Edge Case Inconsistency

**`pkg/agent/instance.go:259-261`**:
```go
if candidate.Provider == "" {
    return ""  // Returns empty string
}
```

**`pkg/providers/fallback.go:45-51`**:
```go
func (c FallbackCandidate) cooldownKey() string {
    if strings.TrimSpace(c.CooldownKey) != "" {
        return c.CooldownKey
    }
    return c.Provider  // What if Provider is also empty?
}
```

If both are empty, an empty string is returned, potentially causing all empty-provider models to share cooldown state.

**Recommendation**: Add defensive check or ensure this case never occurs.

---

#### 5. Config Validation Error Message Inconsistency

Error message says `per_model` but validation normalizes to `per-model`:

```go
return fmt.Errorf("cooldown_strategy must be one of: provider, model, per-model, per_model")
```

**Recommendation**: Clarify the message:
```go
return fmt.Errorf("cooldown_strategy must be one of: provider, model, per-model (or per_model)")
```

---

### 🟢 Minor Suggestions

#### 6. Missing Test Coverage

Consider adding tests for:
- Light model per-model cooldown scenario
- Mixed scenario (some models use per-model, others use provider)
- Empty CooldownStrategy default behavior

#### 7. Documentation Could Be More Detailed

```go
// CooldownStrategy controls the scope of cooldown tracking for this model.
// - "provider" (default): cooldown is shared across all models in the same provider
// - "model" or "per-model": cooldown is isolated to this specific model
// Use per-model when different models in the same provider have independent rate limits.
CooldownStrategy string `json:"cooldown_strategy,omitempty"`
```

---

### Summary

| Severity | Count |
|----------|-------|
| 🔴 Critical | 2 |
| 🟡 Medium | 3 |
| 🟢 Minor | 2 |

**Recommendation**: Please address the critical issues (especially the O(n×m) complexity and light_model test) before merging. The medium/minor issues can be addressed in a follow-up PR if needed.

Overall, this is a well-designed feature. Thank you for the clean implementation! 🙏

Alix-007:
Addressing the requested changes from the March 17, 2026 review:

- Reworked cooldown resolution to pre-build a `modelKey -> strategy` lookup map once at agent creation time, so candidate cooldown assignment is now O(n+m) instead of O(n×m).
- Added a light-model routing regression test that resolves `light_model` via alias and verifies the per-model cooldown key still resolves correctly.
- Moved cooldown strategy normalization into `config.NormalizeCooldownStrategy()` so config validation and agent resolution use the same canonicalization logic.
- Tightened the fallback cooldown key edge case so empty-provider candidates do not silently collapse onto the same shared key.
- Clarified the validation error text for `cooldown_strategy` aliases.

Local verification:
- `go test ./pkg/agent -run 'TestNewAgentInstance_(ResolveCandidatesFromModelListAlias|ResolvePerModelCooldownKeys|ResolveLightModelPerModelCooldownKeys)$'`
- `go test ./pkg/config -run 'Test(ModelConfig_Validate|NormalizeCooldownStrategy|Config_ValidateModelList)$'`
- `go test ./pkg/providers -run 'TestFallback_(CooldownSkip|PerModelCooldownDoesNotSkipSiblingModel|DefaultProviderCooldownSkipsSiblingModel)$'`
- `go test ./pkg/agent`
- `go test ./pkg/config`
- `go test ./pkg/providers`

Alix-007:
Rebased this branch onto the current `main` and force-updated the PR head.

The reviewer-requested changes from Mar 17 are still in place, and I added one follow-up compatibility fix during the rebase so the new provider-shared cooldown behavior does not break existing multi-key failover chains.

Local verification after the refresh:
- `go test ./pkg/agent -run 'TestNewAgentInstance_(ResolveCandidatesFromModelListAlias|ResolvePerModelCooldownKeys|ResolveLightModelPerModelCooldownKeys)$'`
- `go test ./pkg/config -run 'Test(ModelConfig_Validate|NormalizeCooldownStrategy|Config_ValidateModelList)$'`
- `go test ./pkg/providers`

GitHub PR checks are green again, so this should be ready for another look.

Alix-007:
Thanks again for the detailed review. The requested items are addressed in the current head, the branch has been refreshed onto `main`, and the GitHub checks are green again. When you have time, I would appreciate another look.

yinwm (approved):
LGTM

Alix-007:
Addressed the follow-up gap on the runtime switch path:  now reapplies cooldown-key scoping before replacing candidates, and I added a regression test covering per-model cooldown keys after a switch. Targeted agent tests are green, and the PR checks are green as well.

Alix-007:
Addressed the follow-up gap on the runtime switch path: switch model now reapplies cooldown-key scoping before replacing candidates, and I added a regression test covering per-model cooldown keys after a switch. Targeted agent tests are green, and the PR checks are green as well.

Alix-007:
This PR is back to a clean state and all checks are green after the latest refresh. It is currently approved and ready to merge when convenient. Thanks for the review.

Alix-007:
Rebased this branch onto the latest `main` to resolve the new conflicts.

Current head is `d4a19cf53eba5994667e9c3c160fecd8e4115bc6`. The PR scope stays limited to per-model cooldown handling, with the refresh preserving:
- cooldown-key scoping for the routed light-model path
- cooldown-key scoping after `/switch model`
- multi-key isolation behavior

Targeted verification on the refreshed head:
- `go test ./pkg/agent -run 'TestNewAgentInstance_(ResolvePerModelCooldownKeys|ResolveLightModelPerModelCooldownKeys|ResolveMultiKeyCooldownKeys)$|TestProcessMessage_SwitchModelPreservesPerModelCooldownKeys'`\n- `go test ./pkg/config -run 'Test(ModelConfig_Validate|NormalizeCooldownStrategy|Config_ValidateModelList)$'`\n- `go test ./pkg/providers -run 'TestFallback_(CooldownSkip|PerModelCooldownDoesNotSkipSiblingModel|DefaultProviderCooldownSkipsSiblingModel)$'`\n\nThe focused checks are green again. Please recheck when convenient.

Alix-007:
Applied a narrow follow-up for the new CI regression on the current head: the reasoning publish goroutine was inheriting a turn context that gets canceled as soon as the turn returns, which could race and drop reasoning-channel output. The fix now launches that best-effort publish with context.WithoutCancel(turnCtx) so turn-scoped values are preserved without the premature cancel edge.\n\nRechecked targeted coverage on this head:\n- go test -tags goolm,stdjson ./pkg/agent -run 'TestNewAgentInstance_(ResolvePerModelCooldownKeys|ResolveLightModelPerModelCooldownKeys|ResolveMultiKeyCooldownKeys)$|TestProcessMessage_SwitchModelPreservesPerModelCooldownKeys' -count=1\n- go test -tags goolm,stdjson ./pkg/agent -run TestProcessMessage_PublishesReasoningContentToReasoningChannel -count=200

PR #1060, opened 2026-03-04
fix(cron): publish agent response for deliver=false cron tasks

AaronJny:
## 📝 Description

Fix `deliver=false` cron tasks silently discarding the LLM response.

`ExecuteJob` calls `ProcessDirectWithChannel` which returns the LLM response,
but the response was discarded with `_ = response`. The comment claimed
"Will be sent by AgentLoop" but `ProcessDirectWithChannel` bypasses the
`Run()` inbound loop that would normally publish the response.

Now the response is published via MessageBus explicitly, matching the pattern
used by the `deliver=true` and `command` execution paths.

```diff
-// Response is automatically sent via MessageBus by AgentLoop
-_ = response // Will be sent by AgentLoop
+// Send agent response to channel (ProcessDirectWithChannel does not
+// publish on its own — only the Run() inbound loop does that).
+if response != "" {
+    pubCtx, pubCancel := context.WithTimeout(context.Background(), 5*time.Second)
+    defer pubCancel()
+    t.msgBus.PublishOutbound(pubCtx, bus.OutboundMessage{
+        Channel: channel,
+        ChatID:  chatID,
+        Content: response,
+    })
+}
```

## 🗣️ Type of Change
- [x] 🐞 Bug fix (non-breaking change which fixes an issue)

## 🤖 AI Code Generation
- [x] 🛠️ Mostly AI-generated (AI draft, Human verified/modified)

## 🔗 Related Issue

Fixes #1058

## 📚 Technical Context (Skip for Docs)
- **Reasoning:** `ProcessDirectWithChannel` → `processMessage` → `runAgentLoop`
  with `SendResponse: false`. The normal `Run()` loop publishes the response
  after `processMessage` returns, but `ProcessDirectWithChannel` bypasses `Run()`
  entirely, so the response must be published by the caller.

## 🧪 Test Environment
- **Hardware:** Linux server (runtime), MacBook Apple Silicon (development)
- **OS:** Ubuntu Linux (runtime), macOS (development & testing)
- **Model/Provider:** GPT-5.2, GPT-5.3-codex
- **Channels:** Discord

## ☑️ Checklist
- [x] My code/docs follow the style of this project.
- [x] I have performed a self-review of my own changes.
- [x] I have updated the documentation accordingly.

PR #1699, opened 2026-03-17
fix(cron): publish deliver=false agent replies

### LINKED ISSUE #1058: [BUG] Cron deliver=false tasks silently discard LLM response — messages never reach channel

Alix-007:
## Description

When a cron job runs with `deliver=false`, PicoClaw bypasses the normal inbound loop and `ProcessDirectWithChannel` returns the final agent text directly. The old path discarded that return value, so the user never saw the reply.

This change publishes the returned agent text to the outbound message bus when it is non-empty, while keeping blank/whitespace-only responses silent.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor

## AI Code Generation
- [ ] Fully AI-generated
- [x] Mostly AI-generated
- [ ] Mostly Human-written

## Related Issue
Fixes #1058

## Technical Context
- publish the `ProcessDirectWithChannel` return value for `deliver=false` cron jobs
- keep blank replies silent to avoid noisy empty messages
- add `ExecuteJob` regression coverage for both behaviors

## Test Environment
- Hardware: PC
- OS: Linux
- Model/Provider: N/A
- Channels: N/A

## Evidence
- `go test ./pkg/tools -run 'TestCronTool_'`
- `go test ./pkg/tools -run 'TestCron'`
- `go test ./cmd/picoclaw/internal/cron`

## Checklist
- [x] I reviewed the change.
- [x] I updated or added tests.

### DIFF
--- a/pkg/tools/cron.go
+++ b/pkg/tools/cron.go
@@ -374,7 +374,18 @@ func (t *CronTool) ExecuteJob(ctx context.Context, job *cron.CronJob) string {
 		return fmt.Sprintf("Error: %v", err)
 	}
 
-	// Response is automatically sent via MessageBus by AgentLoop
-	_ = response // Will be sent by AgentLoop
+	// ProcessDirectWithChannel bypasses the normal inbound loop, so publish the
+	// returned final text here when the agent produced one.
+	if strings.TrimSpace(response) == "" {
+		return "ok"
+	}
+
+	pubCtx, pubCancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer pubCancel()
+	t.msgBus.PublishOutbound(pubCtx, bus.OutboundMessage{
+		Channel: channel,
+		ChatID:  chatID,
+		Content: response,
+	})
 	return "ok"
 }
--- a/pkg/tools/cron_test.go
+++ b/pkg/tools/cron_test.go
@@ -12,6 +12,39 @@ import (
 	"github.com/sipeed/picoclaw/pkg/cron"
 )
 
+type stubJobExecutor struct {
+	response string
+	err      error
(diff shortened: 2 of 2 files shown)


Alix-007:
Closing this to reduce duplicate review load in the crowded cron-response lane. The linked issue #1058 already has multiple open carriers (#1060, #1839), and this branch has stayed unreviewed without becoming the clear maintainer-preferred path. If this exact narrow fix is needed later, it should come back from a fresh clean base.

PR #1700, opened 2026-03-17
fix(spawn): use target agent model

### LINKED ISSUE #1322: [BUG] spawn ignores target agent model and always uses caller model

Alix-007:
## Description

When `spawn` receives an `agent_id`, the spawned subagent should run with that target agent's configured model. The previous path always reused the caller agent's default model because `SubagentManager` had no way to resolve the target agent before entering the tool loop.

This change adds a target-agent model resolver hook to `SubagentManager` and wires it from the agent registry so async `spawn` requests use the target agent model when one is configured, while still falling back to the caller/default model if the target cannot be resolved.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor

## AI Code Generation
- [ ] Fully AI-generated
- [x] Mostly AI-generated
- [ ] Mostly Human-written

## Related Issue
Fixes #1322

## Technical Context
- add `SetAgentModelResolver` to `SubagentManager`
- resolve `agent_id -> target model` from the agent registry when registering spawn tools
- add regression coverage for manager-level and `spawn` async execution paths

## Test Environment
- Hardware: PC
- OS: Linux
- Model/Provider: mock provider
- Channels: N/A

## Evidence
- `go test ./pkg/tools -run 'Test(SubagentManager_SetLLMOptions_AppliesToRunToolLoop|SubagentManager_RunTask_UsesResolvedTargetAgentModel|SpawnTool_ExecuteAsync_UsesTargetAgentModel|SpawnTool_Execute_ValidTask|SpawnTool_Execute_EmptyTask|SpawnTool_Execute_NilManager)$'`
- `go test ./pkg/agent -run 'TestHandleCommand'`

## Checklist
- [x] I reviewed the change.
- [x] I added regression tests.

### DIFF
--- a/pkg/agent/loop.go
+++ b/pkg/agent/loop.go
@@ -244,6 +244,13 @@ func registerSharedTools(
 			// spawn_status which are added below — preventing recursive
 			// subagent spawning.
 			subagentManager.SetTools(agent.Tools.Clone())
+			subagentManager.SetAgentModelResolver(func(targetAgentID string) (string, bool) {
+				targetAgent, ok := registry.GetAgent(targetAgentID)
+				if !ok || targetAgent == nil || targetAgent.Model == "" {
+					return "", false
+				}
+				return targetAgent.Model, true
+			})
 			if spawnEnabled {
 				spawnTool := tools.NewSpawnTool(subagentManager)
 				currentAgentID := agentID
--- a/pkg/tools/spawn_test.go
+++ b/pkg/tools/spawn_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"strings"
 	"testing"
+	"time"
 )
 
 func TestSpawnTool_Execute_EmptyTask(t *testing.T) {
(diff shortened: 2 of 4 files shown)


Alix-007:
Thanks for the careful review and the LGTM. I noted the simplification suggestion and the optional fallback test ideas. I kept this PR scoped to the minimal fix for #1322 for now, but I can fold in follow-up coverage if a maintainer wants it in this branch.

Alix-007:
Closing this to reduce duplicate review load. Current main already resolves the target agent model in the active spawned subturn path, so the primary user-facing fix from this PR is effectively superseded. The remaining non-duplicate part is a much narrower legacy SubagentManager fallback parity change, which should come back later as a fresh tiny follow-up if it is still needed.

PR #1710, opened 2026-03-17
fix(cron): update test to use OutboundChan instead of removed SubscribeOutbound

liuy:
The SubscribeOutbound method was removed in commit 9c31b0c (`fix: Fixed the bug where the bus was closed...`) but pkg/tools/cron_test.go was not updated to use the new OutboundChan() API.

This PR updates the test to use the new API:
- `SubscribeOutbound(ctx)` → `OutboundChan()` + select

Test: All tests in pkg/tools pass.

### DIFF
--- a/pkg/tools/cron_test.go
+++ b/pkg/tools/cron_test.go
@@ -226,9 +226,12 @@ func TestCronTool_ExecuteJobPublishesErrorWhenExecDisabled(t *testing.T) {
 	ctx, cancel := context.WithTimeout(context.Background(), time.Second)
 	defer cancel()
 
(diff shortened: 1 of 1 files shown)


afjcjsbx:
Thanks for the fix! LGTM

Orgmar:
@liuy Thanks for catching the stale test after the SubscribeOutbound removal. Migrating to OutboundChan() with the select pattern keeps things in sync with the current API. Appreciate you making sure the test suite stays green.

We have a **PicoClaw Dev Group** on Discord for contributors to collaborate more directly. If you'd like to join, send an email to `support@sipeed.com` with subject `[Join PicoClaw Dev Group] liuy` and we'll send you the invite!

PR #1725, opened 2026-03-18
feat: add /reload to gateway api and command

cytown:
## 📝 Description

add /reload to gateway api and command

## 🗣️ Type of Change
- [X] ✨ New feature (non-breaking change which adds functionality)
- [ ] 📖 Documentation update

## 🤖 AI Code Generation
- [X] 🛠️ Mostly AI-generated (AI draft, Human verified/modified)

- **Reference URL:**
- **Reasoning:**

## 🧪 Test Environment
- **Hardware:** 
- **OS:** 
- **Model/Provider:** 
- **Channels:** 

## 📸 Evidence (Optional)
<details>

ccc8fe71960d8550719b9f880b41001a
48833813346244c608f3843712c68f7c

</details>

## ☑️ Checklist

### DIFF
--- a/pkg/commands/cmd_reload.go
+++ b/pkg/commands/cmd_reload.go
@@ -14,7 +14,7 @@ func reloadCommand() Definition {
 			if err := rt.ReloadConfig(); err != nil {
 				return req.Reply("Failed to reload configuration: " + err.Error())
 			}
-			return req.Reply("Configuration reloaded successfully!")
+			return req.Reply("Config reload triggered!")
 		},
 	}
 }
--- a/pkg/gateway/gateway.go
+++ b/pkg/gateway/gateway.go
@@ -7,6 +7,7 @@ import (
 	"os/signal"
 	"path/filepath"
 	"sync"
+	"sync/atomic"
 	"syscall"
 	"time"
 
@@ -55,6 +56,7 @@ type services struct {
 	DeviceService    *devices.Service
 	HealthServer     *health.Server
 	manualReloadChan chan struct{}
+	reloading        atomic.Bool
 }
 
 type startupBlockedProvider struct {
@@ -122,11 +124,16 @@ func Run(debug bool, configPath string, allowEmptyStartup bool) error {
 	manualReloadChan := make(chan struct{}, 1)
 	runningServices.manualReloadChan = manualReloadChan
 	reloadTrigger := func() error {
+		if !runningServices.reloading.CompareAndSwap(false, true) {
+			return fmt.Errorf("reload already in progress")
+		}
 		select {
 		case manualReloadChan <- struct{}{}:
 			return nil
 		default:
-			return fmt.Errorf("reload already in progress")
+			// Should not happen, but reset flag if channel is full
+			runningServices.reloading.Store(false)
+			return fmt.Errorf("reload already queued")
 		}
 	}
 	runningServices.HealthServer.SetReloadFunc(reloadTrigger)
@@ -158,7 +165,11 @@ func Run(debug bool, configPath string, allowEmptyStartup bool) error {
 			shutdownGateway(runningServices, agentLoop, provider, true)
 			return nil
 		case newCfg := <-configReloadChan:
-			err := handleConfigReload(ctx, agentLoop, newCfg, &provider, runningServices, msgBus, allowEmptyStartup)
+			if !runningServices.reloading.CompareAndSwap(false, true) {
+				logger.Warn("Config reload skipped: another reload is in progress")
+				continue
+			}
+			err := executeReload(ctx, agentLoop, newCfg, &provider, runningServices, msgBus, allowEmptyStartup)
 			if err != nil {
 				logger.Errorf("Config reload failed: %v", err)
 			}
@@ -167,13 +178,15 @@ func Run(debug bool, configPath string, allowEmptyStartup bool) error {
 			newCfg, err := config.LoadConfig(configPath)
 			if err != nil {
 				logger.Errorf("Error loading config for manual reload: %v", err)
+				runningServices.reloading.Store(false)
 				continue
 			}
 			if err = newCfg.ValidateModelList(); err != nil {
 				logger.Errorf("Config validation failed: %v", err)
+				runningServices.reloading.Store(false)
 				continue
 			}
-			err = handleConfigReload(ctx, agentLoop, newCfg, &provider, runningServices, msgBus, allowEmptyStartup)
+			err = executeReload(ctx, agentLoop, newCfg, &provider, runningServices, msgBus, allowEmptyStartup)
 			if err != nil {
 				logger.Errorf("Manual reload failed: %v", err)
 			} else {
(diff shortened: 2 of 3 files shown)


pkg/commands/cmd_reload.go
```diff
@@ -0,0 +1,20 @@
...
+func reloadCommand() Definition {
+	return Definition{
+		Name:        "reload",
+		Description: "Reload the configuration file",
+		Usage:       "/reload",
+		Handler: func(_ context.Context, req Request, rt *Runtime) error {
+			if rt == nil || rt.ReloadConfig == nil {
+				return req.Reply(unavailableMsg)
+			}
+			if err := rt.ReloadConfig(); err != nil {
```
wj-xiao, line 14: The `/reload` command reports success too early.
`ReloadConfig()` only enqueues a reload request; the actual `LoadConfig` / validation / provider recreation happens later in the gateway loop. As a result, users can receive “Configuration reloaded successfully!” even when the reload later fails.

pkg/gateway/gateway.go
```diff
@@ -117,6 +118,20 @@ func Run(debug bool, configPath string, allowEmptyStartup bool) error {
 		return err
 	}
 
+	// Setup manual reload channel for /reload endpoint
+	manualReloadChan := make(chan struct{}, 1)
+	runningServices.manualReloadChan = manualReloadChan
+	reloadTrigger := func() error {
```
wj-xiao, line 126: The `reload already in progress` guard does not protect the whole reload lifecycle. Once the main loop reads from manualReloadChan, the channel is empty again while `handleConfigReload(...)` is still running, so additional reload requests can still be queued and trigger repeated stop/restart cycles.

pkg/health/server.go
```diff
@@ -12,11 +12,13 @@ import (
...
-	ready     bool
-	checks    map[string]Check
-	startTime time.Time
+	server     *http.Server
+	mu         sync.RWMutex
+	ready      bool
+	checks     map[string]Check
+	startTime  time.Time
+	reloadFunc func() error
+	reloadMu   sync.Mutex
```
wj-xiao: The Server struct already houses a mu `sync.RWMutex` which is used for data access concurrency control. However, this PR introduces a dedicated reloadMu `sync.Mutex` solely to protect the getter/setter for reloadFunc.
You can eliminate data structure redundancy by reusing the existing `s.mu.Lock()` to protect reloadFunc.

pkg/gateway/gateway.go
```diff
@@ -426,17 +477,16 @@ func restartServices(
 	}
 
 	addr := fmt.Sprintf("%s:%d", cfg.Gateway.Host, cfg.Gateway.Port)
-	runningServices.HealthServer = health.NewServer(cfg.Gateway.Host, cfg.Gateway.Port)
+	// Reuse existing HealthServer to preserve reloadFunc
+	if runningServices.HealthServer == nil {
+		runningServices.HealthServer = health.NewServer(cfg.Gateway.Host, cfg.Gateway.Port)
```
wj-xiao, line 493: The `HealthServer` internally holds an `s.server` which is bound to the old `Gateway.Host` and `Gateway.Port`. If the new configuration alters the Host/Port, the old `HealthServer` will retain the stale address internally.
While currently its methods simply execute `RegisterOnMux` and it never actually calls its own `ListenAndServe`, no real listening exception is triggered, this is still not a good architectural design.

pkg/gateway/gateway.go
```diff
@@ -143,10 +165,35 @@ func Run(debug bool, configPath string, allowEmptyStartup bool) error {
 			shutdownGateway(runningServices, agentLoop, provider, true)
 			return nil
 		case newCfg := <-configReloadChan:
+			if !runningServices.reloading.CompareAndSwap(false, true) {
+				logger.Warn("Config reload skipped: another reload is in progress")
+				continue
+			}
 			err := handleConfigReload(ctx, agentLoop, newCfg, &provider, runningServices, msgBus, allowEmptyStartup)
```
wj-xiao: If `handleConfigReload` encounters a severe panic or unhandled exception, `runningServices.reloading.Store(false)` will never be executed.
This permanently deadlocks the `reloading` state, meaning any subsequent attempts to trigger a reload will continuously be rejected with "reload already in progress".
Wrap the code block safeguarded by the `reloading` atomic lock inside a dedicated function and use `defer runningServices.reloading.Store(false)` to reliably clear the lock.

pkg/gateway/gateway.go
```diff
@@ -143,10 +165,37 @@ func Run(debug bool, configPath string, allowEmptyStartup bool) error {
...
 			}
+			go func() {
+				defer runningServices.reloading.Store(false)
+				err := handleConfigReload(ctx, agentLoop, newCfg, &provider, runningServices, msgBus, allowEmptyStartup)
+				if err != nil {
+					logger.Errorf("Config reload failed: %v", err)
+				}
+			}()
+		case <-manualReloadChan:
+			go func() {
```
wj-xiao: `Gateway.Run` heavily utilizes a single-threaded `for { select { ... } }` loop pattern. Moving [handleConfigReload](file:///Users/xiao/github/picoclaw/pkg/gateway/gateway.go#306-367) into a concurrent `go func()` shatters its single-threaded safety guarantees and introduces severe concurrency bugs:

1. **Concurrent Interface Reads/Writes:**
   Inside [handleConfigReload](file:///Users/xiao/github/picoclaw/pkg/gateway/gateway.go#306-367), there's an assignment: `*providerRef = newProvider`. This mutates the local `provider` interface variable within `Gateway.Run`. In Go, **interface assignment is NOT an atomic operation**. If a user presses `Ctrl+C` (sending a `SIGINT`/`SIGTERM`) while the reload is executing, the main loop will instantly handle the `case <-sigChan:` and pass `provider` to [shutdownGateway()](file:///Users/xiao/github/picoclaw/pkg/gateway/gateway.go#288-305). Reading the interface while it is being partially written across two words causes a massive Go runtime Panic (e.g., `invalid memory address or nil pointer dereference` or `interface conversion error`), crashing the entire Gateway process.

2. **Concurrent Mutation of Internal Service Pointers:**
   [handleConfigReload](file:///Users/xiao/github/picoclaw/pkg/gateway/gateway.go#306-367) restarts services (which mutates pointers like `runningServices.ChannelManager`). Simultaneously, a concurrent [shutdownGateway](file:///Users/xiao/github/picoclaw/pkg/gateway/gateway.go#288-305) sequence might read these exact same pointers to invoke `.StopAll()`. This is another classic Data Race scenario waiting to crash the application structure.

As discussed previously, the HTTP request sent to `/reload` is already structurally **non-blocking** (because `reloadTrigger()` instantaneously pushes a signal to `manualReloadChan` and completes). Thus, there is **absolutely no need** to spawn a `go func()` within the main `Gateway.Run` loop.

Simply **remove the `go func() { ... }()` wrapper** and execute [handleConfigReload](file:///Users/xiao/github/picoclaw/pkg/gateway/gateway.go#306-367) synchronously within the `select` loop:
- Even if the synchronous reload takes a couple of seconds, making the main loop briefly unresponsive guarantees configurations are safely swapped.
- If a user presses `Ctrl+C` during this period, the system's `SIGINT` will effectively be buffered inside `sigChan`. Once the synchronous reload operation finishes, the loop will roll over, parse the buffered signal, and gracefully shut down with complete multi-thread immunity.

pkg/gateway/gateway.go
cytown: healthserver should never reload, if you need configuration for this, you must restart gateway, not hot reload.

it's the core service of gateway, so we can not change on fly.

PR #1786, opened 2026-03-19
fix(config): model_list inherits api_key/api_base from providers

### LINKED ISSUE #1635: [BUG] model_list does not inherit api_key/api_base from providers (even with matching protocol prefix)

sliverp:
## Summary

When users configure credentials in the `providers` section and define models in `model_list`, the model entries do **not** inherit `api_key`/`api_base` from the matching provider. Users must duplicate credentials in every `model_list` entry, violating DRY and causing silent failures when credentials are only set in `providers`.

## Problem

As reported in #1635:

```json
{
  "providers": {
    "deepseek": {
      "api_key": "sk-my-key",
      "api_base": "https://api.deepseek.com/v1"
    }
  },
  "model_list": [
    {
      "model_name": "deepseek-chat",
      "model": "deepseek/deepseek-chat"
    }
  ]
}
```

The model entry has no `api_key` — even though `providers.deepseek` has one. `CreateProviderFromConfig` sees an empty `cfg.APIKey` and returns an error. Users expect the protocol prefix (`deepseek/`) to match the provider automatically.

## Solution

Add `InheritProviderCredentials()` that runs during `LoadConfig`, **after** the existing `ConvertProvidersToModelList` migration. For each `model_list` entry:

1. Extract the protocol prefix from the `Model` field (e.g. `deepseek` from `deepseek/deepseek-chat`)
2. Look up the matching provider in `ProvidersConfig` via a protocol→provider mapping
3. Fill in empty `api_key`, `api_base`, `proxy`, and `request_timeout` from the provider

**Explicit `model_list` values always take precedence** — only empty fields are filled.

### What changed
- **`migration.go`**: New `InheritProviderCredentials()` function + `protocolProviderMapping` covering all 25 supported protocols
- **`config.go`**: `LoadConfig()` calls `InheritProviderCredentials()` after provider migration
- **`migration_test.go`**: 6 new tests

### What did NOT change
- `ConvertProvidersToModelList` logic is unchanged
- `CreateProviderFromConfig` factory is unchanged
- No new dependencies

## Tests

6 new tests added:

| Test | Description |
|------|-------------|
| `FillsMissingAPIKey` | Empty model_list entry inherits api_key + api_base from matching provider |
| `ExplicitValuesTakePrecedence` | Model entry with explicit api_key is NOT overridden |
| `MultipleModels` | Mixed: some inherit, some keep explicit values |
| `NoMatchingProvider` | Unknown protocol prefix — no inheritance, no error |
| `EmptyProviders` | All providers empty — nothing to inherit |
| `InheritsRequestTimeout` | Ollama model inherits api_base + request_timeout |

## Breaking Changes

None. This is purely additive. Existing configs that already set `api_key` on every `model_list` entry are unaffected (explicit values always win).

Closes #1635

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -914,6 +914,15 @@ func LoadConfig(path string) (*Config, error) {
 		cfg.ModelList = ConvertProvidersToModelList(cfg)
 	}
 
+	// Inherit credentials from providers to model_list entries (#1635).
+	// When both providers and model_list are present, model_list entries
+	// whose api_key/api_base are empty will inherit from the matching
+	// provider (matched by protocol prefix).  Explicit model_list values
+	// always take precedence.
+	if cfg.HasProvidersConfig() {
+		InheritProviderCredentials(cfg.ModelList, cfg.Providers)
+	}
+
 	// Validate model_list for uniqueness and required fields
 	if err := cfg.ValidateModelList(); err != nil {
 		return nil, err
--- a/pkg/config/migration.go
+++ b/pkg/config/migration.go
@@ -468,3 +468,84 @@ func ConvertProvidersToModelList(cfg *Config) []ModelConfig {
 
 	return result
 }
+
+// protocolProviderMapping maps a model protocol prefix (the part before "/" in
+// the Model field) to a function that extracts the corresponding ProviderConfig
+// from the legacy ProvidersConfig.  Used by InheritProviderCredentials.
+var protocolProviderMapping = map[string]func(p ProvidersConfig) ProviderConfig{
+	"openai":         func(p ProvidersConfig) ProviderConfig { return p.OpenAI.ProviderConfig },
+	"anthropic":      func(p ProvidersConfig) ProviderConfig { return p.Anthropic },
+	"litellm":        func(p ProvidersConfig) ProviderConfig { return p.LiteLLM },
+	"openrouter":     func(p ProvidersConfig) ProviderConfig { return p.OpenRouter },
+	"groq":           func(p ProvidersConfig) ProviderConfig { return p.Groq },
+	"zhipu":          func(p ProvidersConfig) ProviderConfig { return p.Zhipu },
+	"vllm":           func(p ProvidersConfig) ProviderConfig { return p.VLLM },
+	"gemini":         func(p ProvidersConfig) ProviderConfig { return p.Gemini },
+	"nvidia":         func(p ProvidersConfig) ProviderConfig { return p.Nvidia },
+	"ollama":         func(p ProvidersConfig) ProviderConfig { return p.Ollama },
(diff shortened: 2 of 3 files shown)


Orgmar:
@sliverp Solid work on the credential inheritance. Having model_list entries silently fail because they don't inherit api_key from the matching provider is a frustrating gotcha for users. The protocol-to-provider mapping covering all 25 protocols is thorough, and the 6 test cases cover the edge cases well. Clean fix for #1635.

We're running a **PicoClaw Dev Group** on Discord for contributors to collaborate more directly. If you'd like to join, send an email to `support@sipeed.com` with subject `[Join PicoClaw Dev Group] sliverp` and we'll send you the invite!

PR #1818, opened 2026-03-20
fix(agent): separate empty-response and tool-limit fallbacks

### LINKED ISSUE #1815: Misleading defaultResponse suggests max_tool_iterations when model returns empty
## Description

The `defaultResponse` constant in `pkg/agent/loop.go:70` says:

> I've completed processing but have no response to give. Increase `max_tool_iterations` in config.json.

This message fires when the model returns an empty response — not when tool iterations are actually exhausted. The suggestion to increase `max_tool_iterations` is misleading because the root cause is typically:

- Model returned empty content (common with some providers on edge cases)
- Provider returned an error that was swallowed as empty content
- Token limit hit mid-generation

The message leads operators to increase `max_tool_iterations` which doesn't fix the actual problem.

## Suggested Fix

Either:
1. Change the message to something accurate: `"The model returned an empty response. This may indicate a provider error or token limit."` 
2. Or differentiate between "iterations exhausted" and "empty response" with separate messages.

## Environment

Observed on picoclaw running against DashScope (qwen3.5-plus) and Gemini providers.

Alix-007:
## Summary
- use an accurate fallback message when the model returns an empty response
- keep a dedicated fallback message for turns that actually exhaust `max_tool_iterations`
- add regression coverage for both paths in `pkg/agent/loop_test.go`

## Testing
- `go test ./pkg/agent -run 'TestAgentLoop_(EmptyModelResponseUsesAccurateFallback|ToolLimitUsesDedicatedFallback|ContextExhaustionRetry)$' -count=1`
- `go test ./pkg/agent -run 'TestProcessMessage_(UsesRouteSessionKey|CommandOutcomes|SwitchModelRoutesSubsequentRequestsToSelectedProvider)$' -count=1`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.10.1 run ./pkg/agent/...`

## Notes
- closes #1815
- `go test ./pkg/agent -count=1` still hits the existing `TestGlobalSkillFileContentChange` environment-related failure in this workspace; this PR does not touch that path.

### DIFF
--- a/pkg/agent/loop.go
+++ b/pkg/agent/loop.go
@@ -70,7 +70,8 @@ type processOptions struct {
 }
 
 const (
-	defaultResponse           = "I've completed processing but have no response to give. Increase `max_tool_iterations` in config.json."
+	defaultResponse           = "The model returned an empty response. This may indicate a provider error or token limit."
+	toolLimitResponse         = "I've reached `max_tool_iterations` without a final response. Increase `max_tool_iterations` in config.json if this task needs more tool steps."
 	sessionKeyAgentPrefix     = "agent:"
 	metadataKeyAccountID      = "account_id"
 	metadataKeyGuildID        = "guild_id"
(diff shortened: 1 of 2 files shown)


afjcjsbx:
LGTM!

afjcjsbx (approved):

PR #1863, opened 2026-03-21
Feat/hook manager

alexhoshina:
## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

- **Reference URL:**
- **Reasoning:**

## 🧪 Test Environment
- **Hardware:** 
- **OS:** 
- **Model/Provider:** 
- **Channels:** 

## 📸 Evidence (Optional)
<details>

</details>

## ☑️ Checklist

### DIFF
--- a/config/config.example.json
+++ b/config/config.example.json
@@ -511,6 +511,14 @@
   "voice": {
     "echo_transcription": false
   },
+  "hooks": {
+    "enabled": true,
+    "defaults": {
+      "observer_timeout_ms": 500,
+      "interceptor_timeout_ms": 5000,
+      "approval_timeout_ms": 60000
+    }
+  },
   "gateway": {
     "host": "127.0.0.1",
     "port": 18790
--- a/docs/hooks/README.md
+++ b/docs/hooks/README.md
@@ -1,0 +1,679 @@
+# Hook System Guide
+
+This document describes the hook system that is implemented in the current repository, not the older design draft.
+
+The current implementation supports two mounting modes:
+
+1. In-process hooks
+2. Out-of-process process hooks (`JSON-RPC over stdio`)
+
+The repository no longer ships standalone example source files. The Go and Python examples below are embedded directly in this document. If you want to use them, copy them into your own local files first.
+
+## Supported Hook Types
+
+| Type | Interface | Stage | Can modify data |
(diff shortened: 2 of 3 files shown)


pkg/agent/hook_process.go
```diff
@@ -0,0 +1,511 @@
...
+	select {
+	case err := <-done:
+		if err != nil {
+			return fmt.Errorf("write process hook %q message: %w", ph.name, err)
+		}
+		return nil
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
```
afjcjsbx, line 394: If `ctx.Done()` fires, the function returns and defer `ph.writeMu.Unlock()` is executed. The background goroutine is still blocked trying to write to `ph.stdin`. When a subsequent hook request arrives, it acquires `writeMu` and starts a new write. The original goroutine might unblock simultaneously, resulting in interleaved, corrupted JSON being pushed to the external process's standard input.

If an IPC write times out, the stream state is permanently compromised. You cannot safely recover. You must tear down the process.

```go
case <-ctx.Done():
    // The pipe is hopelessly blocked. We must terminate the hook.
    go ph.Close() 
    return ctx.Err()
```

pkg/agent/hook_process.go
```diff
@@ -0,0 +1,511 @@
...
+		respCh, ok := ph.pending[msg.ID]
+		if ok {
+			delete(ph.pending, msg.ID)
+		}
+		ph.pendingMu.Unlock()
+		if ok {
+			respCh <- msg
+			close(respCh)
+		}
+	}
```
afjcjsbx, line 422: If the external process returns a JSON payload larger than 1MB (e.g., a response containing large base64 images or massive context arrays), `scanner.Scan()` will return `false`, and the loop will simply exit. The `ProcessHook` instance remains registered, but it is no longer reading responses. All pending and future requests to this hook will hang until they hit their context timeouts.

check `scanner.Err()` and proactively tear down the hook so the system knows it has failed.

```go
for scanner.Scan() { ... }

if err := scanner.Err(); err != nil {
    logger.ErrorCF("hooks", "Process hook read loop failed", map[string]any{"error": err})
    ph.failPending(err)
    go ph.Close()
}
```

PR #1877, opened 2026-03-22
feat(tools): add tool argument schema validation before execution

asabya:
## Summary

Adds runtime validation of tool call arguments against each tool's declared `Parameters()` JSON Schema **before** `Execute()` is called. This closes a significant security gap where LLM-generated tool calls were passed to tools without any type checking, required-field enforcement, or extra-property rejection.

**Roadmap item:** Security Hardening > Tool abuse prevention via strict parameter validation

## Problem

Every PicoClaw tool declares its parameter schema via `Parameters() map[string]any` (JSON Schema format), but these schemas were **only sent to the LLM as guidance** — never enforced at execution time. `ExecuteWithContext()` passed `map[string]any` arguments directly to `tool.Execute()` without validation.

This meant:
- A confused LLM could call tools with **wrong argument types** (e.g., passing a number where a string is expected)
- **Extra properties** injected into tool arguments (e.g., `{"path": "/tmp/x", "__command": "rm -rf /"}`) passed through unchecked
- **Missing required fields** caused unpredictable tool-level errors instead of clear validation failures
- MCP tools, skill-provided tools, and core tools were all equally unprotected

### Reference implementations

- **OpenClaw** uses AJV JSON Schema validation + type-safe parameter readers, returning `ToolInputError` on validation failure
- **Paperclip** uses AJV for config validation but does NOT enforce schemas at tool execution time — a known gap

## Solution

A single `validateToolArgs()` function inserted at the `ExecuteWithContext()` choke point in `pkg/tools/registry.go`, between tool lookup (`Get()`) and execution (`Execute()`):

```
RunToolLoop (pkg/tools/toolloop.go)
    → NormalizeToolCall()          [existing]
    → registry.ExecuteWithContext()
        → registry.Get(name)      [existing - tool lookup]
        → validateToolArgs(...)   [NEW - schema validation]
        → tool.Execute(ctx, args) [existing - execution]
```

### What it validates

| Check | Example | Error message |
|---|---|---|
| Required fields present | `exec` tool needs `command` | `missing required property "command"` |
| Type matching | `path` expects `string`, gets `123` | `property "path": expected string, got float64` |
| Extra properties rejected | `{"path": "x", "__inject": "y"}` | `unexpected property "__inject"` |
| `enum` membership | `cron` action must be `add\|list\|remove\|...` | `property "action": value destroy is not in enum` |
| Nested object validation | Recursive type checking for `object` properties | Error includes nested path |
| Array element types | `items: {type: "string"}` with mixed elements | `tags[1]: expected string, got float64` |
| `additionalProperties: true` | MCP tools that accept passthrough params | Extra args allowed |

### What it deliberately skips

- `$ref`, `oneOf`, `anyOf`, `allOf` — not used by any PicoClaw tool
- `pattern` (regex on strings) — not used by any PicoClaw tool
- `minimum`/`maximum` — tools already clamp values internally (e.g., `web_search` clamps `count`)
- `format` — not used by any PicoClaw tool

### Design decisions

- **Zero new dependencies** — uses only Go standard library (`fmt`, `math`), keeping the binary lightweight
- **Always on** — no config toggle. Schema validation is a correctness guarantee, not a policy
- **Non-blocking** — validation failure returns `ToolResult{IsError: true}` (same pattern as "tool not found"), allowing the LLM to self-correct on the next turn
- **MCP-compatible** — handles `[]any` for `required` fields (from JSON deserialization), and respects `additionalProperties: true` for external tool schemas
- **No tool modifications** — validation happens at the registry level, not inside individual tools. Every registered tool is protected automatically
- **float64 edge case** — Go's `json.Unmarshal` turns all JSON numbers into `float64`. The validator accepts `float64` with no fractional part as valid `integer` values

## Files changed

| File | Change |
|---|---|
| `pkg/tools/validate.go` | New file (209 lines) — `validateToolArgs()` + 5 helper functions |
| `pkg/tools/validate_test.go` | New file (465 lines) — 28 unit tests + 9 real-schema tests + 1 integration test |
| `pkg/tools/registry.go` | 8 lines added to `ExecuteWithContext()` |
| `.gitignore` | Added `docs/superpowers/` |

## Test plan

- [x] 28 table-driven unit tests covering all validation paths (type matching, required fields, enum, nested objects, arrays, additionalProperties, float64/integer edge cases, `[]any` vs `[]string` required)
- [x] 9 tests using schemas copied from real PicoClaw tools (`ExecTool`, `CronTool`, `WebSearchTool`)
- [x] 1 integration test exercising the full `ToolRegistry.Execute()` → `validateToolArgs()` → tool execution path
- [x] All 24 existing registry tests pass with zero regressions (nil args with minimal schemas handled correctly)
- [x] `go vet ./pkg/tools/` clean
- [x] Full `go test ./pkg/tools/` suite passes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,6 +40,7 @@ tasks/
 
 # Plans
 docs/plans/
+docs/superpowers/
 
 # Editors
 .vscode/
--- a/pkg/tools/registry.go
+++ b/pkg/tools/registry.go
@@ -180,6 +180,14 @@ func (r *ToolRegistry) ExecuteWithContext(
 		return ErrorResult(fmt.Sprintf("tool %q not found", name)).WithError(fmt.Errorf("tool not found"))
 	}
 
+	// Validate arguments against the tool's declared schema.
+	if err := validateToolArgs(tool.Parameters(), args); err != nil {
+		logger.WarnCF("tool", "Tool argument validation failed",
+			map[string]any{"tool": name, "error": err.Error()})
+		return ErrorResult(fmt.Sprintf("invalid arguments for tool %q: %s", name, err)).
+			WithError(fmt.Errorf("argument validation failed: %w", err))
+	}
+
 	// Inject channel/chatID into ctx so tools read them via ToolChannel(ctx)/ToolChatID(ctx).
 	// Always inject — tools validate what they require.
 	ctx = WithToolContext(ctx, channel, chatID)
--- a/pkg/tools/validate.go
+++ b/pkg/tools/validate.go
@@ -1,0 +1,209 @@
+package tools
+
+import (
+	"fmt"
+	"math"
+)
+
+// validateToolArgs validates args against a JSON Schema-like map.
+// schema is expected to have optional keys: "properties", "required", "additionalProperties".
+func validateToolArgs(schema map[string]any, args map[string]any) error {
+	if len(schema) == 0 {
+		return nil
+	}
+
+	if args == nil {
+		args = map[string]any{}
+	}
+
+	if err := checkRequired(schema, args); err != nil {
+		return err
+	}
+
+	propsRaw, ok := schema["properties"]
+	if !ok {
+		return nil // no properties defined — accept any args
+	}
+
+	props, ok := propsRaw.(map[string]any)
+	if !ok {
+		return nil
+	}
+
+	additional := allowsAdditional(schema)
+
+	for key, val := range args {
+		propSchemaRaw, known := props[key]
+		if !known {
+			if !additional {
+				return fmt.Errorf("unexpected property %q", key)
+			}
+			continue
+		}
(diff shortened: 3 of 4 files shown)


Orgmar:
@asabya This is a really thorough security hardening contribution. Validating tool arguments against JSON Schema at the registry level before execution is a solid architectural choice, and doing it with zero new dependencies keeps the binary lean. The coverage of edge cases like float64/integer conversion, MCP additionalProperties, and nested object validation shows real attention to detail. 28 unit tests plus the real-schema tests give good confidence.

We have a **PicoClaw Dev Group** on Discord where contributors connect and collaborate. If you'd like to join, send an email to `support@sipeed.com` with subject `[Join PicoClaw Dev Group] asabya` and we'll send you the invite link!

PR #1910, opened 2026-03-23
fix: avoid duplicate outbound replies after message tool on named agents

### LINKED ISSUE #1909: Named agents can send duplicate replies after using the message tool
## Summary
When a non-default agent handles a turn and uses the `message` tool, the agent loop can still publish an extra final outbound response. This results in duplicate replies to the user.

## Impact
This affects multi-agent routing setups where a named agent uses `message` to send output directly. The user may receive:
- the direct message from the `message` tool
- a second duplicate reply from the main agent loop

## Root Cause
The per-turn dedupe state is stored on each agent's own `MessageTool` instance (`sentInRound`).

However, final outbound publishing checks the `message` tool state only on the default agent, not on the agent that actually handled the current session/turn. In routed turns, those are different tool instances, so the dedupe check can miss that a named agent already sent a message.

## Reproduction
1. Configure more than one agent.
2. Route a conversation to a non-default agent.
3. Let that agent call the `message` tool during a turn.
4. Return a non-empty final response from the turn.
5. Observe that the user receives both the tool-sent message and a second outbound reply.

## Expected Behavior
The final outbound publish path should check the `message` tool state for the agent associated with the current session, not always the default agent.

## Notes
A session-scoped or turn-scoped regression test should cover the named-agent path to prevent this from regressing.

FyZhu97:
## Description
This change fixes a duplicate-reply bug in the agent loop when a non-default agent uses the `message` tool during a turn.

Previously, outbound dedupe checked the `message` tool state only on the default agent. In multi-agent routing, that could miss the actual tool instance that handled the turn, causing the loop to publish an extra final response after the tool had already sent one.

This patch changes response publishing to resolve the agent from the current session key and read the `message` tool state from that agent instead.

## Type of Change
- Bug fix

## AI Code Generation
- 🛠️ Mostly AI-generated

## Related Issue
- Fixes #1909

## Technical Context
- Root cause: `publishResponseIfNeeded()` used `GetDefaultAgent()` instead of the agent associated with the current session.
- In routed turns, `ResetSentInRound()` and `sentInRound=true` were applied to the named agent's `MessageTool`, but final dedupe read state from the default agent's separate tool instance.
- Fix: use `agentForSession(sessionKey)` inside `publishResponseIfNeeded()` and pass the relevant `sessionKey` from call sites.
- Added a regression test covering named-agent behavior.
- Rebasing onto the latest `main` required a small test-only follow-up commit to align the regression test with the current `AgentDefaults` field names.

## Test Environment
- OS: macOS
- Shell: zsh
- Base branch checked: latest `main` at `cf9e049`
- Feature branch: `codex/fix-message-tool-dedupe`

## Evidence
Passed:
```bash
gofmt -w pkg/agent/loop.go pkg/agent/loop_test.go
go test ./pkg/agent -run "TestPublishResponseIfNeeded_UsesSessionAgentMessageState|TestProcessMessage_IncludesCurrentSenderInDynamicContext|TestProcessMessage_UseCommandLoadsRequestedSkill|TestHandleCommand_UseCommandRejectsUnknownSkill|TestProcessMessage_UseCommandArmsSkillForNextMessage"
```

Relevant result:
```bash
ok   github.com/sipeed/picoclaw/pkg/agent   1.069s
```

Full pre-PR check:
```bash
make check GOLANGCI_LINT=$(go env GOPATH)/bin/golangci-lint
```

Status:
- Blocked by an unrelated existing `vet` failure on current `main`
- Existing failing check:
```bash
pkg/config/security_integration_test.go:23:3: struct field privateField has json tag but is not exported
```

This failure appears during `make check` before package tests for this PR are reached.

## Checklist
- [x] I branched from `main`
- [x] The change is focused and limited to one bug fix
- [x] I reviewed the code and understand the change
- [x] I added a regression test
- [x] Local targeted tests for the changed area pass
- [x] Full `make check` passes locally
- [x] The remaining failing check appears unrelated to this PR

### DIFF
--- a/pkg/agent/loop.go
+++ b/pkg/agent/loop.go
@@ -458,7 +458,7 @@ func (al *AgentLoop) Run(ctx context.Context) error {
 				if target == nil {
 					cancelDrain()
 					if finalResponse != "" {
-						al.publishResponseIfNeeded(ctx, msg.Channel, msg.ChatID, finalResponse)
+						al.publishResponseIfNeeded(ctx, msg.SessionKey, msg.Channel, msg.ChatID, finalResponse)
 					}
 					return
 				}
@@ -518,7 +518,7 @@ func (al *AgentLoop) Run(ctx context.Context) error {
 				}
 
 				if finalResponse != "" {
-					al.publishResponseIfNeeded(ctx, target.Channel, target.ChatID, finalResponse)
+					al.publishResponseIfNeeded(ctx, target.SessionKey, target.Channel, target.ChatID, finalResponse)
 				}
 			}()
 		default:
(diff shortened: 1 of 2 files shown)


FyZhu97:
@lxowalle please review this when you are free. thx.

PR #1971, opened 2026-03-24
feat(config): add command pattern detection tool in exec settings

uiYzzi:
## Summary

Adds a new API endpoint `/api/config/test-command-patterns` that tests a command against configured whitelist and blacklist patterns, along with a frontend UI component to interactively test patterns.

## 📝 Description

This PR adds a pattern detection tool to the exec settings page, allowing users to test whether a specific command would be allowed or blocked by their current whitelist/blacklist configuration. The detection logic runs on the backend using Go's regexp package, ensuring complete consistency with the runtime enforcement.

图片

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 📚 Technical Context
- **Reasoning:** The frontend detection tool now calls the backend API endpoint instead of using JavaScript regex, ensuring the same Go regexp behavior used by ExecTool at runtime.

## ☑️ Checklist
- [x] Backend tests pass: `go test ./api/... -run TestHandleTestCommandPatterns`
- [x] Frontend build succeeds: `npm run build`

### DIFF
--- a/web/frontend/src/components/config/config-sections.tsx
+++ b/web/frontend/src/components/config/config-sections.tsx
@@ -218,14 +218,16 @@ export function ExecSection({ form, onFieldChange }: ExecSectionProps) {
       return
     }
 
-    const denyPatterns = form.customDenyPatternsText
-      .split("\n")
-      .map((p) => p.trim())
-      .filter((p) => p.length > 0)
     const allowPatterns = form.customAllowPatternsText
       .split("\n")
       .map((p) => p.trim())
       .filter((p) => p.length > 0)
+    const denyPatterns = form.enableDenyPatterns
+      ? form.customDenyPatternsText
+          .split("\n")
+          .map((p) => p.trim())
+          .filter((p) => p.length > 0)
+      : []
 
     setIsLoading(true)
     try {


web/frontend/src/components/config/config-sections.tsx
```diff
@@ -201,6 +203,54 @@ interface ExecSectionProps {
...
+  } | null>(null)
+  const [isLoading, setIsLoading] = useState(false)
+
+  const testPatterns = async () => {
+    if (!testCommand.trim()) {
+      setTestResult(null)
+      return
+    }
+
+    const denyPatterns = form.customDenyPatternsText
```
wj-xiao: The new pattern tester does not respect `enable_deny_patterns`.

The frontend tester always sends `customDenyPatternsText` to the backend, even when the blacklist toggle is disabled.
That does not match the real exec behavior: when `enable_deny_patterns` is false, runtime execution skips deny-pattern checks entirely.

This becomes incorrect in a common case where a user previously saved custom deny patterns, then later disables the blacklist. The saved patterns still exist in config, so the tester can report a command as blocked even though the actual exec guard would allow it.

PR #1972, opened 2026-03-24
    refactor(cron): optimize job execution by eliminating redundant l…

hkhere:
…ookups

    Refactor checkJobs() to collect job copies during initial search and pass
    them directly to executeJob(), removing O(n×m) complexity from repeated
    job ID lookups.

    Key changes:
    - Collect []CronJob values (not IDs) in single pass through jobs
    - Rename executeJobByID() → executeJob() accepting *CronJob
    - Remove redundant job search before execution
    - Use struct values instead of pointers to prevent data races

    Performance: O(n+m) vs O(n×m) - significant improvement with many jobs
    Code size: -25 lines net, cleaner logic
    Thread-safety: Maintained with value semantics

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

- **Reference URL:**
- **Reasoning:**

## 🧪 Test Environment
- **Hardware:** 
- **OS:** 
- **Model/Provider:** 
- **Channels:** 

## 📸 Evidence (Optional)
<details>

</details>

## ☑️ Checklist

### DIFF
--- a/pkg/cron/service.go
+++ b/pkg/cron/service.go
@@ -179,24 +179,14 @@ func (cs *CronService) checkJobs() {
 	}
 
 	now := time.Now().UnixMilli()
-	var dueJobIDs []string
+	var dueJobs []CronJob
 
-	// Collect jobs that are due (we need to copy them to execute outside lock)
+	// Collect copies of due jobs and reset their next run times
 	for i := range cs.store.Jobs {
 		job := &cs.store.Jobs[i]
 		if job.Enabled && job.State.NextRunAtMS != nil && *job.State.NextRunAtMS <= now {
-			dueJobIDs = append(dueJobIDs, job.ID)
-		}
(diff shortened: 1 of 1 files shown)


wj-xiao:
This change is not behavior-preserving. `checkJobs()` now captures a `CronJob` snapshot and later passes that snapshot to `executeJob()`, instead of reloading the job by ID right before execution.

That means if a due job is updated or removed after batching but before its turn in the execution loop, `onJob` can still run with stale payload/command data. Under the previous implementation, removed jobs were skipped and updated jobs were executed with their latest state. Since due jobs are executed sequentially, a long-running earlier job can make this window large enough to be user-visible.

The optimization is fine, but we still need to re-read the latest job state before invoking the callback, or otherwise preserve the old semantics explicitly.

PR #1981, opened 2026-03-24
Feat/tool read_file by lines

### LINKED ISSUE #1974: [Proposal] Refactor ReadFileTool to use line-based pagination instead of byte-based
## Background
Currently, our `ReadFileTool` relies on byte-based pagination, using `offset` and `length` parameters to chunk file reading. This approach works well for arbitrary data streams and is very efficient ($O(1)$ seek time for seekable files).

## The Problem
Since this tool is consumed by an LLM agent, byte-based pagination introduces a significant UX and context issue: **content gets truncated at arbitrary points**. 
When reading a chunk, a byte limit might cut off right in the middle of a variable name, a JSON object, or a critical function definition. This makes it harder for the LLM to understand the context and practically impossible to accurately reference line numbers for subsequent editing or patching tools.

## The Proposal
I would like to open a discussion on migrating the `ReadFileTool` from byte-based reading to **line-based reading**. 

Instead of `offset` and `length`, the tool would accept parameters like:
* `start_line` (integer, default 1)
* `max_lines` (integer, replacing `length`)

**Pros:**
* **LLM-friendly:** Code and text are never split mid-word or mid-statement.
* **Better context:** The LLM knows exactly which line numbers it is looking at, making integration with file-editing tools much more reliable.
* **Natural pagination:** "Read lines 100 to 150" is a much more natural reasoning step for an agent than calculating byte offsets.

**Cons / Discussion Points:**
1. **Performance:** Seeking to a specific line in a large text file requires scanning for newline characters (`\n`), which is $O(N)$ instead of $O(1)$. Given typical file sizes in our use case, is this performance hit acceptable?
2. **Long lines:** How should we handle minified files or files with exceptionally long single lines (which could still blow up the LLM context)?
3. **Backwards Compatibility:** Should we completely replace the byte-based approach, or support both (e.g., keeping bytes as a fallback)? 

## Code Context
This change would mostly affect the `Execute` method of `ReadFileTool`, replacing the `io.Seeker` byte-seeking logic with a buffered reader (like `bufio.Scanner` or `bufio.Reader`) that counts lines. We already have a 512-byte sniff mechanism to detect binary files, which we can keep to avoid scanning lines on non-text files.

What does everyone think about this transition?

afjcjsbx:
## 📝 Description

This PR implements [#1974](https://github.com/sipeed/picoclaw/issues/1974) by splitting `read_file` into two mutually exclusive implementations selected via configuration:

- `tools.read_file.mode = "bytes"`: legacy byte-based pagination with `offset` + `length`
- `tools.read_file.mode = "lines"`: text-oriented line-based pagination with `start_line` (1-indexed) + `max_lines`

Additional changes included in this PR:
- Keeps backward compatibility by defaulting to `bytes`
- Adds token-estimated head/tail truncation for line mode
- Updates configuration examples and documentation for the new mode selection

## 🗣️ Type of Change
- [x] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue

Fixes #1974

- **Reference URL:** https://github.com/sipeed/picoclaw/issues/1974
- **Reasoning:** 

## 🧪 Test Environment
- **Hardware:** PC (local development machine)
- **OS:** macOS
- **Model/Provider:** N/A (tooling/config change)
- **Channels:** N/A

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -319,12 +319,12 @@ Text-oriented behavior, optimized for source files, markdown, logs, and configs.
 Parameters:
 
 * `path` (required): File path
-* `start_line` (optional): Starting line number, 1-indexed and inclusive, default `0`
+* `start_line` (optional): Starting line number, 1-indexed and inclusive, default `1`
 * `max_lines` (optional): Maximum number of lines to read, default = all remaining lines until EOF or byte budget
 
 Behavior notes:
 
-* Binary-looking files are rejected with guidance to use `read_file`
+* Binary-looking files are rejected with guidance to switch `read_file` to `mode = bytes`
 * Extremely long single lines are truncated rather than skipped
 
 Use `mode = lines` when:
--- a/pkg/tools/filesystem.go
+++ b/pkg/tools/filesystem.go
@@ -572,7 +572,7 @@ func (t *ReadFileLinesTool) Execute(ctx context.Context, args map[string]any) *T
 	}
 	sample = sample[:sampleN]
 	if isBinaryReadFileData(sample) {
-		return ErrorResult("file appears to be binary; use read_file for byte-based inspection")
+		return ErrorResult("file appears to be binary; switch read_file mode to 'bytes' for byte-based inspection")
 	}
 
 	reader := bufio.NewReaderSize(io.MultiReader(bytes.NewReader(sample), file), 32*1024)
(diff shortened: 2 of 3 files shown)


afjcjsbx:
hi @yinwm, thanks for the careful evaluation, sorry, I was working on PR before but I had to leave, I addressed the requested points, PTAL 🙏

PR #2000, opened 2026-03-25
docs: add cron job behavior guide

### LINKED ISSUE #1868: [Feature] Missing Cronjob Documentation
## 🎯 The Goal / Use Case
Cron Documentation
## 💡 Proposed Solution
Here is docs.

# PicoClaw Cron Reference

## How Cron Jobs Execute

When a cron job fires, `ExecuteJob` in `pkg/tools/cron.go` evaluates the payload in this order:

1. **`command` is set** -- The command runs directly via `ExecTool`, and its stdout/stderr is published to the target channel. No agent involvement. This is the most reliable path.
2. **`deliver: true` (no command)** -- The literal `message` text is sent to the channel verbatim. No agent, no tools, no execution. Use this only for static reminder text.
3. **`deliver: false` (no command)** -- The message is routed through the agent via `ProcessDirectWithChannel`. The agent has full tool access (exec, skills, web, etc.) and can decide what to do. **Known bug in v0.2.3**: the agent's final response is silently dropped (`_ = response`) because `ProcessDirectWithChannel` bypasses the main Run loop that normally publishes responses. Tool feedback messages (e.g. tool call previews) ARE sent during execution, but the final summary is lost.

## Payload Fields

| Field     | Type   | Description |
|-----------|--------|-------------|
| `kind`    | string | Always `"agent_turn"` |
| `message` | string | Description of the job (shown in logs; sent verbatim when `deliver: true`) |
| `command` | string | Shell command to execute directly (takes priority over deliver/agent paths) |
| `deliver` | bool   | `true` = send message as-is; `false` = route through agent or use command |
| `channel` | string | Target channel (e.g. `"telegram"`) |
| `to`      | string | Target chat/user ID |

## Rules for Creating Jobs

- **To run a script and send output**: set `command` to the shell command, set `deliver: false`. The command output goes straight to the channel.
- **To send a static reminder**: set `deliver: true`, leave `command` empty. The `message` text is sent as-is.
- **To let the agent decide**: set `deliver: false`, leave `command` empty. Be aware of the v0.2.3 delivery bug above -- the agent runs but its final answer may not reach the channel.
- **Command scheduling from external channels** (Telegram, Discord, etc.) is blocked by a security gate (`IsInternalChannel` check). Commands can only be scheduled from internal channels or by editing `jobs.json` directly while the gateway is stopped.
- The gateway holds cron state in memory and writes it back to `jobs.json`. To edit jobs manually, **stop the gateway first** (`kill <pid>`), edit the file, then restart (`picoclaw gateway`).

## Schedule Types

| Kind    | Field          | Example |
|---------|----------------|---------|
| `cron`  | `expr`         | `"45 8 * * *"` (daily at 8:45 AM) |
| `every` | `everyMs`      | `7200000` (every 2 hours) |
| `at`    | `atMs`         | Unix ms timestamp for one-shot; job auto-disables or deletes after firing |

## Config Requirements

In `config.json`, the following must be enabled for command-based cron jobs:

```json
"tools": {
  "cron": { "enabled": true, "allow_command": true },
  "exec": { "enabled": true }
}
```

- [ ] This is a Core Feature

Alix-007:
## Summary
- add a dedicated cron behavior guide under `docs/cron.md`
- link the new guide from the main README and tools configuration docs
- document only behavior verified against current code paths

## Verification
- verified documented behavior against `pkg/tools/cron.go`, `pkg/cron/service.go`, `pkg/gateway/gateway.go`, `cmd/picoclaw/internal/cron/add.go`, and `pkg/config/defaults.go`
- ran `git diff --check`

Closes #1868

### DIFF
--- a/docs/cron.md
+++ b/docs/cron.md
@@ -56,7 +56,7 @@ The current CLI `picoclaw cron add` command does not expose a `command` flag.
 
 `tools.cron.enabled` controls whether the agent-facing `cron` tool is registered. Default: `true`.
 
-If you disable `tools.cron`, users can no longer create or manage jobs through the agent tool. The gateway still starts the cron service and the CLI still uses the same job store, so keep `tools.cron.enabled` on if you expect scheduled agent-turn or command jobs to execute through the gateway.
+If you disable `tools.cron`, users can no longer create or manage jobs through the agent tool. The gateway still starts `CronService`, but it does not install the job execution callback. As a result, due jobs do not actually run; one-time jobs may be deleted and recurring jobs may be rescheduled without executing their payload. The CLI still uses the same job store.
 
 `tools.cron.exec_timeout_minutes` sets the timeout used for scheduled command execution. Default: `5`. Set `0` for no timeout.
 


Alix-007:
Maintainer review requested when you have bandwidth: this is a docs-only PR for cron behavior guidance (Closes #1868), and current checks are all green (Linter, Security Check, Tests, license/cla). Scope is intentionally narrow to documentation aligned with existing code paths.

PR #2020, opened 2026-03-25
fix(subturn): inherit runtime tool registry for spawned subagents

### LINKED ISSUE #1713: [BUG] Subagents spawned via spawn tool have no tools
## Quick Summary

Subagents spawned via the `spawn` tool do not receive any tools in the API request to the LLM backend. The `tools` field is missing entirely from the JSON payload sent for subagent completions, causing the model to respond with "I cannot execute commands" instead of making tool calls. The parent agent (e.g., Shark) works correctly and includes `tools` in its requests.

## Environment & Tools
- **Go Version:** go1.25.7
- **AI Model & Provider:** Qwen 3.5 (9B) via llama-server (local, OpenAI-compatible endpoint)
- **Operating System:** Linux arm64 (Raspberry Pi)
- **Channels:** Telegram

## 📸 Steps to Reproduce

1. Configure picoclaw with `spawn` and `subagent` enabled, using any OpenAI-compatible backend that supports tool calling (e.g., llama-server with Qwen 3.5).
2. Send a message that triggers a subagent spawn, e.g.: "Engage Blowfish and tell him to SSH to root@example.com and run uptime"
3. Observe the HTTP traffic between picoclaw and the LLM backend (e.g., `sudo tcpdump -i any -A port 8033`).

## ❌ Actual Behavior

The parent agent's request includes the full `"tools"` array and the model responds with `finish_reason: "tool_calls"` — working correctly.

However, when a subagent is spawned, the request payload looks like this:

```json
{
  "max_tokens": 128000,
  "messages": [
    {"role": "system", "content": "You are a subagent. Complete the given task independently..."},
    {"role": "user", "content": "You are Blowfish ... SysAdmin. Your task is to..."}
  ],
  "model": "qwen3.5",
  "temperature": 0.7
}
```

**No `"tools"` field is present.** The model has no tools available and generates text-only responses like "I cannot execute SSH commands" or fabricates simulated terminal output that was never actually executed.

## ✅ Expected Behavior

Subagent requests should include the `"tools"` array, just like the parent agent's requests. The subagent should be able to call tools (e.g., `exec`, `read_file`, `write_file`) to actually perform its assigned task.

Expected subagent payload:

```json
{
  "max_tokens": 128000,
  "messages": [...],
  "model": "qwen3.5",
  "temperature": 0.7,
  "tools": [
    {"type": "function", "function": {"name": "exec", ...}},
    {"type": "function", "function": {"name": "read_file", ...}},
    ...
  ]
}
```

## 💬 Additional Context

**Root cause analysis:** In `pkg/tools/subagent.go`, `runTask()` reads `sm.tools` and passes it to `RunToolLoop()`. The tools field is present in the code path, but `sm.tools` appears to be nil or empty at runtime. This is confirmed by the tcpdump capture: the parent agent's requests (built by `pkg/agent/loop.go`) include 14 tools, while subagent requests include zero.

The issue is likely in the initialization of `SubagentManager` — the `tools` field is either never populated, or populated with an empty registry. It may also be a timing issue where `SubagentManager` is constructed before tools are registered in the agent loop.

**Evidence from tcpdump:** Full packet capture attached showing side-by-side comparison of parent agent requests (with tools) vs subagent requests (without tools) to the same llama-server endpoint on port 8033.

**Workaround:** None currently. The parent agent can use `exec` with SSH directly (as observed in the capture — Shark successfully ran `ssh root@... "echo 'John doe' > /root/test.txt"` via its own `exec` tool), but delegated subagents cannot perform any tool-based actions.

Alix-007:
## Summary

Fixes #1713.

This PR keeps spawned/subagent child turns aligned with the parent turn's **runtime** tool registry, instead of pre-snapshotting tools into `SubTurnConfig.Tools`.

- `spawn` / `subagent` now leave `SubTurnConfig.Tools` unset (`nil`) so child turns inherit at execution time from the parent turn context.
- `spawnSubTurn` now prioritizes cloning `baseAgent.Tools` when available, preserving registry semantics (including hidden tools and TTL gating).
- Explicit `cfg.Tools` is only used as a fallback when the parent registry is unavailable and the explicit slice is non-empty.
- Empty explicit tool slices no longer clear inherited tools.

## Why

The bug report describes spawned subagents missing tool definitions in provider requests. The robust fix is to inherit from the live parent registry so child turns receive the same tool visibility and runtime additions as the parent, while keeping hidden/TTL behavior intact.

## Validation

Focused tests were added/updated to assert provider-level tool definitions and edge cases:

- `TestSpawnSubTurn_EmptyExplicitToolsStillInheritParentRegistry`
- `TestSpawnSubTurn_InheritsRuntimeAddedTools`
- `TestSpawnSubTurn_PreservesHiddenTTLSemantics`
- `TestSpawnSubTurn_UsesExplicitToolsWhenParentRegistryUnavailable`
- `TestSpawnTool_Execute_LeavesToolsUnsetForRuntimeInheritance`
- `TestSubagentTool_Execute_LeavesToolsUnsetForRuntimeInheritance`

Commands run:

- `go test ./pkg/tools -run 'TestSpawnTool_Execute_(ValidTask|LeavesToolsUnsetForRuntimeInheritance)|TestSubagentTool_Execute_(Success|LeavesToolsUnsetForRuntimeInheritance)' -count=1`\n- `go test ./pkg/agent -run 'TestSpawnSubTurn_(EmptyExplicitToolsStillInheritParentRegistry|InheritsRuntimeAddedTools|PreservesHiddenTTLSemantics|UsesExplicitToolsWhenParentRegistryUnavailable)' -count=1`\n- `go test ./pkg/tools -run 'TestSpawnTool_|TestSubagentTool_' -count=1`\n- `go test ./pkg/agent -run 'TestSpawnSubTurn_' -count=1`

### DIFF
--- a/pkg/agent/subturn.go
+++ b/pkg/agent/subturn.go
@@ -354,12 +354,12 @@ func spawnSubTurn(
 	ephemeralStore := newEphemeralSession(nil)
 	agent := *baseAgent // shallow copy
 	agent.Sessions = ephemeralStore
-	if cfg.Tools != nil {
-		// Explicit tool slice from caller has highest priority for sub-turns.
-		agent.Tools = toolRegistryFromSlice(cfg.Tools)
-	} else if baseAgent.Tools != nil {
-		// Otherwise inherit the parent's tool snapshot.
+	if baseAgent.Tools != nil {
+		// Inherit the parent's tool registry snapshot so hidden/TTL metadata is preserved.
 		agent.Tools = baseAgent.Tools.Clone()
+	} else if len(cfg.Tools) > 0 {
+		// Fallback path for callers that provide explicit tool slices without a parent registry.
+		agent.Tools = toolRegistryFromSlice(cfg.Tools)
 	} else {
 		agent.Tools = tools.NewToolRegistry()
 	}
--- a/pkg/agent/subturn.go
+++ b/pkg/agent/subturn.go
@@ -197,6 +197,16 @@ func (al *AgentLoop) generateSubTurnID() string {
 	return fmt.Sprintf("subturn-%d", al.subTurnCounter.Add(1))
 }
 
+func toolRegistryFromSlice(toolSlice []tools.Tool) *tools.ToolRegistry {
+	registry := tools.NewToolRegistry()
+	for _, tool := range toolSlice {
+		if tool != nil {
+			registry.Register(tool)
+		}
+	}
+	return registry
+}
+
 // ====================== Core Function: spawnSubTurn ======================
 
 // AgentLoopSpawner implements tools.SubTurnSpawner interface.
@@ -344,10 +354,14 @@ func spawnSubTurn(
 	ephemeralStore := newEphemeralSession(nil)
 	agent := *baseAgent // shallow copy
 	agent.Sessions = ephemeralStore
-	// Clone the tool registry so child turn's tool registrations
-	// don't pollute the parent's registry.
-	if baseAgent.Tools != nil {
+	if cfg.Tools != nil {
+		// Explicit tool slice from caller has highest priority for sub-turns.
+		agent.Tools = toolRegistryFromSlice(cfg.Tools)
+	} else if baseAgent.Tools != nil {
+		// Otherwise inherit the parent's tool snapshot.
 		agent.Tools = baseAgent.Tools.Clone()
+	} else {
+		agent.Tools = tools.NewToolRegistry()
 	}
 
 	// Create processOptions for the child turn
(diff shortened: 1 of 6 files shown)


Alix-007:
Maintainers, this PR is ready for first review.

- Fixes #1713 (spawned subagents missing tools) by inheriting the parent turn’s runtime tool registry.
- All checks are green (`Linter`, `Security Check`, `Tests`, `license/cla`).
- Scope is narrow to subturn tool inheritance plus focused regression tests.

Could you please take a first pass when you have bandwidth?

PR #2021, opened 2026-03-25
fix(web): include non-pico channel sessions in history API

### LINKED ISSUE #1996: Web UI: Chat history only shows pico (web) sessions, hides all other channels
## Summary

The web launcher's session history API (`GET /api/sessions`) only lists sessions from the built-in "pico" web channel. Sessions from all other channels (Telegram, Discord, Slack, Matrix, LINE, etc.) are invisible in the web UI, even though their `.jsonl` files exist in the same `sessions/` directory.

## Current Behavior

The `handleListSessions` function in `web/backend/api/session.go` filters session files by the hardcoded prefix `agent_main_pico_direct_pico_`. Any session file that doesn't match this prefix is silently skipped:

```go
const (
    picoSessionPrefix          = "agent:main:pico:direct:pico:"
    sanitizedPicoSessionPrefix = "agent_main_pico_direct_pico_"
)
```

For example, a Telegram session file like `agent_main_telegram_direct_1200880918.jsonl` is present on disk but never appears in the API response.

## Expected Behavior

The web UI's chat history should be able to show sessions from **all connected channels**, not just the built-in web chat. This would let users:

- Review conversation history from any channel in one place
- See what the agent has been doing across Telegram, Discord, etc.
- Debug agent behavior across channels without SSH-ing into the container to read raw JSONL files

## Possible Approaches

1. **Generalize the prefix filter** — Instead of matching only `agent_main_pico_direct_pico_*`, parse any `agent_<id>_<channel>_direct_<peer>` pattern and extract a session identifier. Add a `channel` field to the session list response so the frontend can group/filter by channel.

2. **Add a channel query parameter** — `GET /api/sessions?channel=telegram` (or `channel=all`) to let the frontend request sessions from specific channels or all channels.

3. **Keep pico-only as default, add an "all channels" toggle** — Minimal UI change: a toggle in the chat history sidebar that switches between "Web only" and "All channels".

Any of these would work. Happy to submit a PR if someone confirms the preferred approach.

Alix-007:
## Summary
- include non-pico session keys in GET /api/sessions instead of filtering to pico-only keys
- keep backward compatibility for existing pico short IDs while allowing full non-pico keys for GET /api/sessions/{id} and DELETE /api/sessions/{id}
- add regression tests for non-pico JSONL and legacy JSON session history flows

Fixes #1996.

## Verification
- Command: go test ./web/backend/api
- Result: ok   github.com/sipeed/picoclaw/web/backend/api	(cached)

### DIFF
--- a/web/backend/api/session.go
+++ b/web/backend/api/session.go
@@ -224,10 +224,6 @@ func (h *Handler) readJSONLSessionByKey(dir, sessionKey string) (sessionFile, er
 	}, nil
 }
 
-func (h *Handler) readLegacySession(dir, sessionID string) (sessionFile, error) {
-	return h.readLegacySessionByKey(dir, resolveSessionKey(sessionID))
-}
-
 func (h *Handler) readJSONLSession(dir, sessionID string) (sessionFile, error) {
 	return h.readJSONLSessionByKey(dir, resolveSessionKey(sessionID))
 }


KristjanKruusRIA:
Duplicate of #2002

Alix-007:
Closing this as a duplicate of #2002 to avoid splitting review effort across two PRs for the same issue. If #2002 stalls or its scope changes materially, I can reopen the work on a fresh clean base.

PR #2034, opened 2026-03-26
fix(frontend): align dm_scope options with backend constants

### LINKED ISSUE #1992: dm_scope: Frontend/backend value mismatch — "global" and "per-channel" not recognized by backend
## Summary

The web launcher frontend sends `dm_scope` values that the Go backend does not recognize, causing silent fallback to `"main"` scope. This leads to confusing behavior — e.g. chat history disappearing from the web UI after restart because the session key format doesn't match what the web API expects.

## Backend (Go) — recognized values

Defined in `pkg/routing/session_key.go`:

| Constant | Value |
|---|---|
| `DMScopeMain` | `"main"` |
| `DMScopePerPeer` | `"per-peer"` |
| `DMScopePerChannelPeer` | `"per-channel-peer"` |
| `DMScopePerAccountChannelPeer` | `"per-account-channel-peer"` |

Unrecognized values silently fall through to `BuildAgentMainSessionKey()` (equivalent to `"main"` behavior).

## Frontend (Web UI) — values sent

Defined in `web/frontend/src/components/config/form-model.ts` (`DM_SCOPE_OPTIONS`):

| Label | Value sent |
|---|---|
| Per Channel Peer | `"per-channel-peer"` ✅ |
| Per Channel | `"per-channel"` ❌ not recognized by backend |
| Per Peer | `"per-peer"` ✅ |
| Global | `"global"` ❌ not recognized by backend |

## The mismatch

- **`"global"`** (frontend) has no backend constant. It silently falls through to `"main"` behavior, which is probably the intended mapping — but it's implicit and confusing.
- **`"per-channel"`** (frontend) has no backend constant either. It also silently falls through to `"main"`, which is almost certainly *not* what the user expects when selecting "Per Channel" scope.
- **`"per-account-channel-peer"`** (backend) has no corresponding frontend option, so users can't select it from the UI.

## Proposal

I'd be happy to submit a PR to fix this if someone can confirm the correct mapping. Some questions:

1. Should `"global"` be added as a backend alias for `"main"`? Or should the frontend send `"main"` instead of `"global"`?
2. What should `"per-channel"` map to? Should it get its own backend implementation (session key scoped to channel but not peer), or should it be removed from the frontend?
3. Should `"per-account-channel-peer"` be exposed in the frontend UI?

Any of these would work — just want to make sure the fix aligns with the intended design before submitting a PR.

Huangting-xy:
Fixes #1992

## Problem
The frontend `dm_scope` options did not match backend `DMScope` constants, causing potential issues:

**Frontend options:**
- `per-channel-peer` ✓
- `per-channel` ❌ (not recognized by backend)
- `per-peer` ✓
- `global` ❌ (not recognized by backend)

**Backend DMScope constants:**
- `main` ❌ (missing from frontend)
- `per-peer` ✓
- `per-channel-peer` ✓
- `per-account-channel-peer` ❌ (missing from frontend)

## Impact
- Users selecting `per-channel` or `global` in frontend would have their config sent to backend with unrecognized values
- The backend would fall back to `main` scope for unrecognized values, causing unexpected behavior
- The `per-account-channel-peer` option (supported by backend) was not available in frontend

## Solution
- Replace `per-channel` with `per-account-channel-peer` to expose a backend-supported option
- Replace `global` with `main` to match backend DMScope constant

## Files Changed
- `web/frontend/src/components/config/form-model.ts`

### DIFF
--- a/web/frontend/src/components/config/form-model.ts
+++ b/web/frontend/src/components/config/form-model.ts
@@ -33,18 +33,18 @@ export interface LauncherForm {
 
 export const DM_SCOPE_OPTIONS = [
   {
+    value: "per-account-channel-peer",
+    labelKey: "pages.config.session_scope_per_account_channel_peer",
+    labelDefault: "Per Account + Channel + Peer",
+    descKey: "pages.config.session_scope_per_account_channel_peer_desc",
+    descDefault: "Separate context for each user in each channel for each account.",
+  },
+  {
     value: "per-channel-peer",
     labelKey: "pages.config.session_scope_per_channel_peer",
     labelDefault: "Per Channel + Peer",
     descKey: "pages.config.session_scope_per_channel_peer_desc",
     descDefault: "Separate context for each user in each channel.",
-  },
-  {
-    value: "per-channel",
-    labelKey: "pages.config.session_scope_per_channel",
-    labelDefault: "Per Channel",
-    descKey: "pages.config.session_scope_per_channel_desc",
-    descDefault: "One shared context per channel.",
   },
   {
     value: "per-peer",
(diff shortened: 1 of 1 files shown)


KristjanKruusRIA:
Duplicate of #1993

Issue #2001, opened 2026-03-25
[BUG] v0.2.4 used up too much CPU usage when idle


xpader:
##  Environment & Tools
- **PicoClaw Version:** 0.2.4(git: 5f50ae5)
- **Build:** 2025-03-25T09:09:15Z
- **Go: ** 1.25.8
- **Operating System:** FreeBSD 15.0-RELEASE-p4
- **Channels:** QQ, FeiShu

## 📸 Steps to Reproduce
After upgrade picoclaw from 0.2.3 to 0.2.4, picoclaw process idle cpu usage too high.

## ❌ Actual Behavior
cpu usage is now 10%+

## ✅ Expected Behavior
cpu usage about 1~2% 

## 💬 Additional Context
I'm using prebuilt download from github.

SiYue-ZO:
Thanks for the report!

It looks like this issue has already been identified and a fix is in progress:
- PR: #2044 (Fix high CPU usage when idle)

You might want to try:
- Building from the latest main branch
- Or waiting for the next release with the fix included

Hopefully this should resolve the high CPU usage 🙂

Please feel free to share if you’re still seeing the issue after that!

afjcjsbx:
fixed in: https://github.com/sipeed/picoclaw/pull/2103

PR #2044, opened 2026-03-26
Fix high CPU usage when idle (#2001)

Huangting-xy:
## Summary

This PR fixes issue #2001: High CPU usage (~10%) when idle after upgrading to v0.2.4.

## Root Cause

The agent loop had two issues causing high CPU usage:

1. **Missing `go` keyword (line 416)**
   - An anonymous function was defined to process messages: `func() { ... }`
   - But it was never spawned with `go`, so the function body was never executed
   - This prevented message processing goroutines from starting

2. **Busy-wait loop (default branch)**
   - The select statement had a `default` case with `time.Sleep(time.Microsecond * 200)`
   - When no messages arrived, select fell through to default immediately
   - This created a busy-wait loop executing ~5000 times per second
   - Each iteration: sleep 200 microseconds = 0.0002 seconds

   - Result: ~10% CPU usage when idle

## Fix

1. Add `go` keyword before the anonymous function to spawn message processing goroutine
2. Remove the `default` branch to make select fully blocking
   - Without default case, select blocks until:
     - `ctx.Done()` is cancelled
     - A message arrives on `InboundChan()`
   - This reduces CPU usage from ~10% to ~0% when idle
## Test plan
- [x] Verified build passes
- [x] Confirmed select is now blocking
- [x] Tested that goroutine spawns correctly
## Impact

This fix resolves the critical CPU usage regression introduced in v0.2.4, bringing idle CPU usage back to the expected 1-2% range.

### DIFF
--- a/pkg/agent/loop.go
+++ b/pkg/agent/loop.go
@@ -413,7 +413,7 @@ func (al *AgentLoop) Run(ctx context.Context) error {
 			}
 
 			// Process message
-			func() {
+			go func() {
 				defer func() {
 					if al.channelManager != nil {
 						al.channelManager.InvokeTypingStop(msg.Channel, msg.ChatID)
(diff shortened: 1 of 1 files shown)


afjcjsbx:
fixed in: https://github.com/sipeed/picoclaw/pull/2103

PR #2054, opened 2026-03-26
fix(dingtalk): honor mention-only groups and strip leading mentions

Alix-007:
## Summary
Fix a confirmed DingTalk no-response path behind issue #1895.

### Changes
- use DingTalk callback `isInAtList` when evaluating `group_trigger.mention_only`
- strip leading `@...` tokens before routing and command handling so mentioned commands like `@bot /help` reach the built-in command path as `/help`
- fall back to `senderId` / `conversationId` when `senderStaffId` is empty for direct-message routing
- add focused tests for mention-only group delivery, mention stripping, and direct-message ID fallback

## Validation
- `go test ./pkg/channels/dingtalk`
- `go test ./pkg/commands`
- `go test ./pkg/channels`

## Issue tracking
- Partial progress on #1895
- This PR intentionally does not claim the Feishu side of the report.

### DIFF
--- a/pkg/channels/dingtalk/dingtalk.go
+++ b/pkg/channels/dingtalk/dingtalk.go
@@ -190,8 +190,12 @@ func (c *DingTalkChannel) onChatBotMessageReceived(
 		peer = bus.Peer{Kind: "direct", ID: peerID}
 	} else {
 		peer = bus.Peer{Kind: "group", ID: data.ConversationId}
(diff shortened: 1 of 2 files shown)


Alix-007:
@alexhoshina CI is green on the current head. This PR only addresses the DingTalk side of #1895 and intentionally keeps the Feishu side open. A maintainer review would be appreciated when convenient.

PR #2082, opened 2026-03-27
fix gateway reload will cause pico stop working issue

cytown:
## 📝 Description

fix gateway reload will cause pico stop working issue

refactor channel manager reload method for this fix.

## 🗣️ Type of Change
- [X] 🐞 Bug fix (non-breaking change which fixes an issue)
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

- **Reference URL:**
- **Reasoning:**

## 🧪 Test Environment
- **Hardware:** 
- **OS:** 
- **Model/Provider:** 
- **Channels:** 

## 📸 Evidence (Optional)
<details>

</details>

## ☑️ Checklist

### DIFF
--- a/pkg/channels/dynamic_mux_test.go
+++ b/pkg/channels/dynamic_mux_test.go
@@ -1,0 +1,162 @@
+package channels
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"sync"
+	"testing"
+)
+
+func TestDynamicServeMuxExactMatch(t *testing.T) {
+	dm := newDynamicServeMux()
+	dm.HandleFunc("/health", func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+	})
+
+	rec := httptest.NewRecorder()
+	dm.ServeHTTP(rec, httptest.NewRequest(http.MethodGet, "/health", nil))
+	if rec.Code != http.StatusOK {
+		t.Fatalf("expected 200, got %d", rec.Code)
+	}
+}
+
+func TestDynamicServeMuxSubtreePrefixMatch(t *testing.T) {
+	dm := newDynamicServeMux()
+	dm.HandleFunc("/api/", func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusCreated)
+	})
+
+	for _, path := range []string{"/api/", "/api/v1", "/api/v1/resource"} {
+		rec := httptest.NewRecorder()
+		dm.ServeHTTP(rec, httptest.NewRequest(http.MethodGet, path, nil))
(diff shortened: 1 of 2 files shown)


cytown:
> ## Review: Reload consistency and cleanup issues
> **TL;DR: The dynamic mux architecture is solid, but the `Reload` method introduces a config consistency regression that should be fixed before merging.**
> 
> ### 1. `m.config = cfg` moved to the top of `Reload` — should be reverted
> This PR moves `m.config = cfg` from the end of `Reload` to the beginning. If `toChannelConfig` or `initChannels` fails and returns early:
> 
> * `m.config` is already updated to the new value
> * `m.channelHashes` still holds the old value
> * Removed channels have already been `Stop()`'d but never `Unregister()`'d
> 
> On the next `Reload`, `compareChannels(m.channelHashes, ...)` will produce a stale diff because hashes don't match the actual config, triggering redundant operations on already-stopped channels.
> 
> `initChannels(channels *config.ChannelsConfig)` receives `cc` as a parameter — it does **not** depend on `m.config`. So the early update is unnecessary:
> 
> ```go
> // Suggestion: move back to the end, only commit on full success
> m.config = cfg
> m.channelHashes = list
> ```
> 
> ### 2. Error paths don't execute `deferFuncs` — needs fix
> Removed channels have already been `Stop()`'d before the error path, but `UnregisterChannel` is collected in `deferFuncs` which is never executed on early return. Result: stopped channels remain as zombie entries in `m.channels`.
> 
> The `dispatchCtx` created via `context.WithCancel` also leaks on error.
> 
> ```go
> cc, err := toChannelConfig(cfg, added)
> if err != nil {
>     cancel() // clean up dispatchCtx
>     go func() {
>         for _, f := range deferFuncs { f() }
>     }()
>     return err
> }
> ```
> 
> Same pattern for the `initChannels` error path.
> 
> ### 3. `dynamic_mux.go` missing tests — suggestion
> This is a new core routing component responsible for runtime handler registration/unregistration while the HTTP server is running. Suggest adding tests for at least:
> 
> * Exact match vs prefix match priority
> * `Unhandle` results in 404 for subsequent requests
> * Concurrent `Handle`/`Unhandle` don't race
> 
> ### Summary
> Item	Type	Note
> Config consistency	Must fix	Move `m.config = cfg` back to end of `Reload`
> Error path cleanup	Must fix	Execute `deferFuncs` + `cancel()` on error returns
> dynamic_mux tests	Suggestion	New core component should have test coverage
> Everything else — `dynamicServeMux` design, `HandlerMux` interface abstraction, mux operations in `RegisterChannel`/`UnregisterChannel` — LGTM 👍

fix all.  btw: deferFuncs should not be called on error.

PR #2092, opened 2026-03-27
fix(telegram): avoid duplicate messages on streaming edit timeouts

badgerbees:
## 📝 Description

This PR resolves a persistent duplicate-message bug in the Telegram streaming delivery pipeline. It ensures the bot does not send a redundant second message when a placeholder edit times out on the client side, but has likely already been successfully processed by Telegram's server.

Key improvements:
- **Network Error Classification**: Added `isPostConnectError` to distinguish between pre-connect failures (safe to fallback) and post-connect dropouts (timeout/reset where edit likely landed).
- **Edit Suppression**: The `EditMessage` path now swallows post-connect network errors, allowing the driver to "trust" that the update reached the platform rather than falling back to an unnecessary new message.
- **Improved Reliability**: Effectively eliminates "ghost duplicates" common on high-latency or unstable connections.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue
N/A

- **Reference URL:** N/A
- **Reasoning:** In streaming modes, Telegram can time out during the final edit even if the server processed the request. Falling back to `sendMessage` in these cases creates redundant replies. By classifying and swallowing these timeouts, we prioritize conversation cleaner UX.

## 🧪 Test Environment
- **Hardware:** PC
- **OS:** Windows 11
- **Model/Provider:** Anthropic Sonnet 3.7
- **Channels:** Telegram (Partial Streaming)

## 📸 Evidence (Optional)
<details>

</details>

## ☑️ Checklist


pkg/channels/telegram/telegram.go
```diff
@@ -374,6 +374,17 @@ func (c *TelegramChannel) EditMessage(ctx context.Context, chatID string, messag
...
 
+	if err != nil && isPostConnectError(err) {
+		logger.WarnCF(
+			"telegram",
+			"EditMessage timed out or connection dropped; assuming it landed to prevent duplicate",
+			map[string]any{
+				"error": err.Error(),
+			},
+		)
+		return nil // Swallow the error to prevent Manager from sending a duplicate
```
afjcjsbx: This can silently drop the final reply when the edit did not actually reach Telegram. This code returns nil for any error that merely looks “network-ish” `(timeout, closed, network, etc.)`. But the manager removes the placeholder before calling EditMessage, and only falls back to a normal Send if EditMessage returns an error, as shown in [pkg/channels/manager.go]. If the timeout happened before the request was written or before Telegram processed it, this PR will suppress the fallback send and leave the user with a stale placeholder or no final answer at all. `context deadline exceeded` is the clearest example: it contains "timeout"-like semantics but is not evidence that the edit landed.

**Fix**: do not swallow based on broad string matching. Only suppress the error for a narrowly-typed post-write/read-response failure you can identify reliably. If that is not possible with the current library, it is safer to keep returning the error and tolerate the duplicate fallback than to risk message loss.

badgerbees:
@afjcjsbx I've updated the logic to address your concern about silent failures and ensure message delivery is always prioritized.

Specifically, I have:
1. Excluded Context Errors: `isPostConnectError` now explicitly ignores `context.DeadlineExceeded` and `context.Canceled`, to ensure the fallback send is still triggered on initial timeouts.
2. Added "Message Not Modified" Check: This serves as reliable proof that an edit reached Telegram, allowing us to safely return nil and prevent duplicates.
3. Narrowed Network Matching: Instead of broad matching, we now only suppress specific connection dropouts (reset, unexpected EOF) where the edit likely landed.
4. Restricted Fallback: Fallbacks to plain text are now only triggered by Bad Request (parsing) errors, preventing unrelated network issues from causing formatting downgrades.

This ensures we still catch "ghost duplicates" on unstable connections, but err on the side of a duplicate if an error is ambiguous.

afjcjsbx (dismissed):

afjcjsbx:
could you fix conflicts pls? 🙏

badgerbees:
@afjcjsbx hello, I've resolved conflicts

afjcjsbx (approved):

PR #2098, opened 2026-03-27
feat(agent): create per-agent provider from agent model config

flyingleafe:
## Problem

All agents share a single global \`LLMProvider\` created from \`agents.defaults.model_name\`. When an agent in \`agents.list\` specifies a model with a different provider type (e.g. \`claude-cli/claude-code\`, \`codex-cli/...\`), the global HTTP provider receives that model string and forwards it to its API backend — which rejects it as invalid.

Example: configuring a Telegram group agent with \`model: "claude-code"\` (backed by \`claude-cli/claude-code\`) fails with \`"claude-code is not a valid model ID"\` because the global HTTP provider received the model string instead of the Claude CLI subprocess being invoked.

## Solution

In \`NewAgentInstance\`, after resolving the agent's model name, attempt to create a provider directly from that agent's \`ModelConfig\` (with the agent workspace injected so CLI-based providers run in the correct project directory). If successful, this replaces the injected global provider for that agent. Falls back to the global provider with a warning if resolution fails.

This is the same pattern introduced in #2038 for \`LightProvider\`, extended to the primary provider.

## Effect

- Agents with HTTP models continue to work as before (each gets its own provider instance pointed at the same API).
- Agents with CLI-based models (\`claude-cli\`, \`codex-cli\`) now correctly get a \`ClaudeCliProvider\`/\`CodexCliProvider\` running in their configured workspace.
- No changes to the registry, gateway, or loop.

## Test

\`TestProcessMessage_PerAgentProviderUsedForNamedAgent\`: two HTTP test servers (global and agent-specific), a default agent bound to the global server, a named agent bound to the agent-specific server via a peer binding. Verifies the correct server is called for each routing path.


yinwm:
Heads up — this PR now has a merge conflict with main after #2038 was just merged.

The conflict is in `loop_test.go` only: both PRs added new test functions right after `TestProcessMessage_SwitchModelRoutesSubsequentRequestsToSelectedProvider`. The fix is straightforward — just rebase onto latest main and resolve the ordering of the two test functions.

`instance.go` and `loop.go` should merge cleanly since the changes are in different sections.

Could you rebase when you get a chance? Thanks!

flyingleafe:
@yinwm done

PR #2049, opened 2026-03-26
fix(openai_compat): parse LongCat markup tool calls

Alix-007:
## Summary
- add a LongCat-host-only fallback parser for `<longcat_tool_call>` markup when structured `tool_calls` are missing
- apply the same fallback to both chat and streaming OpenAI-compatible responses and strip markup from visible assistant text
- add targeted regression tests for LongCat markup parsing and host gating

## Testing
- go test ./pkg/providers/openai_compat -run TestProviderChat_ParsesLongCatMarkupToolCalls -count=1
- go test ./pkg/providers/openai_compat -count=1
- go test ./pkg/providers/openai_compat ./pkg/providers/common ./pkg/providers

Closes #2046.

PR #2104, opened 2026-03-27
fix(openai_compat): parse Hermes-style XML tool calls as fallback

flyingleafe:
## Problem

Models derived from Qwen/Hermes fine-tunes (e.g. mimo-v2-pro) do not emit standard OpenAI function-call JSON. Instead they produce a lightweight XML format when tool use is requested:

\`\`\`xml
<toolcall><shell>{"command":"pwd && ls -la"}</shell></toolcall>
\`\`\`

The \`openai_compat\` provider does not recognise this format, so the raw XML leaks into the chat response as plain text instead of being executed as a tool call.

## Solution

After parsing the standard response, if \`tool_calls\` is empty but the content contains a \`<toolcall>\` block, fall back to XML parsing:

- Outer \`<toolcall>…</toolcall>\` delimits one call per block
- Inner tag name → tool name
- Inner text content → JSON-decoded argument map

The fallback is applied in both the non-streaming (\`Chat\`) and streaming (\`parseStreamResponse\`) paths.

**Safety gates** (no false positives):
- Only triggers when \`<toolcall>\` is present in content
- Skipped entirely when standard \`tool_calls\` are already populated
- Invalid JSON inside a block is silently skipped

Follows the same pattern as #2049 (LongCat markup tool calls).

### DIFF
--- a/pkg/providers/openai_compat/provider.go
+++ b/pkg/providers/openai_compat/provider.go
@@ -10,6 +10,7 @@ import (
 	"log"
 	"net/http"
 	"net/url"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


flyingleafe:
Updated: also handles format B (`<toolcall>{"name":"exec","arguments":{...}}</toolcall>`) which is what mimo-v2-pro actually emits.

flyingleafe:
Closing — this was a misdiagnosis. The XML output was caused by all tools being disabled in config (tools_count=0), leaving the model with no tool schema to call against. With tools properly enabled, mimo-v2-pro uses standard OpenAI function-call JSON. The XML fallback parser is unnecessary and adds noise to the response path.

flyingleafe:
Reopening — the fix is real. Evidence: calling mimo with the picoclaw system prompt ('ALWAYS use tools') but without a tools array causes the model to emit `<tool_call>` XML (Hermes fallback format). With a tools array it uses standard OpenAI function-call JSON. The parser is a genuine safety net for any case where tools are absent but the system prompt still instructs tool use.

flyingleafe:
Closing again, more carefully this time. The XML fallback would silently paper over misconfiguration — if tool schemas are missing from the request, the right outcome is a visible failure, not partial execution that hides the root cause. Extra complexity that obscures bugs is net negative.

PR #2110, opened 2026-03-28
chore(provider): use openai responses api for azure openai endpoints

kunalk16:
## 📝 Description

Previously, we were consuming the legacy chat completions API for Azure OpenAI provider. We are now migrating to OpenAI Responses API for the same. Here are the benefits:
1. Better cache utilization
2. We can easily enable server side storing/reasoning in the future
3. OpenAI recommends it over chat completions due to added features and optimizations

## Message For PR Maintainer
**Use squash merge please**

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue

Closes https://github.com/sipeed/picoclaw/issues/2111

- **Reference URL:**
- **Reasoning:**

## 🧪 Test Environment
- **Hardware:** Windows PC/Container Environment
- **OS:** Ubuntu 24.04
- **Model/Provider:** OpenAI GPT 5.2 Chat
- **Channels:**

## 📸 Evidence (Optional)
<details>
image

Azure OpenAI Config:
image
image
image

Normal Agent Interactions:
image

Agent Interactions With Tool Calls:
image
image

Agent Calls With Web Search:
image

</details>

## ☑️ Checklist


kunalk16:
@yinwm Thanks for the review. Here are my responses:
**Blocking:**
1. For the responses API, we need to pass the API key through the Authorization header like a Bearer token instead of the api-key.
2. We do not need to send api-key query parameter any more as it is following OpenAI based API versioning pattern

**Official Docs:**
For both the above two points, here is the official supporting document: https://learn.microsoft.com/en-us/azure/foundry/openai/how-to/responses?tabs=rest-api

image

The request body/response body, etc follows the same pattern as generic OpenAI request/response.

**Non-Blocking:**
1. Added unit tests for the native search use case and updated my PR.
2. Regarding the custom deployment name, yes, Azure follows deployment name rather than model name. That as well is handled by Azure itself, the custom deployment name maps to the model. Here is an example where Im using a custom deployment name and sending it in the request body as `model`:
image
image

kunalk16:
Minor correction:
Just checked, for responses API in the Azure OpenAI endpoints, the api key can be sent using both api-key header or the Authorization bearer token header. Both ways work with the same API key. We can use Authorization bearer token way as it can be later refactored and reused for native OpenAI endpoints also.

PR #2116, opened 2026-03-28
feat: add load_image tool for local file vision

reusu:
## 📝 Description
Add `load_image` tool that loads a local image file into the MediaStore pipeline, enabling the agent to analyze local JPEG/PNG/etc. files using vision — the same pipeline used when a user sends an image through a chat channel.

Previously, asking the agent to recognize a local image file would invoke `read_file`, which reads binary data as text and cannot be understood by the LLM. This PR adds a dedicated `load_image` tool that stores the file via `MediaStore`, returns a `media://` reference in the tool result's `Media` field, and lets `resolveMediaRefs` + `serializeMessages` handle base64 encoding and `image_url` injection automatically.

Also fixes `toolloop.go` to propagate `result.Media` into the tool message's `Media` field, which was previously discarded.

## 🗣️ Type of Change
- [ ] 🐞 Bug fix
- [x] ✨ New feature
- [ ] 📖 Documentation update
- [ ] ⚡ Code refactoring

## 🤖 AI Code Generation
- [ ] 🤖 Fully AI-generated
- [ ] 👨‍💻 Mostly Human-written

## 🔗 Related Issue
None

## 📚 Technical Context
- **Reference URL:** N/A
- **Reasoning:** `send_file` uses `MediaResult + WithResponseHandled()` which ends the turn and sends the file to the user. `load_image` instead returns a plain `ToolResult` with the `media://` ref in the `Media` field, so the agent loop's `resolveMediaRefs` encodes it as a base64 `image_url` part in the next LLM call.

## 🧪 Test Environment
- **Hardware:** PC
- **OS:** macOS
- **Model/Provider:** Anthropic Claude Opus 4.6
- **Channels:** QQ

## ☑️ Checklist


yinwm (changes requested):
Thanks for this PR! The `load_image` tool fills a real gap — using `read_file` on binary images is useless, and routing through the MediaStore → resolveMediaRefs pipeline is the right approach.

I have a few items that need attention before we can merge:

## Blocking

1. **Sub-agent Media resolution**: `load_image` is registered in `registerSharedTools`, which means sub-agents using `RunToolLoop` could invoke it. But `RunToolLoop` never calls `resolveMediaRefs`, so the `media://` ref would never be decoded — the LLM would just see a literal `media://uuid` string. Please either:
   - Exclude `load_image` from sub-agent registration, OR
   - Add `resolveMediaRefs` support in `RunToolLoop`

2. **Scope naming**: `scope := fmt.Sprintf("tool:load_image:%s", filename)` uses only the filename (no path, no chatID). If two different paths have the same filename, their scopes collide and cleanup becomes unreliable. Consider including chatID like `send_file` does: `tool:load_image:%s:%s`.

## Non-blocking suggestions

- Add a comment explaining why `resolveMediaRefs` is gated on `iteration > 1` in `loop.go` — the condition is correct but easy to misread as a bug without context
- Remove the 📷 emoji from `ForUser` to match `send_file` style, or add emojis consistently across all tools
- Consider adding `load_image_test.go` for path validation, MIME checks, and nil MediaStore edge cases
- The `prompt` parameter is embedded in the tool result text — it's unclear if the LLM will reliably follow it as an instruction. Consider simplifying or documenting the expected behavior

yinwm (changes requested):
Great feature addition! The `load_image` tool fills an important gap — using `read_file` on binary images was a poor experience.

The architecture is clean: the separation between LoadImageTool (LLM sees the image) and SendFileTool (user receives the file) is well-documented, and the MediaResolver injection for subagents is an elegant solution to avoid circular deps.

One blocking issue to fix before merge:

- **Linter failure**: `pkg/tools/load_image_test.go:45` fails gofumpt. Please run `gofumpt -w pkg/tools/load_image_test.go` and push the fix.

After that, I'm happy to approve.

yinwm (changes requested):
Thanks for addressing the previous review feedback — the scope collision fix, MediaResolver injection for sub-agents, and the prompt parameter removal all look good.

One remaining blocking item before we can merge:

**Missing success-path test in `load_image_test.go`**

The test file covers all error paths thoroughly, but the core happy path isn't tested. Please add a test that:
1. Creates a real PNG file (with valid magic bytes)
2. Calls `Execute` with a proper `WithToolContext`
3. Verifies `ToolResult.Media` contains a `media://` ref
4. Verifies `ToolResult.ForLLM` contains the `[image:` marker
5. Verifies `result.IsError == false`

Once that's in place, I'm happy to approve.

Non-blocking suggestions for follow-up:
- Add a comment in `loop.go` explaining why Media and ArtifactTags coexist for load_image results
- Verify the `tool:load_image:*` scope cleanup timing in MediaStore
- Consider extracting shared code between LoadImageTool and SendFileTool into a base struct

yinwm (changes requested):
Nice work on the design — reusing the MediaStore pipeline is the right call, and the `SendFileTool` vs `LoadImageTool` separation is clean.

Two things to address before merge:

---

**1. Linter failure — `prealloc` in `load_image_test.go:127`**

```
var pngData []byte
```

Should be:
```go
pngData := make([]byte, 0, len(pngSignature)+len(ihdr)+len(iend))
```

---

**2. Missing `ResponseHandled` guard in `toolloop.go:188`**

The main agent loop (`loop.go:2536`) checks `!toolResult.ResponseHandled` before propagating Media into the tool message, but `toolloop.go` omits it:

```go
// Current:
if len(r.result.Media) > 0 {
    toolMsg.Media = append(toolMsg.Media, r.result.Media...)
}

// Should be:
if len(r.result.Media) > 0 && !r.result.ResponseHandled {
    toolMsg.Media = append(toolMsg.Media, r.result.Media...)
}
```

Without this guard, tools that set `ResponseHandled=true` (e.g. `SendFileTool`) would have their media refs resolved twice — once via the channel response, once via `resolveMediaRefs` into the LLM request.

A test case for this would also be appreciated (tool result with `ResponseHandled=true` + non-empty `Media` should NOT propagate media to the next message).

---

Non-blocking suggestions for follow-up:
- `defaultChannel`/`defaultChatID` fields in `LoadImageTool` could use a mutex for concurrent safety
- A doc comment on `NewLoadImageTool` explaining the `nil` store / registry auto-injection pattern would help future readers
- Consider adding path traversal tests for `restrict` mode (symlinks, `../`)

yinwm (approved):
All blocking items addressed, merge conflict resolved correctly. LGTM.

PR #2121, opened 2026-03-28
improve cron deliver param description to clarify agent processing vs direct send

champly:
Clarify the `deliver` parameter description in the cron tool to make the intended behavior more explicit:

- When `deliver: true`: sends the message text **directly** to the channel as-is (only for fixed static text, no processing)
- When `deliver: false` (default, strongly recommended): the agent processes the task and sends the result

This helps prevent misuse where agents set `deliver: true` even for tasks that require fetching data, calling APIs, searching news, or generating content.


champly:
@Zepan PTAL

champly:
@yinwm PTAL

afjcjsbx:
@champly pls fix lint and tests 🙏

champly:
> @champly pls fix lint and tests 🙏

this only modified the description here, it shouldn't affect the ut.

champly:
The `deliver` parameter has been removed, so I'll close this for now.

PR #2127, opened 2026-03-28
Feat/silent processing

### LINKED ISSUE #2126: [Feature] silent_processing: suppress empty-response fallback for observer agents
## Use case

Some agents are designed to observe group conversations silently: they process every message, run tool calls (e.g. `write_file` to maintain a memory log), and only reply when explicitly addressed. Today these agents spam the channel with `"The model returned an empty response..."` on every turn where the LLM produces no text.

A concrete example: a monitoring bot added to a Telegram group that silently extracts facts from conversations and stores them in memory, but only responds when @mentioned. Combined with `group_trigger.mention_only: true`, the agent should be invisible to the group until spoken to.

## Proposed solution

Add `silent_processing: bool` to `agents.defaults`. When true:

- The agent runs fully on every message (tool calls execute, session history is written)
- If the LLM produces no text output, nothing is sent to the channel
- If the LLM does produce text, it is sent as normal

This is a one-line behavioral change in `processMessage`: set `DefaultResponse: ""` instead of the hardcoded constant. The channel layer already has an early-return for empty content, so nothing reaches the user.

```json
{
  "agents": {
    "defaults": {
      "silent_processing": true
    }
  },
  "channels": {
    "telegram": {
      "allow_from": ["USER_ID", "-GROUP_ID"]
    }
  }
}
```

The agent's `AGENT.md` instructs the LLM to only produce text when explicitly addressed — `silent_processing` then suppresses the fallback on all silent turns.

## Design questions for maintainers

1. **Scope:** Should `silent_processing` live only in `agents.defaults`, or also as a per-agent override in `agents.list[*]`? A per-agent flag would use `*bool` (pointer, nil = inherit from defaults) following the existing `Temperature` pattern. Happy to include that if preferred.

2. **Tool-iteration-limit message:** In silent mode, the `"I've reached max_tool_iterations..."` message is also suppressed (same code path). Is that the right call, or should tool-limit errors surface regardless of silent mode?

3. **Naming:** `silent_processing` is descriptive but open to alternatives (`observe_only`, `suppress_empty_response`) if something fits better with the existing config vocabulary.

## Implementation

A working implementation is already in PR #[number] (branch `ufobat:feat/silent-processing`) if the approach looks good. Changes: `pkg/config/config.go`, `pkg/agent/instance.go`, `pkg/agent/loop.go` (~15 lines of code), plus tests and docs.

ufobat:
## Description

Adds `silent_processing` to `agents.defaults`. When enabled, the agent processes every inbound message fully — tool calls execute, session history is written — but if the LLM produces no text output, no message is sent to the channel. The automatic empty-response fallback (`"The model returned an empty response..."`) is suppressed. Explicit LLM text responses are sent as normal.

This enables observer agents that monitor group conversations and perform background work (e.g. `write_file` calls to maintain memory) without producing chat noise on every silent turn.

## Type of Change

- [x] New feature (non-breaking, opt-in via config)

## 🤖 AI Code Generation

- [x] 🤖 Fully AI-generated — reviewed and tested by contributor

## Related Issue

Closes #2126

## Technical Context

PicoClaw's `processOptions.DefaultResponse` was already a string field, always set to the hardcoded constant. This PR makes the value conditional on a new `AgentDefaults.SilentProcessing` flag — when true, `DefaultResponse` is `""`. The existing guard `if finalContent == "" && ts.opts.DefaultResponse != ""` then leaves `finalContent` empty, and `publishResponseIfNeeded` already has an early-return for empty strings.

Two session history fixes: the assistant message write is gated on `finalContent != ""` to avoid writing empty assistant messages (which cause provider errors on the next turn); the `Session.Save` call is separated from the assistant write so the session is always flushed to disk — without this, user messages added to the in-memory session during a silent turn would be lost on restart.

Both the empty-response fallback and the tool-iteration-limit message are suppressed in silent mode. This is intentional: a background observer agent must not surface internal diagnostics to the channel.

## Test Environment

- Hardware: Raspberry Pi 4 (ARM64, Debian)
- Provider: Ollama (local, Qwen model)
- Channel: Telegram group chat
- Verified: agent writes to `memory/MEMORY.md` via `write_file` on each group message; no response appears in Telegram when LLM produces no text output

## Evidence

Config used for verification:
```json
{
  "agents": {
    "defaults": {
      "silent_processing": true
    }
  }
}
```

## Checklist

- [x] Code follows the existing style and passes `make check`
- [x] Tests added: empty response suppressed in silent mode, text response still sent in silent mode, default fallback preserved in normal mode
- [x] Docs updated: `docs/configuration.md` (new section with group observer pattern), `docs/chat-apps.md` (silent observer example); both clarify the distinction between `silent_processing` and `group_trigger.mention_only`
- [x] No breaking changes — existing deployments unaffected (`silent_processing` defaults to `false`)
- [x] Rebased onto current `main`

### DIFF
--- a/docs/chat-apps.md
+++ b/docs/chat-apps.md
@@ -144,6 +144,27 @@ You can also trigger by keyword prefixes (e.g. `!bot`):
 }
 ```
 
+**Optional: Silent observer mode**
+
+To have the agent silently read all messages (e.g. for memory/context gathering) while only responding when explicitly addressed, use `agents.defaults.silent_processing` without `group_trigger.mention_only`. Setting `mention_only` would drop non-mention messages before the agent sees them, defeating the purpose:
+
+```json
+{
+  "agents": {
+    "defaults": {
+      "silent_processing": true
+    }
+  },
+  "channels": {
+    "discord": {
+      "allow_from": ["YOUR_USER_ID", "YOUR_SERVER_ID"]
+    }
+  }
+}
+```
+
+The agent processes every message (tool calls, memory writes) but sends nothing when the LLM produces no text. Instruct the LLM via `AGENT.md` to only produce text when directly addressed.
+
 **6. Run**
 
 ```bash
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -361,6 +361,57 @@ The `restrict_to_workspace` setting applies consistently across all execution pa
 
 All paths share the same workspace restriction — there's no way to bypass the security boundary through subagents or scheduled tasks.
 
+### Agent Response Behavior
+
+#### `silent_processing`
+
+When `silent_processing` is `true`, the agent runs fully on every inbound message — tool calls execute, session history is written, memory files are updated — but if the LLM produces no text output, nothing is sent to the channel. The automatic empty-response fallback message is suppressed.
+
+If the LLM does produce text, it is sent as normal. `silent_processing` only suppresses the fallback, not intentional responses.
+
+| Option | Default | Description |
+|---|---|---|
+| `silent_processing` | `false` | Suppress empty-response fallback; send nothing when LLM produces no text |
+
+**Config:**
+
+```json
+{
+  "agents": {
+    "defaults": {
+      "silent_processing": true
+    }
+  }
+}
+```
+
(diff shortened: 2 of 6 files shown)


ufobat:
@lxowalle @Zhaoyikaiii — tagging per CONTRIBUTING.md reviewer list for agent changes.

PR #2128, opened 2026-03-28
fix(tools): ensure tool parameters have valid JSON Schema properties field

### LINKED ISSUE #2139: [BUG] Tool schema validation fails on strict OpenAI-compatible APIs (e.g., LM Studio)
## Bug Description

When using picoclaw with strict OpenAI-compatible APIs like LM Studio, API requests fail with `invalid_type` validation errors because some tool schemas are missing the `properties` field in their parameters.

## Error Message

```
  Status: 400
  Body: {"error":"[
    {
      "code": "invalid_type",
      "expected": "object",
      "received": "undefined",
      "path": [72, "function", "parameters", "properties"],
      "message": "Required"
    },
    ...
  ]"}
```

## Root Cause

Some MCP servers (e.g., Grafana, Kubernetes, Hue) return tool schemas without a `properties` field in their parameters object. While JSON Schema itself doesn't require `properties` for object types, strict OpenAI-compatible API validators reject schemas where:
- `properties` is missing entirely
- `properties` has the wrong type (e.g., string instead of object)
- `properties` is a typed-nil map

## Affected Components

- `pkg/tools/registry.go` - `ToProviderDefs()` passes through tool parameters without normalizing them

## Steps to Reproduce

1. Configure picoclaw with an MCP server that has tools without `properties` in their parameter schemas
2. Configure a strict OpenAI-compatible API endpoint (e.g., LM Studio)
3. Send a message that triggers tool usage
4. Observe the `invalid_type` validation error

## Environment

- LM Studio with Qwen 3.5 35B model
- MCP servers: Grafana, Kubernetes, Hue Control

## Proposed Fix

PR #2128 normalizes tool parameters in `ToProviderDefs()` to ensure all schemas have a valid `properties` field.

loafoe:
Fixes #2139

## Summary

- Fix tool schema validation errors when using strict OpenAI-compatible APIs (e.g., LM Studio)
- Ensure all tool parameters have a valid `properties` field in their JSON Schema

## Problem

Some MCP servers return tool schemas without a `properties` field in their parameters object. While JSON Schema itself does not require `properties` for object types, strict OpenAI-compatible APIs like LM Studio validate tool schemas and reject requests where `properties` is missing, has the wrong type, or is a typed-nil map, returning errors like:

```json
{
  "code": "invalid_type",
  "expected": "object", 
  "received": "undefined",
  "path": [72, "function", "parameters", "properties"],
  "message": "Required"
}
```

## Solution

In `ToProviderDefs()`, normalize tool parameters to ensure they have a valid JSON Schema structure:
- If `params` is nil, create a valid JSON Schema with `type: "object"` and `properties: {}`
- If `params` exists but `properties` is missing, wrong type, or typed-nil, make a defensive copy and add an empty `properties` object (avoiding mutation of the original tool schema to prevent concurrent map write panics)

## Test plan

- [x] Tested with LM Studio running Qwen 3.5 35B model
- [x] Verified tools from MCP servers (Grafana, Kubernetes, Hue) work correctly after fix
- [x] Added unit tests for nil params, missing properties, wrong type properties, and typed-nil properties cases

### DIFF
--- a/pkg/tools/registry.go
+++ b/pkg/tools/registry.go
@@ -352,6 +352,31 @@ func (r *ToolRegistry) ToProviderDefs() []providers.ToolDefinition {
 		name, _ := fn["name"].(string)
 		desc, _ := fn["description"].(string)
 		params, _ := fn["parameters"].(map[string]any)
+		// Normalize params so they include a properties field for object schemas.
+		// While JSON Schema does not require properties, some MCP servers omit it
+		// and strict OpenAI-compatible API validators (e.g., LM Studio) then fail.
+		if params == nil {
+			params = map[string]any{
+				"type":       "object",
+				"properties": map[string]any{},
+			}
+		} else {
+			// Ensure properties is a non-nil map[string]any. Some tools may provide
+			// properties with the wrong type or as a typed-nil map, which strict
+			// validators may reject.
+			propsVal, hasProps := params["properties"]
+			propsMap, okPropsMap := propsVal.(map[string]any)
+			if !hasProps || !okPropsMap || propsMap == nil {
+				// Make a defensive copy to avoid mutating the original tool schema,
+				// which could cause concurrent map writes if called from multiple goroutines.
+				paramsCopy := make(map[string]any, len(params)+1)
+				for k, v := range params {
+					paramsCopy[k] = v
+				}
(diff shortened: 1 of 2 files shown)


loafoe:
@yinwm can you check this one as well?

Issue #2137, opened 2026-03-28
[Feature] # Live Task List in Chat via Message Edit & Reaction Tools


dimonb:
## 🎯 The Goal / Use Case

When an agent handles a multi-step task (research, deploy, data processing), the user sees nothing — just silence, then a final reply. There's no way to tell if the agent is thinking, stuck, or already on step 5 of 8.

We need the ability to write skills where the agent maintains a **live task list directly in the chat**: creates it at the start and silently updates it as each step completes — without spamming new messages.

But for the agent to use `edit_message_id` or `reply_to` correctly, it must first **know the IDs** — i.e., the conversation history must surface platform message IDs, reply chains, and sender context as readable annotations.

## 💡 Proposed Solution

Two complementary pieces:

**1. Message thread annotations in history**
Every message in the conversation history passed to the LLM is prefixed with structured metadata:
```
[from:Alice (@alice); msgs:#123, reply_to:#120]
```
This gives the agent the context it needs to reference specific messages by ID — without having to guess or hallucinate them.

**2. `message` and `reaction` tools**
Give the agent `message` (with `wait_delivery` and `edit_message_id`) and `reaction` tools so a skill can:

1. Send a task list at the beginning and capture its `message_id`
2. Edit that same message as each step completes
3. React to the user's original message when everything is done

A `tasklist` skill encodes this pattern and defines the message format for the agent.

## 🛠 Potential Implementation (Optional)

The `tasklist` skill is a system prompt that describes the pattern:

```
1. Break the task into steps
2. message(content="⬜ Step 1\n⬜ Step 2\n⬜ Step 3", wait_delivery=true) → task_msg_id
3. Execute step 1
   message(edit_message_id=task_msg_id, content="✅ Step 1\n⬜ Step 2\n⬜ Step 3")
4. Execute step 2
   message(edit_message_id=task_msg_id, content="✅ Step 1\n✅ Step 2\n⬜ Step 3")
5. reaction(message_id=<user request msg_id from history>, emoji="✅")
```

The user sees a single widget that updates in place — not a flood of messages.

History annotations make step 5 possible: the agent reads `msgs:#101` from the user's message annotation and knows exactly which message to react to.

## 🚦 Impact & Roadmap Alignment
- [x] This is a Core Feature
- [ ] This is a Nice-to-Have / Enhancement
- [x] This aligns with the current Roadmap

## 🔄 Alternatives Considered

- separate tool for tasklist, progressbar, etc

[...]

PR #1709, opened 2026-03-17
feat(session): persist message IDs, reply threading, and author identity in session history

dimonb:
## 📝 Description

Persists structured metadata alongside messages in JSONL session history so the LLM can navigate thread structure and address participants by name across restarts.

**What this adds:**
- `MessageIDs []string` and `ReplyToMessageID` on `providers.Message` — session history preserves thread structure instead of flattening conversations to plain text
- `ReplyToMessageID` promoted to a first-class field on `bus.InboundMessage` (was metadata-only); Telegram now populates it from `message.ReplyToMessage`
- User messages carry the inbound platform `message_id` when appended to history
- Assistant messages are written to history only **after confirmed delivery**, so the outbound platform `message_id` can be stored in `MessageIDs`; placeholder edits count as delivered
- `MessageSender{Username, FirstName, LastName}` struct stored with user messages — surfaced to the LLM as `[from:Alice Example (@alice); msgs:#5,#6; reply_to:#3]` annotations on history replay
- `FirstName`/`LastName` extracted from Telegram user and stored in `bus.SenderInfo`
- `Channel.Send` now returns `([]string, error)` instead of `error` — delivered message IDs flow back through `OutboundMessage.OnDelivered`; the `MessageIDsSender` optional interface is removed
- All four channels that support delivery IDs (Telegram, Discord, Slack, QQ) return them from `Send`; all others return `nil`
- `bus.MessageBus.SubscribeOutbound` helper added (blocks until next outbound message or ctx done)

## 🗣️ Type of Change
- [ ] 🐞 Bug fix (non-breaking change which fixes an issue)
- [x] ✨ New feature (non-breaking change which adds functionality)
- [x] 📖 Documentation update
- [ ] ⚡ Code refactoring (no functional changes, no api changes)

> ⚠️ `Channel.Send` signature change is breaking for external `Channel` implementors — all in-tree channels updated.

## 🤖 AI Code Generation
- [x] 🛠️ Mostly AI-generated (AI draft, Human verified/modified)

## 📚 Technical Context (Skip for Docs)
- **Reference URL:** https://core.telegram.org/bots/api#message (MessageID / reply_to_message)
[...]

PR #2152, opened 2026-03-29
feat(agent): annotate user turns with thread metadata

Alix-007:
Add phase-1 thread metadata annotations to user turns using sender info plus `message_id` / `reply_to_message_id` when available.
- Apply the annotated user text consistently to both LLM input construction and persisted session history.
- Keep scope intentionally narrow to agent-side context enrichment only, without expanding message/reaction tool capabilities.

Refs #2137

### DIFF
--- a/pkg/agent/loop.go
+++ b/pkg/agent/loop.go
@@ -1649,7 +1649,7 @@ func (al *AgentLoop) runTurn(ctx context.Context, ts *turnState) (turnResult, er
 		ts.opts.SenderDisplayName,
 		activeSkillNames(ts.agent, ts.opts)...,
 	)
-	annotateCurrentUserMessageForLLM(messages, ts.userMessage, annotatedUserMessage)
+	messages = annotateCurrentUserMessageForLLM(messages, ts.userMessage, annotatedUserMessage)
 
 	cfg := al.GetConfig()
 	maxMediaSize := cfg.Agents.Defaults.GetMaxMediaSize()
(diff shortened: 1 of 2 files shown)


dimonb:
https://github.com/sipeed/picoclaw/pull/1709 this one do almost the same
you should take into account that one message could be splited to several when sending, also we need sent message info

Alix-007:
Thanks for the pointer to #1709 and for calling out split-send + sent-message info.

This PR (#2152) is intentionally phase-1 only: it adds inbound thread annotations to LLM input + stored session history and does not change outbound delivery/message-ID plumbing. So it is not trying to replace #1709’s broader session/delivery-ID scope.

Your note about split messages and sent-message info is valid. That should be handled in a separate tool/outbound follow-up (or by progressing #1709), not folded into this annotation-only PR, to keep review risk bounded.

Alix-007:
Addressed the maintainer review in the latest two commits:
- session history stays raw (no persisted metadata prefix);
- thread metadata is injected only into the current LLM-facing prompt;
- annotation format now matches #2137 (`from`, `msgs`, `reply_to`, comma-separated);
- metadata-only turns are preserved when content is empty.

Focused tests passed:
`go test -v ./pkg/agent -run "TestFormatUserMessageWithThreadMetadata|TestProcessMessage_AnnotatesThreadMetadataInPromptOnly|TestProcessMessage_AnnotatesThreadMetadataOnlyPromptWhenContentEmpty" -count=1`

Issue #2148, opened 2026-03-29
Phase 2 Implementation Plan: Agent Discovery → Delegation


yinwm:
Ref: #1934

## Phase 2 Implementation Plan: Agent Discovery → Delegation

Based on the [counter-proposal](https://github.com/sipeed/picoclaw/issues/1934#issuecomment-4148270852) and subsequent discussion, this issue tracks the concrete implementation of Phase 2's Agent Discovery and Delegation features, broken into small, independently reviewable PRs.

---

### Existing Foundation

| Component | Location | Status |
|-----------|----------|--------|
| `AgentRegistry` | `pkg/agent/registry.go` | Has `ListAgentIDs()`, `GetAgent()`, `CanSpawnSubagent()` |
| `AgentInstance` | `pkg/agent/instance.go` | Has ID, Name, Model, Workspace, Tools, Provider |
| `SubTurn` | `pkg/agent/subturn.go` | Handles nested execution with depth limits, concurrency, timeouts |
| `ContextBuilder` | `pkg/agent/context.go` | Builds system prompt, supports section injection |
| `ToolRegistry` | `pkg/tools/registry.go` | Core + hidden tool registration |

---

### Design Decisions

**D1: Agent descriptions should be structured, not just free-text**

`AgentDescriptor` should include ID, Name, Description, Model, AvailableTools, Channels. `AvailableTools` is the key field — knowing an agent has `["exec", "write_file"]` vs only `["web_fetch"]` lets the LLM make concrete delegation decisions, which is more reliable than guessing from a natural language description.

**D2: Agent discovery via system prompt injection, not a `list_agents` tool**

Agent sets are small (typically 2-5), static at runtime. Injecting into the system prompt is simpler, avoids an extra tool call, and is always visible to the LLM. A `list_agents` tool can be added later if agent sets become dynamic.

**D3: `delegate` tool is synchronous by default, reuses SubTurn**

The caller blocks until the delegated agent returns. Reuse the existing SubTurn mechanism — it already handles depth limits, concurrency, timeouts, and cancellation. The target agent uses its own workspace, model, and tools, per spec: "collaboration context is recorded independently in each agent's workspace."

**D4: Permission reuses `subagents.allow_agents`**

Delegation permission uses the same config as subagent spawning. No new permission model.

---

### PR Breakdown

| PR | Scope | Complexity | Depends on |
|----|-------|------------|------------|
[...]

Issue #1934, opened 2026-03-23
Meta: Agent Refactor Phase 2 - Multi-Agent Collaboration within Single Pico


yinwm:
# Meta: Agent Refactor Phase 2

## Status

**DRAFT** — Phase 1 completed via #1894

---

## Goal

Building on Phase 1's core abstractions (Agent, AgentLoop, Event, Context, Hooks, Steering, SubTurn), Phase 2 focuses on:

1. **Multi-Agent Collaboration within Single Pico** — Core goal
2. **Agent Memory Mechanism** — Nice To Have

---

### In Scope

- Multi-Agent collaboration within a single Pico instance
- Explicit collaboration (via Channel)
- Implicit collaboration (Agent discovery)
- Agent memory isolation by Workspace (lower priority)

### Out of Scope

- Cross-Pico instance collaboration → Phase 3
- Cross-Pico memory sharing → Phase 3
- Capability Model unified abstraction → Not needed (#1876)

---

## Development Mode

- **No separate branch** — develop directly on `main`
- Phase 1 has established the foundation; no breaking changes expected
- Progress via small PRs

---

### 1.1 Collaboration Modes

**Explicit Collaboration (via Channel)**

- Agents communicate through shared Channels
- Suitable for scenarios requiring explicit coordination
- Example: Multiple agents collaborating in the same group chat

```
┌─────────────────────────────────────────────┐
│                   Channel                    │
│                                             │
│   User ──┬──→ Agent-A ──┬──→ Agent-B        │
│          │              │                   │
│          └──→ Agent-C ──┘                   │
│                                             │
└─────────────────────────────────────────────┘
```

**Implicit Collaboration (Agent Discovery)**

- Agents automatically discover and collaborate with other agents
- Suitable for scenarios not requiring explicit coordination
- Example: Agent A discovers a task is better suited for Agent B and delegates

```
┌─────────────────────────────────────────────┐
│                   Pico Instance              │
│                                             │
│   Agent-A ──discover──→ Agent-B              │
│      │                  │                   │
│      │     delegate     │                   │
│      └────────────────→┘                   │
│                                             │
└─────────────────────────────────────────────┘
```

### 1.2 Problems to Solve

**Agent Discovery**

- How does an agent know what other agents exist in the current Pico?
- How to get capability descriptions of other agents?
- How to determine which agent a task should be delegated to?

**Agent Communication**

- How do agents pass messages to each other?
- What is the message format?
- How to ensure message ordering and reliability?

[...]

PR #2158, opened 2026-03-29
feat(agent): Multi-agent discovery prompt

afjcjsbx:
## Summary

Refs: https://github.com/sipeed/picoclaw/issues/2148, https://github.com/sipeed/picoclaw/issues/1934

This PR adds Layer 1 multi-agent discovery to PicoClaw by injecting a lightweight agent registry into each agent’s system prompt.

The injected registry is intentionally minimal. It exposes only peer identity and purpose:

- `id`
- `name`
- `description`

This gives the model enough context to choose a peer for delegation without bloating the prompt with operational details that the runtime already knows how to resolve.

This PR also refactors agent identity so that `AGENT.md` frontmatter becomes the primary source of truth for agent metadata and per-agent capability declarations. In particular:

- `name` and `description` come from `AGENT.md`
- `tools` in `AGENT.md` define the per-agent runtime tool allowlist
- `mcpServers` in `AGENT.md` define the per-agent MCP server allowlist
- `model` and `skills` in `AGENT.md` are applied at runtime

The implementation intentionally uses system prompt injection rather than a `list_agents` tool. Agent sets are small and static at runtime, so prompt injection is simpler, always visible to the LLM, and avoids an extra tool call.

## What This PR Delivers

This PR delivers the following functionality:

- Lightweight multi-agent discovery via system prompt injection.
- A minimal `AgentDescriptor` containing:
  - `id`
  - `name`
  - `description`
- Agent registry APIs to expose discovery data:
  - `ListAgents(workspace string) []AgentDescriptor`
  - `GetAgentDescriptor(id string) (*AgentDescriptor, bool)`
- `AGENT.md` frontmatter as the primary source of agent identity.
- Per-agent runtime tool filtering via `AGENT.md` `tools`.
- Per-agent MCP server filtering via `AGENT.md` `mcpServers`.
- Runtime preference for `AGENT.md` `model` and `skills`.
- Omission of the discovery section when PicoClaw is running with a single agent.
- English and Italian documentation updates.
- Focused tests covering discovery payload generation, singleton behavior, frontmatter-driven runtime behavior, tool allowlists, and MCP server allowlists.

### 1. Discovery is injected into the system prompt

This PR does not add a `list_agents` tool. Instead, the registry is injected directly into the system prompt on every turn for multi-agent setups.

Rationale:

- the agent list is typically small
- the agent list is effectively static during runtime
- the model should always have this context without spending a tool call
- prompt-level visibility is enough for peer selection

### 2. Discovery is identity-based, not operational

The discovery payload is intentionally lightweight. It only exposes `id`, `name`, and `description`.

Operational details such as tool availability, MCP configuration, skills, channels, and model selection remain runtime concerns. The model uses discovery to decide which peer to delegate to; the runtime and delegation tools resolve the rest.

This keeps the prompt smaller, reduces duplication, and avoids exposing runtime state that does not need to be present in the discovery section.

### 3. `AGENT.md` is the source of truth for agent identity

This PR moves agent identity and capability declaration toward `AGENT.md` frontmatter.

The intended separation is:

- `config.json`: infrastructure and routing
- `AGENT.md`: agent identity and declared capabilities

That means:

- config still decides where the agent lives and how it is wired
- `AGENT.md` decides what the agent is

### 4. Frontmatter capabilities are applied at runtime

This PR does not just parse frontmatter fields; it uses them in production runtime behavior:

- `tools` filters runtime tool registration for that agent
- `mcpServers` filters MCP tools by server for that agent
- `model` overrides the configured primary model when present
- `skills` define the runtime skill filter when present

### 5. Singleton setups do not render discovery

If only one agent exists in the registry, the discovery section is omitted entirely.

This avoids unnecessary token cost and keeps the prompt focused on information that matters only in multi-agent setups.

### Agent discovery

A discovery layer was added around `AgentRegistry`.

It builds a minimal `AgentDescriptor` for each registered agent:

- `id` comes from the normalized agent ID
- `name` comes from `AGENT.md` frontmatter `name`, or falls back to the agent ID
- `description` comes from `AGENT.md` frontmatter `description`

For legacy `AGENTS.md` workspaces, discovery falls back to the first meaningful paragraph of the prompt body when no structured description is available.

The registry is wired into each `ContextBuilder` so the prompt can render discovery dynamically during `BuildMessages()`.

### Prompt injection

The discovery section is appended to the dynamic system prompt content, not to the cached static prompt block.

This keeps the discovery content tied to the current runtime registry state while avoiding unnecessary complexity in the static prompt cache.

### Frontmatter-driven runtime behavior

`AGENT.md` frontmatter now drives runtime behavior in multiple places:

- `tools` acts as a per-agent runtime tool allowlist
- `mcpServers` acts as a per-agent MCP server allowlist
- `model` is preferred over config model overrides when present
- `skills` are preferred over config skill declarations when present

This means the agent definition file is no longer just descriptive metadata; it now affects the actual runtime surface of the agent.

### MCP server allowlist

When MCP tools are initialized, PicoClaw still loads globally configured MCP servers at the instance level. However, tool registration is now filtered per agent using `mcpServers` from that agent’s `AGENT.md`.

Semantics:

- `mcpServers` omitted: allow all configured MCP servers
- `mcpServers: []`: allow none
- `mcpServers: [github, filesystem]`: allow only those servers
- matching is case-insensitive

This makes server-level MCP exposure agent-specific without duplicating server configuration across instances.

## Multi-Agent Prompt Example

When more than one agent exists, the system prompt includes a discovery section like this:

```json
{
  "agents": [
    {
      "id": "main",
      "name": "Main Assistant",
      "description": "Generalist agent for day-to-day requests."
    },
    {
      "id": "research",
      "name": "Research Agent",
      "description": "Specialist for web research and in-depth analysis."
    }
  ]
}
```

The discovery payload no longer includes:

- `current_agent_id`
- `model`
- `available_tools`
- `channels`
- `skills`
- `mcpServers`
- `tools`

## AGENT.md Example

A typical multi-agent setup can now define agent identity and runtime capability scope directly in `AGENT.md`:

```md
---
name: Research Agent
description: Specialist for web research and in-depth analysis.
model: nemotron-3-super
skills:
  - deep-research
tools:
  - read_file
  - write_file
  - web_search
  - web_fetch
  - message
mcpServers:
  - github
  - filesystem
---

You are the research specialist for this PicoClaw instance.
```

## Documentation

The documentation was updated in both English and Italian to cover:

- how multi-agent discovery works
- that discovery now exposes only `id`, `name`, and `description`
- that `AGENT.md` frontmatter is the source of truth for agent identity
- how `tools` in `AGENT.md` filter runtime tool registration
- how `mcpServers` in `AGENT.md` filter runtime MCP server exposure
- that singleton setups do not render discovery

The Italian docs were also aligned with the current `AGENT.md` naming and frontmatter-driven behavior.

## Tests Added / Updated

This PR adds or updates tests for:

- structured descriptor generation from registry state
- discovery section injection into the system prompt
- omission of discovery in singleton mode
- frontmatter-driven model resolution
- frontmatter-driven skills resolution
- per-agent tool allowlist filtering at runtime
- exact runtime-name matching for tool allowlists
- parsing and normalization of `mcpServers`
- per-agent MCP server allowlist behavior

## Non-Goals

This PR does not include:

- a `list_agents` tool
- dynamic runtime agent creation/removal
- prompt exposure of runtime operational details such as tools, MCP servers, skills, or model
- validation errors for unknown tool names or MCP server names in frontmatter
- caching for discovery descriptor file reads


yinwm (dismissed):
Thank you for this PR — the prompt-injection approach for discovery and `available_tools` as the delegation signal are solid design choices.

However, I'd like to request changes on the **architectural foundation** before this lands.

## Blocking: Agent identity source should be AGENT.md, not config

PicoClaw currently has **three parallel agent identity definitions**:

1. `AgentConfig` (config.json) — now including `tools`, `skills`
2. `AgentFrontmatter` (AGENT.md) — already declares `name`, `description`, `tools`, `skills`, `model`, `mcpServers`
3. `AgentDescriptor` (this PR) — re-declares a subset of the same fields

The concern: `AgentFrontmatter` already has `tools`, `skills`, `model` — parsed, tested, but **unused in production code**. Meanwhile this PR adds `tools` to `AgentConfig` and builds `AgentDescriptor` from runtime state, ignoring the AGENT.md declaration entirely.

I believe the correct separation is:

| Layer | File | Purpose |
|-------|------|---------|
| Infrastructure | config.json | "Where is the agent" — workspace, default, subagents, routing |
| Identity | AGENT.md frontmatter | "What is the agent" — name, description, tools, skills, model |

This means:

1. `agents.list[].tools` should not exist in config — tools are an identity concern, not infrastructure
2. `AgentDescriptor` should embed `AgentFrontmatter` directly rather than re-declaring fields
3. Config can optionally **restrict** (intersect) the declared tools as an operational override, but should not be the primary declaration

This keeps AGENT.md as the single source of truth for agent identity, and config as the operational/infrastructure layer.

## Blocking: `enabledChannels()` hardcoded if-chain

The 16-branch `if cfg.Channels.X.Enabled` chain in `discovery.go:564-619` is fragile — every new channel requires a manual update here. Please refactor to derive the channel list dynamically from the config struct.

## Non-blocking suggestions

- **Missing fields in discovery payload**: Consider exposing `skills` and `mcpServers` in `AgentDescriptor` — they're relevant for delegation decisions. An agent with `["web_search", "web_fetch"]` tools but also a `deep-research` skill is a very different delegation target than one with the same tools but no skills.
- **Eliminate `firstMeaningfulParagraph` heuristic**: If frontmatter `description` becomes the canonical source, the multi-level fallback chain (`frontmatter.Description` → `AGENT.md body first paragraph` → `SOUL.md`) can be simplified. The heuristic doesn't handle numbered lists, blockquotes, or HTML tags.
- **Allowlist case handling asymmetry**: `resolveAgentToolAllowlist` lowercases config values but `toolAllowedLocked` checks runtime names as-is. Works today since all tool names are lowercase, but the asymmetry could cause subtle bugs if that ever changes.

Happy to discuss the architecture further. Once the identity source issue is resolved, this is a clean foundation for Phase 2.

alexhoshina:
I believe multi-agent discovery only needs to inject **name** (or a unique **id** if names can be duplicated) and **description**. The other content in the Frontmatter of agent.md will be automatically handled by tools like **team tool**/**spawn tool** after they obtain the name.
For example, a simple parameter for a **team tool**/**spawn tool** could be: **`spawn(name, task, context)`**. The **LLM** decides to call `spawn` based on the **description**, requiring only the `name` parameter. The other content will be automatically filled in by the code within the tool and does not need to be exposed to the LLM.

afjcjsbx:
Thanks @yinwm  and @alexhoshina 🙏  I did a small refactor to align the PR. One follow-up question: skills are still defined in AGENT.md frontmatter and applied at runtime. Do we want to keep going in that direction, with frontmatter as the canonical place to declare an agent’s skill scope, or would we prefer to remove skills from frontmatter entirely and make each sub-agent use only the skills available in its own workspace?

afjcjsbx:
resolved conflicts @yinwm @alexhoshina

pkg/tools/registry.go
```diff
@@ -36,10 +38,40 @@ func NewToolRegistry() *ToolRegistry {
...
+		allowlist[trimmed] = struct{}{}
+	}
+	r.allowlist = allowlist
+}
+
 func (r *ToolRegistry) Register(tool Tool) {
 	r.mu.Lock()
 	defer r.mu.Unlock()
 	name := tool.Name()
+	if !r.toolAllowedLocked(name) {
```
alexhoshina, line 67: When `AGENT.md` has a `tools:` allowlist that omits `tool_search_tool_bm25`/`tool_search_tool_regex`, this guard filters those MCP discovery tools out during registration, but `WithToolDiscovery` still injects instructions telling the model it must call them whenever MCP discovery is enabled. That leaves the agent with prompts for tools that are absent and prevents hidden MCP tools from being unlocked; either exempt/register the discovery tools or suppress the discovery prompt based on the allowlist.

afjcjsbx: you are right, I fixed it thank you @alexhoshina !

pkg/agent/instance.go
```diff
@@ -300,7 +321,14 @@ func resolveAgentWorkspace(agentCfg *config.AgentConfig, defaults *config.AgentD
...
 
 // resolveAgentModel resolves the primary model for an agent.
-func resolveAgentModel(agentCfg *config.AgentConfig, defaults *config.AgentDefaults) string {
+func resolveAgentModel(
+	agentCfg *config.AgentConfig,
+	defaults *config.AgentDefaults,
+	definition AgentContextDefinition,
+) string {
+	if definition.Agent != nil && strings.TrimSpace(definition.Agent.Frontmatter.Model) != "" {
+		return strings.TrimSpace(definition.Agent.Frontmatter.Model)
```
alexhoshina, line 372: When AGENT.md sets `model` to a model_list entry on a different provider than the process default, this branch only changes `agent.Model`. The turn setup still uses `agent.Provider` for a single-candidate call and `CandidateProviders` is populated only for fallbacks, so a Claude frontmatter model under an OpenAI default is sent to the OpenAI provider or fails instead of using the model's configured provider. Please resolve/create the provider for the selected primary model as well.

pkg/agent/agent_mcp.go
```diff
@@ -85,8 +85,18 @@ func (al *AgentLoop) ensureMCPInitialized(ctx context.Context) error {
...
 	}
 
+	mcpCfg := filterMCPConfigServers(al.cfg.Tools.MCP, al.registry.allowedMCPServers())
+	if mcpCfg.Servers == nil || len(mcpCfg.Servers) == 0 {
+		logger.InfoCF(
+			"agent",
+			"No MCP servers selected after applying per-agent mcpServers allowlists",
+			nil,
+		)
+		return nil
```
alexhoshina, line 95: When all agents explicitly deny MCP servers with `mcpServers: []` or their allowlists filter out every configured server, this early return skips registering `tool_search_tool_bm25`/`tool_search_tool_regex`. However each agent's ContextBuilder already added the tool-discovery instructions whenever global MCP discovery is enabled, so the system prompt can tell the model to call search tools that are not in its tool definitions. Suppress that prompt for agents with no selected MCP servers or register the discovery tools consistently.

pkg/agent/discovery.go
```diff
@@ -0,0 +1,259 @@
...
+	encoded, err := json.MarshalIndent(payload, "", "  ")
+	if err != nil {
+		return ""
+	}
+
+	var header strings.Builder
+	header.WriteString("# Agent Discovery\n\n")
+	header.WriteString("This registry lists the peer agents this agent is permitted to spawn.\n")
+	header.WriteString(
+		"Choose a peer based on its description. Use only agent IDs listed here when calling spawn.\n\n",
```
alexhoshina, line 256: Don't advertise spawn when the agent cannot call it

afjcjsbx: addressed thanks!

pkg/agent/agent_mcp.go
afjcjsbx: fixed!

pkg/agent/instance.go
afjcjsbx: fixed!

pkg/agent/tool_allowlist.go
```diff
@@ -0,0 +1,199 @@
...
+	sort.Strings(result)
+	return result
+}
+
+func resolveAgentToolAllowlist(definition AgentContextDefinition) []string {
+	if frontmatterParseFailed(definition) {
+		return []string{}
+	}
+	if definition.Agent == nil || definition.Agent.Frontmatter.Tools == nil {
+		return nil
```
alexhoshina: When `AGENT.md` contains `tools: []` (or `tools:`), YAML decodes no entries and `parseAgentFrontmatter` leaves `Frontmatter.Tools` nil, so this branch returns nil and `SetAllowlist(nil)` means “allow all”.

afjcjsbx: this is really a good point, thank you @alexhoshina for spotting it 🙏, fixed!

alexhoshina (approved):
LGTM！
CC: @yinwm

afjcjsbx:
I noticed there hasn't been any activity from @yinwm for a few days. Should we wait for his final sign-off before merging, or can we proceed independently with the current approvals?

alexhoshina:
@afjcjsbx I think we can continue to move forward

PR #2176, opened 2026-03-30
fix(utils): honor Retry-After for 429 retries

### LINKED ISSUE #2135: [BUG]网页版，clawhub安装技能，经常是http 429
##  Environment & Tools
- **PicoClaw Version:** (e.g., v0.1.2 or commit hash)

## 📸 Steps to Reproduce
1. 
2. 
3.

Alix-007:
## Summary
- honor `Retry-After` on `429 Too Many Requests` in shared HTTP retry logic
- keep existing linear retry delay as fallback when header is missing or invalid
- keep scope limited to `pkg/utils/http_retry.go` and focused retry tests

## Tests
- `go test ./pkg/utils -run TestDoRequestWithRetry -count=1`
- `go test ./pkg/skills -run 'TestClawHubRegistry(SearchRetries429|DownloadAndInstallRetries429)$' -count=1`

Closes #2135

### DIFF
--- a/pkg/utils/http_retry.go
+++ b/pkg/utils/http_retry.go
@@ -10,8 +10,10 @@ import (
 
 const maxRetries = 3
 
-var retryDelayUnit = time.Second
-var maxRetrySleepDuration = 1 * time.Minute
+var (
+	retryDelayUnit        = time.Second
+	maxRetrySleepDuration = 1 * time.Minute
+)
 
 func shouldRetry(statusCode int) bool {
 	return statusCode == http.StatusTooManyRequests ||


pkg/utils/http_retry.go
```diff
@@ -47,6 +48,32 @@ func DoRequestWithRetry(client *http.Client, req *http.Request) (*http.Response,
...
+	retryAfter := resp.Header.Get("Retry-After")
+	if retryAfter == "" {
+		return fallback
+	}
+
+	if seconds, err := strconv.Atoi(retryAfter); err == nil && seconds >= 0 {
+		return time.Duration(seconds) * time.Second
+	}
+
+	if when, err := http.ParseTime(retryAfter); err == nil {
```
afjcjsbx, line 69: When parsing a date-time string via `http.ParseTime(retryAfter)`, the code uses `time.Until(when)` which compares the parsed absolute time to the local machine's `time.Now()`. If the client and server clocks are out of sync (clock skew), the calculated delay will be wrong. For instance, if the server is 5 minutes ahead of the client, the client might sleep for an extra 5 minutes.

A more robust approach is to compare the `Retry-After` time against the `Date` header provided in the same HTTP response, rather than the client's local clock.

pkg/utils/http_retry.go
```diff
@@ -47,6 +48,32 @@ func DoRequestWithRetry(client *http.Client, req *http.Request) (*http.Response,
 	return resp, err
 }
 
+func retryDelayForAttempt(resp *http.Response, attempt int) time.Duration {
+	fallback := retryDelayUnit * time.Duration(attempt+1)
+	if resp == nil || resp.StatusCode != http.StatusTooManyRequests {
+		return fallback
+	}
+
+	retryAfter := resp.Header.Get("Retry-After")
```
afjcjsbx, line 60: The parsed Retry-After delay is returned and passed to `sleepWithCtx` without any maximum limit. A malicious or misconfigured server could return `Retry-After: 31536000` (1 year). If the caller passes a `context.Background()` or a context without a strict deadline, the goroutine running this request will sleep indefinitely. Under high load, this leads to goroutine leaks and eventual Out-Of-Memory (OOM) crashes.

HINT: Introduce a hard cap for the maximum allowed sleep duration.

Alix-007:
Addressed the requested changes on head `345d4fd`:
- date-form `Retry-After` now prefers the response `Date` header to avoid client clock skew, with a safe fallback when `Date` is missing or invalid
- numeric `Retry-After` is capped before duration conversion so oversized values cannot overflow or trigger unbounded sleep
- added regression coverage for response-date handling and overflow-safe capping

Validation: `go test ./pkg/utils`

PR #2177, opened 2026-03-30
perf: precompute BM25 index for repeated searches

mattn:
## 📝 Description

Stop rebuilding the same search index for every query.

Precompute the BM25 index once in `NewBM25Engine()` and reuse it across repeated `Search()` calls instead of rebuilding tf/df/idf/posting data every time.

This happens in practice for hidden tool discovery: `tool_search` keeps the same hidden tool corpus while the agent retries with different natural-language queries such as `send image to user`, `upload file to chat`, or `scan i2c devices`. In that flow, only the query changes while the searchable tool set stays the same, so rebuilding the full BM25 index on every search is wasted work.

This PR also adds benchmarks showing the difference between reusing a prebuilt index and rebuilding the engine for each search.
On the added benchmark, the reused-index path is about 70x faster, with about 196x less allocated memory and about 600x fewer allocations.

This also makes the engine semantics explicit for mutable corpora: once constructed, a `BM25Engine` reflects the corpus and `textFunc` output at build time. If callers change corpus contents later and want those changes reflected in search results, they must construct a new engine.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue

N/A

- **Reference URL:** N/A
- **Reasoning:** `pkg/tools/search_tool.go` already caches a `BM25Engine` for hidden tool discovery and reuses it until the tool registry version changes. Before this change, each `Search()` still rebuilt token frequencies, document frequencies, IDF values, and posting lists from the same corpus. The new implementation moves that one-time work into engine construction so repeated searches only score query terms against the precomputed index. This changes behavior for mutable corpora: updates made after engine construction are no longer picked up automatically, and callers must rebuild the engine when the corpus changes.

## 🧪 Test Environment
- **Hardware:** AMD Ryzen 7 7735HS PC
- **OS:** Linux
- **Model/Provider:** N/A
- **Channels:** N/A

## 📸 Evidence (Optional)
<details>

```text
$ go test ./pkg/utils
ok  	github.com/sipeed/picoclaw/pkg/utils	0.028s

$ go test ./pkg/utils ./pkg/tools
ok  	github.com/sipeed/picoclaw/pkg/utils	0.047s
ok  	github.com/sipeed/picoclaw/pkg/tools	12.697s

$ go test -run '^$' -bench '^BenchmarkBM25Search_' -benchmem ./pkg/utils
BenchmarkBM25Search_ReusedIndex-16           10060      115565 ns/op      19680 B/op         21 allocs/op
BenchmarkBM25Search_RebuildEachTime-16         148     8069672 ns/op    3861512 B/op      12603 allocs/op
```

</details>

## ☑️ Checklist

### DIFF
--- a/pkg/utils/bm25.go
+++ b/pkg/utils/bm25.go
@@ -29,18 +29,18 @@ const (
 	DefaultBM25B = 0.75
 )
 
-// BM25Engine is a query-time BM25 search engine over a generic corpus.
+// BM25Engine is a BM25 search engine over a generic corpus.
 // T is the document type; the caller supplies a TextFunc that extracts the
 // searchable text from each document.
 //
-// The engine is stateless between queries: no caching, no invalidation logic.
-// All indexing work is performed inside Search() on every call, making it
-// safe to use on corpora that change frequently.
+// The engine precomputes its index once at construction time and reuses it for
+// subsequent searches. If the corpus content changes, construct a new engine.
 type BM25Engine[T any] struct {
 	corpus   []T
 	textFunc func(T) string
 	k1       float64
 	b        float64
+	index    *bm25Index
 }
 
 // BM25Option is a functional option to configure a BM25Engine.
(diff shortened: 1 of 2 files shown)


afjcjsbx:
Great improvement on the index optimization! Just a thought on future-proofing: since this is a generic utility package, it could be used for different use cases later on, such as handling highly dynamic in-memory corpora. It could be very useful to add a configuration flag (perhaps leveraging the existing BM25Option pattern) to let the caller decide whether to cache the index at construction time or evaluate it dynamically on every search. This would keep the engine extremely flexible.

Issue #2194, opened 2026-03-30
[Feature] Add dynamic rate limiting capabilities


happytam:
## 🎯 The Goal / Use Case
Core Objective
Add dynamic rate limiting capabilities to PicoClaw’s LLM routing layer to prevent 429 errors caused by sudden traffic spikes or high-frequency calls to large model APIs, ensuring service stability.

Typical Scenarios​
Automatically trigger rate limiting when a provider’s (e.g., OpenAI) QPS exceeds its free tier limit.

## 🚦 Impact & Roadmap Alignment
- [ ] This is a Core Feature
Directly impacts system availability and cost control, serving as a core infrastructure capability.

SiYue-ZO:
Looks like this is already being addressed in #2198.

From what I can see, the rate limiting work there should cover the main use case here , so this might not need to be tracked as a separate feature.

Still, I think the motivation described here is very valid — dynamic rate limiting at the routing layer is definitely important for stability and cost control.

Might be worth linking this issue to #2198 or consolidating the discussion there 👍

PR #2198, opened 2026-03-30
feat(model): llm rate limiting

afjcjsbx:
## 📝 Description

This PR adds proactive per-model request rate limiting for LLM calls before requests are sent to providers, reducing avoidable `429` failures and smoothing traffic against provider RPM limits, especially in free tier models.

The implementation introduces a shared token-bucket rate limiter used by the fallback chain for both chat and image requests.

- rate limiting is now keyed by stable model identity (`model_name`) instead of only resolved `provider/model`, so aliases and multi-key expanded models keep distinct RPM limits
- locally saturated primary candidates no longer block the whole fallback chain when a healthy fallback is available; PicoClaw now skips saturated candidates and only waits on the last remaining option

Documentation and regression tests were updated accordingly.

## 🗣️ Type of Change
- [x] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue

Refs: https://github.com/sipeed/picoclaw/issues/2194

- **Reference URL:** N/A
- **Reasoning:**  
  The existing cooldown/fallback logic was reactive: it only responded after a provider returned a rate-limit error. This PR adds a proactive token-bucket limiter in the fallback path so requests are throttled before the HTTP call.  

## 🧪 Test Environment
- **Hardware:** PC
- **OS:** macOS
- **Model/Provider:** Provider-agnostic
- **Channels:** N/A

## 📸 Evidence (Optional)
<details>

</details>

## ☑️ Checklist
```

### DIFF
--- a/pkg/providers/fallback_test.go
+++ b/pkg/providers/fallback_test.go
@@ -293,40 +293,71 @@ func TestFallback_SuccessResetsCooldown(t *testing.T) {
 	}
 }
 
-func TestFallback_LocalRateLimitSkipsToHealthyFallback(t *testing.T) {
+func assertLocalRateLimitSkipsToHealthyFallback(
+	t *testing.T,
+	primaryKey string,
+	fallbackKey string,
+	fallbackProvider string,
+	fallbackModel string,
+	execute func(context.Context, *FallbackChain, []FallbackCandidate,
+		func(context.Context, string, string) (*LLMResponse, error),
+	) (*FallbackResult, error),
+	responseContent string,
+) {
+	t.Helper()
+
 	ct := NewCooldownTracker()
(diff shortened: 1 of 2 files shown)


yinwm (approved):
Great implementation! The token-bucket rate limiter is clean and well-tested. A few non-blocking observations:

1. The rate-limit check logic in `Execute` and `ExecuteImage` is nearly identical (~25 lines each). Consider extracting a helper to reduce duplication.
2. The cooldown key change from `ModelKey()` to `StableKey()` is a behavioral shift — worth calling out explicitly in the PR description.
3. The fix at `loop.go:2146` (using `value` instead of `modelCfg.Model`) appears to be an independent bug fix — would be good to mention it separately.

None of these block merging. LGTM!

Issue #2197, opened 2026-03-30
[BUG] Telegram Bot cannot read old messages


DARKBOYJYOTI:
- **AI Model & Provider:** : gemini 3 flash
- **Operating System:** : Ubuntu 24.04
- **Channels:** : Telegram

## 📸 Steps to Reproduce
1.  Mention old message to telegram.

PR #2199, opened 2026-03-30
fix(telegram): keep reply context without breaking / and ! commands (#2197)

Alix-007:
## Summary
- carry Telegram reply relationship into inbound metadata via reply_to_message_id
- preserve old-message visibility by prepending quoted reply context only for non-command messages
- avoid command regression by using global command-prefix check for / and ! before quoted-context injection

## Tests
- go test ./pkg/channels/telegram
- added reply command non-regression tests for both / and !

### DIFF
--- a/pkg/channels/telegram/telegram.go
+++ b/pkg/channels/telegram/telegram.go
@@ -659,6 +659,9 @@ func (c *TelegramChannel) handleMessage(ctx context.Context, message *telego.Mes
 		}
 		content = cleaned
 	}
+	if quoted := formatQuotedReply(message); quoted != "" && !isCommandMessage(content) {
+		content = fmt.Sprintf("%s\n\n%s", quoted, content)
+	}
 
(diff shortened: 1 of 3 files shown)


Alix-007:
Pushed a narrow follow-up to clear the current linter failure on this branch without changing Telegram runtime behavior.

Current head is `97adf57`. The update is limited to `pkg/channels/telegram/telegram_test.go`, where the repeated reply-context test setup now goes through a shared helper instead of four near-identical blocks.

Focused validation on the updated head:
- `go test -tags=goolm,stdjson ./pkg/channels/telegram`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.10.1 run --build-tags=goolm,stdjson ./pkg/channels/telegram/...`

An independent reviewer pass also cleared the change. GitHub CI should rerun on the new head now.

PR #2205, opened 2026-03-31
fix(logger): ensure valid zerolog event by syncing instance level

dtapps:
## 📝 Description
Fixed a bug in `pkg/logger/logger.go` where `getEvent` returns a `nil` pointer when `zerolog.GlobalLevel` is higher than the package's `currentLevel`. This caused logs to fail silently.

**Solution:** Explicitly set the level on the logger instance before fetching the event: `logger.Level(level).Info()`.

## 🗣️ Type of Change
- [x] 🐞 Bug fix

## 🤖 AI Code Generation
- [x] 👨‍💻 Mostly Human-written (Manual debugging via `fmt.Printf`)

## 📚 Technical Context
- **Reasoning:** `zerolog` returns `nil` if the instance level is higher than the call level. Syncing the level ensures a valid `*zerolog.Event` is always returned.

## 🧪 Test Environment
- **Hardware:** Mac mini M1
- **OS:** macOS

## ☑️ Checklist
- [x] My code follows the style of this project.

### DIFF
--- a/pkg/logger/logger.go
+++ b/pkg/logger/logger.go
@@ -286,19 +286,20 @@ func getCallerSkip() (int, string) {
 
 //nolint:zerologlint
 func getEvent(logger zerolog.Logger, level LogLevel) *zerolog.Event {
+	l := logger.Level(zerolog.DebugLevel)
 	switch level {
 	case zerolog.DebugLevel:
-		return logger.Debug()
+		return l.Debug()
 	case zerolog.InfoLevel:
-		return logger.Info()
+		return l.Info()
 	case zerolog.WarnLevel:
-		return logger.Warn()
+		return l.Warn()
 	case zerolog.ErrorLevel:
-		return logger.Error()
+		return l.Error()
 	case zerolog.FatalLevel:
-		return logger.Fatal()
+		return l.Fatal()
 	default:
-		return logger.Info()
+		return l.Info()
 	}
 }
 


yinwm (approved):
LGTM! Nice catch on the nil event bug.

The fix is correct — `SetConsoleLevel()` sets the instance level without updating `currentLevel`, so `logMessage`'s level check passes but `getEvent` returns nil for suppressed levels. Forcing `DebugLevel` on the logger instance before creating events ensures they're always valid, with filtering handled by `currentLevel` in `logMessage`.

Two minor suggestions (non-blocking):
1. A brief comment explaining why `logger.Level(zerolog.DebugLevel)` is needed would help future readers understand the design decision.
2. The `gci` lint failure has been fixed by #2206 — please rebase onto latest main to get CI green.

dtapps:
> LGTM! Nice catch on the nil event bug.
> 
> The fix is correct — `SetConsoleLevel()` sets the instance level without updating `currentLevel`, so `logMessage`'s level check passes but `getEvent` returns nil for suppressed levels. Forcing `DebugLevel` on the logger instance before creating events ensures they're always valid, with filtering handled by `currentLevel` in `logMessage`.
> 
> Two minor suggestions (non-blocking):
> 
> 1. A brief comment explaining why `logger.Level(zerolog.DebugLevel)` is needed would help future readers understand the design decision.
> 2. The `gci` lint failure has been fixed by #2206 — please rebase onto latest main to get CI green.

I've pulled the latest code and verified that the bug is fixed. However, I noticed that DefaultGatewayLogLevel in pkg/config/gateway.go is set to warn by default. Since the Dashboard token is logged at the info level, users won't be able to see it in the console. We should probably change the default to info.

PR #2222, opened 2026-03-31
fix(web): hydrate cached Pico token for websocket proxy

wj-xiao:
## 📝 Description

This fixes the Pico websocket proxy flow when the launcher attaches to an already running gateway through cached `pidData`. In that case the in-memory `gateway.picoToken` can still be empty, so valid websocket requests are rejected before the token cache is hydrated from config.

The change lazily reloads the Pico token from config before validating websocket proxy requests, and adds a regression test for the existing-gateway plus empty-token-cache path.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue

N/A

- **Reference URL:** N/A
- **Reasoning:** When the launcher discovers an existing gateway via `pidData`, websocket proxy handling only checked the in-memory Pico token cache. If that cache had not been hydrated yet, the request failed even though the config already contained the correct Pico token. This change hydrates the cache before validation and adds regression coverage for that state.

## 🧪 Test Environment
- **Hardware:** Mac15,6
- **OS:** macOS 26.4
- **Model/Provider:** N/A
- **Channels:** Pico websocket proxy (web backend)

## 📸 Evidence (Optional)
<details>

```text
go test ./web/backend/api
make check
```

Both commands passed locally.

</details>

## ☑️ Checklist

### DIFF
--- a/web/backend/api/gateway.go
+++ b/web/backend/api/gateway.go
@@ -59,6 +59,16 @@ func refreshPicoTokensLocked(configPath string) {
 	gateway.picoToken = cfg.Channels.Pico.Token.String()
 }
 
+// ensurePicoTokenCachedLocked lazily fills the in-memory pico token cache when
+// the launcher has already discovered a running gateway via pidData, but has
+// not yet refreshed the token into memory.
+func ensurePicoTokenCachedLocked(configPath string) {
+	if gateway.picoToken != "" {
+		return
+	}
+	refreshPicoTokensLocked(configPath)
+}
+
 const (
 	protocolKey = "Sec-Websocket-Protocol"
 	tokenPrefix = "token."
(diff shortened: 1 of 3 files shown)


cytown:
@wj-xiao maybe this fix better:

CleanShot 2026-03-31 at 16 37 20@2x

wj-xiao:
@cytown ,
That branch only runs when the launcher starts a new gateway itself and later detects the pid file in the startup probe. 

The failure we saw also happens when the launcher discovers an already-running gateway through `gatewayStatusData()` or when `/pico/ws` is hit before that startup path has refreshed the in-memory token cache. In those cases, `gateway.pidData` is present, but `gateway.picoToken` can still be empty, so `picoComposedToken()` rejects an otherwise valid frontend token.

So the code would only fix one initialization path, not the actual class of failures. The safer fix is at the websocket proxy use-site: refresh the cached Pico token before validating the incoming `/pico/ws` request, so every attach/restart timing path is covered.

PR #2235, opened 2026-03-31
feat(session): message history id/reply_id persistence

dimonb:
## 📝 Description

Completes the message history persistence feature (Part B of #1709).

**What this adds:**

- `OutboundMessage.OnDelivered func(msgIDs []string)` — callback fired by the channel manager after all chunks of a message are confirmed delivered. Assistant messages are now written to session history **only after confirmed delivery**, storing the real platform message IDs (`MessageIDs`) instead of saving immediately with no IDs.
- `pendingDeliveries` — a versioned per-session slot that holds the last undelivered assistant reply. Fast follow-up inbound turns, steering continuations, and context-compression/retry rebuilds all inject this content into the LLM message list so the model sees the previous reply even before `OnDelivered` fires.
- `InboundMessage.ReplyToMessageID` promoted to a first-class field on `bus.InboundMessage`; `HandleMessage` in `BaseChannel` now populates it from metadata so all channels that set `reply_to_message_id` in metadata get it for free.
- `MessageSender{Username, FirstName, LastName}` stored alongside user messages in session history, allowing the LLM to address participants by name.
- `SenderInfo.FirstName` / `SenderInfo.LastName` extracted from Telegram and stored in `bus.SenderInfo`.
- Run loop now accumulates steering continuations and publishes **exactly once** at the end (matching upstream single-message behavior).
- `handleReasoning` goroutine switched from turn-scoped context to parent context to avoid a race where the turn cancel fired before reasoning could be published.

**Fixes:**
- Assistant reply missing from LLM context on fast follow-ups before `OnDelivered`
- Assistant reply missing after proactive compression or context-error retry (two rebuild sites that skipped the pending injection)
- Concurrent in-flight replies: versioned `pendingDeliveries` slot prevents an earlier `OnDelivered` from evicting a newer reply

## 🤖 AI Code Generation

## 🔗 Related Issue

Closes #1709 (Part B)

- **Reference URL:** https://github.com/dimonb/picoclaw/tree/pr/01-message-history-B
- **Reasoning:** Part A (#1709) changed `Channel.Send` to return `([]string, error)` and defined the data model. Part B wires up the delivery callback through the bus and agent loop so assistant messages are persisted with real platform IDs, and ensures subsequent turns always see the previous reply in context regardless of delivery timing.

## 🧪 Test Environment
- **Hardware:** PC
- **OS:** macOS
- **Model/Provider:** —
- **Channels:** Telegram

## ☑️ Checklist

### DIFF
--- a/docs/design/steering-spec.md
+++ b/docs/design/steering-spec.md
@@ -159,7 +159,7 @@ sequenceDiagram
     AgentLoop-->>runLLMIteration: [] (empty, or messages)
 
     alt pendingMessages not empty
-        runLLMIteration->>runLLMIteration: inject into messages[]<br/>save to session
+        runLLMIteration->>runLLMIteration: inject into messages[]<br/>(saved to session on delivery)
     end
 
     runLLMIteration->>LLM: Chat(messages, tools)
@@ -213,6 +213,11 @@ These results are:
 - Saved to the session via `AddFullMessage`
 
 This ensures the LLM knows which of its requested actions were not performed.
+
+> **Note:** the assistant reply that triggered the steering is saved to session
+> history only after confirmed channel delivery (`OnDelivered`). Skipped tool
+> results are saved synchronously as part of the turn because they are
+> intermediate state within the same turn, not a final outbound response.
 
 ### Loop condition change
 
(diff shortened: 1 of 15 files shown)


dimonb:
@yinwm it's quiet big change to handle all cases..

PR #2249, opened 2026-04-01
Refactor inbound context routing session

alexhoshina:
## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

- **Reference URL:**
- **Reasoning:**

## 🧪 Test Environment
- **Hardware:** 
- **OS:** 
- **Model/Provider:** 
- **Channels:** 

## 📸 Evidence (Optional)
<details>

</details>

## ☑️ Checklist


yinwm (changes requested):
Thanks for this substantial refactoring — the InboundContext abstraction and structured session scope are well-designed improvements.

However, I'm requesting changes for two blocking issues:

---

**1. `ResolveSessionKey` performance regression (blocking)**

`pkg/memory/jsonl.go:228-283`

This function is called on **every** session operation (`AddMessage`, `GetHistory`, `SetSummary`, `TruncateHistory`, `Save`), but it always performs a full `os.ReadDir` + reads ALL `.meta.json` files **twice**. The docstring says "first checks direct key existence, then scans metadata aliases on miss" but the implementation does the opposite — `sessionExists()` is called last as a fallback.

With N sessions, every message read/write triggers 2N+ file reads. At 100+ sessions this will cause noticeable latency.

Minimum fixes:
- Short-circuit with `sessionExists(sessionKey)` before `os.ReadDir`
- Merge the two directory passes into one
- Tolerate individual meta file read/decode failures (skip + log) instead of failing the entire resolution

---

**2. Missing `bindings` config migration path (blocking)**

The `bindings` field is removed from the config schema with no migration. Users with existing `bindings` configurations will **silently fall back to the default agent** after upgrading — their routing rules simply stop working with no warning.

Please add at minimum a deprecation warning when a `bindings` field is detected in the config file, and ideally an automatic migration from `bindings` → `dispatch.rules`.

---

Additional non-blocking suggestions (can address in follow-ups):
- `agentForSession` (`steering.go:307-341`) iterates every agent and calls `GetSessionScope` which internally does a full directory scan — consider centralizing scope lookup
- `PublishInbound/PublishOutbound` validate `isZero()` before calling `Normalize*` — consider normalizing first, then validating
- `normalizeKind` (`inbound_context.go:67-74`) is a no-op — all switch branches return the same value

yinwm:
## Code Review

Thanks for this comprehensive refactoring — unified `InboundContext`, structured session scope, opaque keys with alias migration, and ordered dispatch rules are solid improvements over the previous bindings model. Legacy compatibility path (`legacy_bindings.go`, alias promotion) is well thought out.

A few issues that should be addressed before merging:

---

### 🔴 Blocking

**1. `requeueInboundMessage` publishes to the wrong bus channel** (`loop.go:1703-1713`)

`drainBusToSteering` calls `requeueInboundMessage` to requeue non-steering inbound messages, but it calls `PublishOutbound` instead of `PublishInbound`. This means non-steering messages are:
- **Never processed as inbound** (silently lost)
- **Sent to channel adapters as bot responses** (garbage content in chats)

This affects any scenario with concurrent messages across different scopes.

**2. Merge conflict** — `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY`. Needs rebase onto latest main.

**3. `EnsureSessionMetadata` alias history promotion has a TOCTOU race** (`jsonl_backend.go:62-140`)

The multi-step read-then-write sequence isn't covered by a single lock. Between steps 2→4, another goroutine's `AddMessage` can insert new data, and then `SetHistory` at step 4 overwrites it — permanent data loss.

**4. `NormalizeOutboundMessage` unconditionally overwrites `ReplyToMessageID`** (`outbound_context.go:37`)

`msg.ReplyToMessageID = msg.Context.ReplyToMessageID` has no empty-check guard, unlike `Channel`/`ChatID` which use `if msg.X == ""`. If a caller sets a `ReplyToMessageID` but `Context.ReplyToMessageID` is empty, the reply target is silently zeroed out.

---

### 🟡 Suggestions

**5. `resolveSessionKey` silently falls back to original key on error** (`jsonl_backend.go:38-52`)

Transient I/O errors cause the system to operate on non-canonical keys — duplicate sessions or fragmented history. Consider surfacing the error on first resolution failure.

**6. `ChatType` hardcoded to `"direct"` in `normalizeProcessOptions`** (`dispatch_request.go:96-98`)

When constructing `InboundContext` from legacy fields, `ChatType` is set to `"direct"` if both `Channel` and `ChatID` are non-empty. Callers passing a group `ChatID` without `ChatType` will get `chat=direct:<group_id>` instead of `chat=group:<group_id>`.

**7. Dispatch rules first-match-wins semantics are undocumented**

Legacy bindings were sorted by specificity (direct > guild > team > account > wildcard). Migrated rules follow array order — overlapping rules (e.g., channel + sender) behave differently depending on config order. Please document this.

---

### 📝 Minor

- `resolveLinkedPeerID` (`session/key.go`) and `resolveLinkedDispatchID` (`routing/route.go`) are near-identical — consider extracting a shared utility.
- Please fill in the PR description — this is a significant architectural change that deserves motivation, migration strategy, and test coverage notes.

yinwm:
## Follow-up Review

Great work on the fixes! All three blocking issues from the previous review have been addressed:

- ✅ `requeueInboundMessage` now uses `PublishInbound`
- ✅ Merge conflict resolved
- ✅ `PromoteAliasHistory` with `lockSessionPair` — excellent atomic design

A few remaining items (all non-blocking):

---

### 1. `NormalizeOutboundMessage` — `ReplyToMessageID` unconditional overwrite (`outbound_context.go:37`)

```go
msg.ReplyToMessageID = msg.Context.ReplyToMessageID  // no empty-check guard
```

`Channel` and `ChatID` have `if msg.X == ""` guards, but `ReplyToMessageID` doesn't. If a caller sets `ReplyToMessageID = "67890"` but `Context.ReplyToMessageID` is empty, the reply target is silently zeroed out.

Fix: `if msg.ReplyToMessageID == "" { msg.ReplyToMessageID = msg.Context.ReplyToMessageID }`

### 2. `ChatType` hardcoded to `"direct"` in two places

- `dispatch_request.go:96-98` — legacy field → InboundContext construction
- `steering.go:304` — `continueWithSteeringMessages`

Both unconditionally set `ChatType = "direct"` when `Channel` and `ChatID` are non-empty. If the message comes from a group, this produces `chat=direct:<group_id>` instead of `chat=group:<group_id>`, routing to the wrong session partition.

### 3. `promoteAliasHistoryLocked` write ordering (`memory/jsonl.go:390-394`)

Currently writes meta first, then JSONL. If `writeMeta` succeeds but `rewriteJSONL` fails, the canonical session enters an unrecoverable inconsistent state: meta says `Count = N` but the JSONL file is empty. `sessionHasVisibleContentLocked` only checks meta, so subsequent promote attempts are blocked.

The project's own `SetHistory` (see its comment) documents the correct pattern: write meta first (crash-safe because old file is intact), then rewrite JSONL. But for promotion, the canonical JSONL file doesn't exist yet — so a failed rewrite leaves nothing. Consider swapping the order or adding a rollback.

### 4. Dead fallback promote path (`session/jsonl_backend.go:98-150`)

The fallback in `EnsureSessionMetadata` (when store doesn't implement `aliasPromotingStore`) still has the original TOCTOU race. Since `JSONLStore` now implements `aliasPromotingStore`, this path is dead code. Consider removing it or marking as deprecated to prevent future store implementations from hitting the race.

yinwm (approved):
All 9 original issues resolved across 4 review rounds. The latest commit is a main branch merge (channel config refactor) that does not affect the PR's core changes.

Note: The main merge introduces a separate issue (frontend sends "channels" but backend expects "channel_list") — this is a main branch regression, not introduced by this PR.

PR #2251, opened 2026-04-01
feat(channels): add Grafana Alertmanager webhook channel

loafoe:
## Summary
- Add a new `grafana_alertmanager` input-only channel that exposes a webhook endpoint for receiving alerts from Grafana Alertmanager
- Parse Alertmanager webhook payloads and format them into readable messages
- Support triggering specific skills when alerts are received via the `skill` config option
- Add `forced_skills` metadata key support in the agent loop for channel-based skill triggering

## Features
- Webhook endpoint at `/webhook/grafana-alertmanager` (configurable via `webhook_path`)
- HMAC-SHA256 signature verification for security (via `secret` config)
- Parses Grafana Alertmanager v4/v5 webhook payload format
- Formats alert data into readable messages with:
  - Alert status and title
  - Labels and annotations (summary, description)
  - Value strings
  - Dashboard/Panel/Silence URLs
- Supports forced skill triggering via config

## Configuration
```json
{
  "channels": {
    "grafana_alertmanager": {
      "enabled": true,
      "webhook_path": "/webhook/grafana-alertmanager",
      "secret": "your-webhook-secret",
      "allow_from": ["*"],
      "skill": "alert-handler",
      "chat_id": "alerts"
    }
  }
}
```

## Test plan
- [ ] Verify channel compiles with project build tags
- [ ] Test webhook endpoint receives POST requests correctly
- [ ] Test signature verification with valid and invalid signatures
- [ ] Test alert payload parsing and message formatting
- [ ] Test skill triggering via metadata

### DIFF
--- a/pkg/channels/grafana_alertmanager/grafana_alertmanager.go
+++ b/pkg/channels/grafana_alertmanager/grafana_alertmanager.go
@@ -1,0 +1,312 @@
+package grafana_alertmanager
+
+import (
+	"context"
+	"crypto/hmac"
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+	"time"
+
+	"github.com/sipeed/picoclaw/pkg/bus"
+	"github.com/sipeed/picoclaw/pkg/channels"
+	"github.com/sipeed/picoclaw/pkg/config"
+	"github.com/sipeed/picoclaw/pkg/logger"
+)
+
+const (
+	// maxWebhookBodySize limits request body to prevent memory exhaustion.
+	maxWebhookBodySize = 1 << 20 // 1 MiB
+)
+
+// Alert represents a single alert from Grafana Alertmanager.
+type Alert struct {
+	Status       string            `json:"status"`
+	Labels       map[string]string `json:"labels"`
+	Annotations  map[string]string `json:"annotations"`
+	StartsAt     time.Time         `json:"startsAt"`
+	EndsAt       time.Time         `json:"endsAt"`
+	GeneratorURL string            `json:"generatorURL"`
+	Fingerprint  string            `json:"fingerprint"`
+	SilenceURL   string            `json:"silenceURL,omitempty"`
+	DashboardURL string            `json:"dashboardURL,omitempty"`
+	PanelURL     string            `json:"panelURL,omitempty"`
+	Values       map[string]any    `json:"values,omitempty"`
+	ValueString  string            `json:"valueString,omitempty"`
+}
+
+// WebhookPayload represents the payload sent by Grafana Alertmanager.
+type WebhookPayload struct {
+	Receiver          string            `json:"receiver"`
+	Status            string            `json:"status"`
+	Alerts            []Alert           `json:"alerts"`
+	GroupLabels       map[string]string `json:"groupLabels"`
+	CommonLabels      map[string]string `json:"commonLabels"`
+	CommonAnnotations map[string]string `json:"commonAnnotations"`
+	ExternalURL       string            `json:"externalURL"`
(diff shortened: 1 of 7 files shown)


yinwm (changes requested):
Thanks for the PR! The Grafana Alertmanager integration direction is solid, but I found several critical issues that need to be fixed before merging.

**Blocking:**

1. **Channel can't start**: `initChannels` in `manager.go` has no `GrafanaAlertmanager` initialization block. The channel will never be created even with `enabled: true` — it's completely non-functional. Please add the init logic similar to other channels.

2. **`manager_channel.go` missing branches**: Both `hiddenValues` and `updateKeys` need a `case "grafana_alertmanager"` entry. Without this, hot-reload will break the config and secrets may be mishandled.

3. **Wrong signature header name**: The code reads `X-Grafana-Alertmanager-Signature` but Grafana sends `X-Grafana-Alerting-Signature` (note: "Alerting" not "Alertmanager"). If `secret` is configured, ALL valid webhook requests get 403'd. If `secret` is not configured, verification is skipped entirely — the security feature is ineffective either way.

4. **`Send()` returns error causing retry storm**: `Send()` returns `fmt.Errorf("grafana_alertmanager is an input-only channel")` which triggers Manager's retry logic 3x with exponential backoff on every agent reply. Return `nil, nil` or `nil, channels.ErrNotRunning` instead to skip retries.

5. **`forced_skills` metadata injection risk**: The new `metadataKeyForcedSkills` handling in `loop.go` can be triggered by ANY channel, not just grafana_alertmanager. Any channel that passes through metadata with this key can force arbitrary skill execution. This needs either a trust boundary check (only allow from internal webhook channels) or at minimum a validation that skill names exist in the registered skill list.

6. **Linter fails**: 4 formatting issues (gci + golines) need fixing before merge.

**Non-blocking suggestions:**
- Add unit tests — especially for HMAC verification, payload parsing, and message formatting
- Remove redundant `bus` field (already in embedded BaseChannel)
- Implement HealthChecker interface (mentioned in PR description but not in code)
- Clarify `AllowFrom` semantics — currently filters sender ID, not source IP as users might expect
- Consider async processing in `ServeHTTP` (return 200 first, process in goroutine) to avoid blocking HTTP connection
- Sort map keys in `formatAlertMessage` for deterministic output
- Unify `MetadataKeyForcedSkills` constant into `bus` package (currently duplicated across packages)

Please address these and ping for re-review!

loafoe:
Thanks for the thorough review @yinwm! I've addressed all the blocking issues and Copilot's feedback in the latest commit:

**Blocking issues:**
1. ✅ Channel init block added in `manager.go:initChannels`
2. ✅ Added `case "grafana_alertmanager"` to `hiddenValues` and `updateKeys` in `manager_channel.go`
3. ✅ Fixed signature header to `X-Grafana-Alerting-Signature`
4. ✅ `Send()` now returns `nil, nil` to avoid retry storm
5. ✅ Removed the `forced_skills` feature entirely - agreed it was too invasive for proper trust boundaries
6. ✅ Linter issues fixed

**Additional improvements:**
- Improved HMAC verification to hex decode and compare bytes (constant-time)
- Added validation: require `secret` when `allow_from` is restrictive (since `allow_from` can't meaningfully restrict webhook callers)
- Added comprehensive unit tests for the webhook handler (405, 413, 403, 400, 200 responses, signature verification, message formatting)
- Removed redundant `bus` field and `Skill` config option

I'll do some testing and refinement before requesting re-review.

imguoguo:
@loafoe Hi! Since there's been no activity after the earlier reminder, I'm closing this Grafana Alertmanager channel PR for now to keep the queue tidy. The work looked solid, so if you'd like to pick it back up, feel free to reopen anytime and we'll gladly continue the review.

PR #2259, opened 2026-04-01
fix(provider): use native Anthropic SDK for anthropic protocol with API key

deuxksy:
## Summary

- When using the `"anthropic"` protocol with an API key (non-OAuth), the factory incorrectly created an `HTTPProvider` (OpenAI-compatible) instead of the native Anthropic SDK provider (`ClaudeProvider`)
- This caused requests to be sent in OpenAI `/v1/chat/completions` format to the Anthropic API endpoint, resulting in failures
- Now uses `NewClaudeProviderWithBaseURL` which sends requests via the native Anthropic Messages API format

## Root Cause

In `CreateProviderFromConfig` (`pkg/providers/factory_provider.go`), the `case "anthropic":` branch with API key auth was calling `NewHTTPProviderWithMaxTokensFieldAndRequestTimeout` — the same OpenAI-compatible HTTP client used for `openai`, `groq`, `deepseek`, etc.

## Fix

Replace the HTTPProvider instantiation with `NewClaudeProviderWithBaseURL`, which wraps the native Anthropic Go SDK provider.

## Test Plan

- [x] `go build ./pkg/providers/...` passes
- [x] `go test ./pkg/providers/...` passes (all packages OK)
- [ ] Manual test: configure `anthropic/claude-sonnet-4.6` with API key and verify native Messages API format is used

### DIFF
--- a/pkg/providers/factory_provider.go
+++ b/pkg/providers/factory_provider.go
@@ -266,22 +266,11 @@ func CreateProviderFromConfig(cfg *config.ModelConfig) (LLMProvider, string, err
 			}
 			return provider, modelID, nil
 		}
-		// Use API key with HTTP API
-		apiBase := cfg.APIBase
-		if apiBase == "" {
-			apiBase = "https://api.anthropic.com/v1"
-		}
+		// Use API key with native Anthropic SDK (not OpenAI-compatible HTTPProvider)
 		if cfg.APIKey() == "" {
 			return nil, "", fmt.Errorf("api_key is required for anthropic protocol (model: %s)", cfg.Model)
 		}
-		return NewHTTPProviderWithMaxTokensFieldAndRequestTimeout(
-			cfg.APIKey(),
-			apiBase,
-			cfg.Proxy,
-			cfg.MaxTokensField,
-			cfg.RequestTimeout,
-			cfg.ExtraBody,
-		), modelID, nil
+		return NewClaudeProviderWithBaseURL(cfg.APIKey(), cfg.APIBase), modelID, nil
 
 	case "anthropic-messages":
 		// Anthropic Messages API with native format (HTTP-based, no SDK)


yinwm (changes requested):
Thanks for catching this bug! The direction is correct — `anthropic/` protocol should use the native SDK, not the OpenAI-compatible HTTPProvider.

However, I found issues that need to be addressed before merging:

## Critical: Authentication header mismatch

`NewProviderWithBaseURL` (at `pkg/providers/anthropic/provider.go:46-47`) uses `option.WithAuthToken(token)` which sends `Authorization: Bearer <key>`, but Anthropic API keys should be sent via `X-Api-Key` header (i.e., `option.WithAPIKey(token)`).

This function was originally designed for OAuth tokens where Bearer auth is correct. But this PR routes API keys through the same path, which will likely cause authentication failures for all API key users.

## Warning: Proxy, RequestTimeout, ExtraBody configs silently dropped

The old `NewHTTPProviderWithMaxTokensFieldAndRequestTimeout` accepted `cfg.Proxy`, `cfg.RequestTimeout`, `cfg.ExtraBody`, and `cfg.MaxTokensField`. The new `NewClaudeProviderWithBaseURL(cfg.APIKey(), cfg.APIBase)` doesn't accept any of these:

- **Proxy**: Users behind corporate proxies or in regions requiring proxy to access `api.anthropic.com` will break after this change.
- **RequestTimeout**: Long-running inference tasks may be interrupted by SDK default timeout. Note that the `anthropic-messages` protocol correctly passes this config.
- **ExtraBody**: Users relying on `extra_body` for custom parameters (e.g., beta feature flags) will lose them silently.

## Suggested fix

1. Add an API-key-specific constructor (e.g., `NewProviderWithAPIKey`) that uses `option.WithAPIKey()` instead of `option.WithAuthToken()`.
2. Pass `cfg.Proxy` via `option.WithHTTPClient()` with a configured proxy client.
3. Pass `cfg.RequestTimeout` via context timeout or SDK timeout options.
4. Either support `cfg.ExtraBody` or document that it's not supported with the native SDK path.

Report generated by `/pr-analyze`

JivinDotL:
I have tested this commit. but still failed.  There is no Authorization header in the request.

POST /v1/chat/completions HTTP/1.1\r
Host: 127.0.0.1:12345\r


17:33:34 INF eventbus pkg/agent/loop.go:962 > Agent event: error error="API request failed:\n  Status: 401\n  Body:   {\"error\": \"\\u7f3a\\u5c11 API Key\", \"request_id\": \"20260408173334272294\"}" agent_id=main event_kind=error iteration=1 session_key=agent:main:main source=runTurn stage=llm trace=turn.error turn_id=main-turn-1
17:33:34 ERR agent pkg/agent/loop.go:2178 > LLM call failed error="API request failed:\n  Status: 401\n  Body:   {\"error\": \"\\u7f3a\\u5c11 API Key\", \"request_id\": \"20260408173334272294\"}" agent_id=main iteration=1 model=Qwen3.5-397B-A17B-FP8
17:33:34 INF eventbus pkg/agent/loop.go:962 > Agent event: turn_end agent_id=main duration_ms=146 event_kind=turn_end final_len=0 iteration=1 iterations_total=1 session_key=agent:main:main source=runTurn status=error trace=turn.end turn_id=main-turn-1
Error: LLM call failed after retries: API request failed:
  Status: 401
  Body:   {"error": "\u7f3a\u5c11 API Key", "request_id": "20260408173334272294"}

PR #2244, opened 2026-04-01
feat(channels): add teams_webhook channel

loafoe:
## Summary
- Add a new Microsoft Teams webhook channel using `github.com/atc0005/go-teams-notify/v2`
- This is an **output-only** channel for sending notifications to Teams via Power Automate workflow webhooks
- **Multiple webhook targets** supported via map-based configuration (similar to MCP servers pattern)
- Target selection via `chat_id` at send time

## Configuration
```json
{
  "channels": {
    "teams_webhook": {
      "enabled": true,
      "webhooks": {
        "alerts": {
          "webhook_url": "https://prod-XX.westus.logic.azure.com/workflows/...",
          "title": "PicoClaw Alerts"
        },
        "logs": {
          "webhook_url": "https://prod-YY.westus.logic.azure.com/workflows/...",
          "title": "PicoClaw Logs"
        }
      }
    }
  }
}
```

## Usage
Agents can route messages to specific Teams webhooks using the `message` tool with `chat_id` as the target selector:

```json
{
  "tool": "message",
  "parameters": {
    "content": "Error: Database connection failed",
    "channel": "teams_webhook",
    "chat_id": "alerts"
  }
}
```

## Test plan
- [x] Unit tests pass (`go test ./pkg/channels/teams_webhook/...`)
- [x] Build succeeds with all tags (`go build -tags goolm,stdjson,bedrock ./...`)
- [x] Gateway starts with teams_webhook enabled
- [x] Unknown target returns helpful error with available targets
- [x] Verify message delivery to Teams webhook (manual test)

🤖 Generated with [Claude Code](https://claude.ai/code)

PR #2270, opened 2026-04-02
fix(config): handle non-addressable SecureString values in collectSensitive

loafoe:
## Summary
- Fix panic when `collectSensitive` iterates over map values containing `SecureString`
- Go reflection returns non-addressable values for map lookups, causing `v.Addr()` to panic
- Now checks `v.CanAddr()` and creates addressable copies when needed
- Adds regression tests to verify the fix

## Problem
When using channels with map-based configurations containing `SecureString` values (like `TeamsWebhookConfig.Webhooks`), the gateway would panic during sensitive data filtering:

```
panic: reflect.Value.Addr of unaddressable value
```

This happened because map values in Go are not addressable, but the code unconditionally called `v.Addr()` to invoke the `String()` method on `*SecureString`.

## Test plan
- [x] New regression tests for map values with SecureString
- [x] All existing config tests pass
- [x] Manual testing with teams_webhook channel

### DIFF
--- a/pkg/config/security.go
+++ b/pkg/config/security.go
@@ -250,6 +250,31 @@ func (sec *Config) collectSensitiveValues() []string {
 	return values
 }
 
+// secureStringValue extracts the string value from a SecureString reflect.Value.
+// Uses typed call when possible for better performance and type safety.
+func secureStringValue(v reflect.Value) string {
+	// Prefer typed call over reflective MethodByName for performance and type safety
+	if v.CanInterface() {
+		if ss, ok := v.Interface().(SecureString); ok {
+			return ss.String()
+		}
+	}
+	// Fallback to reflection for unexported/non-interfaceable values
+	var ptrVal reflect.Value
+	if v.CanAddr() {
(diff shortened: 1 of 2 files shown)


loafoe:
@yinwm this one is ready for review / merge, it fixes a crashing bug that was discovered during testing of #2244

yinwm:
Thanks for the fix! The implementation looks clean — extracting `secureStringValue` with the type-assertion fast path is a nice improvement over the original reflective-only approach. Tests cover the key scenarios well (map, nested map, empty map, nil map).

One blocking issue: there's a merge conflict with main. Please rebase and I'll approve.

Minor note (non-blocking): `initSensitiveCache` has a pre-existing race condition where the `sensitiveCache == nil` check is outside `sync.Once\ protection — worth tracking as a separate issue.

loafoe:
@yinwm resolved the merge conflict 👍🏻

loafoe:
@yinwm  can we merge this one?

loafoe:
@yinwm bumping this one, thx

Issue #1919, opened 2026-03-23
[Feature] Seahorse - Biologically-inspired Memory System for AI Agents


liuy:
# Seahorse - a biologically-inspired memory system for AI agents

> Named after the **hippocampus** (seahorse in Greek), the brain's critical structure for memory formation and consolidation.
> **Our Goal:** Give AI agents a human-like memory system — short-term for lossless context compaction, long-term for cross-session persistent knowledge.

---

## 0. The Memory Challenge

AI agents face four fundamental memory challenges:

### 0.1 Limited Context Space

LLMs have finite context windows (~200k tokens). You cannot stuff infinite conversation history into them.

**Seahorse Solution:** `short_compact()` + `short_assemble()` — intelligent compression that maximizes utility of limited space.

### 0.2 Context Compaction Loses Detail

Compression is inherently lossy. When you summarize 10 turns into 1 paragraph, details vanish.

**Seahorse Solution:** 
- LCM-style "lossless" compression — original messages are preserved
- `short_expand()` — trace back to original content when needed

### 0.3 Context Rot

Context degradation over time manifests in several ways:

| Rot Type | Description | Example |
|----------|-------------|---------|
| **Error Propagation** | Early mistakes get compacted into summaries → subsequent reasoning builds on wrong premises | Agent misremembers a config value → all future decisions are wrong |
| **Stale Information** | Outdated facts remain in context | "Peter is on vacation" → Peter returned 2 weeks ago |
| **Compaction Drift** | Each compression layer introduces errors → "telephone game" effect | Original: "Peter suggested indexing user_id" → Final: "Peter mentioned something about databases" |
| **Noise Pollution** | Irrelevant content dilutes useful signal | Long debugging session drowns out earlier architectural decisions |

**Seahorse Solution:**
| Rot Type | Solution |
|----------|----------|
| Error Propagation | `long_retain()` with confidence scores; `reflect()` detects contradictions |
| Stale Information | `long_forget()` access-based decay; high-salience facts override old ones |
| Compaction Drift | `short_expand()` traces to original; Facts are independent of Summaries |
| Noise Pollution | `short_compact()` filters by importance; `recall()` targets specific memories |

### 0.4 Lack of Long-Term Memory

Sessions end, memories vanish. No cross-session knowledge persistence.

**Seahorse Solution:** Complete `long_retain/recall/reflect/forget` system for persistent knowledge storage.

---

### 1.1 The Dual Memory Architecture

```
[...]

PR #2285, opened 2026-04-02
feat(seahorse): implement short-term memory engine (LCM)

liuy:
Implements short-term memory (LCM) for AI agents based on issue #1919. Uses SQLite with
DAG-based summary hierarchy for lossless context compression, FTS5 for full-text search,
and budget-aware assembly with fresh-tail protection.

### Context Management
- Message ingestion and automatic context compaction
- DAG-based summary tree (leaf → condensed hierarchy)
- Budget-aware context assembly with 32-message fresh tail protection
- Async condensed compaction to avoid blocking agent responses

### Tools for Agents
- **short_grep** - Full-text search across compressed context (FTS5 with bm25 ranking)
- **short_expand** - Expand summary to original messages

## Configuration

Enable seahorse via config:

```yaml
agents:
  defaults:
    context_manager: seahorse
```

Or via environment variable:

```bash
PICOCLAW_AGENTS_DEFAULTS_CONTEXT_MANAGER=seahorse
```

## Database

SQLite database location: `~/.picoclaw/workspace/{agent_id}/sessions/seahorse.db`

Each agent has its own database file containing:
- conversations
- messages (with parts for structured content)
- summaries (DAG hierarchy)
- context_items (ordinal gap numbering for efficient inserts)
- FTS5 virtual tables (summaries_fts, messages_fts) with trigram tokenizer

## Migration from Legacy

Switching from legacy context manager to seahorse is automatic. Seahorse bootstraps
by reading existing session history from JSONL files and syncing to SQLite. No manual
migration required.

## Technical Details

- FTS5 with trigram tokenizer for CJK support
- Per-session mutex sharding (256 shards)
- Budget fallback (50%) when MaxTokens >= budget
- Startup FTS5 compatibility check

### DIFF
--- a/pkg/agent/context_seahorse.go
+++ b/pkg/agent/context_seahorse.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"strings"
 
 	"github.com/sipeed/picoclaw/pkg/logger"
 	"github.com/sipeed/picoclaw/pkg/providers"
@@ -112,42 +111,13 @@ func (m *seahorseContextManager) Assemble(ctx context.Context, req *AssembleRequ
 		return nil, fmt.Errorf("seahorse assemble: %w", err)
 	}
 
-	// Convert to provider messages
 	history := seahorseToProviderMessages(result)
 
-	// Format ALL summaries as XML with metadata (depth, kind, etc.)
-	// This allows the LLM to understand the hierarchical structure
-	var summaryParts []string
-	for _, sum := range result.Summaries {
-		if sum.Content != "" {
-			// Get parent IDs for condensed summaries
-			parentIDs := getParentSummaryIDs(ctx, m.engine.GetRetrieval().Store(), sum.SummaryID)
-			xml := seahorse.FormatSummaryXML(&sum, parentIDs)
-			summaryParts = append(summaryParts, xml)
-		}
-	}
-	var summary string
-	if len(summaryParts) > 0 {
-		summary = strings.Join(summaryParts, "\n\n")
-	}
-
+	// Summary is already formatted as XML with system prompt addition by assembler
 	return &AssembleResponse{
 		History: history,
-		Summary: summary,
+		Summary: result.Summary,
 	}, nil
-}
-
-// getParentSummaryIDs retrieves parent summary IDs for a summary.
-func getParentSummaryIDs(ctx context.Context, store *seahorse.Store, summaryID string) []string {
-	parents, err := store.GetSummaryParents(ctx, summaryID)
-	if err != nil {
-		return nil
-	}
-	ids := make([]string, len(parents))
-	for i, p := range parents {
-		ids[i] = p.SummaryID
-	}
-	return ids
 }
 
 // Compact compresses conversation history via seahorse summarization.
@@ -212,7 +182,7 @@ func providerToSeahorseMessage(msg protocoltypes.Message) seahorse.Message {
 		Role:             msg.Role,
 		Content:          msg.Content,
 		ReasoningContent: msg.ReasoningContent,
-		TokenCount:       tokenizer.EstimateMessageTokens(providers.Message{Content: msg.Content}),
+		TokenCount:       tokenizer.EstimateMessageTokens(msg),
 	}
 
 	// Convert ToolCalls → MessageParts
@@ -249,8 +219,6 @@ func providerToSeahorseMessage(msg protocoltypes.Message) seahorse.Message {
 }
 
 // seahorseToProviderMessages converts a seahorse.AssembleResult to []providers.Message.
-// NOTE: Summaries are already included in Messages as XML-formatted messages by the assembler.
-// We do NOT convert Summaries separately to avoid double injection.
 func seahorseToProviderMessages(result *seahorse.AssembleResult) []protocoltypes.Message {
 	messages := make([]protocoltypes.Message, 0, len(result.Messages))
 
@@ -276,6 +244,9 @@ func seahorseToProviderMessages(result *seahorse.AssembleResult) []protocoltypes
 			}
 			if part.Type == "tool_result" {
 				pm.ToolCallID = part.ToolCallID
+				if pm.Content == "" && part.Text != "" {
+					pm.Content = part.Text
+				}
 			}
 			if part.Type == "media" && part.MediaURI != "" {
 				pm.Media = append(pm.Media, part.MediaURI)
--- a/pkg/agent/context_seahorse.go
+++ b/pkg/agent/context_seahorse.go
@@ -1,0 +1,295 @@
+package agent
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/sipeed/picoclaw/pkg/logger"
+	"github.com/sipeed/picoclaw/pkg/providers"
+	"github.com/sipeed/picoclaw/pkg/providers/protocoltypes"
+	"github.com/sipeed/picoclaw/pkg/seahorse"
+	"github.com/sipeed/picoclaw/pkg/session"
+	"github.com/sipeed/picoclaw/pkg/tokenizer"
+)
+
+// seahorseContextManager adapts seahorse.Engine to agent.ContextManager.
+type seahorseContextManager struct {
+	engine   *seahorse.Engine
+	sessions session.SessionStore // for startup bootstrap
+}
+
+// newSeahorseContextManager creates a seahorse-backed ContextManager.
+func newSeahorseContextManager(_ json.RawMessage, al *AgentLoop) (ContextManager, error) {
+	if al == nil {
+		return nil, fmt.Errorf("seahorse: AgentLoop is required")
+	}
+
+	// Resolve workspace for DB path
+	// DB stores session data, so it goes in sessions/ directory
+	agent := al.registry.GetDefaultAgent()
+	dbPath := agent.Workspace + "/sessions/seahorse.db"
+
+	// Create CompleteFn from provider
+	completeFn := providerToCompleteFn(agent.Provider, agent.Model)
+
+	// Create engine
+	engine, err := seahorse.NewEngine(seahorse.Config{
+		DBPath: dbPath,
+	}, completeFn)
+	if err != nil {
+		return nil, fmt.Errorf("seahorse: create engine: %w", err)
+	}
+
+	mgr := &seahorseContextManager{
+		engine:   engine,
+		sessions: agent.Sessions,
+	}
+
+	// Register seahorse tools with the agent's tool registry
+	retrieval := mgr.engine.GetRetrieval()
+	al.RegisterTool(seahorse.NewGrepTool(retrieval))
+	al.RegisterTool(seahorse.NewExpandTool(retrieval))
+
+	// Bootstrap all existing sessions at startup
+	if agent.Sessions != nil {
+		ctx := context.Background()
+		for _, sessionKey := range agent.Sessions.ListSessions() {
+			mgr.bootstrapSession(ctx, sessionKey)
+		}
+	}
+
+	return mgr, nil
+}
+
+// providerToCompleteFn wraps providers.LLMProvider as a seahorse.CompleteFn.
+func providerToCompleteFn(provider providers.LLMProvider, model string) seahorse.CompleteFn {
+	return func(ctx context.Context, prompt string, opts seahorse.CompleteOptions) (string, error) {
+		resp, err := provider.Chat(
+			ctx,
+			[]providers.Message{{Role: "user", Content: prompt}},
+			nil, // no tools for summarization
+			model,
+			map[string]any{
+				"max_tokens":       opts.MaxTokens,
+				"temperature":      opts.Temperature,
+				"prompt_cache_key": "seahorse",
+			},
+		)
+		if err != nil {
+			return "", err
+		}
+		return resp.Content, nil
+	}
+}
+
+// Assemble builds budget-aware context from seahorse SQLite.
+func (m *seahorseContextManager) Assemble(ctx context.Context, req *AssembleRequest) (*AssembleResponse, error) {
+	if req == nil {
+		return nil, fmt.Errorf("seahorse assemble: nil request")
+	}
+
+	budget := req.Budget
+	if budget <= 0 {
+		budget = 100000
+	}
+
+	// Reserve space for model response (spec lines 1400-1410)
+	effectiveBudget := budget - req.MaxTokens
+	if effectiveBudget <= 0 {
+		// MaxTokens >= budget is a configuration problem
+		// Use 50% as minimum to avoid guaranteed overflow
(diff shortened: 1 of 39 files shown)


pkg/seahorse/schema.go
```diff
@@ -0,0 +1,175 @@
...
+
+// runSchema creates or upgrades the database schema.
+// All migrations are idempotent (safe to run multiple times).
+func runSchema(db *sql.DB) error {
+	// Check FTS5 support before creating tables
+	if err := checkFTS5Support(db); err != nil {
+		return fmt.Errorf("FTS5 check: %w", err)
+	}
+
+	stmts := []string{
```
sky5454, line 36: SQL is more suitable as an external file, which is easy for the tool to manage and avoids being Imitated by others later

pkg/seahorse/short_assembler.go
```diff
@@ -0,0 +1,254 @@
...
+	return resolvedItem{
+		ordinal:    item.Ordinal,
+		itemType:   item.ItemType,
+		tokenCount: item.TokenCount,
+	}, nil
+}
+
+// FormatSummaryXML formats a summary as XML for LLM context.
+// This is exported so context managers can format summaries consistently.
+func FormatSummaryXML(s *Summary, parentIDs []string) string {
```
sky5454, line 232: XML should be processed using an official library https://golang.org/pkg/encoding/xml/

pkg/seahorse/tool_expand.go
```diff
@@ -0,0 +1,129 @@
...
+
+func NewExpandTool(engine *RetrievalEngine) *ExpandTool {
+	return &ExpandTool{engine: engine}
+}
+
+func (t *ExpandTool) Name() string {
+	return "short_expand"
+}
+
+func (t *ExpandTool) Description() string {
```
sky5454, line 24: Long text suggested with variables

pkg/tokenizer/estimator.go
```diff
@@ -0,0 +1,91 @@
...
+		totalChars += len(d.Function.Name) + len(d.Function.Description)
+
+		if d.Function.Parameters != nil {
+			if paramJSON, err := json.Marshal(d.Function.Parameters); err == nil {
+				totalChars += len(paramJSON)
+			}
+		}
+
+		// Per-tool overhead: type field, JSON structure, separators.
+		totalChars += 20
```
sky5454, line 87: Seems like a magic number, need to wrap it in variables

pkg/seahorse/tool_grep.go
```diff
@@ -0,0 +1,172 @@
...
+
+func (t *GrepTool) Execute(ctx context.Context, args map[string]any) *tools.ToolResult {
+	pattern, ok := args["pattern"].(string)
+	if !ok || pattern == "" {
+		return tools.ErrorResult("Missing required 'pattern' argument. Example: {\"pattern\": \"authentication\"}")
+	}
+
+	input := GrepInput{Pattern: pattern}
+
+	if scope, ok := args["scope"].(string); ok && scope != "" {
```
sky5454, line 122: A better way:
```
	scope, _ := args["scope"].(string)
	role, _  := args["role"].(string)
	last, _  := args["last"].(string)
	input := GrepInput{Pattern: pattern, Scope: scope, Role: role, Last: last}

```

pkg/seahorse/types_test.go
```diff
@@ -0,0 +1,54 @@
...
+		t.Errorf("expected SummaryKindLeaf = 'leaf', got %q", SummaryKindLeaf)
+	}
+	if SummaryKindCondensed != "condensed" {
+		t.Errorf("expected SummaryKindCondensed = 'condensed', got %q", SummaryKindCondensed)
+	}
+}
+
+func TestConstants(t *testing.T) {
+	// Ordinal gap step
+	if OrdinalStep != 100 {
```
sky5454, line 18: Confusing code

sky5454 (changes requested):
Request changes

sky5454 (changes requested):
Some code needs to be optimized

afjcjsbx:
Hi, I saw that steering messages are currently not ingested in Seahorse in loop.go, is this a design choice or an oversight?

pkg/agent/context_seahorse.go
```diff
@@ -0,0 +1,295 @@
...
+	if err := m.engine.Bootstrap(ctx, sessionKey, msgs); err != nil {
+		logger.WarnCF("seahorse", "bootstrap", map[string]any{
+			"session": sessionKey,
+			"error":   err.Error(),
+		})
+	}
+}
+
+// providerToSeahorseMessage converts a providers.Message to a seahorse.Message.
+func providerToSeahorseMessage(msg protocoltypes.Message) seahorse.Message {
```
afjcjsbx, line 181: `providerToSeahorseMessage` leaves text in Content, but as soon as Parts exist the ingest goes from `AddMessageWithParts` which saves empty messages.content. In assemble you rebuild tool metadata from Parts but never rebuild text, so `tool_results` and any assistant message with text plus tool calls go back to provider with empty Content.

pkg/agent/context_seahorse.go
```diff
@@ -0,0 +1,295 @@
...
+	result := seahorse.Message{
+		Role:             msg.Role,
+		Content:          msg.Content,
+		ReasoningContent: msg.ReasoningContent,
+		TokenCount:       tokenizer.EstimateMessageTokens(providers.Message{Content: msg.Content}),
```
afjcjsbx: `TokenCount` is estimated using Content only. This discards ReasoningContent, ToolCalls, ToolCallID, and Media, so Seahorse underestimates precisely the most expensive messages and assembly and compaction thresholds will be triggered too late.

pkg/agent/context_seahorse.go
```diff
@@ -0,0 +1,295 @@
...
+		Budget: effectiveBudget,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("seahorse assemble: %w", err)
+	}
+
+	// Convert to provider messages
+	history := seahorseToProviderMessages(result)
+
+	// Format ALL summaries as XML with metadata (depth, kind, etc.)
```
afjcjsbx: The adapter rebuilds all the summaries in `AssembleResponse.Summary`, but the assembler already has them in `result.Messages` as system messages. Then `BuildMessages` puts Summary back into the system prompt, so the same summary goes into the final prompt twice, consuming budget and duplicating instructions.

afjcjsbx:
remembering that pico is a lightweight assistant for resource-poor systems it might be useful to also include a mechanism that avoids indefinitely growing the size of the sqlite database, maybe settable by the user

afjcjsbx:
in any case, congrats for the PR, the idea seems truly brilliant!

pkg/agent/context_seahorse.go
liuy: sanitizeHistoryForProvider will delete summary msg with XML and it is leftover dead code while I fixed no summary issue in the system message. Anyway, good catch, thanks ;) Will fix it

pkg/agent/context_seahorse.go
liuy: Good Catch. Fixed

pkg/agent/context_seahorse.go
liuy: Good catch. Fixed

pkg/seahorse/schema.go
liuy: there is a config DBPath for users who need this

liuy:
> Hi, I saw that steering messages are currently not ingested in Seahorse in loop.go, is this a design choice or an oversight?

Good catch. My oversight. Will fix it, thank you

liuy:
> remembering that pico is a lightweight assistant for resource-poor systems it might be useful to also include a mechanism that avoids indefinitely growing the size of the sqlite database, maybe settable by the user

Yes, I think we should add TTL-based auto-cleanup for old messages/summaries or Configurable max DB size with oldest-first pruning. Both need to add configs though someone might need infinite persistent sessions.  You are very welcome to help add it once this is merged.

yinwm (changes requested):
Great work on this PR. The seahorse architecture is well-designed — clean separation between Engine/Assembler/Compaction/Retrieval/Store, solid test coverage, and good data roundtrip consistency across ToolCalls, Media, and Content paths. The FTS5 trigram support for CJK and the budget-aware assembly with fresh tail protection are particularly nice touches.

I found one blocking bug and several issues that should be addressed:

---

## Blocking Issues

### 1. `resequenceContextItemsTx` swallows scan errors (store.go:849-850)

```go
var scanErr error
if scanErr = rows.Scan(&i.ordinal, &i.itemType, &sid, &mid, &i.tokenCount); scanErr != nil {
    return err  // BUG: err is nil here (QueryContext succeeded), should be scanErr
}
```

When ordinal gap exhaustion triggers resequence and `rows.Scan` fails, the function returns `nil` error with incomplete data. Since resequence is on the critical path for compaction, this silently corrupts `context_items`.

**Fix**: `return scanErr` instead of `return err`.

### 2. `sql.NullString` for `INTEGER` column (store.go:847)

`message_id` is `INTEGER` in schema but scanned as `sql.NullString` then parsed via `strconv.ParseInt`. While modernc.org/sqlite currently tolerates this, it's not guaranteed behavior and is a type safety violation.

**Fix**: Use `sql.NullInt64` directly.

### 3. `compactCondensed` fallback may delete non-candidate items (short_compaction.go:388-405)

`selectShallowestCondensationCandidate` (fallback path) doesn't require ordinal continuity among candidates, but `ReplaceContextRangeWithSummary` deletes the entire ordinal range. Messages or other summaries between two candidate summaries would be silently deleted.

**Fix**: Before calling `ReplaceContextRangeWithSummary`, verify the range contains only candidate items, or switch to a per-item delete + insert approach.

---

## Non-blocking Suggestions

### 4. Compact Phase 2 budget calculation (short_compaction.go:63)

```go
budget = int(float64(tokensBefore) * ContextThreshold) // tokensBefore * 0.75
```

When `input.Budget` is nil, `budget = tokensBefore × 0.75`, which means `tokensBefore > budget` is always true. The async condensed goroutine is always triggered, creating unnecessary compaction overhead. Budget should be based on context window size, not current token count.

### 5. `Assemble` lazy initialization without mutex — data race (short_engine.go:310-312)

```go
if e.assembler == nil {
    e.assembler = &Assembler{store: e.store, config: e.config}
}
```

Multiple goroutines (agent loop + compact goroutine) can call `Assemble` concurrently. This check-then-create pattern is a data race (`go test -race` would flag it). Compare with `Compact` (line 327) which uses `initCompactionOnce()` with `sync.Once`.

Actual impact is low — `Assembler` is stateless (only store/config pointers), so duplicate creation won't corrupt data. But it should use `sync.Once` for consistency with the `Compact` pattern.

### 6. `appendContextItems` read-then-write not in transaction (store.go:710-740)

`GetMaxOrdinal` and subsequent INSERT are not in the same transaction. While `Ingest` has session mutex protection, the condensed goroutine uses an independent `shutdownCtx` and may concurrently operate on the same conversation.

### 7. `DeleteMessagesAfterID` not transactional (store.go:646-685)

Each msgID triggers 4 sequential DELETEs (context_items, message_parts, summary_messages, messages_fts) without a transaction. Partial failure leaves the database inconsistent. Also, the FTS manual delete is redundant — the trigger already handles it.

### 8. `selectShallowestCondensationCandidate` assumes continuous depths (short_compaction.go:447)

`for depth := 0; ; depth++` iterates depthGroups map assuming continuous keys. If depths are non-contiguous (e.g., 0, 1, 3 missing 2), depth=2 breaks and depth=3 is never selected. Should iterate sorted map keys instead.

### 9. File size (store.go — 1369 lines)

Consider splitting by functional domain: `conversation_ops.go`, `message_ops.go`, `summary_ops.go`, `context_ops.go`, `search_ops.go`.

### 10. `GetAllSessionStatuses` is O(N×M) (store.go:112-137)

Each session calls `GetSessionStatus` (4 SQL queries). For large databases, consider a JOIN-based batch query.

### 11. `GetMessages` N+1 query (store.go:286-292)

Each message triggers a separate Parts query. Consider batch loading by message_id range.

---

Requesting changes for items 1-3. The rest are suggestions for a follow-up PR.

liuy:
All review issues fixed:

**Blocking (1-3)**: ✅ Fixed in 615c617a, a2f72f45
**Non-blocking (4-8)**: ✅ Fixed in 971b992e, 9ace4567, 668ce6bc

**Optimizations (9-11)**: Deferred to future

All tests pass, lint clean.

liuy:
@yinwm 3 of the 4 follow-up issues are fixed, 1 does not need a fix:

### ✅ Fixed

1. **formatMessagesForSummary / truncateSummary ignores Parts** — `AddMessageWithParts` now derives readable content from Parts via `partsToReadableContent` and stores it in the Content field. `formatMessagesForSummary` and `truncateSummary` also fall back to Parts when Content is empty. 09bd5e6b

2. **FTS5/LIKE search cannot find tool_use/tool_result messages** — Same root cause as above. Since `AddMessageWithParts` now populates Content with readable text derived from Parts, the FTS5 trigger and LIKE queries work correctly. 09bd5e6b

3. **CompactUntilUnder has no iteration cap** — Added `MaxCompactIterations = 20`. If exceeded, a warning is logged and the function returns gracefully. Rationale: each iteration reduces ~4x tokens via leaf (8:1) or condensed (4:1) compaction; with a 200k context window and 75% threshold, 20 iterations covers any realistic scenario. 09bd5e6b

### ⏭ Not needed

**ExpandMessages silently skips missing messages** — No fix needed:
- Message IDs come from `short_grep` results, so they always exist in the DB
- Even in the extreme case where a message gets deleted between grep and expand, the caller (LLM agent) can detect fewer messages returned than requested by simple count comparison
- Adding a `MissingIDs` field would be redundant — the caller can diff input IDs vs returned message IDs

yinwm (approved):
All review issues verified fixed in f40e989. The `partsToReadableContent` approach cleanly resolves the Content/Parts roundtrip gaps across search, summary, and truncate paths. `CompactUntilUnder` iteration cap also in place. LGTM.

yinwm (approved):
All review issues verified fixed in f40e989. The `partsToReadableContent` approach cleanly resolves the Content/Parts roundtrip gaps across search, summary, and truncate paths. `CompactUntilUnder` iteration cap also in place with good test coverage. LGTM.

yinwm (approved):
All review issues verified fixed in f40e989. The partsToReadableContent approach cleanly resolves the Content/Parts roundtrip gaps across search, summary, and truncate paths. CompactUntilUnder iteration cap also in place with good test coverage. LGTM.

yinwm (approved):
All review issues verified fixed in f40e989. The partsToReadableContent approach cleanly resolves the Content/Parts roundtrip gaps across search, summary, and truncate paths. CompactUntilUnder iteration cap also in place with good test coverage. LGTM.

easyteacher:
Does the feature require many tokens? By setting context_manager I get errors like "max token exceed 97280".

PR #2317, opened 2026-04-03
feat(launcher): replace token-in-logs auth with standard HTTP login flow, and Enhanced security for login

sky5454:
## Problem

Previously users had to find the one-time token from console logs or log files to access the dashboard - a non-standard, error-prone workflow with no clear path for changing credentials.

### Auth flow (new)
1. First run: browser opens, session guard detects uninitialized state, redirects to /launcher-setup
2. User sets a password (min 8 chars) via POST /api/auth/setup {password, confirm}, bcrypt(cost=12) hash stored in ~/.picoclaw/launcher-auth.db (SQLite)
3. Subsequent logins: POST /api/auth/login {password}, HttpOnly cookie picoclaw_launcher_auth (HMAC-SHA256 signed, 7-day expiry)
4. 401 on any API call, frontend redirects to /launcher-login
5. Logout: POST /api/auth/logout, cookie cleared, redirect to login

### Backend changes
- web/backend/api/auth.go: renamed Token to Password; added handleSetup; launcherAuthStatusResponse now includes Initialized bool; PasswordStore interface wires bcrypt store into handlers
- web/backend/dashboardauth/: new package - Store with New(dir) / Open(path); SetPassword (bcrypt cost=12), VerifyPassword, IsInitialized
  - sql.go: all DB-layer constants (DBFilename, sqliteDriver, bcryptCost, four SQL query strings) - compile-time constants, zero runtime overhead
- web/backend/middleware/launcher_dashboard_auth.go: /launcher-setup and /api/auth/setup added to public paths
- web/backend/main.go:
  - dashboardauth.New(picoHome) replaces manual path construction
  - maskSecret(): suffix only revealed when >=5 chars hidden (length >= 12), preventing 8-char minimum passwords from leaking their tail
- web/backend/main_test.go: TestMaskSecret updated with boundary cases

### Forward-compatibility: pkg/credential integration

If the dashboard password is later reused as the enc:// passphrase, the bcrypt hash in launcher-auth.db becomes an offline oracle. Recommended mitigation (not yet implemented): derive two independent subkeys via HKDF before use:

  bcrypt(HKDF(password, info="picoclaw-dashboard-login-v1"))  stored in DB
  HKDF(password, info="picoclaw-credential-enc-v1")           passed to PassphraseProvider

This isolates the two domains: cracking the bcrypt hash yields only the login subkey, which is computationally independent of the enc:// subkey.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

- **Reference URL:**
- **Reasoning:**

## 🧪 Test Environment
- **Hardware:** 
- **OS:** 
- **Model/Provider:** 
- **Channels:** 

## 📸 Evidence (Optional)
<details>

</details>

## ☑️ Checklist

### DIFF
--- a/web/backend/i18n.go
+++ b/web/backend/i18n.go
@@ -24,8 +24,6 @@ const (
 	AppTooltip         TranslationKey = "AppTooltip"
 	MenuOpen           TranslationKey = "MenuOpen"
 	MenuOpenTooltip    TranslationKey = "MenuOpenTooltip"
-	MenuCopyToken      TranslationKey = "MenuCopyToken"
-	MenuCopyTokenHint  TranslationKey = "MenuCopyTokenHint"
 	MenuAbout          TranslationKey = "MenuAbout"
 	MenuAboutTooltip   TranslationKey = "MenuAboutTooltip"
 	MenuVersion        TranslationKey = "MenuVersion"
@@ -49,8 +47,6 @@ var translations = map[Language]map[TranslationKey]string{
 		AppTooltip:         "%s - Web Console",
 		MenuOpen:           "Open Console",
 		MenuOpenTooltip:    "Open PicoClaw console in browser",
-		MenuCopyToken:      "Copy dashboard token",
-		MenuCopyTokenHint:  "Copy the current web console access token to the clipboard",
 		MenuAbout:          "About",
 		MenuAboutTooltip:   "About PicoClaw",
 		MenuVersion:        "Version: %s",
@@ -68,8 +64,6 @@ var translations = map[Language]map[TranslationKey]string{
 		AppTooltip:         "%s - Web Console",
 		MenuOpen:           "打开控制台",
 		MenuOpenTooltip:    "在浏览器中打开 PicoClaw 控制台",
-		MenuCopyToken:      "复制控制台口令",
-		MenuCopyTokenHint:  "将当前 Web 控制台访问口令复制到剪贴板",
 		MenuAbout:          "关于",
 		MenuAboutTooltip:   "关于 PicoClaw",
 		MenuVersion:        "版本: %s",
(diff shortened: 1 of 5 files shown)


sky5454:
#  About password store
 
 
 ## How bcrypt verification works

When a password is **set**, bcrypt generates a random salt, runs the password through a cost-based key derivation function (Blowfish cipher iterated 2^cost times), and stores the result as a single self-contained string:

```
$2a$12$<22-char salt><31-char hash>
```

The cost factor (`12` here) is embedded in the string itself.

When a password is **verified**, the process is:

1. Pull the stored hash string from the DB
2. Extract the salt and cost from it (no separate lookup needed — they're in the string)
3. Re-run the same KDF on the candidate password using that salt and cost
4. Compare the output directly against the stored hash

The key property: **you never decrypt the hash**. You re-derive and compare. If they match, the password is correct.

---

## This is exactly how Linux `/etc/shadow` works

Linux stores passwords in `/etc/shadow` in the same format:

```
$6$rounds=5000$<salt>$<hash>
```

- `$6$` = SHA-512 crypt (analogous to `$2a$` for bcrypt)
- The salt and iteration count are embedded in the stored string

When you run `login` or `sudo`, PAM calls `crypt(candidate, stored_hash)` — it reads the algorithm/salt/rounds straight out of the stored hash, re-derives, and compares. No decryption, no separate salt table.

---

## Key similarities

| Aspect | Linux shadow | picoclaw dashboardauth |
|---|---|---|
| Storage format | `$algo$cost$salt$hash` | `$2a$12$salt+hash` |
| Salt location | Embedded in hash string | Embedded in hash string |
| Verification | Re-derive → compare | `bcrypt.CompareHashAndPassword()` |
| No plaintext ever stored | ✓ | ✓ |
| Cost tunable | `rounds=N` | `bcryptCost = 12` |
| Single DB field needed | one column in shadow | one `bcrypt_hash` column |

The only practical difference is that bcrypt uses the Blowfish cipher internally (intentionally slow, memory-hard-ish), while `$6$` uses SHA-512 with a simpler PBKDF. Both defend against offline dictionary attacks by making each guess computationally expensive.


the example likes
image

sky5454:
the pr could fix those problem: https://github.com/sipeed/picoclaw/issues/2319

Issue #2319, opened 2026-04-04
[BUG] v0.2.5   WebSocket  connect fail


axwfae:
## Quick Summary
 WebSocket  connect fail 

##  Environment & Tools
- **PicoClaw Version:**   v.0.2.5
- **Go Version:** (e.g., go 1.22)
- **AI Model & Provider:** (e.g., GPT-4o via OpenAI / DeepSeek via SiliconFlow)
- **Operating System:** (e.g., Ubuntu 22.04 / macOS / Android Termux)
- **Channels:** (e.g., Discord, Telegram, Feishu, ...)

## 📸 Steps to Reproduce
1.  v0.2.4 WebSocket  connect ok 
2.  set  launcher_token  fix to  'clawtestoken"
3.  v0.2.5 WebSocket  connect fail 
4. 
## ❌ Actual Behavior
# PicoClaw Chrome Extension Connection Issue - Bug Report

## Environment
- PicoClaw Version: 0.2.5
- Chrome Extension Version: 1.0.56
- NAS Docker: Running on local network (192.168.1.200)
- Gateway Port: 18790
- Launcher/API Port: 18800

## Current Configuration
- Dashboard Token (launcher_token): clawtestoken
- Pico Token (from /api/pico/token): b9825e6b35e5a67444e795c2547a5110
- Gateway PID Token (from ~/.picoclaw/.picoclaw.pid): 5713189781afdcf865a26f273fbcf1fe

## Problem Description
The Chrome extension cannot connect to the PicoClaw WebSocket server (both via Gateway port 18790 and Launcher port 18800). All authentication attempts fail with various HTTP errors.

### Test 1: Direct to Gateway (18790) - Simple Token
```
GET /pico/ws
Sec-Websocket-Protocol: token.b9825e6b35e5a67444e795c2547a5110
```
Result: **401 Unauthorized**

### Test 2: Direct to Gateway (18790) - Composed Token
```
GET /pico/ws
Sec-Websocket-Protocol: token.pico-5713189781afdcf865a26f273fbcf1feb9825e6b35e5a67444e795c2547a5110
```
Result: **400 Bad Request**

### Test 3: Via Launcher (18800) - Simple Token
```
GET /pico/ws
Authorization: Bearer clawtestoken
Sec-Websocket-Protocol: token.b9825e6b35e5a67444e795c2547a5110
```
Result: **400 Bad Request**

### Test 4: Via Launcher (18800) - Composed Token
```
GET /pico/ws
[...]

PR #2339, opened 2026-04-04
feat(launcher): standard HTTP login/setup/logout flow for dashboard, frontend and backend impl. and fix windows pid lock for ws

sky5454:
## 📝 Description

This can effectively solve the problem of reducing the difficulty of user operation, and realize the standard web login process.

After initializing the boot, you will see the setup page, enter the password twice to confirm, the password will be used for the login page.
In most platforms and the vast majority of users can not or can not view the log use scenarios, will effectively solve the log problem

b5b7a90bcd81fa6b81ed281cb2b5b5ff
a8ffc914f7505c9930e4df8148ba47dd

## Problem

Previously users had to find the one-time token from console logs or log files to access the dashboard - a non-standard, error-prone workflow with no clear path for changing credentials.

### Auth flow (new)
1. First run: browser opens, session guard detects uninitialized state, redirects to /launcher-setup
2. User sets a password (min 8 chars) via POST /api/auth/setup {password, confirm}, bcrypt(cost=12) hash stored in ~/.picoclaw/launcher-auth.db (SQLite)
3. Subsequent logins: POST /api/auth/login {password}, HttpOnly cookie picoclaw_launcher_auth (HMAC-SHA256 signed, 7-day expiry)
4. 401 on any API call, frontend redirects to /launcher-login
5. Logout: POST /api/auth/logout, cookie cleared, redirect to login

### Backend changes
- web/backend/api/auth.go: renamed Token to Password; added handleSetup; launcherAuthStatusResponse now includes Initialized bool; PasswordStore interface wires bcrypt store into handlers
- web/backend/dashboardauth/: new package - Store with New(dir) / Open(path); SetPassword (bcrypt cost=12), VerifyPassword, IsInitialized
  - sql.go: all DB-layer constants (DBFilename, sqliteDriver, bcryptCost, four SQL query strings) - compile-time constants, zero runtime overhead
- web/backend/middleware/launcher_dashboard_auth.go: /launcher-setup and /api/auth/setup added to public paths
- web/backend/main.go:
  - dashboardauth.New(picoHome) replaces manual path construction
  - maskSecret(): suffix only revealed when >=5 chars hidden (length >= 12), preventing 8-char minimum passwords from leaking their tail
- web/backend/main_test.go: TestMaskSecret updated with boundary cases

### Forward-compatibility: pkg/credential integration

If the dashboard password is later reused as the enc:// passphrase, the bcrypt hash in launcher-auth.db becomes an offline oracle. Recommended mitigation (not yet implemented): derive two independent subkeys via HKDF before use:

  bcrypt(HKDF(password, info="picoclaw-dashboard-login-v1"))  stored in DB
  HKDF(password, info="picoclaw-credential-enc-v1")           passed to PassphraseProvider

This isolates the two domains: cracking the bcrypt hash yields only the login subkey, which is computationally independent of the enc:// subkey.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

- **Reference URL:**
- **Reasoning:**

## 🧪 Test Environment
- **Hardware:** 
- **OS:** 
- **Model/Provider:** 
- **Channels:** 

## 📸 Evidence (Optional)
<details>

</details>

## ☑️ Checklist

### DIFF
--- a/web/frontend/src/routes/__root.tsx
+++ b/web/frontend/src/routes/__root.tsx
@@ -1,6 +1,6 @@
 import { Outlet, createRootRoute, useRouterState } from "@tanstack/react-router"
 import { TanStackRouterDevtools } from "@tanstack/react-router-devtools"
-import { useEffect } from "react"
+import { useEffect, useState } from "react"
 
 import { getLauncherAuthStatus } from "@/api/launcher-auth"
 import { AppLayout } from "@/components/app-layout"
@@ -30,6 +30,8 @@ const RootLayout = () => {
       (m) => m.routeId === "/launcher-login" || m.routeId === "/launcher-setup",
     )
 
+  const [authError, setAuthError] = useState<string | null>(null)
+
   // Session guard: proactively check auth status on every page load.
   // This catches the case where ?token= auto-login bypassed the login/setup UI.
   useEffect(() => {
@@ -43,12 +45,16 @@ const RootLayout = () => {
         }
       })
       .catch((err: unknown) => {
-        // For real HTTP errors (e.g. 503 when the auth store is unavailable),
-        // redirect to login so the user can re-authenticate rather than being
-        // silently stranded on the dashboard.  Network failures (no response)
-        // are left alone — launcherFetch handles the 401 redirect on real calls.
-        if (err instanceof Error && /^status \d+$/.test(err.message)) {
+        // On 401/403, redirect to login — the session is invalid.
+        // On 5xx (e.g. 503 when the auth store is unavailable) or network errors,
+        // do NOT redirect: a subsequent successful login would loop straight back here.
+        // launcherFetch handles 401 on real API calls regardless.
+        if (err instanceof Error && /^status 40[13]$/.test(err.message)) {
           globalThis.location.assign("/launcher-login")
+        } else {
+          setAuthError(
+            err instanceof Error ? err.message : "Auth service unavailable, please try to delete the launcher-auth.db at picoclaw home directory and restart the application.",
+          )
         }
       })
   }, [isAuthPage])
@@ -70,10 +76,24 @@ const RootLayout = () => {
   }
 
   return (
-    <AppLayout>
-      <Outlet />
-      {import.meta.env.DEV ? <TanStackRouterDevtools /> : null}
-    </AppLayout>
+    <>
+      {authError && (
+        <div className="bg-destructive text-destructive-foreground fixed inset-x-0 top-0 z-[100] flex items-center justify-between px-4 py-2 text-sm shadow-md">
+          <span>Auth service error: {authError}</span>
+          <button
+            className="ml-4 opacity-70 hover:opacity-100"
+            onClick={() => setAuthError(null)}
+            aria-label="Dismiss"
+          >
+            ✕
+          </button>
+        </div>
+      )}
+      <AppLayout>
+        <Outlet />
+        {import.meta.env.DEV ? <TanStackRouterDevtools /> : null}
+      </AppLayout>
+    </>
   )
 }
 


sky5454:
#  About password store
 
 
 ## How bcrypt verification works

When a password is **set**, bcrypt generates a random salt, runs the password through a cost-based key derivation function (Blowfish cipher iterated 2^cost times), and stores the result as a single self-contained string:

```
$2a$12$<22-char salt><31-char hash>
```

The cost factor (`12` here) is embedded in the string itself.

When a password is **verified**, the process is:

1. Pull the stored hash string from the DB
2. Extract the salt and cost from it (no separate lookup needed — they're in the string)
3. Re-run the same KDF on the candidate password using that salt and cost
4. Compare the output directly against the stored hash

The key property: **you never decrypt the hash**. You re-derive and compare. If they match, the password is correct.

---

## This is exactly how Linux `/etc/shadow` works

Linux stores passwords in `/etc/shadow` in the same format:

```
$6$rounds=5000$<salt>$<hash>
```

- `$6$` = SHA-512 crypt (analogous to `$2a$` for bcrypt)
- The salt and iteration count are embedded in the stored string

When you run `login` or `sudo`, PAM calls `crypt(candidate, stored_hash)` — it reads the algorithm/salt/rounds straight out of the stored hash, re-derives, and compares. No decryption, no separate salt table.

---

## Key similarities

| Aspect | Linux shadow | picoclaw dashboardauth |
|---|---|---|
| Storage format | `$algo$cost$salt$hash` | `$2a$12$salt+hash` |
| Salt location | Embedded in hash string | Embedded in hash string |
| Verification | Re-derive → compare | `bcrypt.CompareHashAndPassword()` |
| No plaintext ever stored | ✓ | ✓ |
| Cost tunable | `rounds=N` | `bcryptCost = 12` |
| Single DB field needed | one column in shadow | one `bcrypt_hash` column |

The only practical difference is that bcrypt uses the Blowfish cipher internally (intentionally slow, memory-hard-ish), while `$6$` uses SHA-512 with a simpler PBKDF. Both defend against offline dictionary attacks by making each guess computationally expensive.


the example likes
image

sky5454:
the pr could fix those problem: https://github.com/sipeed/picoclaw/issues/2319

sky5454:
The term of the previous two prs:
 - https://github.com/sipeed/picoclaw/pull/2318
 - https://github.com/sipeed/picoclaw/pull/2317

pkg/pid/pidfile_windows.go
sky5454, line 1: Windows's pid handle is no same as UNIX, so fix it by this way.

nglmercer:
* **The Issue:** The code was checking the `err` return from `syscall.LazyProc.Call()`. In Go, this error is always non-nil (constructed from `GetLastError`), even if the call succeeds. This caused the system to falsely believe the process wasn't running, leading to the deletion of the PID file and subsequent **503 errors**.

* **The Fix:** The code now ignores the error return and strictly validates the primary return values (`handle != 0` or `ret != 0`) as per Win32 semantics.

That seems right

web/backend/api/auth.go
```diff
@@ -141,3 +174,66 @@ func (h *launcherAuthHandlers) handleStatus(w http.ResponseWriter, r *http.Reque
 	}
 	_, _ = w.Write(enc)
 }
+
+// handleSetup sets or changes the dashboard password.
+//
+// Rules:
+//   - If the store has no password yet, the endpoint is open (no session required).
+//   - If a password is already set, the caller must hold a valid session cookie.
+func (h *launcherAuthHandlers) handleSetup(w http.ResponseWriter, r *http.Request) {
```
wj-xiao, line 218: **Blocking**: there is a lockout path when the password store cannot be opened.
If `dashboardauth.New()` fails, `PasswordStore` is nil, `/api/auth/status` reports `initialized=false`, the frontend redirects to `/launcher-setup`, but `/api/auth/setup` returns `501 password store not configured`.

Result: users cannot complete setup and cannot recover via login UI.
Please add a safe fallback behavior (e.g., keep token login flow visible when store is unavailable, or fail startup hard with a clear fatal error).

web/backend/api/auth.go
```diff
@@ -141,3 +174,66 @@ func (h *launcherAuthHandlers) handleStatus(w http.ResponseWriter, r *http.Reque
...
+	w.Header().Set("Content-Type", "application/json")
+
+	if h.store == nil {
+		w.WriteHeader(http.StatusNotImplemented)
+		_, _ = w.Write([]byte(`{"error":"password store not configured"}`))
+		return
+	}
+
+	initialized := h.isStoreInitialized(r.Context())
```
wj-xiao: `isStoreInitialized()` currently converts store errors into `false (err == nil && ok)`.
This makes runtime DB errors look like “not initialized”, which can unintentionally relax auth behavior (fallback path / setup gating).
Please switch to fail-closed handling: surface/store the error and return 5xx for auth status/setup decisions instead of silently treating errors as uninitialized.

wj-xiao:
This PR combines multiple concerns in one change set, which makes review and rollback harder.
Suggest splitting into smaller PRs:

1. Windows PID fix
2. Backend auth core and frontend auth flow
3. Tray/token cleanup

This would reduce risk and make each PR independently verifiable.

sky5454:
windows now is broken, for a broken thing's that's  no risk.
If the fixed key and tray-related functions are not removed, it may interfere with web authorization and cause unexpected strange problems.

web/backend/api/auth.go
sky5454: todo: An error message will be added to ask users to delete the password database manually.

web/backend/api/auth.go
sky5454: todo: Optimize and use explicit errors

wj-xiao:
Thanks a lot for the contribution and for the detailed work on this PR.

Just a quick note: the Windows PID fix in `pkg/pid/pidfile_windows.go` is already being handled in PR #2267 , which we’re merging separately.
Could you please rebase and remove that duplicated part from this PR, so we can focus review here on the launcher auth/setup changes only?

Really appreciate your effort on improving the auth flow.

wj-xiao:
When the password store is unavailable at startup, backend behavior is inconsistent across auth endpoints:

1. `POST /api/auth/login` explicitly allows fallback to token auth when storeErr != nil and can return success.
> Reference: auth.go (line 120), auth.go (line 122)

2. `GET /api/auth/status` does not use the same fallback and returns 503 for the same store error.
> Reference: auth.go (line 194), auth.go (line 196)

3. Frontend treats non-OK status as error and redirects to login from root guard, so users can be bounced back to `/launcher-login` even after successful login.
> Reference: launcher-auth.ts (line 28), __root.tsx (line 50)

This is blocking because it can produce an authentication loop and prevent normal dashboard access in a recoverable failure mode.

Most practical fix: make `/api/auth/status` follow the same fallback policy as `/api/auth/login` when storeErr != nil (treat as legacy-token mode and return `initialized=false` instead of 503), so frontend routing remains consistent and login success is stable.

sky5454:
> When the password store is unavailable at startup, backend behavior is inconsistent across auth endpoints:
> 
> 1. `POST /api/auth/login` explicitly allows fallback to token auth when storeErr != nil and can return success.
> 
> > Reference: auth.go (line 120), auth.go (line 122)
> 
> 2. `GET /api/auth/status` does not use the same fallback and returns 503 for the same store error.
> 
> > Reference: auth.go (line 194), auth.go (line 196)
> 
> 3. Frontend treats non-OK status as error and redirects to login from root guard, so users can be bounced back to `/launcher-login` even after successful login.
> 
> > Reference: launcher-auth.ts (line 28), __root.tsx (line 50)
> 
> This is blocking because it can produce an authentication loop and prevent normal dashboard access in a recoverable failure mode.
> 
> Most practical fix: make `/api/auth/status` follow the same fallback policy as `/api/auth/login` when storeErr != nil (treat as legacy-token mode and return `initialized=false` instead of 503), so frontend routing remains consistent and login success is stable.

5xx It is a definition of an internal server error, which is mainly used to report errors, rather than the normal processing process. When an error occurs, it indicates that the internal system is unstable at this time, which is a temporary or irrecoverable state.

wj-xiao:
> 5xx It is a definition of an internal server error, which is mainly used to report errors, rather than the normal processing process. When an error occurs, it indicates that the internal system is unstable at this time, which is a temporary or irrecoverable state.

I agree with your 5xx definition.

The problem is that in the same store-failure state, `/api/auth/login` may still return success (token fallback), while `/api/auth/status` returns 503. That creates a contradictory contract and causes the post-login redirect loop.

If we treat store failure as an internal unstable state, the behavior should be consistent across endpoints.

Most practical fix (aligned with your 5xx stance):
When password-store initialization/check fails, make `/api/auth/login` return 503 as well (no token fallback in that state), and let the frontend show a recovery error screen/message for 503 instead of redirecting to `/launcher-login`.

sky5454:
# https://github.com/sipeed/picoclaw/pull/2339/commits/13c9a2087a4483239207ecb6d166a4228d17b3a4 Focus: Redirect Loop When the Store is Broken

| Steps | Before PR Fix (Bug Existing) | Current Behavior |
|---|---|---|
| Dashboard Loading → `GET /api/auth/status` | Returns 503 | Returns 503 ✓ |
| Session Guard Catch | Redirects `/launcher-login` | **Displays banner, no redirect** ✓ |
| User logs in on the login page using startup token | Returns 200 (token fallback) → Jumps to dashboard | Same as above |
| Dashboard session guard triggers again | Another 503 → Jumps to login page again → Infinite loop 💥 | Banner reappears, no redirect ✓ |

**Redirect loop issue fixed.**

PR #2353, opened 2026-04-05
feat(membench): add LOCOMO memory benchmark for retrieval evaluation

liuy:
Part of #1919

## Summary

Add `cmd/membench/` — a LOCOMO-based benchmark comparing legacy session store vs seahorse short memory retrieval quality. Runs end-to-end with `make mem`.

## Usage

```bash
make mem
```

This will:
1. Build the membench binary
2. Download the LOCOMO dataset (10 conversations, ~300 turns each) if not cached
3. Ingest all conversations into both legacy store and seahorse SQLite
4. Evaluate 1986 QA pairs across 5 categories
5. Print a comparison table

## Current Results (budget=4000 tokens)

```
--- No LLM generation ---
Mode       HitRate  F1       C1      C2      C3      C4      C5
---------------------------------------------------------------------
legacy     0.2196   0.0017   0.1660  0.2103  0.2063  0.2319  0.2399
seahorse   0.6997   0.0066   0.3998  0.7775  0.3489  0.7632  0.7892
```

## Interpreting Results

- **HitRate** (primary metric): Measures whether the evidence passages referenced by each question were successfully retrieved. Seahorse achieves 0.70 vs legacy 0.22 — a **3.2x improvement**, demonstrating that keyword-based FTS5/BM25 search with message expansion significantly outperforms naive budget truncation of raw history.

- **F1** (supplementary): Token overlap F1 between the full retrieved context and the gold answer. Both modes score near zero because we compare raw multi-turn message text against concise gold answers without an LLM to extract/summarize. This metric is only meaningful with LLM-generated answers.

- **Per-category (C1-C5)**: Per-category HitRate breakdown. Seahorse shows strong retrieval on multi-hop (C2: 0.78), temporal (C4: 0.76), and adversarial (C5: 0.79) questions. Single-hop (C1: 0.40) and open-ended (C3: 0.35) are relatively weaker, suggesting room for improvement in keyword extraction for these question types.

## Why No LLM Generation

The official LOCOMO evaluation pipeline uses an LLM to generate answers from retrieved context, then scores those answers against gold. We skip this step because it requires ~4000 sequential LLM API calls (one per QA pair), taking hours to complete. For now we focus on **retrieval quality** (HitRate) which can be measured without an LLM. LLM-based answer generation and scoring will be added in a follow-up.

## Limitations

1. **Token estimation**: Uses `len(text)/4`, a rough approximation for English text only.
2. **Keyword extraction**: Simple stopword removal without n-gram support. Multi-word entities (e.g. "New York") are split into individual keywords.
3. **Evidence matching**: Substring match between turn text and retrieved content. Minor formatting differences could cause false negatives.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -349,6 +349,25 @@ build-macos-app:build-launcher
 	@./scripts/build-macos-app.sh $(PLATFORM)-$(ARCH)
 	@echo "macOS .app bundle created: $(BUILD_DIR)/PicoClaw.app"
 
+## mem: Build membench, download LOCOMO data (if needed), run benchmark, and show results
+mem:
+	@echo "Building membench..."
+	@mkdir -p $(BUILD_DIR)
+	@$(GO) build -o $(BUILD_DIR)/membench ./cmd/membench
+	@echo "Build complete: $(BUILD_DIR)/membench"
+	@if [ ! -f $(BUILD_DIR)/memdata/locomo10.json ]; then \
+		echo "Downloading LOCOMO dataset..."; \
+		mkdir -p $(BUILD_DIR)/memdata; \
+		curl -sfL "https://raw.githubusercontent.com/snap-research/locomo/main/data/locomo10.json" \
+			-o $(BUILD_DIR)/memdata/locomo10.json && [ -s $(BUILD_DIR)/memdata/locomo10.json ] || { echo "Error: LOCOMO download failed"; exit 1; }; \
+		echo "Download complete"; \
+	else \
+		echo "LOCOMO dataset already exists, skipping download"; \
+	fi
+	@echo "Running benchmark..."
+	@rm -rf $(BUILD_DIR)/memout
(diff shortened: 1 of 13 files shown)


yinwm (approved):
LGTM! Clean benchmark tool with good test coverage and clear separation of concerns.

A few minor suggestions for follow-up (non-blocking):

1. **Resource cleanup**: `ir.Engine.Close()` is never called in `runEval`/`runAll`. Consider adding `defer ir.Engine.Close()` after `IngestSeahorse` in the seahorse case of `runEval` (the test in `ingest_test.go:36` already does this correctly).

2. **Invalid mode validation**: `modesFromFlag` silently accepts invalid values like `--mode foobar` and returns `["foobar"]`, which then matches no switch case and silently does nothing. Consider validating the mode value early.

3. **Silent sample skipping**: When a sample's conversation ID is missing from `ConvMap`, it's silently skipped without generating an EvalResult. Consider logging a summary of skipped samples or returning a zero-filled result.

PR #2364, opened 2026-04-05
fix: avoid restoring stale sessions with dangling tool calls

changeworldBT:
## Summary
- drop dangling tool-call tails when restoring session history
- avoid restoring stale runtime state after restart
- prevent Telegram sessions from getting stuck in typing/steering mode

## Problem
If a session is persisted while an assistant tool-call turn is still unfinished, restarting picoclaw can restore that broken tail as valid history. New Telegram messages may then get redirected into the steering queue instead of starting a fresh turn, leaving the bot stuck in typing without a final reply.

## Fix
Sanitize recovered session history before it is reused by the agent. If the history ends with an assistant message containing tool calls that do not have a complete set of matching tool results, trim that dangling suffix and keep only the safe, completed history.

## Notes
This keeps completed tool-call sequences intact while preventing stale unfinished execution state from surviving process restart.

### DIFF
--- a/pkg/session/manager.go
+++ b/pkg/session/manager.go
@@ -97,7 +97,7 @@ func (sm *SessionManager) GetHistory(key string) []providers.Message {
 
 	history := make([]providers.Message, len(session.Messages))
 	copy(history, session.Messages)
-	return history
+	return sanitizeRecoveredHistory(history)
 }
 
 func (sm *SessionManager) GetSummary(key string) string {
--- a/pkg/session/recovery.go
+++ b/pkg/session/recovery.go
@@ -1,0 +1,52 @@
+package session
+
+import "github.com/sipeed/picoclaw/pkg/providers"
+
+// sanitizeRecoveredHistory drops any dangling tail that starts with an
+// assistant tool-call message whose tool results were never fully persisted.
+//
+// This prevents a restarted agent from restoring an unfinished runtime state
+// (assistant tool calls plus later steering/user messages) as if it were valid
+// history. We keep completed tool-call sequences intact and only trim the
+// incomplete suffix.
+func sanitizeRecoveredHistory(history []providers.Message) []providers.Message {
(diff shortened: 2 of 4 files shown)


pkg/session/recovery.go
```diff
@@ -0,0 +1,52 @@
...
+import "github.com/sipeed/picoclaw/pkg/providers"
+
+// sanitizeRecoveredHistory drops any dangling tail that starts with an
+// assistant tool-call message whose tool results were never fully persisted.
+//
+// This prevents a restarted agent from restoring an unfinished runtime state
+// (assistant tool calls plus later steering/user messages) as if it were valid
+// history. We keep completed tool-call sequences intact and only trim the
+// incomplete suffix.
+func sanitizeRecoveredHistory(history []providers.Message) []providers.Message {
```
afjcjsbx, line 12: `sanitizeRecoveredHistory` not only removes the incomplete assistant/tool block: with `return history[:i]` it truncates the entire suffix starting with the assistant with `ToolCalls`. This means that any user/steering message persisted after that point is lost altogether on reboot. It is also in contrast to the sanitizer already used by the `ContextBuilder`, which discards the broken tool-call round but preserves subsequent non-tool messages. Here we risk permanent loss of real input instead of limiting the fix to only incomplete runtime state.

pkg/session/manager.go
afjcjsbx, line 89: The helper is called inside `SessionManager.GetHistory`, so it doesn't just apply to history retrieved from disk after a restart: it changes the meaning of each session read in memory. This also breaks the stated `SessionStore` contract (GetHistory returns the full message history) and causes runtime callers to see a filtered view instead of the actual persisted log. If recovery is the problem, sanitization should be applied in the session bootstrap/reopen or materialized once, not hidden behind each `GetHistory`.

afjcjsbx (changes requested):

PR #2406, opened 2026-04-07
Fix tzdata not installed issue in termux env

A1311981684:
## 📝 Description

In termux env, tzdata is not installed or can be installed by pkg -i command. In the latest version 0.2.5, TZ info is acquired via env param, but since tzdata is not installed in the system, picoclaw fails to work with time correctly. The cron feature is baddly affected by this issue.

## 🗣️ Type of Change
- [√ ] 🐞 Bug fix (non-breaking change which fixes an issue)
- [ ] 📖 Documentation update

## 🤖 AI Code Generation
- [ √] 👨‍💻 Mostly Human-written (Human lead, AI assisted or none)

## 🔗 Related Issue

None.

- **Reference URL:** none
- **Reasoning:** none

## 🧪 Test Environment
- **Hardware:** 
- **OS:** 
- **Model/Provider:** 
- **Channels:** 

## 📸 Evidence (Optional)
<details>
No need to provide.
</details>

## ☑️ Checklist
- [ √] My code/docs follow the style of this project.
- [ √] I have performed a self-review of my own changes.
- [ ×] I have updated the documentation accordingly.

### DIFF
--- a/cmd/picoclaw/main.go
+++ b/cmd/picoclaw/main.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"os"
 	"time"
+	_ "time/tzdata"
 
 	"github.com/spf13/cobra"
 


cmd/picoclaw/main.go
```diff
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"os"
 	"time"
+	_ "time/tzdata"
```
sky5454, line 13: https://go.dev/src/time/tzdata/tzdata.go
According to the annotations in the official library
```go
// Importing this package will increase the size of a program by about 450 KB.
```

For the Android architecture, use Bionic instead of glibc.

Another problem with finding time zones is more likely to be the TZ variable, which is not available on Android


Tips: You could see the `/system/usr/share/zoneinfo/tzdata` at Android termux


And I recommend you to try the GUI at Android: https://github.com/sipeed/picoclaw_fui/releases

sky5454 (changes requested):

cmd/picoclaw/main.go
sky5454: Tips: You could see the `/system/usr/share/zoneinfo/tzdata` at Android termux

sky5454 (changes requested):
TZDATA, It exists on Android, but isn't recognized

sky5454:
I found a method that might help you, and it's the standard Android variable read solution
https://github.com/golang/go/issues/20455#issuecomment-342287698



image

PR #2410, opened 2026-04-07
feat(tool): add browser automation via Chrome DevTools Protocol (CDP)

Yourdaylight:
## Summary

Implement a **BrowserTool** that provides browser automation capabilities through direct CDP (Chrome DevTools Protocol) WebSocket connection. This addresses **Issue #293** (Browser Automation) from the [project roadmap](https://github.com/sipeed/picoclaw/blob/main/ROADMAP.md#-4-advanced-capabilities-from-chatbot-to-agentic-ai).

**Key design choices:**
- **Go-native CDP client** using `gorilla/websocket` (already in go.mod — zero new dependencies)
- **Optional via `//go:build cdp`** — default builds have zero browser code; `go build -tags cdp` enables full support
- **Borrowed [N]-indexed DOM state** from [opencli](https://github.com/jackwener/opencli) for LLM-friendly structured output at zero vision token cost
- **13 stealth anti-detection patches** to avoid website bot detection

### Features

| Action | Description |
|--------|-------------|
| `navigate` | Open URL, wait for load |
| `state` | Extract [N]-indexed interactive DOM elements (core action) |
| `click [N]` | Click element by index |
| `type [N] "text"` | Type into element |
| `fill [N] "text"` | Clear + type into element |
| `select [N] "option"` | Select dropdown option |
| `screenshot` | Capture PNG, deliver via MediaStore to user |
| `get_text [N]` | Extract element text/value |
| `scroll up/down` | Scroll page |
| `keys <key>` | Keyboard input (Enter, Tab, etc.) |
| `evaluate <js>` | Run JavaScript (opt-in via config) |
| `close` | Close browser session |

### Security
- SSRF protection via `net.ParseIP` — blocks private/loopback/link-local/metadata IPs
- JS `evaluate` disabled by default (`allow_evaluate: false`)
- Concurrent-safe lazy CDP connection (`sync.Mutex`)
- Screenshot temp files cleaned up properly
- Safe JS string embedding via `json.Marshal` (not `fmt.Sprintf %q`)

### Browsing History
Tracks visited pages and includes compact history summary in `state` output:
```
Browsing history (3 pages):
  1. Example Domain — https://example.com/
  2. Sign in to GitHub — https://github.com/login
> 3. Hacker News — https://news.ycombinator.com/

Current page: Hacker News
URL: https://news.ycombinator.com/

Interactive elements (25):
[0] a "Hacker News"
[1] a "new"
...
```

### Files

**New (7 files, +2230 lines):**
- `pkg/tools/browser.go` — BrowserTool core (Tool interface, 12 actions)
- `pkg/tools/browser_cdp.go` — CDP WebSocket client
- `pkg/tools/browser_cdp_util.go` — Chrome path auto-detection (macOS/Linux/Windows)
- `pkg/tools/browser_stealth.go` — 13 anti-detection JS patches
- `pkg/tools/browser_stub.go` — Stub for non-CDP builds
- `pkg/tools/browser_test.go` — Unit tests (8 tests)
- `pkg/tools/browser_integration_test.go` — Integration tests (9 tests)

**Modified (2 files):**
- `pkg/config/config.go` — `BrowserToolConfig` struct + `IsToolEnabled("browser")`
- `pkg/agent/instance.go` — Browser tool registration

## Test plan

- [x] `go build` (default) compiles without browser code
- [x] `go build -tags cdp` compiles with full browser support
- [x] `go vet` passes for both build tag variants
- [x] Unit tests pass (8/8): Chrome detection, URL validation, stealth JS, params
- [x] Integration tests pass (9/9): navigate, state, screenshot, click, type, evaluate, stealth, scroll, SSRF blocking, temp file cleanup
- [x] E2E verified against Chrome 146 headless: example.com, GitHub login page, Hacker News
- [x] GitHub login form correctly detected: username input, password input, submit button
- [x] Stealth: `navigator.webdriver === false` confirmed
- [x] SSRF: localhost, 127.0.0.1, 10.x, 192.168.x, 172.16.x, [::1], 169.254.169.254 all blocked

Ref: #293

### DIFF
--- a/pkg/agent/instance.go
+++ b/pkg/agent/instance.go
@@ -100,6 +100,16 @@ func NewAgentInstance(
 		}
 	}
 
+	if cfg.Tools.IsToolEnabled("browser") {
+		browserTool, err := tools.NewBrowserTool(cfg.Tools.Browser)
+		if err != nil {
+			logger.WarnCF("agent", "Browser tool unavailable (optional)",
+				map[string]any{"error": err.Error()})
+		} else {
+			toolsRegistry.Register(browserTool)
+		}
+	}
+
 	if cfg.Tools.IsToolEnabled("edit_file") {
 		toolsRegistry.Register(tools.NewEditFileTool(workspace, restrict, allowWritePaths))
 	}
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -901,7 +901,7 @@ type SearchCacheConfig struct {
 type BrowserToolConfig struct {
 	ToolConfig  `envPrefix:"PICOCLAW_TOOLS_BROWSER_"`
 	CDPEndpoint string `json:"cdp_endpoint"    yaml:"cdp_endpoint,omitempty"  env:"PICOCLAW_TOOLS_BROWSER_CDP_ENDPOINT"`
-	Timeout     int    `json:"timeout_seconds"  yaml:"timeout_seconds,omitempty" env:"PICOCLAW_TOOLS_BROWSER_TIMEOUT"`
+	Timeout     int    `json:"timeout_seconds"  yaml:"timeout_seconds,omitempty" env:"PICOCLAW_TOOLS_BROWSER_TIMEOUT_SECONDS"`
 	Stealth     bool   `json:"stealth"          yaml:"stealth,omitempty"       env:"PICOCLAW_TOOLS_BROWSER_STEALTH"`
 	AllowEval   bool   `json:"allow_evaluate"   yaml:"allow_evaluate,omitempty" env:"PICOCLAW_TOOLS_BROWSER_ALLOW_EVALUATE"`
 }
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -875,6 +875,7 @@ type ToolsConfig struct {
 	Subagent        ToolConfig         `json:"subagent"          yaml:"-"                                                       envPrefix:"PICOCLAW_TOOLS_SUBAGENT_"`
 	WebFetch        ToolConfig         `json:"web_fetch"         yaml:"-"                                                       envPrefix:"PICOCLAW_TOOLS_WEB_FETCH_"`
 	WriteFile       ToolConfig         `json:"write_file"        yaml:"-"                                                       envPrefix:"PICOCLAW_TOOLS_WRITE_FILE_"`
+	Browser         BrowserToolConfig  `json:"browser"           yaml:"browser,omitempty"`
 }
 
 // IsFilterSensitiveDataEnabled returns true if sensitive data filtering is enabled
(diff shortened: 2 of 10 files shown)


yinwm (changes requested):
Thanks for this solid contribution! The CDP build tag approach, stealth patches, and zero-new-dependency design are well thought out. However, I found several issues that need to be addressed before merging.

## Critical Issues

### 1. Screenshot file deleted after MediaStore.Store (`browser.go:575`)

`MediaStore.Store` records the file path without copying it — it only creates a mapping (`s.refs[ref] = mediaEntry{path: localPath, ...}`). The subsequent `os.Remove(tmpFile)` makes the media reference unresolvable: any later `Resolve(ref)` returns a path that no longer exists, so users never receive the screenshot. Meanwhile `ResponseHandled: true` tells the framework the media was delivered.

**Fix:** Remove the `os.Remove(tmpFile)` call. Let MediaStore manage the file lifecycle via its `CleanupPolicy`. Or copy the file to a MediaStore-managed directory before removing the temp.

### 2. SSRF protection has multiple bypass vectors (`validateBrowserURL`)

Three bypass vectors exist:

- **Non-standard IP formats:** `http://0/`, `http://127.1/`, `http://0x7f.0.0.1/`, `http://2130706433/` — `net.ParseIP()` doesn't recognize these (returns nil), so they fall through to DNS resolution where behavior is unpredictable.
- **DNS rebinding (TOCTOU):** `validateBrowserURL` resolves DNS at check time, but Chrome navigates later. An attacker's DNS can return a public IP during validation and `127.0.0.1` during actual connection. `WebFetchTool` mitigates this with dial-level re-checking via `newSafeDialContext`; BrowserTool can't intercept Chrome's connections due to CDP architecture.
- **Missing address ranges:** `100.64.0.0/10` (CGNAT) and `198.18.0.0/15` (benchmark) return `false` from `IsPrivate()` and are not blocked. These ranges may host internal services in cloud environments.

**Fix:** Reject numeric-only hostnames. Document the DNS rebinding limitation as inherent to the CDP approach. Consider adding explicit CIDR checks for the missing ranges.

### 3. Concurrent access to mutable state (`browser.go`)

`history`, `cdp`, and `mediaStore` are read/written without synchronization:
- `executeClose()` reads/writes `t.cdp` and `t.history` without locks
- `recordVisit()` and `updateCurrentTitle()` append to `t.history` without locks
- `SetMediaStore()` writes `t.mediaStore` without locks

`cdpMu` only protects `connectIfNeeded()` — not the rest of the mutable state. While the agent loop typically serializes tool calls, there's no framework guarantee (e.g., subagent scenarios).

**Fix:** Introduce `stateMu sync.Mutex` to guard `cdp`, `history`, and `mediaStore` access.

## Other Issues to Fix

4. **`scrollIntoView({behavior: 'instant'})` throws TypeError in Chrome 125+** (`browser.go:364, 414, 459`) — Change to `'auto'`. This affects click, type, and fill actions. Your test plan mentions Chrome 146, which would trigger this.

5. **`json.Unmarshal` errors silently ignored** in `executeType`, `executeFill`, `executeSelect`, `executeGetText` — If CDP returns unexpected data, these handlers may report success incorrectly. Handle errors and return `ErrorResult`.

6. **Merge conflict** — `mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`. Please rebase onto latest `main`.

## Non-blocking Suggestions

- Split `browser.go` (889 lines exceeds the 600-line limit)
- Add element count limit in `executeState` JS (prevent huge output on large SPAs)
- Fix screenshot integration test assertion (expects "Screenshot captured" but gets "Screenshot saved to..." without MediaStore)
- Align `PICOCLAW_TOOLS_BROWSER_TIMEOUT` env var to `*_TIMEOUT_SECONDS` pattern
- Add compile + test CI for both default and `cdp` build tags

PR #2411, opened 2026-04-07
fix(provider): handle split SSE stream chunk parsing

imalasong:
Preserve partial stream payloads only when JSON is incomplete so chunk boundaries no longer drop content, while malformed payloads are discarded to avoid poisoning subsequent valid chunks. Add regression tests for split-payload reassembly and malformed-chunk recovery.

Made-with: Cursor

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

- **Reference URL:**
- **Reasoning:**

## 🧪 Test Environment
- **Hardware:** 
- **OS:** 
- **Model/Provider:** 
- **Channels:** 

## 📸 Evidence (Optional)
<details>

</details>

## ☑️ Checklist

### DIFF
--- a/pkg/providers/openai_compat/provider.go
+++ b/pkg/providers/openai_compat/provider.go
@@ -316,6 +316,7 @@ func parseStreamResponse(
 
 	scanner := bufio.NewScanner(reader)
 	scanner.Buffer(make([]byte, 0, 1024*1024), 10*1024*1024) // 1MB initial, 10MB max
+	var pendingData string
 	for scanner.Scan() {
 		// Check for context cancellation between chunks
 		if err := ctx.Err(); err != nil {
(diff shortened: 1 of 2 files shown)


afjcjsbx:
Generally good fix, weak point: recognizing "truncated JSON" via `strings.Contains(err.Error()`, "unexpected end of JSON input") is fragile (depends on error text); To have more robustness, better to rely on type `*json.SyntaxError / offset`, or at least include unexpected EOF as well.
**Edge case**: if `pendingData` is valued but the next line is a valid non-continuous chunk, today concatenate → parse fails → also discard the valid chunk. **Tip**: when `pendingData != ""` and concatenate parse fails with non-"truncated" error, try doing `json.Unmarshal(data)` "by itself" before dropping.
**Safety**: `pendingData` can grow without limit (sum of multiple rows) beyond max scanner (10MB); I would put a cap and reset over threshold to avoid memory blow-up.

PR #2430, opened 2026-04-08
feat(web): show disabled reasons in tooltips when buttons are disabled

SiYue-ZO:
## 📝 Description

Show specific disabled reasons when buttons or input fields are disabled in the WebUI, improving user experience.

e43c2f8f383d999134be67b28a95a81e

### Changes Made:
- Add disabled reason tooltips for model card actions (set default, delete)
- Add disabled reason tooltips for marketplace skill card install button
- Add disabled reason display for chat input when disabled
- Add internationalization support for all disabled reasons (en/zh)

### Detailed Changes:
- **Model card:** Show specific reasons when set-default or delete buttons are disabled
  - Set-default button: Shows "Setting as default...", "Cannot set unavailable model as default", "Already the default model", "Cannot set virtual model as default"
  - Delete button: Shows "Cannot delete the default model"
- **Marketplace skill card:** Show specific reasons when install button is disabled
  - Install button: Shows "Installing...", "Already installed", "Cannot install: related tool is not enabled"
- **Chat composer:** Show reason text below input when input is disabled
  - Shows "Gateway is not running. Start it to chat." or "No default model configured. Go to Models page to set one."

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 📚 Technical Context
- **Reasoning:** Use button's `title` attribute to show disabled reasons, add hint text below the chat input when disabled

## 🧪 Test Environment
- **Hardware:** PC
- **OS:** Linux
- **Channels:** Web

## ☑️ Checklist
- [x] My code follows the style of this project
- [x] I have performed a self-review of my own changes
- [x] Build passes successfully
- [ ] I have updated the documentation accordingly

### DIFF
--- a/web/frontend/src/components/agent/hub/market-skill-card.tsx
+++ b/web/frontend/src/components/agent/hub/market-skill-card.tsx
@@ -36,6 +36,13 @@ export function MarketSkillCard({
 }) {
   const { t } = useTranslation()
 
+  const installDisabledReason = (() => {
+    if (installPending) return t("pages.agent.skills.marketplace_installDisabled.installing")
+    if (result.installed) return t("pages.agent.skills.marketplace_installDisabled.installed")
+    if (!canInstall) return t("pages.agent.skills.marketplace_installDisabled.cannotInstall")
+    return t("pages.agent.skills.marketplace_install_action")
+  })()
+
   return (
     <Card
       className="group border-border/40 bg-card/40 hover:border-border/80 hover:bg-card relative overflow-hidden transition-all hover:shadow-md"
@@ -92,6 +99,7 @@ export function MarketSkillCard({
               className="shadow-sm transition-all"
               disabled={!canInstall || result.installed || installPending}
               onClick={onInstall}
+              title={installDisabledReason}
             >
               {installPending ? (
                 <IconLoader2 className="size-4 animate-spin" />
--- a/web/frontend/src/components/agent/hub/market-skill-card.tsx
+++ b/web/frontend/src/components/agent/hub/market-skill-card.tsx
@@ -18,6 +18,11 @@ import {
   CardHeader,
   CardTitle,
 } from "@/components/ui/card"
+import {
+  Tooltip,
+  TooltipContent,
+  TooltipTrigger,
+} from "@/components/ui/tooltip"
 
 export function MarketSkillCard({
   result,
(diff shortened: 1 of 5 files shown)


web/frontend/src/components/agent/hub/market-skill-card.tsx
```diff
@@ -92,6 +99,7 @@ export function MarketSkillCard({
               className="shadow-sm transition-all"
               disabled={!canInstall || result.installed || installPending}
               onClick={onInstall}
+              title={installDisabledReason}
```
wj-xiao: `title` is being used to show the disabled reason, but our shared `Button` style includes `disabled:pointer-events-none` (see `components/ui/button.tsx`). Once disabled, the button does not receive hover/pointer events, so the tooltip does not appear. This means the PR’s main UX goal (“show disabled reason”) is not achieved here.
Suggestion: attach the tooltip trigger/title to a non-disabled wrapper (for example, a surrounding `span`) or use a tooltip component mounted on a non-disabled element.

web/frontend/src/components/models/model-card.tsx
```diff
@@ -110,7 +122,7 @@ export function ModelCard({
             size="icon-sm"
             onClick={() => onDelete(model)}
             disabled={model.is_default}
-            title={t("models.action.delete")}
+            title={deleteDisabledReason}
```
wj-xiao: Same issue as above for the model action buttons.

SiYue-ZO: Updated in 7fe727f as well. The model action buttons now use a non-disabled span as the TooltipTrigger wrapper, so the disabled reason still appears even when the inner Button disables pointer events.

web/frontend/src/components/agent/hub/market-skill-card.tsx
SiYue-ZO: Updated in 7fe727f. The disabled install button is now wrapped with a non-disabled span used as the tooltip trigger, so the disabled reason remains hoverable/focusable even though Button applies disabled:pointer-events-none.

SiYue-ZO:
@wj-xiao @yinwm Plz review again Thanks

web/frontend/src/components/models/model-card.tsx
```diff
@@ -81,19 +100,32 @@ export function ModelCard({
...
-            >
-              {settingDefault ? (
-                <IconLoader2 className="size-3.5 animate-spin" />
-              ) : (
-                <IconStar className="size-3.5" />
-              )}
-            </Button>
+            <Tooltip delayDuration={!canSetDefault || settingDefault ? 0 : 700}>
+              <TooltipTrigger asChild>
+                <span
```
wj-xiao, line 108: This introduces an accessibility regression for the icon-only model action controls.

After moving the tooltip trigger to the wrapping `span`, the actual buttons no longer have an accessible name because the previous `title` was removed.
In the enabled state, keyboard focus lands on an unlabeled `<button>`, and in the disabled state it lands on a generic focusable `<span>` with no button semantics.

Please preserve an accessible label on the real button (`aria-label` or `title`), and if the wrapper remains focusable for disabled tooltips, give it matching semantics (`role`, `aria-disabled`, and the same label) or switch to an aria-disabled pattern instead of relying on a native disabled button.

SiYue-ZO: @wj-xiao Thanks for the detailed feedback!
Could you please take another look when you have time? Thanks!

wj-xiao (approved):

PR #2442, opened 2026-04-09
refactor skills registries and add GitHub-backed skill discovery

lxowalle:
### Summary
This PR refactors skills registry configuration into a shared multi-registry model and adds GitHub as a first-class registry for skill discovery and installation.
It also unifies registry handling across the CLI, agent tools, and web API, while preserving compatibility with the legacy tools.skills.github.* configuration.
### What Changed
- Replaced the old fixed registry config shape with a generic registries model based on shared fields plus per-registry parameters
- Added GitHub registry support for:
  - searching for skills via GitHub code search
  - installing skills from owner/repo, repo subpaths, and full GitHub URLs
  - GitHub Enterprise base URL support
- Added shared registry provider/manager plumbing so CLI commands, tools, and the web API use the same registry construction and install flow
- Kept backward compatibility for legacy tools.skills.github.* settings by mapping them into the new registry model at runtime
- Updated ClawHub integration to use the same registry abstraction
- Improved installed skill origin metadata handling and normalization across install paths
- Improved staged skill installation behavior in the web API to avoid cross-device rename issues
- Fixed multiple GitHub URL parsing edge cases, including:
  - slash-containing branch names
  - tree and blob skill URLs
  - registry host and scheme validation
  - correct install directory resolution for blob .../SKILL.md URLs
- Updated examples, docs, and tests for the new registry configuration model
### Compatibility
- Existing legacy tools.skills.github.* config is still supported
- Existing ClawHub behavior is preserved through the new registry abstraction
- The new registries.github configuration is now the preferred approach
### Why This Matters
Before this change, skill registry behavior was split across multiple code paths and GitHub-based skill discovery/install support was limited. This PR makes registry behavior consistent across the product, expands GitHub install/search support, and creates a cleaner foundation for additional registry backends in the future.
### Testing
- Added and updated unit/integration tests for config loading, migration, registry behavior, URL parsing, install metadata, and API flows
- Verified with:
  - make check
  - make lint
 
## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

- **Reference URL:**
- **Reasoning:**

## 🧪 Test Environment
- **Hardware:** 
- **OS:** 
- **Model/Provider:** 
- **Channels:** 

## 📸 Evidence (Optional)
<details>

</details>

## ☑️ Checklist


pkg/config/config.go
```diff
@@ -925,25 +926,70 @@ type SearchCacheConfig struct {
...
-	DownloadPath    string       `json:"download_path"       yaml:"-"                    env:"PICOCLAW_SKILLS_REGISTRIES_CLAWHUB_DOWNLOAD_PATH"`
-	Timeout         int          `json:"timeout"             yaml:"-"                    env:"PICOCLAW_SKILLS_REGISTRIES_CLAWHUB_TIMEOUT"`
-	MaxZipSize      int          `json:"max_zip_size"        yaml:"-"                    env:"PICOCLAW_SKILLS_REGISTRIES_CLAWHUB_MAX_ZIP_SIZE"`
-	MaxResponseSize int          `json:"max_response_size"   yaml:"-"                    env:"PICOCLAW_SKILLS_REGISTRIES_CLAWHUB_MAX_RESPONSE_SIZE"`
+	BaseURL string       `json:"base_url,omitempty" yaml:"-"               env:"PICOCLAW_TOOLS_SKILLS_GITHUB_BASE_URL"`
+	Token   SecureString `json:"token,omitzero"     yaml:"token,omitempty" env:"PICOCLAW_TOOLS_SKILLS_GITHUB_TOKEN"`
+	Proxy   string       `json:"proxy,omitempty"    yaml:"-"               env:"PICOCLAW_TOOLS_SKILLS_GITHUB_PROXY"`
+}
+
+type SkillRegistryConfig struct {
```
wj-xiao, line 876: Environment-variable compatibility for skills registries is broken.

After `SkillsRegistriesConfig` was changed to a slice of `SkillRegistryConfig`, the registry fields are no longer env-bindable. `LoadConfig` still relies on `env.Parse(cfg)`, so variables such as `PICOCLAW_SKILLS_REGISTRIES_CLAWHUB_BASE_URL`, `..._ENABLED`, and `..._AUTH_TOKEN` no longer take effect. With those env vars set, `LoadConfig` still returned the default ClawHub base URL and an empty auth token. This is a regression for env-driven/container deployments.

pkg/skills/config_bridge.go
```diff
@@ -0,0 +1,91 @@
...
+	cfg config.SkillsToolsConfig,
+	registryCfg config.SkillRegistryConfig,
+) config.SkillRegistryConfig {
+	if registryCfg.Name != "github" {
+		return registryCfg
+	}
+	if registryCfg.Param == nil {
+		registryCfg.Param = map[string]any{}
+	}
+	if registryCfg.BaseURL == "" {
```
wj-xiao: Deprecated `tools.skills.github.base_url` compatibility is broken.

The compatibility layer only copies `cfg.Github.BaseURL` when `registryCfg.BaseURL` is empty, but the default config now always pre-populates `registries.github.base_url` with `https://github.com`(pkg/config/defaults.go line 448 ). As a result, a config that only sets the deprecated `tools.skills.github.base_url` can no longer override the default, so GitHub Enterprise setups silently fall back to github.com.

lxowalle: fixed

pkg/skills/installer.go
```diff
@@ -96,46 +214,135 @@ func parseGitHubRef(repo string) (GitHubRef, error) {
...
+		return "", fmt.Errorf("failed to parse repository metadata: %w", err)
+	}
+	if strings.TrimSpace(repository.DefaultBranch) == "" {
+		return "", fmt.Errorf("repository %s/%s did not report a default branch", owner, repo)
+	}
+	return repository.DefaultBranch, nil
+}
 
-	skillName := ref.RepoName
+func githubInstallDirNameWithBaseURL(repo, githubBaseURL string) (string, error) {
```
wj-xiao, line 385: `githubInstallDirNameWithBaseURL` validates non-URL targets, but it does not validate the directory name derived from URL installs.
A crafted URL like `.../tree/main/..` makes `ref.SubPath` become `..`, and `filepath.Base(ref.SubPath)` also returns `..`. The callers then join that with `workspace/skills`, which escapes the skills directory and can point at the workspace root.
In the failure path, the rollback code may then call `RemoveAll(targetDir)` on that escaped path. Please validate the derived install directory for URL targets as well, and explicitly reject `.` / `..` before returning it.

pkg/config/config.go
```diff
@@ -837,11 +837,12 @@ type ExecConfig struct {
 }
 
 type SkillsToolsConfig struct {
-	ToolConfig            `                       yaml:"-"                 envPrefix:"PICOCLAW_TOOLS_SKILLS_"`
-	Registries            SkillsRegistriesConfig `yaml:",inline,omitempty"                                    json:"registries"`
-	Github                SkillsGithubConfig     `yaml:"github,omitempty"                                     json:"github"`
-	MaxConcurrentSearches int                    `yaml:"-"                                                    json:"max_concurrent_searches" env:"PICOCLAW_TOOLS_SKILLS_MAX_CONCURRENT_SEARCHES"`
-	SearchCache           SearchCacheConfig      `yaml:"-"                                                    json:"search_cache"`
+	ToolConfig `                       yaml:"-"                    envPrefix:"PICOCLAW_TOOLS_SKILLS_"`
+	Registries SkillsRegistriesConfig `yaml:"registries,omitempty"                                    json:"registries"`
```
wj-xiao, line 744: This looks like a backward-compatibility regression for existing `.security.yml` users.

Before this change, registry secrets could be stored under the legacy skills-level layout (for example, skills.clawhub.auth_token).
After changing SkillsToolsConfig.Registries to use yaml:"registries,omitempty", those legacy entries no longer map to the registry config.

Unless dropping that format is intentional, this change may silently break existing ClawHub auth_token configuration for upgraded users.

pkg/config/config_struct.go
```diff
@@ -325,3 +326,336 @@ func (v SecureModelList) MarshalYAML() (any, error) {
 
 	return mm, nil
 }
+
+func (v *SkillsRegistriesConfig) UnmarshalJSON(data []byte) error {
```
wj-xiao, line 355: This JSON unmarshal path replaces the entire registry set whenever `tools.skills.registries` is present. 
Since config loading starts from `DefaultConfig()`, a config that only overrides `registries.github` drops the default `clawhub` entry.

Reproduced this with `config.LoadConfig`, which returned only the GitHub registry. Because migrated configs are auto-saved, this can also silently remove unspecified registries during v1-to-v2 migration.

Can we merge by registry name here, like the YAML/security overlay does, instead of treating the JSON object as a full replacement?

pkg/skills/installer.go
wj-xiao, line 385: If blob URLs are intended to be supported, this path is broken; otherwise the parser/tests should reject blob URLs instead of implying support.

This accepts GitHub `/blob/...` URLs(line 196), but `githubInstallDirNameWithBaseURL` derives the install directory from `filepath.Base(ref.SubPath)`.
For a URL like `/blob/main/.agents/skills/pr-review/SKILL.md`, that resolves to `SKILL.md`, so the installer creates `workspace/skills/SKILL.md/` and then treats the file path as a directory in `InstallFromGitHubToDir`/`downloadRaw`.

pkg/skills/installer.go
```diff
@@ -32,110 +34,435 @@ type GitHubRef struct {
...
+		return gitHubTarget{Ref: ref, Endpoints: endpoints}, nil
+	}
+
+	endpoints, err := resolveGitHubEndpoints(githubBaseURL)
+	if err != nil {
+		return gitHubTarget{}, err
 	}
 
 	// Handle shorthand format
 	parts := strings.Split(strings.Trim(repo, "/"), "/")
```
wj-xiao, line 304: `picoclaw skills install <skill-name>`, the `install_skill tool`, and `/api/skills/install` now default to the `github` registry when `registry` is omitted.
That breaks the existing ClawHub flow, because normal ClawHub slugs are now parsed as GitHub targets and fail unless they are in `owner/repo` form.



Please either:

1. Preserve the previous/default install behavior for non-GitHub slugs, or
2. Only infer `github` when the input is clearly a GitHub URL or `owner/repo` style target.

Additionally, please resolve the conflicts in the `README.zh.md` file.

wj-xiao (approved):

PR #2474, opened 2026-04-10
fix(cron): make each job execution use an independent session

srcrs:
## 📝 Description

Previously all executions of the same cron job reused the session key "cron-{jobID}", causing conversation history to accumulate across runs. Now each run gets a unique key "cron-{jobID}-{timestamp}", preventing cross-execution interference.

## 🗣️ Type of Change
- [ x] 🐞 Bug fix (non-breaking change which fixes an issue)
- [ ] 📖 Documentation update

## 🤖 AI Code Generation
- [x ] 👨‍💻 Mostly Human-written (Human lead, AI assisted or none)

- **Reference URL:**
- **Reasoning:** The cron job session key was previously "cron-{jobID}", causing each execution to accumulate in the same conversation history. Now it uses "cron-{jobID}-{timestamp}" to ensure each run gets an independent session.

## ☑️ Checklist
- [x ] My code/docs follow the style of this project.
- [x ] I have performed a self-review of my own changes.

### DIFF
--- a/pkg/tools/cron.go
+++ b/pkg/tools/cron.go
@@ -342,7 +342,7 @@ func (t *CronTool) ExecuteJob(ctx context.Context, job *cron.CronJob) string {
 		return "ok"
 	}
 
-	sessionKey := fmt.Sprintf("cron-%s", job.ID)
+	sessionKey := fmt.Sprintf("cron-%s-%d", job.ID, time.Now().UnixMilli())
 
 	// Call agent with the job message
 	response, err := t.executor.ProcessDirectWithChannel(
--- a/pkg/tools/cron.go
+++ b/pkg/tools/cron.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/google/uuid"
+
 	"github.com/sipeed/picoclaw/pkg/bus"
 	"github.com/sipeed/picoclaw/pkg/config"
 	"github.com/sipeed/picoclaw/pkg/constants"
--- a/pkg/tools/cron.go
+++ b/pkg/tools/cron.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/google/uuid"
 	"github.com/sipeed/picoclaw/pkg/bus"
 	"github.com/sipeed/picoclaw/pkg/config"
 	"github.com/sipeed/picoclaw/pkg/constants"
@@ -342,7 +343,7 @@ func (t *CronTool) ExecuteJob(ctx context.Context, job *cron.CronJob) string {
 		return "ok"
 	}
 
-	sessionKey := fmt.Sprintf("cron-%s-%d", job.ID, time.Now().UnixMilli())
+	sessionKey := fmt.Sprintf("agent:cron-%s-%s", job.ID, uuid.New().String())
 
 	// Call agent with the job message
 	response, err := t.executor.ProcessDirectWithChannel(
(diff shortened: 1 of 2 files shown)


yinwm (changes requested):
Thanks for the fix! The core idea is correct — cron jobs should get independent sessions. However, there are two blocking issues:

**1. Session key is silently ignored by `resolveScopeKey`**

In `pkg/agent/loop.go:1457-1462`, `resolveScopeKey` only honors session keys prefixed with `"agent:"`:

```go
func resolveScopeKey(route routing.ResolvedRoute, msgSessionKey string) string {
    if msgSessionKey != "" && strings.HasPrefix(msgSessionKey, sessionKeyAgentPrefix) {
        return msgSessionKey
    }
    return route.SessionKey
}
```

The cron key `"cron-{jobID}-{timestamp}"` doesn't match this prefix, so it falls back to `route.SessionKey` (derived from channel/chatID). This means the fix may have **no effect** — multiple executions of the same cron job could still share a session.

Could you verify this by checking the actual `scope_key` in logs when a cron job fires? If confirmed, consider either:
- Prefixing cron keys with `"agent:"` (e.g., `"agent:cron-{jobID}-{timestamp}"`)
- Or adding a special case in `processMessage` for cron-sourced messages

**2. Test assertion is hardcoded to old format**

`pkg/tools/cron_test.go:274` still expects `"cron-job-1"` but the new format is `"cron-job-1-{timestamp}"`. This will fail CI. Update to use a prefix match like `strings.HasPrefix(executor.lastKey, "cron-job-1-")`.

**Non-blocking suggestions:**
- Add a TTL/cleanup strategy for cron sessions to prevent unbounded disk and memory growth
- Consider using UUID instead of `time.Now().UnixMilli()` to avoid collision in concurrent scenarios

srcrs:
> Thanks for the fix! The core idea is correct — cron jobs should get independent sessions. However, there are two blocking issues:
> 
> **1. Session key is silently ignored by `resolveScopeKey`**
> 
> In `pkg/agent/loop.go:1457-1462`, `resolveScopeKey` only honors session keys prefixed with `"agent:"`:
> 
> ```go
> func resolveScopeKey(route routing.ResolvedRoute, msgSessionKey string) string {
>     if msgSessionKey != "" && strings.HasPrefix(msgSessionKey, sessionKeyAgentPrefix) {
>         return msgSessionKey
>     }
>     return route.SessionKey
> }
> ```
> 
> The cron key `"cron-{jobID}-{timestamp}"` doesn't match this prefix, so it falls back to `route.SessionKey` (derived from channel/chatID). This means the fix may have **no effect** — multiple executions of the same cron job could still share a session.
> 
> Could you verify this by checking the actual `scope_key` in logs when a cron job fires? If confirmed, consider either:
> 
> * Prefixing cron keys with `"agent:"` (e.g., `"agent:cron-{jobID}-{timestamp}"`)
> * Or adding a special case in `processMessage` for cron-sourced messages
> 
> **2. Test assertion is hardcoded to old format**
> 
> `pkg/tools/cron_test.go:274` still expects `"cron-job-1"` but the new format is `"cron-job-1-{timestamp}"`. This will fail CI. Update to use a prefix match like `strings.HasPrefix(executor.lastKey, "cron-job-1-")`.
> 
> **Non-blocking suggestions:**
> 
> * Add a TTL/cleanup strategy for cron sessions to prevent unbounded disk and memory growth
> * Consider using UUID instead of `time.Now().UnixMilli()` to avoid collision in concurrent scenarios

fixed, testing is normal.

afjcjsbx:
could you fix lint please? 🙏

srcrs:
> could you fix lint please? 🙏

done

afjcjsbx (approved):
LGTM

yinwm (approved):
LGTM. All blocking issues from the previous review have been addressed — the `agent:` prefix ensures `resolveScopeKey` honors the session key, UUID avoids collision, and tests are updated. Clean fix.

PR #2491, opened 2026-04-12
Add session management commands: /status, /compact, /new

linknightmap:
## 📝 Description

Add session management commands: /status, /compact, /new

This PR adds three new slash commands for manual session context management:
- /status (or /s): Display detailed session status with model, token usage, context bar, history count, summary presence, and compaction count
- /compact [instructions]: Manually compress/compact context, optionally with custom instructions
- /new (or /reset): Start a fresh session

Inspired by OpenClaw's session management features.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 📚 Technical Context
- **Reference URL:** https://github.com/sipeed/picoclaw
- **Reasoning:** Inspired by OpenClaw's /status, /compact, /new commands for better session management

## 🧪 Test Environment
- **Hardware:** PC
- **OS:** Windows
- **Model/Provider:** MiniMax-M2.7
- **Channels:** Telegram

## 📸 Evidence

<details>
<summary>Click to view status output</summary>
🦞 PicoClaw dev

🧠 Model: minimax/MiniMax-M2.7
📚 Context: [████░░░░░░░░░░░░░░░░░] 5.9k/131.1k (4.5%)
💬 History: 20 messages
📝 Summary: present
🧹 Compactions: 0

💡 Tokens: estimated from stored session history

🧵 Session: agent:main:telegram:direct:8154728424
⚙️ Runtime: direct · 🤖 Think: off
</details>

## ☑️ Checklist

### DIFF
--- a/pkg/commands/cmd_compact.go
+++ b/pkg/commands/cmd_compact.go
@@ -1,0 +1,52 @@
+package commands
+
+import (
+	"context"
+	"fmt"
+	"strings"
+)
+
+func compactCommand() Definition {
+	return Definition{
+		Name:        "compact",
+		Description: "Compact session context and summarize older messages",
+		Usage:       "/compact [instructions]",
+		Aliases:     []string{"c"},
+		Handler: func(_ context.Context, req Request, rt *Runtime) error {
+			if rt == nil {
+				return req.Reply(unavailableMsg)
+			}
+
+			if rt.CompactContext == nil {
+				return req.Reply("Compaction is not available in the current context.")
+			}
+
+			var instructions string
+			if fields := strings.Fields(req.Text); len(fields) > 1 {
+				instructions = strings.Join(fields[1:], " ")
+			}
+
+			droppedMessages, err := rt.CompactContext(instructions)
+			if err != nil {
+				return req.Reply("Failed to compact context: " + err.Error())
+			}
+
+			if droppedMessages > 0 {
+				label := "(default)"
+				if instructions != "" {
+					label = instructions
+				}
+				return req.Reply(fmt.Sprintf(
+					"Context compacted.\n\n- Messages summarized: %d\n- Instructions: %s\n- Older context was condensed into the session summary.",
+					droppedMessages,
+					label,
+				))
+			}
+
+			if instructions != "" {
+				return req.Reply("Compaction requested, but the session was already small enough that no history was summarized.")
+			}
+			return req.Reply("Context is already compact enough. No summarization was needed.")
+		},
+	}
+}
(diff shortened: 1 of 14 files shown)


pkg/commands/cmd_new.go
```diff
@@ -0,0 +1,42 @@
+package commands
+
+import "context"
+
+func newCommand() Definition {
```
afjcjsbx, line 5: `/new` is not a new session. It is a second clear path that bypasses `ContextManager.Clear`, does not rotate the session key, and therefore breaks `seahorse`: JSONL history is cleared, but the real SQLite context remains and will still be assembled on the next turn.

pkg/commands/cmd_status.go
```diff
@@ -0,0 +1,112 @@
+package commands
+
+import (
+	"context"
+	"fmt"
+	"strings"
+)
+
+func statusCommand() Definition {
```
afjcjsbx, line 9: There is already a shared context-usage path; this patch adds a second token calculator that ignores system prompt cost, tool definitions, and output reserve. That gives users false confidence: “15% used” can still be followed by a context overflow or forced compression. It also overlaps with /context and /show model.

pkg/commands/cmd_compact.go
```diff
@@ -0,0 +1,52 @@
+package commands
+
+import (
+	"context"
+	"fmt"
+	"strings"
+)
+
+func compactCommand() Definition {
```
afjcjsbx, line 9: `/compact` assumes compaction success can be inferred from `len(agent.Sessions.GetHistory(...))`. That is false for `seahorse`, which compacts its own SQLite store and can do condensed compaction asynchronously. Result: the command can report “no summarization was needed” while compaction actually ran or is still running. Make manual compaction return a real manager-specific result. Add a proper result type to the context-manager API, and make seahorse use a synchronous compaction path for explicit `/compact `requests.

pkg/commands/cmd_compact.go
```diff
@@ -0,0 +1,52 @@
...
+	return Definition{
```
afjcjsbx, line 10: Manual compaction bypasses the existing per-session summarization dedup. If automatic summarization is already running, /compact can launch a second summarizer against the same session. Both goroutines read stale history, both truncate, both overwrite summary state; last writer wins. Manual compaction must go through the same per-session lock/dedup path as auto compaction. One session, one summarizer.

afjcjsbx (changes requested):

PR #2495, opened 2026-04-13
feat(agent): /clear clears seahorse DB

liuy:
## Why

Previously `/clear` only cleared the JSONL session history file. When seahorse is the active context manager, the real conversation data lives in a SQLite database — leaving old messages and summaries untouched. On the next assemble, those old messages could reappear in the context window, especially with a large budget.

## What

`/clear` (and its alias `/reset`) now clears all seahorse SQLite session data.

### Implementation notes

- `store.go`: `ClearConversation` deletes all rows from `messages`, `summaries`, `context_items`, `summary_messages`, and `summary_parents` within a transaction, in child→parent order to respect FK constraints. FTS5 trigger bodies (`summaries_fts`, `messages_fts`) are kept in sync automatically.
- `short_engine.go`: `ClearSession` resolves `sessionKey → convID` then calls `ClearConversation`.
- `context_manager.go`: interface gains a `Clear(ctx, sessionKey)` method, implemented by both `legacyContextManager` (JSONL-only) and `seahorseContextManager` (DB + JSONL).
- `loop.go`: `buildRuntime` injects `rt.ClearHistory` via the new interface method instead of a closure.

### Migration: old DBs are upgraded automatically

During development of this feature the FTS5 DELETE trigger body was initially written with incorrect syntax (`INSERT INTO fts VALUES('delete', ...)`). Databases created with that version would encounter `"summaries: SQL logic error"` on `/clear`. The schema now runs `DROP TRIGGER IF EXISTS` before each `CREATE TRIGGER`, so any existing DB is upgraded to the corrected trigger body automatically on next startup — **no manual DB deletion required**.

## Test plan

- [x] `go test ./pkg/seahorse/... ./pkg/agent/... ./pkg/commands/...` — all pass
- [x] `TestStoreClearConversation` — verifies all tables empty after clear
- [x] `TestTriggerMigration` — verifies old DB with buggy trigger is corrected by `runSchema` migration

Part of https://github.com/sipeed/picoclaw/issues/1919

### DIFF
--- a/pkg/agent/context_legacy.go
+++ b/pkg/agent/context_legacy.go
@@ -61,6 +61,16 @@ func (m *legacyContextManager) Ingest(_ context.Context, _ *IngestRequest) error
 	return nil
 }
 
+func (m *legacyContextManager) Clear(_ context.Context, sessionKey string) error {
+	agent := m.al.registry.GetDefaultAgent()
+	if agent == nil || agent.Sessions == nil {
+		return fmt.Errorf("sessions not initialized")
+	}
+	agent.Sessions.SetHistory(sessionKey, []providers.Message{})
+	agent.Sessions.SetSummary(sessionKey, "")
+	return agent.Sessions.Save(sessionKey)
+}
+
 // maybeSummarize triggers summarization if the session history exceeds thresholds.
 // It runs asynchronously in a goroutine.
 func (m *legacyContextManager) maybeSummarize(sessionKey string) {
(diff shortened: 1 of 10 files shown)


yinwm (approved):
LGTM. Clean implementation with correct FK deletion order, proper transaction usage, and thorough trigger migration handling for existing DBs.

Two non-blocking suggestions:
1. `ClearSession` doesn't acquire the session mutex unlike `Ingest` — consider adding `e.getSessionMutex(sessionKey).Lock()` for consistency.
2. Adding a log line in `ClearSession`/`ClearConversation` would help with debugging "clear didn't work" reports.

PR #2502, opened 2026-04-13
feat(agent): add /btw one-off side-question command

lxowalle:
### Summary
- add a new `/btw <question>` built-in command for one-off side questions
- route `/btw` through a priority command path so it can bypass the normal queued turn flow and reply immediately
- keep `/btw` requests and replies out of persisted session history while preserving routed agent/session metadata on immediate replies
- update command runtime/tests and refresh localized chat/configuration docs to include `/btw`

### Why
This adds a lightweight way to ask a quick side question while another turn is busy, without appending the `/btw` request or reply to the active session history.

### Behavior Notes
- `/btw` is a temporary side-question message, not part of the main session turn flow.
- `/btw` does not persist into the active session history.
- `/btw` is independent from `/use`: pending skill overrides armed by `/use <skill>` apply to the next regular session message, not to `/btw`.

### Changes
- register `/btw` in the built-in command registry
- add a dedicated `/btw` command handler and runtime hook
- implement agent-loop support to:
  - detect `/btw` as a priority command in the steering path
  - execute `/btw` immediately instead of enqueueing it behind the current turn
  - keep the `/btw` exchange out of persisted session history
  - preserve the routed agent/session metadata when publishing immediate `/btw` replies
- add tests covering:
  - command execution and usage validation
  - non-persistent history behavior
  - steering tests for priority execution / queue bypass behavior and outbound metadata preservation
- update documentation and Telegram command-menu references in localized docs

### Testing
- added unit tests for `/btw` command handling
- added agent loop tests for no-history persistence behavior
- added steering tests for priority execution / queue bypass behavior and outbound metadata preservation

## 🗣️ Type of Change
- [x] 📖 Documentation update

## 🤖 AI Code Generation

- **Reference URL:**
- **Reasoning:**

## 🧪 Test Environment
- **Hardware:**
- **OS:**
- **Model/Provider:**
- **Channels:**

## 📸 Evidence (Optional)
<details>

</details>

## ☑️ Checklist


pkg/agent/loop.go
```diff
@@ -1529,6 +1541,119 @@ func (al *AgentLoop) ProcessHeartbeat(
...
+	var summary string
+	if opts != nil {
+		channel = opts.Channel
+		chatID = opts.ChatID
+		senderID = opts.SenderID
+		senderDisplayName = opts.SenderDisplayName
+		media = append([]string(nil), opts.Media...)
+		activeSkills = activeSkillNames(agent, *opts)
+
+		if !opts.NoHistory {
```
wj-xiao, line 1568: `/btw` loses the live session context.

This implementation only reassembles the session summary and then calls `BuildMessages` with `nil` history, so the `/btw` request does not see the current conversation state unless it has already been compacted into the summary.

That creates a real behavior gap for the feature as described. A side question such as “what did we decide above?” or “can you expand on the previous point?” may fail during an active turn because the recent turns are still in history, not in the summary. In contrast, the normal turn path includes both history and summary, so `/btw` is now operating on materially less context than the main conversation.

I think `/btw` needs to preserve the same read context as the active session while only skipping persistence of the `/btw` exchange itself. Otherwise this will produce inconsistent and confusing answers in exactly the situations where users are most likely to use the command.

pkg/agent/loop.go
```diff
@@ -1529,6 +1541,119 @@ func (al *AgentLoop) ProcessHeartbeat(
...
+		"temperature":      agent.Temperature,
+		"prompt_cache_key": agent.ID + ":btw",
+	}
+	if agent.ThinkingLevel != ThinkingOff {
+		if tc, ok := activeProvider.(providers.ThinkingCapable); ok && tc.SupportsThinking() {
+			llmOpts["thinking_level"] = string(agent.ThinkingLevel)
+		}
+	}
+
+	callProvider := func(ctx context.Context, provider providers.LLMProvider, model string) (*providers.LLMResponse, error) {
```
wj-xiao, line 1612: `/btw` bypasses the normal tool and hook pipeline.

This path calls `provider.Chat` directly with nil tools, which means `/btw` no longer behaves like a normal agent turn. Side questions cannot use the agent’s configured tools, and any logic enforced in the standard turn pipeline (for example LLM hooks, policy checks, request mutation, or approval gates) is skipped entirely.

That is a much larger semantic change than “do not persist `/btw` in session history”. It effectively introduces a second execution model for the same agent, with different capabilities and different enforcement points. Even if the no-history behavior is intentional, bypassing tools and hooks is likely to create hard-to-debug inconsistencies between a normal prompt and the same prompt sent through `/btw`.

Unless `/btw` is explicitly intended to be a stripped-down raw-model query, I think  this should be routed through the normal agent execution path with a “do not persist” option, rather than calling the provider directly.

pkg/agent/loop.go
```diff
@@ -653,6 +653,18 @@ func (al *AgentLoop) drainBusToSteering(ctx context.Context, activeScope, active
 		// Transcribe audio if needed before steering, so the agent sees text.
 		msg, _ = al.transcribeAudioInMessage(ctx, msg)
 
+		// Handle priority commands (e.g. /btw) immediately instead of queueing them.
+		if handled, response := al.tryHandlePriorityCommand(ctx, msg); handled {
+			if response != "" {
+				al.bus.PublishOutbound(ctx, bus.OutboundMessage{
```
wj-xiao: Immediate `/btw` replies drop reply metadata.

The fast path publishes an outbound message with only `Channel`, `ChatID`, and `Content`. It does not preserve `ReplyToMessageID` or the normalized outbound context used by the regular response path.

On reply-capable channels, that can cause the `/btw` response to be posted as a detached top-level message instead of replying to the triggering message/thread. This looks like a regression in delivery behavior even if the text content itself is correct.

pkg/agent/loop.go
lxowalle: /btw no need to call the tool; it's currently just for quick Q&A.

lxowalle (commented):
发现 1 个需要修正的问题：

空闲场景下 `/btw` 的回复仍然丢失 `agent_id/session_key/scope`，和 busy-turn 的 priority 路径行为不一致。

- busy-turn 场景走 `tryHandlePriorityCommand()`（`pkg/agent/loop.go:4058`），这里会显式构造带 `AgentID/SessionKey/Scope` 的 `OutboundMessage`。
- 但空闲场景仍走普通 `processMessage -> handleCommand` 路径（`pkg/agent/loop.go:1738`），最终在 `Run()` 中通过 `PublishResponseIfNeeded()` 发送（`pkg/agent/loop.go:530`, `pkg/agent/loop.go:690`），这里没有附带 routed metadata。

结果就是同一个 `/btw` 命令会因为系统是否正忙而产出两种不同的 outbound 结构。这和 PR 描述里“preserve routed agent/session metadata on immediate replies”的目标不一致，也可能破坏依赖这些字段的下游逻辑。

建议补一个 idle-path 测试，并让空闲时的 `/btw` 回复也走和 priority path 一致的 metadata 填充逻辑。

PR #2503, opened 2026-04-13
refactor: make agent loop support parallel and update docs

cytown:
## 📝 Description

Base on PR#2481, please review after the PR was merged.

Refactored the AgentLoop in /Users/billy/github/picoclaw/pkg/agent/loop.go to support parallel processing of user messages with correct response routing. Here's what was implemented:

    Key Changes

     1. Worker Pool Architecture (loop.go):
        - Added workerPoolSize and workerSem (semaphore) fields to AgentLoop struct
        - Added activeWorkers counter for monitoring
        - Workers are dispatched via goroutines with semaphore-based concurrency limiting

     2. Enhanced `Run()` Method (loop.go:468-543):
        - Replaced sequential message processing with worker pool dispatcher
        - Messages with different session keys are processed in parallel (up to MaxParallelTurns)
        - Messages with same session key are enqueued to steering queue (preserves conversation integrity)
        - System messages are processed immediately in synchronous mode

     3. New Helper Methods:
        - processMessageSync(): Handles non-routable messages synchronously
        - runTurnWithSteering(): Runs a complete turn and drains steering queue in worker context

     4. Configuration Support (config.go:262):
        - Added MaxParallelTurns field to AgentDefaults struct
        - Default value: 1 (preserves backward compatibility - sequential processing)
        - Environment variable: PICOCLAW_AGENTS_DEFAULTS_MAX_PARALLEL_TURNS

     5. Removed Obsolete Code:
        - Removed drainBusToSteering() - no longer needed with direct steering queue enrollment
        - Removed associated test TestDrainBusToSteering_RequeuesDifferentScopeMessage

     6. Comprehensive Tests (loop_test.go):
        - TestParallelMessageProcessing_DifferentSessionsProcessedConcurrently: Verifies parallel execution (achieved 3
          concurrent turns)
        - TestParallelMessageProcessing_SameSessionProcessedSequentially: Verifies same-session messages are queued

    How It Works

      1 Inbound Message Flow:
      2 ┌─────────────────────────────────────────┐
      3 │  bus.InboundChan()                      │
      4 └──────────────┬──────────────────────────┘
      5                │
      6                ▼
      7 ┌─────────────────────────────────────────┐
      8 │  resolveSteeringTarget(msg)             │
      9 │  Returns: sessionKey, agentID, ok       │
     10 └──────────────┬──────────────────────────┘
     11                │
     12         ┌──────┴──────┐
     13         │             │
     14         ▼             ▼
     15    Not routable   Has session
     16    (system msg)   key
     17         │             │
     18         │             ├─► activeTurnStates.Load(sessionKey)?
     19         │             │         │
     20         │             │    Yes  │  No
     21         │             │         │     │
     22         ▼             │         ▼     ▼
     23  processMessageSync   │   Enqueue   Acquire workerSem
     24                       │   to        slot
     25                       │   steering  │
     26                       │   queue     ▼
     27                       │         runTurnWithSteering()
     28                       │         (processes message +
     29                       │          drains steering queue)

     

## 🗣️ Type of Change
- [x] 📖 Documentation update

## 🤖 AI Code Generation

- **Reference URL:**
- **Reasoning:**

## 🧪 Test Environment
- **Hardware:** 
- **OS:** 
- **Model/Provider:** 
- **Channels:** 

## 📸 Evidence (Optional)
<details>

</details>

## ☑️ Checklist


pkg/agent/loop.go
```diff
@@ -472,127 +491,99 @@ func (al *AgentLoop) Run(ctx context.Context) error {
...
-			if activeScope, activeAgentID, ok := al.resolveSteeringTarget(msg); ok {
-				drainCtx, cancel := context.WithCancel(ctx)
-				drainCancel = cancel
-				go al.drainBusToSteering(drainCtx, activeScope, activeAgentID)
+			// Resolve the session key for this message
+			sessionKey, agentID, ok := al.resolveSteeringTarget(msg)
+			if !ok {
+				// Non-routable message (e.g., system) — dispatch to a worker
+				// goroutine so the main receive loop keeps draining.
+				go al.processMessageSync(ctx, msg)
```
alexhoshina: When multiple **system messages** arrive simultaneously, multiple _goroutines_ will be launched.
It is necessary to evaluate whether there is an issue of concurrent read/write access to the same session history.

pkg/agent/loop.go
```diff
@@ -472,127 +491,99 @@ func (al *AgentLoop) Run(ctx context.Context) error {
...
+			// to prevent a TOCTOU race where multiple messages for the same session
+			// pass the Load check before either registers.
+			// The placeholder ensures GetActiveTurnBySession() never returns nil
+			// during turn setup.
+			if _, loaded := al.activeTurnStates.LoadOrStore(sessionKey, placeholderTurnState); loaded {
+				// Another turn is already active (or reserved) for this session — enqueue
+				if err := al.enqueueSteeringMessage(sessionKey, agentID, providers.Message{
+					Role:    "user",
+					Content: msg.Content,
+					Media:   append([]string(nil), msg.Media...),
```
alexhoshina: Not calling **transcribeAudioInMessage()** prevents voice messages in the **steering phase** from being transcribed

pkg/agent/loop.go
```diff
@@ -472,127 +491,99 @@ func (al *AgentLoop) Run(ctx context.Context) error {
...
+			}
+
+			// Session claimed — spawn a worker goroutine that acquires a semaphore
+			// slot. The goroutine is spawned immediately so the main loop keeps
+			// draining the inbound channel. The goroutine blocks on the semaphore.
+			go func(m bus.InboundMessage) {
+				// Acquire semaphore slot (blocks if at capacity)
+				select {
+				case al.workerSem <- struct{}{}:
+					// Got slot, start worker
```
alexhoshina, line 534: In the case of **worker** parallelism, the state of **message tools** is shared. This results in all **message tools** reading the same **sentTargets** state to determine whether to suppress the final reply.

pkg/agent/loop.go
```diff
@@ -675,6 +666,102 @@ func (al *AgentLoop) drainBusToSteering(ctx context.Context, activeScope, active
...
+		logger.InfoCF("agent", "Continuing queued steering after turn end",
+			map[string]any{
+				"channel":     target.Channel,
+				"chat_id":     target.ChatID,
+				"session_key": target.SessionKey,
+				"queue_depth": al.pendingSteeringCountForScope(target.SessionKey),
+			})
+
+		continued, continueErr := al.Continue(ctx, target.SessionKey, target.Channel, target.ChatID)
+		if continueErr != nil {
```
alexhoshina, line 645: There is a gap period; the **session** lock will have a release gap before the **steering** queue is fully cleared.
If a new message arrives:
1. **Run()** finds no **active turn**
2. It will start a new **worker**
3. The same **session** may have concurrent **workers**.

alexhoshina (changes requested):

pkg/agent/loop.go
cytown: Not a issue:
1. When the steering message is later dequeued by Continue() → continueWithSteeringMessages, it passes through processMessage which calls transcribeAudioInMessage (line 1641).
2. The steering queue stores providers.Message{Content: msg.Content, Media: msg.Media}. When dequeued, this becomes a new bus.InboundMessage that goes through the full processMessage pipeline, including transcription.

pkg/agent/loop.go
cytown: Not a issue:

     1. The placeholder is only deleted by clearActiveTurn(ts) which runs after runTurn returns.
     2. The steering drain loop runs inside runTurnWithSteering, which is still within the same worker goroutine that holds the placeholder.
     3. clearActiveTurn is deferred in runTurn, so the placeholder is replaced by the real turnState until the very end.
     4. Even after clearActiveTurn, the worker is still executing. A new message arriving would hit LoadOrStore after the key is deleted, but it would start a new turn — which is correct behavior since the previous turn is complete.

pkg/agent/loop.go
cytown: Reverted go al.processMessageSync back to synchronous al.processMessageSync — system messages block the receive loop but guarantee session serialization

pkg/agent/loop.go
cytown: Converted MessageTool.sentTargets from a single slice to a map[string][]sentTarget keyed by session key. Updated ResetSentInRound, HasSentInRound, HasSentTo, and PublishResponseIfNeeded to accept sessionKey parameter

alexhoshina (approved):

yinwm (changes requested):
Thanks for the thorough refactor! The worker pool architecture is clean and well-documented. However, I found two critical issues with the shared `placeholderTurnState` singleton that need to be addressed before merging:

## 1. Worker placeholder cleanup can delete another worker's placeholder (breaks session serialization)

In `loop.go:546-553`, the safety-net defer checks `ts.turnID == "pending"` to detect a stale placeholder. Since `placeholderTurnState` is a package-level singleton shared across all sessions, Worker A's cleanup defer can accidentally delete Worker B's freshly-stored placeholder for the *same* session:

```
Worker A finishes → clearActiveTurn deletes session key
Run() main loop → LoadOrStore stores new placeholder → spawns Worker B
Worker A's cleanup defer → Load finds placeholder (Worker B's) → deletes it
Run() main loop → LoadOrStore succeeds again → spawns Worker C
→ Worker B and Worker C run concurrently for the same session ❌
```

**Fix**: Create a unique placeholder instance per `LoadOrStore` call (e.g., `&turnState{turnID: "pending-" + sessionKey}`), or use a unique ID to distinguish "my placeholder" from "someone else's placeholder" in the cleanup defer.

## 2. HardAbort/InterruptHard can mutate the shared placeholder singleton

`HardAbort()` (steering.go:489) does a type assertion `tsInterface.(*turnState)` which succeeds for `placeholderTurnState` since it's also `*turnState`. This calls `ts.Finish(true)` which permanently marks the global singleton as finished, affecting all future sessions that use it. Similarly, `InterruptHard()` via `getAnyActiveTurnState()` can call `requestHardAbort()` on the placeholder.

**Fix**: Add a guard in `HardAbort` and `InterruptHard`:
```go
if ts.turnID == "pending" {
    return fmt.Errorf("turn is still initializing for session %s", sessionKey)
}
```

## Additional non-blocking suggestions

- **Continue() placeholder handling**: In `steering.go:355-357`, `GetActiveTurnBySession` returns non-nil for placeholder (turnID="pending"), causing `Continue()` to return an error instead of gracefully yielding. This can prematurely break the steering drain loop in `runTurnWithSteering`.
- **sentTargets memory leak**: `message.go:68` — `ResetSentInRound` truncates the slice but never `delete`s the map key. Over time with many unique sessions, the map grows unbounded.
- **Dead code**: `drainBusToSteering` is no longer called from `Run()` (only referenced in `steering_test.go`).

These issues don't manifest when `max_parallel_turns=1` (default), but will break session serialization when parallelism is enabled.

cytown:
> Thanks for the thorough refactor! The worker pool architecture is clean and well-documented. However, I found two critical issues with the shared `placeholderTurnState` singleton that need to be addressed before merging:
> 
> ## 1. Worker placeholder cleanup can delete another worker's placeholder (breaks session serialization)
> In `loop.go:546-553`, the safety-net defer checks `ts.turnID == "pending"` to detect a stale placeholder. Since `placeholderTurnState` is a package-level singleton shared across all sessions, Worker A's cleanup defer can accidentally delete Worker B's freshly-stored placeholder for the _same_ session:
> 
> ```
> Worker A finishes → clearActiveTurn deletes session key
> Run() main loop → LoadOrStore stores new placeholder → spawns Worker B
> Worker A's cleanup defer → Load finds placeholder (Worker B's) → deletes it
> Run() main loop → LoadOrStore succeeds again → spawns Worker C
> → Worker B and Worker C run concurrently for the same session ❌
> ```
> 
> **Fix**: Create a unique placeholder instance per `LoadOrStore` call (e.g., `&turnState{turnID: "pending-" + sessionKey}`), or use a unique ID to distinguish "my placeholder" from "someone else's placeholder" in the cleanup defer.
> 
> ## 2. HardAbort/InterruptHard can mutate the shared placeholder singleton
> `HardAbort()` (steering.go:489) does a type assertion `tsInterface.(*turnState)` which succeeds for `placeholderTurnState` since it's also `*turnState`. This calls `ts.Finish(true)` which permanently marks the global singleton as finished, affecting all future sessions that use it. Similarly, `InterruptHard()` via `getAnyActiveTurnState()` can call `requestHardAbort()` on the placeholder.
> 
> **Fix**: Add a guard in `HardAbort` and `InterruptHard`:
> 
> ```go
> if ts.turnID == "pending" {
>     return fmt.Errorf("turn is still initializing for session %s", sessionKey)
> }
> ```
> 
> ## Additional non-blocking suggestions
> * **Continue() placeholder handling**: In `steering.go:355-357`, `GetActiveTurnBySession` returns non-nil for placeholder (turnID="pending"), causing `Continue()` to return an error instead of gracefully yielding. This can prematurely break the steering drain loop in `runTurnWithSteering`.
> * **sentTargets memory leak**: `message.go:68` — `ResetSentInRound` truncates the slice but never `delete`s the map key. Over time with many unique sessions, the map grows unbounded.
> * **Dead code**: `drainBusToSteering` is no longer called from `Run()` (only referenced in `steering_test.go`).
> 
> These issues don't manifest when `max_parallel_turns=1` (default), but will break session serialization when parallelism is enabled.

all done.

yinwm (approved):
LGTM. Both CRITICAL issues from the first round are properly fixed:
- Placeholder is now a unique per-claim instance (makePendingTurnID with sequence number)
- HardAbort guards against pending turns with prefix check
- sentTargets memory leak fixed with delete()
- Dead code (drainBusToSteering, requeueInboundMessage) cleaned up

The extra cleanup (vision retry inline, code reorganization) is a nice bonus.

PR #2517, opened 2026-04-14
fix(web_search): validate missing API key/URL directly in Search methods

lxowalle:
### Summary
Move the API key / URL validation for web search providers from `NewWebSearchTool` into the individual `Search` methods. When a required key or URL is missing, the tool now returns a clear error at execution time instead of silently creating a `nil` tool.
### Changes
- `BraveSearchProvider.Search`: return `no API key provided` when the key pool is empty
- `TavilySearchProvider.Search`: return `no API key provided` when the key pool is empty
- `PerplexitySearchProvider.Search`: return `no API key provided` when the key pool is empty
- `GLMSearchProvider.Search`: return `no API key provided` when the API key is empty
- `BaiduSearchProvider.Search`: return `no API key provided` when the API key is empty
- `SearXNGSearchProvider.Search`: return `no SearXNG URL provided` when `baseURL` is empty
- `NewWebSearchTool`: remove key / URL non-empty guards so the tool can be instantiated even when credentials are missing
- Update `TestWebTool_WebSearch_NoApiKey` to assert the runtime error behavior
- 
## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

- **Reference URL:**
- **Reasoning:**

## 🧪 Test Environment
- **Hardware:** 
- **OS:** 
- **Model/Provider:** 
- **Channels:** 

## 📸 Evidence (Optional)
<details>

</details>

## ☑️ Checklist

### DIFF
--- a/pkg/tools/web.go
+++ b/pkg/tools/web.go
@@ -218,6 +218,10 @@ func (p *BraveSearchProvider) Search(
 	count int,
 	rangeCode string,
 ) (string, error) {
+	if p.keyPool == nil || len(p.keyPool.keys) == 0 {
+		return "", errors.New("no API key provided")
+	}
+
 	searchURL := fmt.Sprintf("https://api.search.brave.com/res/v1/web/search?q=%s&count=%d",
 		url.QueryEscape(query), count)
 	if freshness := mapBraveFreshness(rangeCode); freshness != "" {
@@ -317,6 +321,10 @@ func (p *TavilySearchProvider) Search(
 	count int,
 	rangeCode string,
 ) (string, error) {
+	if p.keyPool == nil || len(p.keyPool.keys) == 0 {
+		return "", errors.New("no API key provided")
+	}
+
 	searchURL := p.baseURL
 	if searchURL == "" {
 		searchURL = "https://api.tavily.com/search"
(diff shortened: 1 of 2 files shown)


pkg/tools/web.go
wj-xiao, line 944: This change does more than move validation from construction time to execution time; it also changes provider selection semantics.

Before this patch, `NewWebSearchTool` selected the first usable provider in priority order by requiring a non-empty key/URL for providers that need credentials.
After this patch, the first enabled provider wins even when it is not configured, so a setup like `PerplexityEnabled=true` with no keys and `DuckDuckGoEnabled=true` will now fail at runtime with "no API key provided" instead of falling back to DuckDuckGo.

The same regression applies to Brave, SearXNG, Tavily, Baidu Search, and GLM Search. We should preserve the existing "first usable provider" fallback behavior and add a regression test for mixed-provider configurations.

lxowalle: I think it is acceptable. Previously, whether to load a provider was determined based on the supplier's key/URL. If the key/URL was not set, the web_search tool would not be loaded into the context. Now, if a provider does not have a key configured, it will intuitively inform the user that the key/URL is not configured when the tool is called. I believe sacrificing the priority selection feature is acceptable.

wj-xiao (approved):

PR #2547, opened 2026-04-16
feat(network): improve network error classification and fallback handling

### LINKED ISSUE #2538: Feature: Network errors (TLS/connection) should trigger fallback retry instead of failing immediately
## Problem

Currently, when a network-level error occurs (such as TLS errors or connection resets), the fallback chain **immediately terminates** instead of trying the next candidate model.

### Examples of affected errors:

1. **TLS errors:**
   ```
   fallback: unclassified error from minimax/minimax-m2.7: failed to send request: Post "https://opencode.ai/zen/go/v1/chat/completions": tls: bad record MAC
   ```

2. **Connection reset:**
   ```
   LLM call failed after retries: fallback: unclassified error from minimax/minimax-m2.7: failed to send request: Post "https://opencode.ai/zen/go/v1/chat/completions": read tcp 10.20.0.1:61279->172.65.90.20:443: read: connection reset by peer
   ```

## Root Cause

In `pkg/providers/fallback.go:207-216`, when `ClassifyError()` returns `nil` (unclassified error), the fallback chain terminates immediately:

```go
if failErr == nil {
    // Unclassifiable error: do not fallback, return immediately.
    return nil, fmt.Errorf("fallback: unclassified error from %s/%s: %w", ...)
}
```

The `ClassifyError` function (`pkg/providers/error_classifier.go:115`) only recognizes:
- HTTP status codes (401, 403, 429, 500, 502, 503, etc.)
- Known error message patterns (rate limit, timeout, auth, billing, context overflow)

Network-level errors like `tls: bad record MAC`, `connection reset by peer`, `connection refused`, `handshake failed`, `certificate expired`, etc. don't match any of these patterns, so they're treated as \"unclassified\" and fallback is skipped.

## Expected Behavior

Network errors are inherently transient - a connection reset or TLS handshake failure on one provider often succeeds on another. These errors should:

1. Be classified as retriable (like `FailoverTimeout`)
2. Trigger fallback to the next candidate in the chain

## Suggested Fix

Add network error pattern matching in `error_classifier.go`, for example:

```go
networkPatterns = []errorPattern{
    substr("tls:"),
    substr("connection reset"),
    substr("connection refused"),
    substr("handshake"),
    substr("certificate"),
    substr("read tcp"),
    substr("write tcp"),
    substr("eof"),
    substr("broken pipe"),
}
```

Then in `classifyByMessage`, these should return `FailoverTimeout` (or a new `FailoverNetwork` type) so the fallback chain continues.

## Impact

This would improve reliability when:
- One provider has temporary network issues
- TLS certificates are temporarily problematic on one endpoint
- Load balancer drops connections intermittently

Currently users see \"LLM call failed after retries\" even when alternative models could have succeeded.

lc6464:
## 📝 Description

This PR improves provider fallback behavior for transport-level failures that were previously left unclassified.

It adds a dedicated `network` failover reason so errors such as TLS failures, DNS lookup failures, EOF/unexpected EOF, connection reset/refused, and other concrete network errors can continue the fallback chain instead of stopping on the first provider.

It also adds regression tests for:
- network-related message patterns
- concrete network error types
- fallback execution when the first candidate fails with a network error

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue
Fixes #2538

## 📚 Technical Context
- **Reference URL:** Issue #2538
- **Reasoning:** Previously, fallback relied on HTTP status extraction and limited message heuristics. Transport-layer failures without a structured HTTP status could remain unclassified, which caused fallback to stop early. This change introduces `FailoverNetwork`, classifies concrete transport errors before message-based heuristics, and marks the new reason as retriable so the next fallback candidate can be attempted.

## 🧪 Test Environment
- **OS:** Ubuntu Server 24
- **Model/Provider:** -
- **Channels:** Web

## 📸 Evidence
<details>

image

image

</details>

## ☑️ Checklist

### DIFF
--- a/pkg/providers/error_classifier.go
+++ b/pkg/providers/error_classifier.go
@@ -2,8 +2,12 @@ package providers
 
 import (
 	"context"
+	"errors"
+	"io"
+	"net"
 	"regexp"
 	"strings"
+	"syscall"
 )
 
 // Common patterns in Go HTTP error messages
@@ -48,6 +52,30 @@ var (
 		substr("timed out"),
 		substr("deadline exceeded"),
 		substr("context deadline exceeded"),
+	}
+
+	networkPatterns = []errorPattern{
+		substr("connection reset"),
+		substr("reset by peer"),
+		substr("connection refused"),
+		substr("connection aborted"),
+		substr("broken pipe"),
+		substr("use of closed network connection"),
+		substr("network is unreachable"),
+		substr("host is unreachable"),
+		substr("no such host"),
+		substr("temporary failure in name resolution"),
+		substr("server misbehaving"),
+		substr("read tcp"),
+		substr("write tcp"),
+		substr("dial tcp"),
+		substr("tls:"),
+		substr("x509:"),
+		substr("certificate"),
+		substr("handshake"),
+		substr("unexpected eof"),
+		substr("read: eof"),
+		substr("write: eof"),
 	}
 
 	billingPatterns = []errorPattern{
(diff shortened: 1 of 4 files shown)


yinwm (approved):
LGTM! Two-layer network error classification (type-based + message heuristic) is well-designed. The timeout/network distinction after Copilot's feedback is correct.

One pre-existing note for a future PR: `ExecuteImage` doesn't use `ClassifyError`, so network errors on the image path lack classification and cooldown. Not blocking this PR, but worth tracking for consistency.

Approved.

PR #2481, opened 2026-04-11
refactor(config):  make config.Channel to multiple instance support

cytown:
## 📝 Description

 refactor(config):  make config.Channel to multiple instance support
    
    add new field type to Channel struct
    config.channels refactor to channel_list
    update config version to 3
    update the docs

## 🗣️ Type of Change
- [ ] 🐞 Bug fix (non-breaking change which fixes an issue)
- [X] ✨ New feature (non-breaking change which adds functionality)
- [ ] 📖 Documentation update
- [x] ⚡ Code refactoring (no functional changes, no api changes)

## 🤖 AI Code Generation
- [ ] 🤖 Fully AI-generated (100% AI, 0% Human)
- [x] 🛠️ Mostly AI-generated (AI draft, Human verified/modified)
- [ ] 👨‍💻 Mostly Human-written (Human lead, AI assisted or none)

## 📚 Technical Context (Skip for Docs)
- **Reference URL:**
- **Reasoning:**

## 🧪 Test Environment
- **Hardware:** 
- **OS:** 
- **Model/Provider:** 
- **Channels:** 

## 📸 Evidence (Optional)
<details>
<summary>Click to view Logs/Screenshots</summary>

</details>

## ☑️ Checklist
- [x] My code/docs follow the style of this project.
- [x] I have performed a self-review of my own changes.
- [x] I have updated the documentation accordingly.

PR #2551, opened 2026-04-16
refactor: standardize channel identification and decouple name from provider type

cytown:
Decouple channel names (config keys) from channel types (provider IDs) to allow multiple instances of the same provider. This ensures robust identification across the message bus and agent dispatch logic.

- Introduce ChannelType in InboundContext for consistent provider tracking
- Refactor IsInternalChannel to use provider type instead of generic names
- Update all channel implementations to pass through their provider type via BaseChannel
- Standardize agent dispatch and heartbeat recording to use ChannelType
- Enhance shell and cron tools with improved execution tracking and fallback logic
- Fix redundant channel field assignments and hardcoded types in providers
- Resolve build issues by adding missing logger imports to tool packages

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

- **Reference URL:**
- **Reasoning:**

## 🧪 Test Environment
- **Hardware:** 
- **OS:** 
- **Model/Provider:** 
- **Channels:** 

## 📸 Evidence (Optional)
<details>

</details>

## ☑️ Checklist

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -474,7 +474,7 @@ func (al *AgentLoop) runAgentLoop(
 	// Record last channel for heartbeat notifications (skip internal channels and cli)
 	if opts.Dispatch.Channel() != "" &&
 		opts.Dispatch.ChatID() != "" &&
-		!constants.IsInternalChannel(opts.Dispatch.Channel()) {
+		!constants.IsInternalChannel(opts.Dispatch.ChannelType()) {
 		channelKey := fmt.Sprintf("%s:%s", opts.Dispatch.Channel(), opts.Dispatch.ChatID())
 		if err := al.RecordLastChannel(channelKey); err != nil {
 			logger.WarnCF(
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -536,7 +536,7 @@ func (al *AgentLoop) runAgentLoop(
 	// Record last channel for heartbeat notifications (skip internal channels and cli)
 	if opts.Dispatch.Channel() != "" &&
 		opts.Dispatch.ChatID() != "" &&
-		!constants.IsInternalChannel(opts.Dispatch.Channel()) {
+		!constants.IsInternalChannel(opts.Dispatch.ChannelType()) {
 		channelKey := fmt.Sprintf("%s:%s", opts.Dispatch.Channel(), opts.Dispatch.ChatID())
 		if err := al.RecordLastChannel(channelKey); err != nil {
 			logger.WarnCF(
--- a/pkg/agent/agent_message.go
+++ b/pkg/agent/agent_message.go
@@ -53,10 +53,11 @@ func (al *AgentLoop) ProcessDirectWithChannel(
 
 	msg := bus.InboundMessage{
 		Context: bus.InboundContext{
-			Channel:  channel,
-			ChatID:   chatID,
-			ChatType: "direct",
-			SenderID: "cron",
+			Channel:     channel,
+			ChannelType: channel, // For direct calls, channel name equals channel type
+			ChatID:      chatID,
+			ChatType:    "direct",
+			SenderID:    "cron",
 		},
 		Content:    content,
 		SessionKey: sessionKey,
@@ -86,10 +87,11 @@ func (al *AgentLoop) ProcessHeartbeat(
 	}
 	if channel != "" || chatID != "" {
 		dispatch.InboundContext = &bus.InboundContext{
-			Channel:  channel,
-			ChatID:   chatID,
-			ChatType: "direct",
-			SenderID: "heartbeat",
+			Channel:     channel,
+			ChannelType: channel, // For heartbeat, channel name equals channel type
+			ChatID:      chatID,
+			ChatType:    "direct",
+			SenderID:    "heartbeat",
 		}
 	}
 	return al.runAgentLoop(ctx, agent, processOptions{
@@ -247,7 +249,7 @@ func (al *AgentLoop) processSystemMessage(
 	// Parse origin channel from chat_id (format: "channel:chat_id")
 	var originChannel, originChatID string
 	if idx := strings.Index(msg.ChatID, ":"); idx > 0 {
-		originChannel = msg.ChatID[:idx]
+		originChannel = msg.ChatID[:idx] // e.g. "telegram"
 		originChatID = msg.ChatID[idx+1:]
 	} else {
 		originChannel = "cli"
--- a/pkg/agent/agent_message.go
+++ b/pkg/agent/agent_message.go
@@ -53,10 +53,11 @@ func (al *AgentLoop) ProcessDirectWithChannel(
 
 	msg := bus.InboundMessage{
 		Context: bus.InboundContext{
-			Channel:  channel,
-			ChatID:   chatID,
-			ChatType: "direct",
-			SenderID: "cron",
+			Channel:     channel,
+			ChannelType: channel, // For direct calls, channel name equals channel type
+			ChatID:      chatID,
+			ChatType:    "direct",
+			SenderID:    "cron",
 		},
 		Content:    content,
 		SessionKey: sessionKey,
@@ -86,10 +87,11 @@ func (al *AgentLoop) ProcessHeartbeat(
 	}
 	if channel != "" || chatID != "" {
 		dispatch.InboundContext = &bus.InboundContext{
-			Channel:  channel,
-			ChatID:   chatID,
-			ChatType: "direct",
-			SenderID: "heartbeat",
+			Channel:     channel,
+			ChannelType: channel, // For heartbeat, channel name equals channel type
+			ChatID:      chatID,
+			ChatType:    "direct",
+			SenderID:    "heartbeat",
 		}
 	}
 	return al.runAgentLoop(ctx, agent, processOptions{
@@ -256,7 +258,7 @@ func (al *AgentLoop) processSystemMessage(
 	// Parse origin channel from chat_id (format: "channel:chat_id")
 	var originChannel, originChatID string
 	if idx := strings.Index(msg.ChatID, ":"); idx > 0 {
-		originChannel = msg.ChatID[:idx]
+		originChannel = msg.ChatID[:idx] // e.g. "telegram"
 		originChatID = msg.ChatID[idx+1:]
 	} else {
 		originChannel = "cli"
--- a/pkg/agent/agent_test.go
+++ b/pkg/agent/agent_test.go
@@ -3987,7 +3987,8 @@ func TestProcessMessage_PicoPublishesReasoningAsThoughtMessage(t *testing.T) {
 	al := NewAgentLoop(cfg, msgBus, provider)
 
 	response, err := al.processMessage(context.Background(), bus.InboundMessage{
-		Channel:  "pico",
+		Context:  bus.InboundContext{Channel: "pico1", ChannelType: "pico"},
+		Channel:  "pico1",
 		SenderID: "user1",
 		ChatID:   "pico:test-session",
 		Content:  "hello",
@@ -4547,7 +4548,8 @@ func TestRun_PicoPublishesAssistantContentDuringToolCallsWithoutFinalDuplicate(t
 	}()
 
 	if err := msgBus.PublishInbound(context.Background(), bus.InboundMessage{
-		Channel:  "pico",
+		Context:  bus.InboundContext{Channel: "pico1", ChannelType: "pico"},
+		Channel:  "pico1",
 		SenderID: "user-1",
 		ChatID:   "session-1",
 		Content:  "run with tools",
--- a/pkg/agent/agent_test.go
+++ b/pkg/agent/agent_test.go
@@ -3911,7 +3911,8 @@ func TestProcessMessage_PicoPublishesReasoningAsThoughtMessage(t *testing.T) {
 	al := NewAgentLoop(cfg, msgBus, provider)
 
 	response, err := al.processMessage(context.Background(), bus.InboundMessage{
-		Channel:  "pico",
+		Context:  bus.InboundContext{Channel: "pico1", ChannelType: "pico"},
+		Channel:  "pico1",
 		SenderID: "user1",
 		ChatID:   "pico:test-session",
 		Content:  "hello",
@@ -4471,7 +4472,8 @@ func TestRun_PicoPublishesAssistantContentDuringToolCallsWithoutFinalDuplicate(t
 	}()
 
 	if err := msgBus.PublishInbound(context.Background(), bus.InboundMessage{
-		Channel:  "pico",
+		Context:  bus.InboundContext{Channel: "pico1", ChannelType: "pico"},
+		Channel:  "pico1",
 		SenderID: "user-1",
 		ChatID:   "session-1",
 		Content:  "run with tools",
--- a/pkg/agent/agent_utils.go
+++ b/pkg/agent/agent_utils.go
@@ -209,6 +209,9 @@ func appendEventContextFields(fields map[string]any, turnCtx *TurnContext) {
 		if inbound.Channel != "" {
 			fields["inbound_channel"] = inbound.Channel
 		}
+		if inbound.ChannelType != "" {
+			fields["inbound_channel_type"] = inbound.ChannelType
+		}
 		if inbound.Account != "" {
 			fields["inbound_account"] = inbound.Account
 		}
--- a/pkg/agent/agent_utils.go
+++ b/pkg/agent/agent_utils.go
@@ -217,6 +217,9 @@ func appendEventContextFields(fields map[string]any, turnCtx *TurnContext) {
 		if inbound.Channel != "" {
 			fields["inbound_channel"] = inbound.Channel
 		}
+		if inbound.ChannelType != "" {
+			fields["inbound_channel_type"] = inbound.ChannelType
+		}
 		if inbound.Account != "" {
 			fields["inbound_account"] = inbound.Account
 		}
--- a/pkg/agent/dispatch_request.go
+++ b/pkg/agent/dispatch_request.go
@@ -27,6 +27,13 @@ func (r DispatchRequest) Channel() string {
 	return r.InboundContext.Channel
 }
 
+func (r DispatchRequest) ChannelType() string {
+	if r.InboundContext == nil {
+		return ""
+	}
+	return r.InboundContext.ChannelType
+}
+
 func (r DispatchRequest) ChatID() string {
 	if r.InboundContext == nil {
 		return ""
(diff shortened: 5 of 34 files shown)


yinwm (changes requested):
Hi @cytown, thanks for the refactoring — decoupling channel name from provider type is the right direction.

I found **two critical issues** that need to be addressed before merging:

### 1. `processSystemMessage` loses originChannel (loop_message.go:250)

```go
// Before: parsed from ChatID
originChannel = msg.ChatID[:idx]  // e.g. "telegram"

// After: reads ChannelType
originChannel = msg.Context.ChannelType  // "" — not set on system messages!
```

System messages published at `loop_turn.go:1121` don't set `ChannelType`, so `originChannel` becomes empty. This causes the response to be published with an empty `Channel`, which gets silently dropped by `dispatchLoop` ("Unknown channel" warning). **Result: subagent async tool results are never delivered to external channels.**

### 2. `RecordLastChannel` format breaks routing in multi-instance mode (loop.go:475)

```go
// Before: ChannelName:ChatID (used for routing)
channelKey := fmt.Sprintf("%s:%s", opts.Dispatch.Channel(), ...)

// After: ChannelType:ChatID
channelKey := fmt.Sprintf("%s:%s", opts.Dispatch.ChannelType(), ...)
```

`dispatchLoop` routes by channel **name** (config key), not channel type. `heartbeat/service.go` and `devices/service.go` read this key via `parseLastChannel`, then use `NewOutboundContext(platform, ...)` which sets `Context.Channel` to the type. But `m.workers[type]` doesn't exist — workers are keyed by config name. **Result: in multi-instance scenarios (the PR's core purpose), heartbeat and device notifications are silently dropped.**

### Additional non-blocking suggestions

- `loop_turn.go:506,843,1221` — 3 calls to `IsInternalChannel(ts.channel)` still use channel name instead of type
- `ProcessDirectWithChannel` and `ProcessHeartbeat` don't set `ChannelType` on InboundContext
- `ts.channel == "pico"` hardcoded comparisons will break in multi-instance mode

alexhoshina:
I have some disagreement about which field should carry the instance identity for multi-instance support.

In #2481, the meaning of `channel` was effectively changed. Previously, `channel` carried the provider type, such as `telegram` / `slack`. During the work to support multiple instances in #2481, a separate `type` field was introduced, and the runtime meaning of `channel` started to shift toward config key / instance name. This PR now introduces `channel_type`, which is essentially adding another field after #2481 to carry the provider semantics that `channel` used to represent before.

I do not think this is the most reasonable direction. `InboundContext` was designed to be the unified carrier for these routing and identity fields. In principle, the whole runtime, including all components inside the AgentLoop flow, should rely on it to propagate address semantics. Since `InboundContext` already has `Account`, the dimension of "which concrete bot instance / which concrete channel instance" should be normalized into `Account`, instead of redefining `Channel` to mean name and then introducing `ChannelType` on top of that just to restore the original provider type.

I would much prefer the following model:

- `Channel`: provider type, such as `telegram` / `slack`
- `Account`: concrete bot instance identity, ideally a stable remote bot/account id; if some providers do not yet expose a stable id, this can temporarily fall back to the configured instance name
- `SpaceID/SpaceType`: workspace / guild / tenant dimension
- `ChatID`: concrete conversation target

That gives us a runtime address model of `(channel, account, chat_id)`. This model is semantically complete, self-consistent, and much closer to the original design of `InboundContext`. By contrast, the current direction becomes:

- `Channel = instance name`
- `ChannelType = provider type`
- `Account = another identity field`

This makes the responsibilities inside `InboundContext` more fragmented: instance semantics move into `Channel`, provider semantics move into `ChannelType`, and `Account` remains as yet another identity field. Overall, this increases the cognitive burden.

So my position is:

1. The instance identity should live in `Account`, rather than rewriting the meaning of `Channel`.
2. `Channel` should continue to represent the provider type, not the config key / instance name.
3. `channel_type` is not extending the existing context model; it is patching over the consequences of changing the meaning of `channel` in #2481.
4. Instead of introducing `channel_type`, we should normalize the semantics of `Account` across all channels. That is already a healthier and more consistent direction by itself.

For these reasons, I do not support continuing down the `Channel = name`, `ChannelType = type` path. I would rather first clarify what address model we actually want: `(channel=name, channel_type, chat_id)` or `(channel=type, account, chat_id)`. In my view, the latter is more consistent with the existing `InboundContext` design and is cleaner overall.

CC: @yinwm

cytown:
> config

All done.

cytown:
> I have some disagreement about which field should carry the instance identity for multi-instance support.
> 
> In #2481, the meaning of `channel` was effectively changed. Previously, `channel` carried the provider type, such as `telegram` / `slack`. During the work to support multiple instances in #2481, a separate `type` field was introduced, and the runtime meaning of `channel` started to shift toward config key / instance name. This PR now introduces `channel_type`, which is essentially adding another field after #2481 to carry the provider semantics that `channel` used to represent before.
> 
> I do not think this is the most reasonable direction. `InboundContext` was designed to be the unified carrier for these routing and identity fields. In principle, the whole runtime, including all components inside the AgentLoop flow, should rely on it to propagate address semantics. Since `InboundContext` already has `Account`, the dimension of "which concrete bot instance / which concrete channel instance" should be normalized into `Account`, instead of redefining `Channel` to mean name and then introducing `ChannelType` on top of that just to restore the original provider type.
> 
> I would much prefer the following model:
> 
> * `Channel`: provider type, such as `telegram` / `slack`
> * `Account`: concrete bot instance identity, ideally a stable remote bot/account id; if some providers do not yet expose a stable id, this can temporarily fall back to the configured instance name
> * `SpaceID/SpaceType`: workspace / guild / tenant dimension
> * `ChatID`: concrete conversation target
> 
> That gives us a runtime address model of `(channel, account, chat_id)`. This model is semantically complete, self-consistent, and much closer to the original design of `InboundContext`. By contrast, the current direction becomes:
> 
> * `Channel = instance name`
> * `ChannelType = provider type`
> * `Account = another identity field`
> 
> This makes the responsibilities inside `InboundContext` more fragmented: instance semantics move into `Channel`, provider semantics move into `ChannelType`, and `Account` remains as yet another identity field. Overall, this increases the cognitive burden.
> 
> So my position is:
> 
> 1. The instance identity should live in `Account`, rather than rewriting the meaning of `Channel`.
> 2. `Channel` should continue to represent the provider type, not the config key / instance name.
> 3. `channel_type` is not extending the existing context model; it is patching over the consequences of changing the meaning of `channel` in [refactor(config):  make config.Channel to multiple instance support #2481](https://github.com/sipeed/picoclaw/pull/2481).
> 4. Instead of introducing `channel_type`, we should normalize the semantics of `Account` across all channels. That is already a healthier and more consistent direction by itself.
> 
> For these reasons, I do not support continuing down the `Channel = name`, `ChannelType = type` path. I would rather first clarify what address model we actually want: `(channel=name, channel_type, chat_id)` or `(channel=type, account, chat_id)`. In my view, the latter is more consistent with the existing `InboundContext` design and is cleaner overall.
> 
> CC: @yinwm

That's a good question.  The multi-channel actually have two kind of implements, first is the PR#2481, second is to refactor channel to add an array names accounts or something like that.  I choose the first implement was because it's the minimum modifications for the whole project, the modification were limited in agent/message, for each channels, they don't need to do more modifications.  If we use the second implements, not only the agent/message need to refactor, all the channel who want to support multi-channel will also need to do huge change for it.

BTW: for account implement, you need to add account everywhere just like I add channelType, if you can look current channel is a account name, you will find there almost no differ.

alexhoshina:
@cytown Thanks for the explanation. I understand why #2481 chose the current shape: using the channel config key as the runtime channel name keeps dispatch mostly unchanged, because `m.channels` / `m.workers` are keyed by that config key.

However, after looking at this again, I still do not think `Channel = instance name` plus `ChannelType = provider type` is the right long-term model.

My understanding is that #2481 made multi-bot support work roughly like this:

- `channel_list.tg_work.type = telegram`
- `tg_work` is the config key / bot instance name
- `telegram` is the provider type
- `Manager` creates a worker keyed by `tg_work`
- runtime `InboundContext.Channel` is set to `tg_work` so outbound dispatch can route back to the same worker

That works for dispatch, but it also changes the meaning of `Channel`. Before this, `Channel` meant provider type, such as `telegram` / `slack` / `discord`. After #2481, `Channel` effectively means config key / instance name. This PR then adds `ChannelType` to recover the provider semantics that `Channel` used to carry.

I think that is the wrong split. It leaves us with:

- `Channel` = dispatch key / instance name
- `ChannelType` = provider type
- `Account` = another identity field whose responsibility becomes unclear

Using `Account` for the instance identity does not seem inherently more complex to me. This PR already has to propagate a new field everywhere: `InboundContext`, normalization, `BaseChannel`, channel constructors, `DispatchRequest`, internal-channel checks, outbound dispatch, and manually constructed contexts. The mechanical plumbing cost of `Account` and `ChannelType` is roughly the same.

The cleaner model would be:

- `Channel` = provider type, such as `telegram`
- `Account` = concrete bot/channel instance, for example `tg_work`; later this can become a stable remote bot/account id where available
- `ChatID` = concrete conversation target

Then multi-bot routing can still work:

```go
workerKey := msg.Context.Account
if workerKey == "" {
    workerKey = msg.Context.Channel // legacy fallback
}
````

So a message from `tg_work` would carry:

```go
InboundContext{
    Channel: "telegram",
    Account: "tg_work",
    ChatID:  "...",
}
```

and dispatch can still route to `m.workers["tg_work"]`.

The advantage is that provider semantics stay in `Channel`, instance identity moves into the already existing `Account` field, and checks like internal-channel detection or provider-specific behavior do not need to switch between `Channel` and `ChannelType`.

I agree this requires normalizing `Account` across channels, and there may be some existing channel-specific uses that need to move to `SpaceID/SpaceType` if they represent workspace / guild / tenant identity. But I see that as cleaning up the existing `InboundContext` model, not as extra complexity compared with adding `ChannelType`.

So my position remains: #2481 made multi-bot support work by putting instance identity into `Channel`; #2551 adds `ChannelType` to compensate for the provider type that was lost there. I would prefer to put instance identity into `Account` instead and keep `Channel` as the provider type.

That said, I do not want to make the final call alone. I would like to defer this design decision to @yinwm. If @yinwm agrees that the current `Channel = instance name`, `ChannelType = provider type` direction is acceptable, then I am okay with letting this PR proceed, and I can open a follow-up PR later to adjust or normalize the address model further.

cytown:
> @cytown Thanks for the explanation. I understand why #2481 chose the current shape: using the channel config key as the runtime channel name keeps dispatch mostly unchanged, because `m.channels` / `m.workers` are keyed by that config key.
> 
> However, after looking at this again, I still do not think `Channel = instance name` plus `ChannelType = provider type` is the right long-term model.
> 
> My understanding is that #2481 made multi-bot support work roughly like this:
> 
> * `channel_list.tg_work.type = telegram`
> * `tg_work` is the config key / bot instance name
> * `telegram` is the provider type
> * `Manager` creates a worker keyed by `tg_work`
> * runtime `InboundContext.Channel` is set to `tg_work` so outbound dispatch can route back to the same worker
> 
> That works for dispatch, but it also changes the meaning of `Channel`. Before this, `Channel` meant provider type, such as `telegram` / `slack` / `discord`. After #2481, `Channel` effectively means config key / instance name. This PR then adds `ChannelType` to recover the provider semantics that `Channel` used to carry.
> 
> I think that is the wrong split. It leaves us with:
> 
> * `Channel` = dispatch key / instance name
> * `ChannelType` = provider type
> * `Account` = another identity field whose responsibility becomes unclear
> 
> Using `Account` for the instance identity does not seem inherently more complex to me. This PR already has to propagate a new field everywhere: `InboundContext`, normalization, `BaseChannel`, channel constructors, `DispatchRequest`, internal-channel checks, outbound dispatch, and manually constructed contexts. The mechanical plumbing cost of `Account` and `ChannelType` is roughly the same.
> 
> The cleaner model would be:
> 
> * `Channel` = provider type, such as `telegram`
> * `Account` = concrete bot/channel instance, for example `tg_work`; later this can become a stable remote bot/account id where available
> * `ChatID` = concrete conversation target
> 
> Then multi-bot routing can still work:
> 
> ```go
> workerKey := msg.Context.Account
> if workerKey == "" {
>     workerKey = msg.Context.Channel // legacy fallback
> }
> ```
> 
> So a message from `tg_work` would carry:
> 
> ```go
> InboundContext{
>     Channel: "telegram",
>     Account: "tg_work",
>     ChatID:  "...",
> }
> ```
> 
> and dispatch can still route to `m.workers["tg_work"]`.
> 
> The advantage is that provider semantics stay in `Channel`, instance identity moves into the already existing `Account` field, and checks like internal-channel detection or provider-specific behavior do not need to switch between `Channel` and `ChannelType`.
> 
> I agree this requires normalizing `Account` across channels, and there may be some existing channel-specific uses that need to move to `SpaceID/SpaceType` if they represent workspace / guild / tenant identity. But I see that as cleaning up the existing `InboundContext` model, not as extra complexity compared with adding `ChannelType`.
> 
> So my position remains: #2481 made multi-bot support work by putting instance identity into `Channel`; #2551 adds `ChannelType` to compensate for the provider type that was lost there. I would prefer to put instance identity into `Account` instead and keep `Channel` as the provider type.
> 
> That said, I do not want to make the final call alone. I would like to defer this design decision to @yinwm. If @yinwm agrees that the current `Channel = instance name`, `ChannelType = provider type` direction is acceptable, then I am okay with letting this PR proceed, and I can open a follow-up PR later to adjust or normalize the address model further.

I think Account/Channel can not replace the Channel/ChannelType combine:  the account were just used for few channels: qq/slack/wecom, not for all the channels, but channel name were used for all the channels, also, the account is a runtime value, which can not be identified in all life time, also, you can not find the config of account in package config.

cytown:
pending review

cytown:
> I think we should settle the address model before continuing this PR.
> 
> The goal of this PR is valid: multi-instance channels need to preserve both the provider type and the concrete channel instance identity. However, I do not think adding `ChannelType` to `InboundContext` is the right long-term model.
> 
> My preferred canonical model is:
> 
> * `Channel`: provider type, for example `feishu`, `telegram`, `slack`, `websocket`
> * `Account`: local channel instance identity, usually the config key, for example `feishu_hr`, `feishu_coder`, `ws_ops`
> * `ChatID`: concrete conversation target
> * `SenderID`: user identity inside the conversation, when available
> * `SpaceID` / `SpaceType`: tenant, workspace, guild, team, or similar scope, when available
> * provider-specific remote IDs such as `bot_id`, `app_id`, `tenant_key`, `team_id` should stay in provider metadata / raw fields / aliases unless we decide to normalize them explicitly
> 
> The key clarification is that `Account` should not mean only a remote `bot_id`. It should be the stable local routing identity for one configured channel instance. Some providers have a natural remote bot/account ID, some do not, and websocket/custom channels may require the user to define the instance identity. That is fine: `Account` can still be the config-backed instance name.
> 
> For example:
> 
> ```go
> // Feishu HR bot
> Channel = "feishu"
> Account = "feishu_hr"
> ChatID  = "oc_xxx"
> // raw/provider metadata may contain bot_id/app_id/tenant_key
> 
> // Feishu coder bot in the same provider family
> Channel = "feishu"
> Account = "feishu_coder"
> ChatID  = "oc_xxx"
> 
> // User-defined websocket instance
> Channel = "websocket"
> Account = "ws_ops"
> ChatID  = "client_123"
> ```
> 
> This also handles the case where two different bot instances are present in the same external channel/group: the provider family is still `Channel`, while the selected configured instance is `Account`.
> 
> For outbound dispatch, we can keep a compatibility path:
> 
> ```go
> workerKey := msg.Context.Account
> if workerKey == "" {
>     workerKey = msg.Context.Channel // legacy fallback
> }
> ```
> 
> So the migration can be incremental:
> 
> 1. Keep `Channel` as the provider type in normalized runtime context.
> 2. Populate `Account` with the configured channel instance name / config key.
> 3. Dispatch by `Account` first, and fall back to `Channel` for legacy configs.
> 4. Normalize adapters gradually so all channel implementations set `Account` consistently.
> 5. Avoid introducing `ChannelType` unless we find a separate concept that cannot be represented by `Channel` + `Account`.
> 
> My concern with the current `Channel = instance name` plus `ChannelType = provider type` approach is that it splits one address model across two fields and makes `InboundContext` harder to reason about. The concrete routing issues already found in this PR, such as system messages losing `originChannel` or heartbeat/device notifications being recorded under provider type but dispatched by channel name, look like symptoms of that split.
> 
> Therefore I would suggest changing the direction of this PR from "add `ChannelType`" to "normalize `Account` as the channel instance identity". If that is too large for this PR, I think it is better to pause/close this one and open a smaller follow-up PR focused on the `Channel` / `Account` migration path.

That's a good advise, but it's not suit for current situation.  The channel_list in config was already refactor to make key as channelName and add channelType for specified type, if we use channel/account combine, we need to refactor the config.json again to make key as account and channelType as channel, that's cause more issues than what I did in this commit.

Actually the channel/channelType combine vs account/channel combine has no difference in deed, and the previous wont affect current account field in few channel, otherwise it's must change to another name, because current account contains runtime info.

alexhoshina:
I agree with @yinwm — I believe there is no need to modify the config itself; only the mapping between the config and the runtime needs to be changed.

This approach is beneficial for long-term maintainability. The key point is that the `Account` field in the current runtime is deprecated and is not actually being used anywhere. Even if we choose to introduce `channel type` now, we will still need to clean up `Account` later, and that cleanup effort will inevitably circle back to the current topic. This is unavoidable.

cytown:
> Thanks, that makes sense from a migration-cost perspective. I agree we should not refactor `config.json` again if it is not necessary.
> 
> But I think we can separate the external config schema from the normalized runtime model.
> 
> The config can stay as it is today:
> 
> ```json
> "channel_list": {
>   "feishu_hr": {
>     "type": "feishu"
>   }
> }
> ```
> 
> At runtime, we can map it into the normalized context as:
> 
> ```go
> Channel = channelConfig.Type // "feishu"
> Account = channelName        // "feishu_hr"
> ```
> 
> So this does not require changing `channel_list` again. The config key remains the configured channel instance name, and `type` remains the provider type. The only change is how we populate the normalized `InboundContext`.
> 
> I also agree that the current `Account` usage in a few channel implementations may already contain runtime/provider-specific information. But I would treat that as the thing we need to clean up, because `InboundContext.Account` is already part of the normalized routing/session model. If a channel needs provider-specific account/bot IDs, those can go into `Raw`, aliases, or a more specific field later. They should not block `Account` from meaning the stable local channel instance identity.
> 
> So the migration can still be incremental:
> 
> 1. Keep config unchanged.
> 2. When constructing inbound context, set:
>    
>    * `Channel = config.type`
>    * `Account = config key / channel name`
> 3. For outbound dispatch, route by `Account` first and fall back to `Channel` for legacy behavior.
> 4. Audit existing channel-specific `Account` usage and move provider runtime values to raw/provider metadata where needed.
> 
> That gives us the same implementation compatibility as `Channel + ChannelType`, but keeps the normalized address model simpler and avoids making `ChannelType` a long-term public field.

That requires 2 previous condition:

1. Someone must cleanup current Account field.
2. Developers will recognize `Channel = config.type` and `Account = config key / channel name`

Why don't we just keep it simple...

cytown:
> I agree with @yinwm — I believe there is no need to modify the config itself; only the mapping between the config and the runtime needs to be changed.
> 
> This approach is beneficial for long-term maintainability. The key point is that the `Account` field in the current runtime is deprecated and is not actually being used anywhere. Even if we choose to introduce `channel type` now, we will still need to clean up `Account` later, and that cleanup effort will inevitably circle back to the current topic. This is unavoidable.

Actually the channel.type was already exists about a month, not new introduce...

cytown:
continue.

PR #2563, opened 2026-04-16
feat(web): download files on frontend

afjcjsbx:
## 📝 Description

This PR adds file attachment support to the `pico` web channel so files sent through tools such as `send_file` can be downloaded directly from the UI.

The change covers the full flow end-to-end: the backend now emits structured attachment metadata, exposes authenticated media download endpoints, persists attachment information in session history, and the frontend renders downloadable file cards and inline image previews. This removes the current limitation where Pico could only return the local file path or paste the content in chat.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue

Closes the Pico Web limitation where `send_file` outputs could not be downloaded from the UI.

- **Reference URL:** N/A
- **Reasoning:** Pico previously did not support attachment delivery in the frontend channel, so tool outputs such as generated files were not usable from the web UI. This PR introduces attachment-aware message payloads, authenticated same-origin download proxying, history persistence, and frontend rendering for files and images.

## 🧪 Test Environment
- **Hardware:** PC
- **OS:** macOS
- **Model/Provider:** N/A
- **Channels:** Pico (Web UI)

## 📸 Evidence (Optional)
<details>

</details>

## ☑️ Checklist


pkg/agent/loop.go
```diff
@@ -1455,6 +1455,24 @@ func inferMediaType(filename, contentType string) string {
 	return "file"
 }
 
+func buildProviderAttachments(store media.MediaStore, refs []string) []providers.Attachment {
```
wj-xiao: Persisting only `media://...` refs here makes session history depend on the in-memory `MediaStore`. After a gateway restart, or once media cleanup evicts the ref, reopened sessions may still show attachment links even though `/pico/media/...` can no longer resolve them.

Should we decide explicitly what behavior we want for handled tool attachments in session history?
If these attachments are expected to remain available when a past session is reopened, we likely need some form of durable storage rather than ephemeral media refs.
If we only want to guarantee attachment availability in the live session, it may be better not to surface them from persisted history, or to make that limitation explicit in the session API/UI.

afjcjsbx: Good catch @wj-xiao! I made the behavior explicit in this PR: persisted session history now only exposes durable attachment URLs.

Concretely, we no longer reconstruct `/pico/media/...` links from persisted `media://...` refs when reopening session history, because those refs depend on the in-memory MediaStore and may stop resolving after a gateway restart or cleanup. So handled tool attachments are effectively live-session-only unless they were persisted with a durable URL.

If we want reopened sessions to preserve handled tool attachments in the future, that would need a separate durable-storage design rather than ephemeral MediaStore refs.

wj-xiao (approved):

PR #2573, opened 2026-04-17
fix(web): keep launcher locale changes from mutating shared web-search routing

### LINKED ISSUE #2572: [BUG] Launcher UI language changes leak into backend web_search auto provider selection
## Quick Summary

Changing the launcher UI language currently mutates backend `web_search` auto-provider selection for the whole process, not just for the client that changed the language.

When both `sogou` and `duckduckgo` are enabled under `provider=auto`, a `POST /api/ui/language` request can change the reported `current_service` and the ambiguous-query fallback behavior seen by other launcher contexts in the same process.

## Environment & Tools
- **PicoClaw Version:** `main` at `9fe678247f13b42b02b256d5db622c0c9b4ced29`
- **Go Version:** go 1.25+
- **AI Model & Provider:** N/A
- **Operating System:** macOS (local validation), but the behavior is process-scoped rather than OS-specific
- **Channels:** Launcher web console

## 📸 Steps to Reproduce
1. Start from a config with `tools.web.provider = "auto"` and both `sogou` + `duckduckgo` enabled.
2. Create two independent launcher handlers / contexts in the same process (I validated this with two handlers backed by different temp config files in a regression test).
3. Query `/api/tools/web-search-config` from handler B and observe `current_service`.
4. Send `POST /api/ui/language` with `{"language":"en"}` to handler A.
5. Query `/api/tools/web-search-config` from handler B again.
6. Repeat with `{"language":"zh"}`.

## ❌ Actual Behavior

A language change in handler A changes handler B's backend-visible `current_service` (`sogou` ↔ `duckduckgo`) even though handler B did not change its own config.

The root cause is that:
- `pkg/tools/web.go` stores preferred language in a package-global value,
- `/api/ui/language` mutates that global,
- `resolveCurrentWebSearchProvider()` and ambiguous-query auto selection both read from that shared state.

## ✅ Expected Behavior

Launcher UI language changes should not mutate backend auto-provider routing for unrelated handlers/clients.

If backend routing needs a language hint, it should come from explicit config or request-scoped state rather than a process-global singleton updated by whichever launcher client changed locale last.

## 💬 Additional Context

Why I think this is a real bug instead of intended behavior:
- PicoClaw already documents explicit isolation concepts for sessions, channels, accounts, and routing.
- The launcher also supports remote / Docker / VM workflows, so backend state leaking across contexts is not just theoretical.
- I am **not** reporting the previously discussed network-fallback behavior here; that path already has upstream issue/PR history and appears intentional.

I have a small fix ready that keeps the scope narrow:
- remove the backend sync on UI locale changes,
- drop the now-unused `/api/ui/language` endpoint,
- make auto-selection fall back to a stable default when the query itself does not provide a script hint.

shaun0927:
## 📝 Description

This PR fixes a validated launcher bug where UI language changes mutate backend `web_search` auto-provider routing for the whole process.

Before this change, the launcher sent every locale change to `/api/ui/language`, which updated a package-global language value in `pkg/tools/web.go`. That same global was then used to:
- resolve `current_service` for `/api/tools/web-search-config`, and
- choose `duckduckgo` vs `sogou` for ambiguous `provider=auto` queries.

As a result, one launcher context could change another context's backend-visible routing behavior without changing its config.

This patch keeps the fix intentionally narrow:
- remove backend sync on launcher locale changes,
- drop the now-unused `/api/ui/language` endpoint and tests,
- keep auto-selection query-driven, and
- use a stable default (`sogou`) when the query itself does not provide a script hint.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue
Fixes #2572

- **Reference URL:** #2572
- **Reasoning:** The previous implementation coupled launcher locale toggles to a process-global backend preference. That made auto provider resolution depend on whichever client changed UI language last. This PR removes that cross-context mutation without redesigning per-user preference storage.

## 🧪 Test Environment
- **Hardware:** Apple Silicon Mac local development machine
- **OS:** macOS 15.x
- **Model/Provider:** N/A, launcher web console / web_search config only
- **Channels:** Launcher web console

## 📸 Evidence (Optional)
<details>

```text
go test ./web/backend/api ./pkg/tools -count=1
ok   github.com/sipeed/picoclaw/web/backend/api   15.138s
ok   github.com/sipeed/picoclaw/pkg/tools         16.631s
```

```text
cd web/frontend && pnpm lint
# completed successfully
```

```text
cd web/frontend && pnpm build
✓ built in 5.17s
```

</details>

## ☑️ Checklist
- [x] Documentation updates are not required for this bug fix.


wj-xiao:
@shaun0927 Thank you for your contribution. There are currently some file conflicts in this PR. Would you mind resolving them first?

shaun0927:
@wj-xiao Rebased onto the latest main and resolved the conflicts. The core fix is unchanged; the conflict area mostly overlapped with #2586's `ResolveWebSearchProviderName` refactor, so I dropped my inline `resolveCurrentWebSearchProvider` rewrite and kept yours, since with the preferred-language global removed `prefersDuckDuckGoQuery("")` now returns false and the helper already yields the stable default. Locally `go test ./web/backend/api ./pkg/tools ./pkg/tools/integration` passes.

wj-xiao (approved):

PR #2581, opened 2026-04-18
Recover Codex output from streamed message events

astrada-c:
## Summary
- capture `response.output_item.done` items while reading the Codex streaming response
- rebuild `response.output` from those streamed items when the completed payload is empty
- add regression tests covering the fallback path and the existing completed-output path

## Why
The ChatGPT Codex OAuth backend can return assistant text in streamed `response.output_item.done` events while leaving the final completed `response.output` array empty. In that case the provider currently drops valid text and downstream callers see an empty assistant response.

This patch keeps the normal completed-response path unchanged and only falls back to streamed items when:
- a completed/incomplete/failed response event exists
- `response.output` is empty
- streamed output items were actually observed

## Validation
- `go test ./pkg/providers/oauth/...`

## Repro context
I hit this in a live Dockerized PicoClaw runtime using the Codex OAuth backend. The stream included the assistant text in `response.output_item.done`, but the final completed payload carried `output: []`, which caused the caller to surface an empty model response. This patch fixes that case without changing behavior for completed responses that already include populated output.

### DIFF
--- a/pkg/providers/oauth/codex_provider.go
+++ b/pkg/providers/oauth/codex_provider.go
@@ -104,8 +104,12 @@ func (p *CodexProvider) Chat(
 	defer stream.Close()
 
 	var resp *responses.Response
+	var outputItems []responses.ResponseOutputItemUnion
 	for stream.Next() {
 		evt := stream.Current()
+		if evt.Type == "response.output_item.done" {
+			outputItems = append(outputItems, evt.Item)
+		}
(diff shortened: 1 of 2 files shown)


astrada-c:
Still interested in getting this reviewed/merged. This fixes Codex streamed output recovery and includes tests. Happy to adjust if maintainers prefer a different fallback shape.

PR #2626, opened 2026-04-22
feat(agent): support native audio input for multimodal LLMs

webhtb:
This PR adds support for native audio input in multimodal LLMs (like Gemini 1.5).

Key changes:
- Added 'Audio' field to 'protocoltypes.Message'.
- Updated 'loop_media.go' to detect audio MIME types and encode them as data URLs in the 'Audio' field instead of just providing a path tag.
- Updated 'SerializeMessages' in 'common.go' to handle 'input_audio' type for OpenAI-compatible providers.

This allows the agent to send audio files directly to models that support it, enabling native audio understanding (intonation, background noise, etc.) instead of relying solely on text transcriptions.


webhtb:
Still here! I've resolved the conflicts locally and I'm ready for the native audio support to be integrated. This is a crucial feature for multimodal models like Gemini 1.5 Pro.

webhtb:
Updated the branch to resolve merge conflicts with the latest master. 

Native audio support is essential for modern multimodal LLMs (like GPT-4o, Gemini 1.5, and various local models). By sending the raw audio stream instead of just text transcriptions, the agent can leverage the model's ability to understand prosody, tone, and environmental context, while also reducing latency by bypassing an extra STT step where the provider supports it natively.

Ready for review!

PR #2641, opened 2026-04-23
  feat(mcp): add show, add, list, remove, test, edit cli commands

afjcjsbx:
## 📝 Description

This PR introduces a comprehensive **MCP (Model Context Protocol) CLI management suite**. Users can now manage their MCP server configurations directly from the terminal without manual JSON editing.

The new command group `picoclaw mcp` provides a full CRUD interface for MCP servers, including live connectivity testing and detailed inspection of exposed tools using a rich, terminal-optimized UI.

## 🗣️ Type of Change

  - [x] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue

Fixes # (Add issue number here if applicable)

  - **Reference URL:** [[Model Context Protocol Specification](https://modelcontextprotocol.io/)](https://modelcontextprotocol.io/)
  - **Reasoning:** Managing MCP servers via JSON is error-prone. This CLI ensures atomic writes, validates the configuration against a schema before saving, and provides immediate feedback on server reachability.

## 🧪 Test Environment

  - **Hardware:** PC
  - **OS:** Ubuntu 22.04 / macOS
  - **Model/Provider:** N/A (CLI logic)
  - **Channels:** CLI

## 📸 Evidence (Optional)

<details>
<summary>Click to view CLI Commands & UI Examples</summary>

### New Commands Added:

  * `picoclaw mcp add`: Supports `stdio`, `http`, and `sse` transports with `--env` and `--header` flags.
  * `picoclaw mcp list`: Displays a summary of configured servers (Fancy UI on wide terminals, ASCII table on narrow ones).
  * `picoclaw mcp show <name>`: Connects to the server and renders a detailed view of all available tools and their parameters.
  * `picoclaw mcp test <name>`: Quick reachability check.
  * `picoclaw mcp remove`: Deletes a server entry.
  * `picoclaw mcp edit`: Opens the config file in the system `$EDITOR`.

image

image

</details>

## ☑️ Checklist

  - [x] I have updated the documentation accordingly (Updated README.md, added `mcp-cli.md`).

### DIFF
--- a/cmd/picoclaw/internal/mcp/add.go
+++ b/cmd/picoclaw/internal/mcp/add.go
@@ -223,6 +223,9 @@ func buildServerConfig(target string, args []string, opts addOptions) (config.MC
 	if err := validateLocalCommandPath(target); err != nil {
 		return config.MCPServerConfig{}, err
 	}
+	if isLocalCommandPath(command) {
+		command = expandHomePath(command)
+	}
 
 	server.Command = command
 	server.Args = commandArgs
--- a/cmd/picoclaw/internal/mcp/add.go
+++ b/cmd/picoclaw/internal/mcp/add.go
@@ -12,6 +12,7 @@ import (
 
 type addOptions struct {
 	Env       []string
+	EnvFile   string
 	Headers   []string
 	Transport string
 	Force     bool
@@ -70,7 +71,8 @@ func newAddCommand() *cobra.Command {
 	}
 
 	flags := cmd.Flags()
-	flags.StringArrayP("env", "e", nil, "Environment variable in KEY=value format (repeatable)")
+	flags.StringArrayP("env", "e", nil, "Environment variable in KEY=value format (repeatable, saved to config)")
+	flags.String("env-file", "", "Path to an env file for stdio servers (recommended for secrets)")
 	flags.StringArrayP("header", "H", nil, "HTTP header in 'Name: Value' or 'Name=Value' format (repeatable)")
 	flags.StringP("transport", "t", "stdio", "Transport type: stdio, http, or sse")
 	flags.BoolP("force", "f", false, "Overwrite an existing server without prompting")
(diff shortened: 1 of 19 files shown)


cmd/picoclaw/internal/mcp/add.go
```diff
@@ -0,0 +1,232 @@
...
+			}
+
+			fmt.Fprintf(cmd.OutOrStdout(), "✓ MCP server %q saved.\n", name)
+			return nil
+		},
+	}
+
+	flags := cmd.Flags()
+	flags.StringArrayP("env", "e", nil, "Environment variable in KEY=value format (repeatable)")
+	flags.StringArrayP("header", "H", nil, "HTTP header in 'Name: Value' or 'Name=Value' format (repeatable)")
```
alexhoshina: nit: Would copying the token from env directly into config cause keys that users do not want to be saved in the config file to be saved?

cmd/picoclaw/internal/mcp/add.go
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	command := target
+	commandArgs := append([]string(nil), args...)
+
+	if err := validateLocalCommandPath(target); err != nil {
+		return config.MCPServerConfig{}, err
+	}
+
+	server.Command = command
```
alexhoshina, line 243: If a user adds a local server as `~/bin/my-mcp`, `validateLocalCommandPath` succeeds because it expands `~` for `os.Stat`, but the saved `server.Command` is still the original `~/...` string. `pkg/mcp/manager.ConnectServer` later executes that value with `exec.CommandContext`, which does not perform **shell expansion**, so every `mcp test`/`gateway start` for such servers fails with no such file or directory.

afjcjsbx: Good catch, thanks @alexhoshina. I fixed this by normalizing local stdio command paths before saving the server config, so `~/...` is now expanded and persisted as an absolute path instead of the raw shell-style string.

I also added a small runtime safeguard in the MCP manager to expand `~` when launching stdio servers, so existing handwritten or previously saved configs keep working too.

Added also tests for both cases.

alexhoshina (approved):
LGTM

cmd/picoclaw/internal/mcp/add.go
afjcjsbx: Good point. I addressed this in two ways:
- clarified the `--env` help text to make it explicit that values are saved to config
- added `--env-file` for stdio MCP servers, which is the safer path for secrets that should stay outside the main config

PR #2656, opened 2026-04-24
Prompt layering

alexhoshina:
## Overview

This PR introduces an internal structured layering mechanism for agent prompts, unifying the originally directly concatenated prompt sources such as system prompts, turn messages, skills, tool discovery, and MCP into prompt parts with `layer`, `slot`, and `source` metadata.

The prompt format visible to the model remains compatible: the final output is still sent as plain text, without introducing XML or new model-visible boundaries. The structured information is used only within the agent runtime to distinguish sources, constrain write positions, support caching, and facilitate future prompt assembly evolution.

## Rationale

Currently, the sources of prompts are gradually increasing, including workspace/bootstrap, memory, skill catalog, active skill, tool discovery, MCP server, SubTurn overlay, steering, interrupt, etc. Continuing to use plain string concatenation will blur boundaries and ownership, making it difficult to determine where a certain prompt segment comes from, which layer it belongs to, and whether it is allowed to be written by a certain source during subsequent loop rotation, caching, hook modification, and debugging.

This PR adopts a relatively conservative five-layer structure: `kernel`, `instruction`, `capability`, `context`, `turn`. Specific capability sources are extended through slots, such as `tooling`, `mcp`, `skill_catalog`, `active_skill`, to avoid prematurely solidifying unpredictable future sources into more top-level layers.

## Changes

- Added structures such as `PromptPart`, `PromptRegistry`, and `PromptContributor`, using `layer / slot / source` to describe prompt segments, and verifying whether the source writes to the allowed position.
- Use `ContextBuilder` as the registration entry for prompt sources/contributors; unregistered sources are allowed through in compatibility mode, but a warning will be logged.
- Change the static system prompt to first build structured parts, then render them into the old text format, preserving existing provider compatibility and caching strategies.
- Include active skill, SubTurn system prompt, user message, steering, SubTurn result, and interrupt hint in the internal prompt metadata.
- Migrate tool discovery to a contributor under the `capability/tooling` slot.
- Migrate the MCP server capability prompt to a contributor under the `capability/mcp` slot, and register it for the agent during MCP initialization.
- Add internal prompt metadata for provider tool definitions; regular tools default to `capability/tooling`, MCP tools belong to `capability/mcp`, and these fields will not be included in the provider JSON.
- Restrict the `BeforeLLM` modification capability of hooks: hooks cannot directly modify the system prompt, nor can they directly modify tool/MCP definitions; when attempting to modify, retain the original content and log a warning.
- Add tests related to structured prompts, MCP/tool metadata, and hook mutation guards.

### DIFF
--- a/pkg/agent/context.go
+++ b/pkg/agent/context.go
@@ -22,13 +22,11 @@ import (
 )
 
 type ContextBuilder struct {
-	workspace          string
-	skillsLoader       *skills.SkillsLoader
-	memory             *MemoryStore
-	toolDiscoveryBM25  bool
-	toolDiscoveryRegex bool
-	splitOnMarker      bool
-	promptRegistry     *PromptRegistry
+	workspace      string
+	skillsLoader   *skills.SkillsLoader
+	memory         *MemoryStore
+	splitOnMarker  bool
+	promptRegistry *PromptRegistry
 
 	// Cache for system prompt to avoid rebuilding on every call.
 	// This fixes issue #607: repeated reprocessing of the entire context.
@@ -50,8 +48,16 @@ type ContextBuilder struct {
 }
 
 func (cb *ContextBuilder) WithToolDiscovery(useBM25, useRegex bool) *ContextBuilder {
-	cb.toolDiscoveryBM25 = useBM25
-	cb.toolDiscoveryRegex = useRegex
+	if useBM25 || useRegex {
+		if err := cb.RegisterPromptContributor(toolDiscoveryPromptContributor{
+			useBM25:  useBM25,
+			useRegex: useRegex,
+		}); err != nil {
+			logger.WarnCF("agent", "Failed to register tool discovery prompt contributor", map[string]any{
+				"error": err.Error(),
+			})
+		}
+	}
 	return cb
 }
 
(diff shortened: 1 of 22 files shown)


pkg/agent/hooks.go
```diff
@@ -367,6 +369,58 @@ func (hm *HookManager) AfterLLM(ctx context.Context, resp *LLMHookResponse) (*LL
 	return current, HookDecision{Action: HookActionContinue}
 }
 
+func (hm *HookManager) applyBeforeLLMControls(
```
afjcjsbx, line 372: `applyBeforeLLMControls()` checks that system messages have not changed using `reflect.DeepEqual` on complete messages. With this PR however `Message` / `ContentBlock` now include internal metadata (`PromptLayer` / `PromptSlot` / `PromptSource`) marked `json:"-"` in `pkg/providers/protocoltypes/types.go`. This can lead to discarding legitimate modifications of process hooks

The problem is that process hooks do round-trip JSON of `LLMHookRequest`, so those fields are lost even when the system prompt is not actually changed. If the process hook changes only lawful parts of the request (e.g., adds/changes a non-system message), `llmHookSystemMessagesUnchanged()` can still return `false`, and the code ends up restoring `next.Messages = current.Messages`, discarding the valid change as well.

The added tests cover only in-process hooks based on `req.Clone()`, so this regression on process hooks is not covered today.

pkg/agent/context.go
```diff
@@ -548,6 +650,20 @@ func (cb *ContextBuilder) BuildMessages(
...
+		Media:             media,
+		Channel:           channel,
+		ChatID:            chatID,
+		SenderID:          senderID,
+		SenderDisplayName: senderDisplayName,
+		ActiveSkills:      append([]string(nil), activeSkills...),
+	})
+}
+
+func (cb *ContextBuilder) BuildMessagesFromPrompt(req PromptBuildRequest) []providers.Message {
```
afjcjsbx, line 666: The new `PromptPart.Cache` is never used: in `BuildMessagesFromPrompt() `contributors/overlays are always issued with `promptContentBlock(..., nil)`, so `PromptCacheEphemeral` remains dead code. In practice the tool discovery rule, which used to live in the cacheable static block, now ends up in an uncached block; same for MCP contributors. It does not break correctness, but it is a concrete cost/performance regression from the refactor goal.

alexhoshina: `PromptPart.Cache` now maps to `CacheControl`, and `PromptCacheEphemeral` actually enters `SystemParts`.

pkg/agent/hooks.go
alexhoshina: **hook**'s system message/tool definition comparison now ignores internal prompt metadata, preventing legitimate non-system message modifications from being misjudged and discarded after process hook JSON round-trips.

afjcjsbx (approved):
LGTM!

PR #2657, opened 2026-04-24
fix(reasoning): persist canonical history for DeepSeek and web chat

### LINKED ISSUE #2615: [BUG] The messages displayed in Web Chat is inconsistent before and after refreshing
## Quick Summary
The tool call summary disappears after the page is refreshed, resulting in a significant inconsistency in the information displayed before and after the refresh.
This issue is essentially the same as #2427, which is fixed by #2449, and it has reappeared now.

## Environment & Tools
- **PicoClaw Version:** 71c877a67fa69c3b8c782144e4d0ff4893159929
- **Go Version:** 1.25.9
- **AI Model & Provider:** Any
- **Operating System:** Ubuntu Server 24
- **Channels:** Web

## 📸 Steps to Reproduce
1. Trigger tool calls during conversations in Web Chat.
2. Everything was observed to be normal.
3. Tool call summaries are lost after refreshing.

## ❌ Actual Behavior
The tool call summary disappears after refreshing.

## ✅ Expected Behavior
The information displayed before and after the refresh should be consistent.

## 💬 Additional Context
This issue is essentially the same as #2427, which is fixed by #2449, and it reappeared after being fixed.

### LINKED ISSUE #2648: [BUG]DeepSeek returns 400 due to misordered reasoning content in history after tool calls
Quick Summary
When using a DeepSeek model with PicoClaw, after a tool call completes and the assistant produces a final reply, the conversation history retains the model's internal thinking (reasoning) content, and the positions of the final reply content and the thinking content within the assistant message(s) are malformed. This leads to a 400 Bad Request error from the DeepSeek API on subsequent requests.

Image

Image

## Environment & Tools
- **PicoClaw Version:** ( latest commit)
- **Go Version:** (go 1.26)
- **AI Model & Provider:** DeepSeek (deepseek-v4-flash) via official DeepSeek API
- **Operating System:** (Ubuntu 22.04)
- **Channels:** any (issue occurs at the AI adapter level)

## 📸 Steps to Reproduce
1. Start a conversation and trigger a tool call through the assistant (e.g., ask a question that requires function calling).
2. Allow the assistant to execute the tool and produce the final answer.
3. Inspect the internally stored conversation history after the turn is complete – note that the thinking/reasoning blocks are retained and **their placement relative to the final `content` of the assistant message is abnormal** (e.g., mixed together, or placed in the wrong field).
4. Send another follow‑up message to continue the conversation.
5. The request to the DeepSeek API now includes the malformed history, triggering a 400 error.

## ❌ Actual Behavior
The DeepSeek API returns a **400 Bad Request** error, typically indicating an invalid message structure. The root cause is twofold:
1. Thinking (reasoning) content from the earlier assistant turn is not cleaned up and remains in the history.
2. The **positions of the final reply content and the thinking content are abnormal**: e.g., they may be merged into a single `content` field, or the thinking tokens appear in the wrong part of the message object, violating the expected schema (where assistants should output *either* a normal `content` *or* a `tool_calls` block with an empty/null `content`, and reasoning must be placed in a dedicated `reasoning_content` field where supported).

## ✅ Expected Behavior
After a tool call is resolved and the assistant delivers its final user‑visible response, the conversation history sent to DeepSeek should:
- Strip out any internal reasoning/thinking blocks that belong to the previous assistant turn, **or** correctly map them to the `reasoning_content` field (if the model/provider supports it).
- Ensure the final reply `content` occupies the correct field and does not leak into other fields or mix with the reasoning text.
- Maintain a message array that conforms to the DeepSeek API spec so that conversations continue without 400 errors.

## 💬 Additional Context
- DeepSeek models (especially `deepseek-reasoner`) may emit `reasoning_content` during streaming, and PicoClaw appears to collect and concatenate these chunks into the assistant message incorrectly, leading to **positional anomalies** between the thinking and the final reply. The thinking text might end up in the `content` field together with the final answer, or appear in an extra assistant message that precedes or follows the actual answer, tipping the structure out of spec.
- A potential fix: during history construction for subsequent API calls, the adapter should separate or discard reasoning segments from the previous assistant turn, and only retain the polished final `content` (and any `tool_calls` information in the correct format). If the target model endpoint expects reasoning to be presented separately, it should be placed in the dedicated `reasoning_content` field **of the assistant message that carries the final response**, not mixed into `content`.
- This may also require careful handling of multi‑turn tool call sequences, where the assistant sends a tool request (no `content`), then the tool result is injected, and finally the assistant replies with textual `content` – any reasoning emitted before or during that final reply must not interfere.

### LINKED ISSUE #2650: [BUG] DeepSeek-V4-Flash开启推理时，调用skill、exec之类的工具之后会报错
## Quick Summary
DeepSeek-V4-Flash开启推理时，调用skill、exec之类的工具之后会报错，清空会话/删除会话文件后重新对话可以继续，但再调用skill会继续触发。或者关闭推理可以继续对话。
##  Environment & Tools
- **PicoClaw Version:** (0.2.7)
- **Go Version:** (e.g., go 1.25.9)
- **AI Model & Provider:** (deepseek-v4-flash 默认开启推理)
- **Operating System:** (fedora 34)
- **Channels:** (Feishu)

## 📸 Steps to Reproduce
1. 新开对话
2. 查天气
3. 再次对话，报错：
```
Status: 400
Body:   {"error":{"message":"The reasoning_content in the thinking mode must be passed back to the API.","type":"invalid_request_error...
```

lc6464:
## 📝 Description

This PR fixes Web Chat refresh inconsistencies and DeepSeek thinking-mode follow-up failures by making assistant history persist and replay in a canonical shape.

It does this by:
- persisting assistant `reasoning_content` together with the corresponding assistant `content` / `tool_calls` in session history instead of dropping it after the turn;
- normalizing OpenAI-compatible request history so non-DeepSeek providers do not replay `reasoning_content`, while DeepSeek providers keep it only for assistant turns that belong to tool-interaction rounds;
- propagating the concrete provider name through the HTTP provider stack so DeepSeek-specific history handling can be applied reliably;
- rebuilding Web session detail responses from canonical assistant records, including assistant thought messages and tool-call summaries, and returning assistant `kind` so the frontend can render refreshed history consistently;
- skipping transient reasoning-only assistant artifacts when reconstructing detail history or preparing upstream requests;
- adding regression coverage for direct replies, tool-call turns, DeepSeek/non-DeepSeek history canonicalization, and Web refresh reconstruction.

This addresses the Web refresh regression from #2615 and the DeepSeek 400 errors from #2648 and #2650 caused by malformed or incomplete replay of thinking-mode history.

During testing, we found that the summary of the `tool_calls` response is modified by `message.update`, but this is not a bug introduced by this PR, so we will not fix it for now.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue
Fixes #2615
Fixes #2648
Fixes #2650

## 📚 Technical Context
- **Reference URL:** https://api-docs.deepseek.com/zh-cn/guides/thinking_mode, https://api-docs.deepseek.com/zh-cn/guides/multi_round_chat
- **Reasoning:** DeepSeek thinking mode still uses a stateless `/chat/completions` flow, so the client must persist and replay prior assistant turns itself. The replay rules differ from generic OpenAI-compatible providers: `reasoning_content` should not be replayed universally, but DeepSeek thinking-mode tool rounds require the matching reasoning history to be passed back in canonical assistant messages. This PR keeps the persisted assistant turn shape consistent, limits reasoning replay to the DeepSeek-specific cases, and uses the same canonical records to rebuild Web Chat detail output after refresh.

## 🧪 Test Environment
- **OS:** Ubuntu Server 24
- **Model/Provider:** DeepSeek V4 Flash
- **Channels:** Web

## 📸 Evidence
<details>

image

</details>

## ☑️ Checklist

### DIFF
--- a/pkg/memory/jsonl.go
+++ b/pkg/memory/jsonl.go
@@ -10,6 +10,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 	"sync"
 	"time"
@@ -406,11 +407,8 @@ func (s *JSONLStore) promoteAliasHistoryLocked(
 }
 
 func (s *JSONLStore) sessionHasVisibleContentLocked(sessionKey string, meta SessionMeta) (bool, error) {
-	if meta.Count-meta.Skip > 0 || strings.TrimSpace(meta.Summary) != "" {
+	if strings.TrimSpace(meta.Summary) != "" {
 		return true, nil
-	}
-	if meta.Count != 0 || meta.Skip != 0 {
-		return false, nil
 	}
 	history, err := readMessages(s.jsonlPath(sessionKey), meta.Skip)
 	if err != nil {
@@ -498,28 +496,44 @@ func readMessages(path string, skip int) ([]providers.Message, error) {
 	return msgs, nil
 }
 
-// countLines counts the total number of non-empty lines in a .jsonl file.
-// Used by TruncateHistory to reconcile a stale meta.Count without
-// the overhead of unmarshaling every message.
-func countLines(path string) (int, error) {
+// scanRetainedMessageLines returns the total number of non-empty raw JSONL
+// lines plus the raw line numbers that survive readMessages filtering.
+// TruncateHistory uses this to compute keepLast against retained messages
+// while preserving the raw-line skip offset stored in metadata.
+func scanRetainedMessageLines(path string) (int, []int, error) {
 	f, err := os.Open(path)
 	if os.IsNotExist(err) {
-		return 0, nil
+		return 0, []int{}, nil
 	}
 	if err != nil {
-		return 0, fmt.Errorf("memory: open jsonl: %w", err)
+		return 0, nil, fmt.Errorf("memory: open jsonl: %w", err)
 	}
 	defer f.Close()
 
-	n := 0
+	rawCount := 0
+	retained := make([]int, 0)
 	scanner := bufio.NewScanner(f)
 	scanner.Buffer(make([]byte, 0, 64*1024), maxLineSize)
 	for scanner.Scan() {
-		if len(scanner.Bytes()) > 0 {
-			n++
+		line := scanner.Bytes()
+		if len(line) == 0 {
+			continue
 		}
+		rawCount++
+
+		var msg providers.Message
+		if err := json.Unmarshal(line, &msg); err != nil {
+			continue
+		}
+		if messageutil.IsTransientAssistantThoughtMessage(msg) {
+			continue
+		}
+		retained = append(retained, rawCount)
 	}
-	return n, scanner.Err()
+	if err := scanner.Err(); err != nil {
+		return 0, nil, err
+	}
+	return rawCount, retained, nil
 }
 
 func (s *JSONLStore) AddMessage(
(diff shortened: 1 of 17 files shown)


pkg/memory/jsonl.go
```diff
@@ -482,6 +483,9 @@ func readMessages(path string, skip int) ([]providers.Message, error) {
 				lineNum, filepath.Base(path), err)
 			continue
 		}
+		if messageutil.IsTransientAssistantThoughtMessage(msg) {
+			continue
+		}
```
alexhoshina, line 486: When an existing JSONL session contains one of the transient assistant thought records this now skips, `meta.Count`/`meta.Skip` still count raw lines. Later `TruncateHistory(keepLast)` computes the skip offset from the raw line count, and `readMessages` then skips that many raw lines plus filters the transient line, so sessions with old thought-only entries can lose more valid history than requested during summarization/compaction. Reconcile the file/counts when dropping these records or base truncation on retained messages.

alexhoshina (approved):
LGTM!Thx！

PR #2669, opened 2026-04-25
feat(agent): add network error retry with configurable max retries and backoff

david1gp:
## Summary

Add network error retry handling to the LLM call pipeline with configurable retry count and backoff. This prevents transient network failures (connection reset, broken pipe, etc.) from immediately failing agent requests.

## Problem

When making LLM API calls, transient network errors like `connection reset by peer` cause immediate failure. The retry logic only handled:
- Vision unsupported errors
- Timeout errors  
- Context window errors

Network errors (ECONNRESET, broken pipe, etc.) were not retried at all, causing agents to crash on what could be temporary connectivity issues.

Example error that would fail immediately:
```
LLM call failed after retries: failed to send request: Post "https://api.minimax.io/v1/chat/completions": read tcp 192.168.43.228:52748->47.252.72.253:443: read: connection reset by peer
```

## Solution

1. **Add `isNetworkError` detection** - Detects patterns like "connection reset", "broken pipe", "read tcp", "write tcp", "eof" in error messages

2. **Add retry logic for network errors** - With exponential backoff between retries

3. **Add configurable options** in `agents.defaults`:
   - `max_llm_retries` - max retry attempts (default: 2 = 3 total attempts)
   - `llm_retry_backoff_secs` - seconds between retries (default: 2)

4. **Make timeout backoff configurable** - Was hardcoded 5s, now uses same configurable backoff

## Configuration Example

```json
{
  "agents": {
    "defaults": {
      "max_llm_retries": 5,
      "llm_retry_backoff_secs": 3
    }
  }
}
```

## Backward Compatibility

- Defaults match previous behavior (2 retries with ~2s average backoff)
- Existing configurations continue to work unchanged

## Testing

The change adds new error classification patterns and retry paths that follow the existing test patterns in `pkg/agent/pipeline_llm.go`.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -275,7 +275,7 @@ type AgentDefaults struct {
 	SplitOnMarker             bool               `json:"split_on_marker"                  env:"PICOCLAW_AGENTS_DEFAULTS_SPLIT_ON_MARKER"` // split messages on <|[SPLIT]|> marker
 	ContextManager            string             `json:"context_manager,omitempty"        env:"PICOCLAW_AGENTS_DEFAULTS_CONTEXT_MANAGER"`
 	ContextManagerConfig      json.RawMessage    `json:"context_manager_config,omitempty" env:"PICOCLAW_AGENTS_DEFAULTS_CONTEXT_MANAGER_CONFIG"`
-	MaxLLMRetries            int                `json:"max_llm_retries,omitempty"        env:"PICOCLAW_AGENTS_DEFAULTS_MAX_LLM_RETRIES"`
+	MaxLLMRetries            int                `json:"max_llm_retries,omitempty"           env:"PICOCLAW_AGENTS_DEFAULTS_MAX_LLM_RETRIES"`
 	LLMRetryBackoffSecs      int                `json:"llm_retry_backoff_secs,omitempty" env:"PICOCLAW_AGENTS_DEFAULTS_LLM_RETRY_BACKOFF_SECS"`
 }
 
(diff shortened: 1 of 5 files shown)


david1gp:
## Motivation / Problem

Running agents that get rate limited by the llm provider -> completely stops their self independent work loop and you have to manually restart them.
Some providers (like minimax.io) are sometimes unreliable and block at the network level if to they are overloaded.

afjcjsbx:
Hi, can you add unit tests to validate the functioinality? 🙏

david1gp:
@afjcjbx added tests, like you wished :)

pkg/config/config.go
```diff
@@ -276,6 +275,8 @@ type AgentDefaults struct {
 	SplitOnMarker             bool               `json:"split_on_marker"                  env:"PICOCLAW_AGENTS_DEFAULTS_SPLIT_ON_MARKER"` // split messages on <|[SPLIT]|> marker
 	ContextManager            string             `json:"context_manager,omitempty"        env:"PICOCLAW_AGENTS_DEFAULTS_CONTEXT_MANAGER"`
 	ContextManagerConfig      json.RawMessage    `json:"context_manager_config,omitempty" env:"PICOCLAW_AGENTS_DEFAULTS_CONTEXT_MANAGER_CONFIG"`
+	MaxLLMRetries            int                `json:"max_llm_retries,omitempty"        env:"PICOCLAW_AGENTS_DEFAULTS_MAX_LLM_RETRIES"`
```
afjcjsbx: it might be useful to add the default values ​​also in the `defaults.go` file and in the `config.exmaple.json` file to make users aware of these new parameters

david1gp: done

afjcjsbx (approved):

PR #2670, opened 2026-04-25
feat(agent): add pretty_print and disable_escape_html options for tool feedback

david1gp:
## Summary
- Add `PrettyPrint` and `DisableEscapeHTML` config options to `ToolFeedbackConfig`
- Add `FormatArgsJSON` helper function with configurable pretty printing and HTML escaping
- Add unit tests for `FormatArgsJSON`

## Motivation
This fixes the issue where '&&' would be displayed as '\u0026' in tool feedback messages and provides optional pretty-printing for better readability.


david1gp:
# Motivation

The main Motivation is for people like me who look at the tool output.
It was not formatted corrently (html escaping), and since its json -> it can be easily pretty printed as well to make reading it easier.

## Before

```
{"action":"run","command":"ls -la /home/david/Coding-Using/plans/ \u0026\u0026 wc -l /home/david/Coding-Using/plans/*.md"}
```

## After

```
{
  "action": "run",
  "command": "ls /home/david/Coding-Using/plans"
}
```

david1gp:
I could use some help in the design decision:

I added to config values to be able to have the old behavior (not escaping html and not pretty printing json). 

Is there a use case there you would like to see the unescaped, not pretty-printed version? 

I think the config values could be removed (and the code simplified)
and the "escaping html characters" + "pretty printing of json" be made the default always-on behavior.

Edit / Update:

I'm rolling with default behavior, without the extra config values. 
There is no reason why someone would like to see un-escaped html or non-pretty json as a human.

afjcjsbx:
This PR introduces the `tool_feedback`.go and `tool_feedback_test`.go files in the repository root, duplicating logic already present in `pkg/utils/`. Keeping two packages with overlapping symbols risks confusing imports, code divergence (drift), and unnecessary increase in CI time. I suggest you remove the files in the root and use the existing implementation in `pkg/utils`.

pkg/utils/tool_feedback.go
```diff
@@ -1,12 +1,29 @@
...
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
 	"strings"
 )
 
 const ToolFeedbackContinuationHint = "Continuing the current task."
 
+func FormatArgsJSON(args map[string]any, prettyPrint, disableEscapeHTML bool) string {
```
afjcjsbx, line 12: `FormatArgsJSON` now hides serialization failures by returning `{}.` The old path used `json.MarshalIndent` and, on error, fell back to `fmt.Sprintf("%v", args)`. The new helper returns "{}" for any encode failure.
If tool args ever contain a non-JSON value (func, chan, cyclic data, custom type with bad marshaling), the feedback shown to users/logs becomes a misleading empty object instead of "something went wrong with these args". That makes debugging and approval flows much harder. Hint: Preserve the old fallback or return an error.

afjcjsbx, line 12: Nil-argument behavior is inconsistent across helpers (null vs {}). `FormatArgsJSON(nil, ...)` returns `null`, but `toolFeedbackArgsPreviewWithOptions` rewrites nil to `{}` first. This is a fragile API contract. Future callers of `FormatArgsJSON` will get a different output than the main tool-feedback path, and tests already codify both behaviors. Pick one representation and enforce it in one place. {} is usually friendlier for "no args".

afjcjsbx (changes requested):

david1gp:
I think I addressed all issues pointed out.
Thanks @afjcjsbx!

afjcjsbx (approved):

david1gp:
still relevant

PR #2677, opened 2026-04-26
Feat/runtime events

alexhoshina:
## 📝 Description

This PR introduces a unified runtime event infrastructure for PicoClaw and migrates agent observability onto it.

The new `pkg/events` package defines a shared event envelope, event kinds, filters, subscription channels, backpressure policies, subscriber stats, and an in-process runtime event bus. Agent, channel, gateway, message bus, and MCP code can now publish structured runtime events through the same read-only observation surface.

It also separates event publishing from log printing. Runtime components publish events to the bus, while the built-in runtime event logger subscribes to that bus and prints only the events selected by `events.logging` configuration. The default keeps existing log behavior by printing `agent.*` events only, while still publishing channel, gateway, bus, and MCP events for hooks, tests, diagnostics, and future UI consumers.

Key changes:

- Add `pkg/events` with event envelope, filtering, fan-out bus, subscriptions, backpressure handling, lifecycle close behavior, and stats.
- Replace the legacy agent event bus with runtime events emitted directly from agent turn, LLM, tool, context, steering, interrupt, and subturn paths.
- Split agent event payload definitions into `event_payloads.go` and bridge agent hook metadata into runtime event scope/correlation fields.
- Wire `AgentLoop` to own or accept a runtime event bus, expose runtime event options, and keep hook observation on the shared bus.
- Add a centralized runtime event logger with configurable `enabled`, `include`, `exclude`, `min_severity`, and `include_payload` settings.
- Make runtime event logging reloadable on config reload, and drain logger subscriptions during shutdown so buffered final events are not dropped.
- Publish runtime events from gateway lifecycle/reload, channel lifecycle/webhook/outbound flows, message bus publish/close paths, and MCP server/tool activity.
- Update tests to observe runtime events instead of the removed legacy agent event bus.
- Add architecture/configuration docs for runtime events and event logging in English and Chinese.

## 🗣️ Type of Change

- [x] 📖 Documentation update

## 🤖 AI Code Generation


pkg/events/subscription.go
```diff
@@ -0,0 +1,404 @@
...
+
+	switch s.opts.Backpressure {
+	case DropOldest:
+		return s.enqueueDropOldest(evt)
+	default:
+		return s.enqueueDropNewest(evt)
+	}
+}
+
+func (s *eventSubscription) enqueueBlocking(ctx context.Context, evt Event) deliveryResult {
```
afjcjsbx, line 372: `enqueueBlock` together with the synchronous publishers in `pkg/agent/events_runtime.go`, `pkg/bus/events.go`, `pkg/channels/events.go`, `pkg/gateway/events.go`, `pkg/mcp/events.go`, and `pkg/tools/integration/mcp_tool.go`: runtime event publication is no longer effectively non-blocking. Core paths now call `Publish()` inline with a `100ms` timeout, and the new bus explicitly supports Backpressure: Block. A single slow/full subscriber can therefore add ~100ms per emitted event to turn execution, tool execution, outbound delivery, startup, reload, and shutdown.
Why: the event system is now on hot paths. With a blocking subscriber or tiny buffer, latency compounds quickly.
Possible Fix: keep core runtime publication strictly non-blocking. The safest options are: add a `PublishNonBlocking/TryPublish` API, force `DropNewest/DropOldest` for runtime observers, or reject `Backpressure: Block` for this bus entirely.

pkg/events/subscription.go
```diff
@@ -0,0 +1,404 @@
...
+	case Keyed:
+		// TODO: replace this with keyed executors when runtime events need
+		// per-scope ordering with cross-scope concurrency.
+		s.handle(ctx, evt)
+	default:
+		s.handle(ctx, evt)
+	}
+}
+
+func (s *eventSubscription) handle(ctx context.Context, evt Event) {
```
afjcjsbx, line 241: `SubscribeOptions.Timeout` is not actually enforced. The code creates a timed context and passes it to the handler, but then waits forever for the handler to return. If the handler ignores `ctx.Done()`, a `Locked` subscription can wedge permanently, `Done()` never closes, and goroutines leak.
Why: the API advertises a timeout, but the implementation is only cooperative.
Possible Fix: either enforce the timeout with a goroutine/select boundary, or document clearly that timeout is advisory only and rename it to avoid a false guarantee.

pkg/events/bus.go
```diff
@@ -0,0 +1,233 @@
+package events
```
afjcjsbx, line 1: `pkg/bus/bus.go` adds `bus.close.*` emissions, but `pkg/gateway/gateway.go` never closes the gateway’s `MessageBus`. `Run()` wires `msgBus.SetEventPublisher(...)`, yet `shutdownGateway()` and reload paths do not call `msgBus.Close().`
Why: the new close lifecycle events are effectively dead in production, so subscribers will never observe them outside tests.
Fix: make the gateway own and close the message bus on full shutdown, and decide whether reload should recreate or intentionally retain the same bus.

pkg/agent/hook_mount.go
```diff
@@ -308,10 +310,50 @@ func processHookObserveKindsFromConfig(observe []string) ([]string, bool, error)
 	return normalized, true, nil
 }
 
-func validHookEventKinds() map[string]struct{} {
-	kinds := make(map[string]struct{}, int(eventKindCount))
-	for kind := EventKind(0); kind < eventKindCount; kind++ {
-		kinds[kind.String()] = struct{}{}
+func validHookEventKinds() map[string]string {
```
afjcjsbx, line 313: `validHookEventKinds` process-hook `observe` validation still only allows agent-domain events. The new runtime bus adds `channel.*, bus.*, gateway.*, and mcp.*`, but external process hooks cannot subscribe to them because config parsing rejects those names.

pkg/agent/runtime_event_logger.go
```diff
@@ -0,0 +1,408 @@
...
+		}
+		if _, exists := fields[key]; exists {
+			fields["attr_"+key] = value
+			continue
+		}
+		fields[key] = value
+	}
+}
+
+func appendRuntimeEventPayloadSummary(fields map[string]any, payload any) {
```
afjcjsbx, line 252: Only agent payload types are summarized. `Channel/bus/gateway/MCP` events log almost no useful details unless `include_payload=true`. For example, `gateway.ready`, `bus.publish.failed`, `channel.message.outbound_failed`, and `mcp.server.failed` won’t expose `duration_ms`, `error`, `server`, `tool`, etc. in the default safe mode.
Why: the feature’s value is observability, but non-agent logs are nearly empty by default, and the docs currently promise those details.
Fix: move shared payload types into `pkg/events`, or have publishers place safe summary fields into `Attrs` so the logger can emit them without importing downstream packages.

pkg/events/bus.go
```diff
@@ -0,0 +1,233 @@
...
+
+	subs := make([]*eventSubscription, 0, len(b.subs))
+	for _, sub := range b.subs {
+		subs = append(subs, sub)
+	}
+	sortSubscriptions(subs)
+	return subs, false
+}
+
+func sortSubscriptions(subs []*eventSubscription) {
```
afjcjsbx, line 231: `snapshotSubscribers` and `sortSubscriptions` on every publish copies the subscriber map and sorts it. With many subscribers or chatty event kinds, this creates avoidable allocation and `O(n log n)` work on every event.
Why: the event bus sits in a central runtime path now.
Fix: maintain a copy-on-write ordered slice updated on subscribe/unsubscribe instead of rebuilding it per publish.

afjcjsbx (changes requested):

alexhoshina:
@afjcjsbx Thanks for the detailed review. I addressed all six threads in 78fd0801: runtime event publishing is now non-blocking on hot paths, subscription timeouts no longer wedge locked subscribers, gateway full shutdown closes the MessageBus, hook observe validation accepts all runtime event kinds, non-agent events include safe default log attrs, and subscriber ordering is maintained on subscribe/unsubscribe rather than sorted on every publish.

afjcjsbx (approved):

PR #2679, opened 2026-04-26
fix: enable ChatGPT subscription (OAuth) in picoclaw

tbeaudouin05:
## Summary
- Always use chatgpt.com/backend-api/codex for OAuth/auth_method=token providers
- Handle response.output_text.delta streaming for Codex backend (fixes empty responses)
- Reconstruct tool calls from stream events when the completed response output is empty

## Problem
- ChatGPT Plus subscriptions use OAuth tokens, which only work on chatgpt.com backend
- The Codex backend returns streaming text in `response.output_text.delta` events, but picoclaw was only parsing `response.completed.output` (which can be empty)
- Similarly, tool calls arrive as `response.function_call_arguments.delta` and `response.output_item.added` events during the stream, but the final completed response may have an empty `output` array — this caused empty tool call results even when the model was generating valid function calls
- This caused empty responses even when the model was generating text or tool calls

## Changes
- Route all OAuth/token auth providers to the Codex backend URL
- Parse `response.output_text.delta` streaming deltas for content
- Parse `response.output_item.added`, `response.function_call_arguments.delta`, and `response.function_call_arguments.done` stream events to reconstruct tool calls when the completed response has no output
- Simplify: remove `ShouldUseCodexBackend()` (always returned true)

## Test
```bash
picoclaw agent --model gpt-5.4-mini -m "Use the write_file tool to create ~/.picoclaw/workspace/test.txt with exactly: hello"
```

### DIFF
--- a/pkg/providers/factory_provider.go
+++ b/pkg/providers/factory_provider.go
@@ -75,6 +75,7 @@ func createClaudeAuthProvider() (LLMProvider, error) {
 }
 
 // createCodexAuthProvider creates a Codex provider using OAuth credentials from auth store.
+// ChatGPT subscription OAuth tokens use the Codex backend for all OpenAI models.
 func createCodexAuthProvider() (LLMProvider, error) {
 	cred, err := getCredential("openai")
 	if err != nil {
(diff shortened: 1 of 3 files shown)


tbeaudouin05:
Don't close

PR #2680, opened 2026-04-26
refactor(pico): unify message kind handling of tool_calls and thought

lc6464:
## 📝 Description

Unify message kind handling for `tool_calls` and `thought`.

This change makes `thought` use the same protocol shape as `tool_calls`, i.e. `payload.kind = "thought"`, instead of the legacy `payload.thought = true`.

This frontend-side breaking change is intentional. In Pico's current architecture, the web UI is served from backend-embedded assets rather than deployed independently, so frontend/backend version skew is not a primary compatibility target in practice. Keeping legacy frontend parsing for `payload.thought === true` would preserve extra protocol branching and maintenance cost for a case that is uncommon in real usage.

Backend-side compatibility is still preserved for Go client/server interaction, since those are separate binaries and may not always be upgraded in strict lockstep.

## 🗣️ Type of Change
- [x] ⚡ Code refactoring (with internal pico protocol breaking changes)

## 🤖 AI Code Generation

## 🧪 Test Environment
- **OS:** Ubuntu Server 24
- **Model/Provider:** DeepSeek V4 Flash
- **Channels:** Web

## 📸 Evidence
<details>

Thought:
image

Tool calls:
image

Web:
image

</details>

## ☑️ Checklist

### DIFF
--- a/pkg/channels/pico/pico.go
+++ b/pkg/channels/pico/pico.go
@@ -323,10 +323,12 @@ func (c *PicoChannel) Send(ctx context.Context, msg bus.OutboundMessage) ([]stri
 
 	payload := map[string]any{
 		PayloadKeyContent: content,
-		PayloadKeyThought: isThought,
 		"message_id":      msgID,
 	}
-	if isToolCalls {
+	switch {
+	case isThought:
+		payload[PayloadKeyKind] = MessageKindThought
+	case isToolCalls:
 		payload[PayloadKeyKind] = MessageKindToolCalls
 		if toolCalls, ok := picoToolCallsPayload(msg); ok {
 			payload[PayloadKeyToolCalls] = toolCalls
@@ -457,7 +459,6 @@ func (c *PicoChannel) SendPlaceholder(ctx context.Context, chatID string) (strin
 	msgID := uuid.New().String()
 	outMsg := newMessage(TypeMessageCreate, map[string]any{
 		PayloadKeyContent: text,
-		PayloadKeyThought: false,
 		"message_id":      msgID,
 	})
 
--- a/pkg/channels/pico/pico.go
+++ b/pkg/channels/pico/pico.go
@@ -328,6 +328,12 @@ func (c *PicoChannel) Send(ctx context.Context, msg bus.OutboundMessage) ([]stri
 	switch {
 	case isThought:
 		payload[PayloadKeyKind] = MessageKindThought
+
+		// This field is kept solely for compatibility with legacy pico clients that
+		// do not yet support the newer "kind" field.
+		// DO NOT use it for any purpose other than legacy client compatibility.
+		payload[PayloadKeyThought] = true
+
 	case isToolCalls:
 		payload[PayloadKeyKind] = MessageKindToolCalls
 		if toolCalls, ok := picoToolCallsPayload(msg); ok {
--- a/web/frontend/src/features/chat/assistant-message-state.ts
+++ b/web/frontend/src/features/chat/assistant-message-state.ts
@@ -19,14 +19,25 @@ export interface AssistantMessageUpdateState {
   toolCalls?: AssistantToolCalls
 }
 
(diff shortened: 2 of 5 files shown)


web/frontend/src/features/chat/assistant-message-state.ts
```diff
@@ -19,14 +19,25 @@ export interface AssistantMessageUpdateState {
...
+  payload: Record<string, unknown>,
+): string | undefined {
+  if (typeof payload.kind !== "string") {
+    return undefined
+  }
+  const kind = payload.kind.trim().toLowerCase()
+  return kind || undefined
+}
+
 function parseAssistantMessageKind(
```
afjcjsbx, line 32: Frontend dropped legacy `payload.thought === true` support, while backend explicitly keeps it for inbound compatibility. After this patch it only recognizes `payload.kind === "thought"`. Meanwhile `pkg/channels/pico/protocol.go` explicitly preserves legacy bool support for inbound compatibility.
During mixed-version deployments, old servers or cached sessions may still send the legacy boolean marker. The Go client will ignore those thought messages correctly, but the web frontend will now misclassify them as "normal". That can surface internal reasoning in the UI or break message grouping/rendering logic.
Fix: Keep legacy bool support in the frontend until the protocol migration is fully complete.

pkg/channels/pico/pico.go
```diff
@@ -323,10 +323,12 @@ func (c *PicoChannel) Send(ctx context.Context, msg bus.OutboundMessage) ([]stri
 
 	payload := map[string]any{
 		PayloadKeyContent: content,
-		PayloadKeyThought: isThought,
 		"message_id":      msgID,
 	}
-	if isToolCalls {
+	switch {
```
afjcjsbx, line 328: `Send()` no longer emits `PayloadKeyThought: true/false`; thought messages are now represented only by `PayloadKeyKind = "thought"`, and normal/placeholder messages omit both markers.
This is fine if every consumer has already been updated. If not, older clients that only understand thought: true will stop recognizing thought messages and may render internal reasoning as normal assistant output. The backend comment only mentions inbound compatibility for old servers; it does not address old clients consuming new messages.
Fix: During migration, emit both markers for thought messages:

afjcjsbx:
New Go tests assert that legacy `PayloadKeyThought: true` is still ignored by the Pico client, but there are no corresponding frontend tests for legacy payload parsing. The missing frontend test is likely why the regression slipped in.  
  **Fix:** Add frontend unit tests for:
  - `payload = { thought: true }` => `"thought"`
  - `payload = { kind: " thought " }` => `"thought"`
  - `payload = { kind: "tool_calls" }` => `"tool_calls"`
  - `payload = { thought: false }` => not thought
  - `hasExplicitAssistantKindPayload({ thought: true }) === true`

afjcjsbx (changes requested):

lc6464:
@afjcjsbx 
Thanks for the review.

The frontend behavior here is intentional, not an accidental regression.

In Pico's current architecture, the web UI is served from backend-embedded assets, so frontend/backend version skew is uncommon in practice and not a primary compatibility target here. In the rare case that versions do not match, the worst likely outcome is a UX issue where a thought message is rendered as a normal assistant message. This does not involve historical replay concerns in this case, and launcher/gateway skew is also not a common deployment path.

Because of that, I think removing legacy frontend support for `payload.thought === true` is an acceptable internal breaking change. Keeping both branches on the frontend would add maintenance complexity for limited practical benefit.

The backend compatibility is kept for a different reason: the Go server and Go client are separate applications, so they are more likely to be upgraded out of sync.

For tests, I agree frontend coverage would be useful in general. But since the frontend currently has no unit test setup, introducing tests from scratch together with CI integration would make this refactor PR significantly broader than intended. I would prefer to keep this PR focused on the protocol cleanup itself.

afjcjsbx (approved):

imguoguo (approved):

PR #2693, opened 2026-04-27
fix: block find / from bypassing workspace sandbox (fixes #2688)

### LINKED ISSUE #2688: Security: find / can enumerate paths outside workspace sandbox
### Summary

The safety guard in picoclaw blocks direct file reads (`cat`, `ls`) outside the workspace directory (`/var/lib/picoclaw/workspace`), but **`find /` commands still work**, allowing path enumeration across the entire filesystem.

### Steps to Reproduce

1. Execute `find / -name "private*" -type f 2>/dev/null` �~@~T returns files outside workspace (e.g., Rust crates with `private_key.rs`, `private_exponent.rs`)
2. Run `find / -name "config*" -path "*picoclaw*"` to enumerate config file locations
3. Run `find / -name "*.pem" -o -name "*.key" -o -name "token*" 2>/dev/null` to probe for credential files

### Impact

- **Information disclosure**: Can check existence of arbitrary files on the system (`/etc/shadow`, `/root/.ssh/id_rsa`, `/etc/ssl/private/`, etc.)
- **Path enumeration**: Can discover system structure, config locations, credential file presence
- While actual file *content* is protected, knowing a file exists is already a leak (e.g., confirming `/etc/shadow` exists is harmless, but confirming `/root/.ssh/id_rsa` exists reveals SSH key presence)

### Suggested Fix

Either:
1. Block `find` commands that traverse outside the workspace root, OR
2. Apply the same sandbox restrictions to `find` as are applied to `ls`/`cat`

### Environment

- Platform: linux riscv64
- Runtime: Go go1.25.9
- Version: picoclaw 0.2.7 (git: 71c877a6)

islobodan:
Fix for #2688 - sandbox bypass via find / and ls /

### DIFF
--- a/pkg/tools/shell.go
+++ b/pkg/tools/shell.go
@@ -95,12 +95,12 @@ var (
 		regexp.MustCompile(`\bssh\b.*@`),
 		regexp.MustCompile(`\beval\b`),
 		regexp.MustCompile(`\bsource\s+.*\.sh\b`),
-		regexp.MustCompile(`\bfind\s+/\b`),       // find / - traverse entire filesystem
-		regexp.MustCompile(`\bls\s+/\b`),         // ls / - list root directory
+		regexp.MustCompile(`\bfind\s+/(\s|[;&|><\n]|$)`), // find / - traverse entire filesystem
+		regexp.MustCompile(`\bls\s+/(\s|[;&|><\n]|$)`),   // ls / - list root directory
 	}
 
 	// absolutePathPattern matches absolute file paths in commands (Unix and Windows).
-	absolutePathPattern = regexp.MustCompile(`[A-Za-z]:\\[^\\\"']+|/(?:[^\s\"']*)?`)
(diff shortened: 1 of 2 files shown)


islobodan:
Fixed the sandbox bypass (#2688):

   - Deny patterns for `find /` and `ls /` now catch shell metacharacters (`;`, `&`, `|`, `&&`, `||`) after `/`
   - `absolutePathPattern` changed to capture bare `/` so `ls /` reaches the check
   - `/` added to `safePaths` so bare root doesn't trigger a false workspace violation

   Tests: 37 cases covering blocked commands, allowed workspace commands, metachar bypasses, mixed case, and deny-disabled behavior.

islobodan:
Force-pushed to fix the golines linter error.

islobodan:
Rebased onto latest main (Go 1.25.10) to resolve false-positive security check — no changes to the fix itself

PR #2701, opened 2026-04-28
feat(provider,web,asr): enhance model management with explicit provider metadata

lc6464:
## 📝 Description

Enhance model management in the Web UI by moving model entries onto explicit provider metadata while preserving compatibility for legacy provider/model configs, including ElevenLabs ASR.

- add a canonical provider catalog and provider/model normalization helpers so model_list entries are stored as explicit provider + model pairs
- expose provider_options and default_model_allowed from /api/models, and align availability/default gating for Bedrock, OAuth-backed providers, local CLI providers, and ASR-only providers such as ElevenLabs
- replace free-form provider inputs in add/edit sheets with provider selection, provider-specific defaults, localized labels/icons, and UI affordances for providers that cannot become the default chat model
- keep provider-omitted updates round-trippable for older clients while migrating legacy ElevenLabs transcription configs and clearing invalid historical default chat-model selections on save
- extend backend, provider, ASR, frontend, and README coverage for canonical provider storage and ElevenLabs transcription compatibility

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 📚 Technical Context
- **Reference URL:** -
- **Reasoning:** Introduce explicit provider metadata end-to-end so the Web UI, gateway readiness checks, default-model eligibility, and ASR detection all consume the same backend-owned provider catalog while preserving legacy config compatibility.

## 🧪 Test Environment
- **OS:** Ubuntu Server 24
- **Model/Provider:** -
- **Channels:** Web

## 📸 Evidence
<details>

DeepSeek with default api base:
image

Gemini with default api base:
image

LM Studio with default api base and empty api key:
image

CLI Providers cannot be selected when the executable file is not exist:
image

</details>

## ☑️ Checklist


web/backend/api/models.go
```diff
@@ -51,6 +51,145 @@ type modelResponse struct {
...
+		}
+		return changed
+	}
+
+	effectiveProvider, modelID := providers.SplitModelProviderAndID(model, "openai")
+	if effectiveProvider == "" {
+		return changed
+	}
+	if mc.Provider != effectiveProvider {
+		mc.Provider = effectiveProvider
```
alexhoshina, line 110: When a stored model has no explicit provider and uses the documented ElevenLabs ASR form elevenlabs/scribe_v1, this defaulting treats elevenlabs as unsupported and rewrites the entry to provider: "openai" whenever any model is added or updated. isElevenLabsTranscriptionModel only recognizes that legacy ASR form while Provider is empty, so the next web save can silently break ElevenLabs transcription or route it through the OpenAI audio path with the ElevenLabs key; skip this migration for unsupported ASR prefixes or handle ElevenLabs explicitly.

lc6464:
@alexhoshina Thanks!
Fixed.

alexhoshina:
@huaaudio This PR involves modifications to the ASR configuration. Please help review it.

This PR modifies the configuration for the old version of the ElevenLabs ASR model.
- It is no longer possible to create new configurations of this type via the API.
- Existing old configurations that have been retained also cannot be saved in the updated UI.

lxowalle (commented):
I think there’s still a compatibility hole here around legacy ElevenLabs ASR configs.

Right now `GET /api/models` shows those entries as `provider: "elevenlabs"` and `model: "scribe_v1"`, which looks fine on the surface. But if a client takes that response and sends the same visible values back through `PUT /api/models/{index}`, the update path rejects it because `elevenlabs` is not treated as a supported general model provider. So the API output is not actually round-trippable for that legacy case.

The same problem shows up in the web UI. The backend returns `provider: "elevenlabs"` for the model, but `provider_options` does not include it. That leaves the edit form in an invalid state and blocks saving, so users can’t even make simple compatibility edits like changing `api_base`.

So at the moment we’re exposing these legacy entries in the new model-management flow, but that flow still can’t fully handle them. This probably needs either full support in both API and UI, or a decision to keep these legacy ASR entries out of the editable flow entirely.

lxowalle (commented):
This PR improves provider metadata and normalizes ElevenLabs ASR configs, but I still see two issues introduced by the new flow.

First, the chat-page default-model selector is not using the new `default_model_allowed` flag. `GET /api/models` now marks ElevenLabs ASR entries as unavailable for default chat-model selection, and the backend correctly rejects `POST /api/models/default` for them. However, `useChatModels` still groups models only by availability/auth type, so an available ElevenLabs ASR entry will still appear in the chat-page model dropdown and can still be selected from the UI. That creates a broken flow where the frontend offers a choice that the backend now rejects.

Second, this PR makes ElevenLabs a first-class provider in model management, but the runtime still ignores the configured model ID. The new UI/API flow lets users save `provider: "elevenlabs"` together with an editable `model` field, yet the ElevenLabs transcriber still hardcodes `model_id = "scribe_v1"` and only checks whether the provider resolves to `elevenlabs`. In other words, the new explicit-provider flow suggests that the stored model ID matters, but the runtime will silently ignore it. I think this needs either validation that pins the model to the only supported value, or transcriber logic that actually uses the configured model field.

lxowalle (approved):

PR #2705, opened 2026-04-29
add MQTT channel support

hehaijunandhenry:
## 📝 Description
Add MQTT channel support, enabling message communication via the MQTT protocol.

## 🗣️ Type of Change
- [x] 📖 Documentation update

## 🤖 AI Code Generation

- **Reference URL:**  
https://github.com/eclipse-paho/paho.mqtt.golang

- **Reasoning:**
Users need to connect IoT devices via MQTT protocol, which is not currently supported as a channel.

- **Design:**
  - Implemented a new MQTT channel adapter following the existing channel interface
  - Supports MQTT and SSL both
  - Uses persistent connection with auto-reconnect on failure
  - Compatible with standard MQTT v3.1.1 / v5.0

## 🧪 Test Environment
- **Hardware:** PC/VM
- **OS:** Ubuntu 20.04
- **Model/Provider:** qwen-plus
- **Channels:** pico/mqtt

## 📸 Evidence (Optional)
<details>

mqtt conversation:

image

image

WebUI configuration:

image

image

image

</details>

## ☑️ Checklist


hehaijunandhenry:
Hi @yinwm  @alexhoshina , could you please take a look at this PR when you have a moment? I've filled out the template and signed the CLA. Thanks a lot! 🙏

alexhoshina:
linter plz

pkg/channels/mqtt/mqtt.go
```diff
@@ -0,0 +1,242 @@
...
+	})
+
+	opts.SetConnectionLostHandler(func(_ pahomqtt.Client, err error) {
+		logger.WarnCF("mqtt", "MQTT connection lost", map[string]any{"error": err.Error()})
+	})
+
+	client := pahomqtt.NewClient(opts)
+	token := client.Connect()
+	if !token.WaitTimeout(10 * time.Second) {
+		return fmt.Errorf("mqtt connect timed out after 10s (broker: %s)", c.cfg.Broker)
```
alexhoshina, line 107: When the broker is unavailable at startup, `SetConnectRetry(true)` keeps Paho's connect goroutine retrying even after `WaitTimeout` returns false. Because this path returns without `Disconnect` and before assigning `c.client`, the failed channel can later connect and subscribe as a zombie while the manager has no worker for it, causing requests/responses to be lost or duplicated after a later restart. Disconnect/cancel the client before returning on timeout, or disable connect retry for the initial connect.

pkg/channels/mqtt/mqtt.go
```diff
@@ -0,0 +1,242 @@
...
+	topic := fmt.Sprintf("%s/%s/+/request", c.topicPrefix(), c.cfg.AgentID)
+	token := client.Subscribe(topic, c.qos, func(_ pahomqtt.Client, msg pahomqtt.Message) {
+		c.handleInbound(msg)
+	})
+	token.Wait()
+	if err := token.Error(); err != nil {
+		logger.ErrorCF("mqtt", "Failed to subscribe", map[string]any{
+			"topic": topic,
+			"error": err.Error(),
+		})
```
alexhoshina, line 169: If the broker accepts the connection but rejects this SUBSCRIBE, for example because ACLs deny the request topic or an invalid QoS is configured, this branch only logs the error; `Start` still sets the channel running and returns nil. The gateway then reports MQTT as started even though it will never receive inbound requests, so the initial subscribe error should be propagated before marking the channel running.

pkg/config/config.go
```diff
@@ -512,6 +512,17 @@ type TeamsWebhookTarget struct {
 	Title      string       `json:"title,omitempty"      yaml:"-"`
 }
 
+type MQTTSettings struct {
+	Broker      string       `json:"broker"                yaml:"-"                        env:"PICOCLAW_CHANNELS_MQTT_BROKER"`
+	AgentID     string       `json:"agent_id"              yaml:"-"                        env:"PICOCLAW_CHANNELS_MQTT_AGENT_ID"`
+	TopicPrefix string       `json:"topic_prefix,omitempty" yaml:"-"                       env:"PICOCLAW_CHANNELS_MQTT_TOPIC_PREFIX"`
+	Username    SecureString `json:"username,omitzero"     yaml:"username,omitempty"       env:"PICOCLAW_CHANNELS_MQTT_USERNAME"`
+	Password    SecureString `json:"password,omitzero"     yaml:"password,omitempty"       env:"PICOCLAW_CHANNELS_MQTT_PASSWORD"`
```
alexhoshina: These new credentials are `SecureString`s, but `toChannelHashes` only hashes real secret values for channels handled by `hiddenValues`, and no MQTT case was added. When `channel_list.mqtt.settings.username` or `password` changes in `.security.yml`, the JSON hash still omits/masks the values, so `Manager.Reload` sees no channel change and keeps the existing MQTT connection authenticated with the old credentials.

pkg/channels/mqtt/mqtt.go
hehaijunandhenry: Fixed. "client.Disconnect(250)" is now called before returning on both the timeout path and the explicit connect-error path.

pkg/channels/mqtt/mqtt.go
hehaijunandhenry: Fixed. The subscribe result is now propagated back to Start() via a sync.Once-guarded channel. If the broker rejects the SUBSCRIBE (ACL denial, invalid QoS, etc.), Start() calls client.Disconnect(250) and returns the error before SetRunning(true) is ever reached. The gateway will correctly report the channel as failed rather than started.

pkg/config/config.go
hehaijunandhenry: Fixed. Added a case "mqtt" block to hiddenValues in manager_channel.go that injects the decoded Username and Password real values into the hash input before computing the MD5.

hehaijunandhenry:
@alexhoshina 

All issues have been addressed and fixed, please take another look when you get a chance.
Many thanks!

alexhoshina:
linter plz

hehaijunandhenry:
@alexhoshina 

Linter done and passed, please help to check the fixes, many thanks.

pkg/channels/mqtt/mqtt.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+
+	opts := pahomqtt.NewClientOptions()
+	opts.AddBroker(c.cfg.Broker)
+	opts.SetClientID(c.clientID)
+	opts.SetKeepAlive(time.Duration(keepAlive) * time.Second)
+	opts.SetAutoReconnect(true)
+	opts.SetConnectRetry(true)
+	opts.SetConnectRetryInterval(5 * time.Second)
+	opts.SetTLSConfig(&tls.Config{InsecureSkipVerify: true}) //nolint:gosec
```
alexhoshina, line 83: Perhaps it would be safer to use a verified TLS configuration by default?

pkg/channels/mqtt/mqtt.go
```diff
@@ -0,0 +1,255 @@
...
+	c.HandleInboundContext(context.Background(), chatID, payload.Text, nil, inboundCtx)
+}
+
+// Stop disconnects from the MQTT broker.
+func (c *MQTTChannel) Stop(_ context.Context) error {
+	logger.InfoC("mqtt", "Stopping MQTT channel")
+	c.SetRunning(false)
+
+	if c.client != nil && c.client.IsConnected() {
+		c.client.Disconnect(500)
```
alexhoshina, line 217: If `Stop` is called while the Paho client is in a disconnected/reconnecting state, `IsConnected()` is false and `Disconnect` is skipped, but `SetAutoReconnect(true)` means the reconnect goroutine can continue and resubscribe after the channel was disabled or reloaded.

pkg/channels/mqtt/mqtt.go
hehaijunandhenry: Indeed, verified TLS is safer and ensures certificate trustworthiness. However, MQTT is most commonly used for internal/private integrations where the broker often uses self-signed certificates or has no domain name at all — making proper TLS verification a little difficult for most users. That's why intentionally to disable certificate verification as the default.

pkg/channels/mqtt/mqtt.go
hehaijunandhenry: The critical issue fixed.

alexhoshina (approved):
LGTM!

PR #2708, opened 2026-04-29
fix(feishu): fix image download with API fallback and post image support

imguoguo:
## 📝 Description

- Add Image.Get API fallback when MessageResource.Get fails (different permission scope: im:resource vs im:message:readonly)
- Extract and download images from post (rich text) messages
- Extract images from interactive card messages
- Deduplicate post image keys across locales
- Add comprehensive tests for new helpers

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

- **Reference URL:**
- **Reasoning:**

## 🧪 Test Environment
- **Hardware:** 
- **OS:** 
- **Model/Provider:** 
- **Channels:** 

## 📸 Evidence (Optional)
<details>

</details>

## ☑️ Checklist

### DIFF
--- a/pkg/agent/agent_media.go
+++ b/pkg/agent/agent_media.go
@@ -11,6 +11,7 @@ import (
 	"encoding/base64"
 	"io"
 	"os"
+	"regexp"
 	"strings"
 
 	"github.com/h2non/filetype"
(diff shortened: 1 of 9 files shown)


lxowalle (commented):
### Comment 1

`pkg/agent/agent_media.go:245`

This change can break Feishu `interactive` and `post` messages again, just at a different layer.

`appendMediaTags()` intentionally leaves those message types unchanged so their raw JSON stays valid for downstream parsing. But `resolveMediaRefs()` now generates `[image:/path]` tags for all user-image refs, and `injectPathTags()` falls back to appending the tag when no generic placeholder exists.

For Feishu cards/posts, the content is raw JSON and does not contain `[image]` placeholders, so the final provider input becomes something like:

```text
{"schema":"2.0",...} [image:/tmp/...]
```

That is no longer valid JSON, which means the model loses the structured card/post payload this PR is trying to preserve. Please add an end-to-end test for `resolveMediaRefs()` with Feishu `interactive`/`post` JSON content plus media refs, and keep those payloads JSON-safe at the agent layer as well.

### Comment 2

`pkg/agent/agent_media.go:35`

This changes the behavior for normal inbound user images from "model can see the image immediately" to "model only receives a file path unless it explicitly calls `load_image`".

Previously, user-supplied image refs were converted into base64 media for the next provider call. After this patch, user messages keep only `[image:/path]` tags, while base64 injection happens only for tool results. That makes image understanding depend on `load_image` being enabled and selected by the model, but `load_image` is still conditionally registered in `agent_init.go`.

So in deployments where `load_image` is disabled, inbound channel images silently regress from vision-capable input to plain text path references. If that tradeoff is intentional, it needs explicit coverage and probably a config/compatibility note; otherwise I think user-role images still need a direct multimodal path.

lxowalle (approved):

PR #2713, opened 2026-04-29
fix(channels): dismiss tool feedback animation when turn ends via ResponseHandled

imguoguo:
## 📝 Description

When a tool sets `ResponseHandled=true` (e.g., `send_file`), the turn ends without producing a final assistant response. This means no outbound message triggers `FinalizeToolFeedbackMessage` on the channel, leaving the animation goroutine running indefinitely — editing the Feishu card every 3 seconds with `.` / `..` suffixes long after the tool had finished.

Root cause: the tool feedback outbound goes through the text worker queue, while the tool's media output goes through the media worker queue. The media worker often finishes first (before the text worker even creates the feedback card), so `SendMedia`'s dismiss logic finds nothing to clean up. The text worker then creates the feedback card and starts the animation goroutine, but since the turn has already ended with no final response, nobody ever calls `FinalizeToolFeedbackMessage` or `DismissToolFeedbackMessage`.

Fix: call `DismissToolFeedback` at the "Tool output satisfied delivery" exit point so the tracker is cleared and the animation goroutine is stopped immediately when the turn ends.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue

Observed on Feishu channel: after calling `send_file`, the tool feedback card kept animating indefinitely with `.`/`..` suffix every 3 seconds.

- **Reference URL:** `pkg/channels/tool_feedback_animator.go` (animation goroutine lifecycle), `pkg/agent/pipeline_execute.go:707` (ResponseHandled turn-end path)
- **Reasoning:** The animation goroutine is started by `ToolFeedbackAnimator.Record()` and stopped by `Take()` / `Clear()`. When a turn ends via ResponseHandled, no outbound flows through `Send()` to trigger finalize/dismiss, so the goroutine is orphaned. Adding an explicit `DismissToolFeedback` call at the turn-end point ensures cleanup regardless of how the turn ends.

## 🧪 Test Environment
- **Hardware:** MacBook Pro (Apple Silicon)
- **OS:** macOS 15.4
- **Model/Provider:** GPT-5.5 via OpenAI-compatible proxy
- **Channels:** Feishu (websocket mode)

## 📸 Evidence (Optional)
<details>

Before fix — animation continues after turn end:
```
05:02:20 INF agent Tool output satisfied delivery; ending turn without follow-up LLM
05:02:23 DBG feishu EditMessage called content_len=210 preview="🔧 `send_file.`..."
05:02:26 DBG feishu EditMessage called content_len=211 preview="🔧 `send_file..`..."
05:02:29 DBG feishu EditMessage called content_len=210 preview="🔧 `send_file.`..."
05:02:32 DBG feishu EditMessage called content_len=211 preview="🔧 `send_file..`..."
(continues forever)
```

After fix: animation stops at turn end, card is dismissed.

</details>

## ☑️ Checklist

### DIFF
--- a/pkg/agent/adapters/channelmanager.go
+++ b/pkg/agent/adapters/channelmanager.go
@@ -44,6 +44,8 @@ func (a *channelManagerAdapter) SendPlaceholder(ctx context.Context, channel, ch
 	return a.inner.SendPlaceholder(ctx, channel, chatID)
 }
 
-func (a *channelManagerAdapter) DismissToolFeedback(ctx context.Context, channel, chatID string, outboundCtx *bus.InboundContext) {
+func (a *channelManagerAdapter) DismissToolFeedback(
+	ctx context.Context, channel, chatID string, outboundCtx *bus.InboundContext,
+) {
 	a.inner.DismissToolFeedback(ctx, channel, chatID, outboundCtx)
 }
--- a/pkg/agent/adapters/channelmanager.go
+++ b/pkg/agent/adapters/channelmanager.go
@@ -44,6 +44,6 @@ func (a *channelManagerAdapter) SendPlaceholder(ctx context.Context, channel, ch
 	return a.inner.SendPlaceholder(ctx, channel, chatID)
 }
 
-func (a *channelManagerAdapter) DismissToolFeedback(ctx context.Context, channel, chatID string) {
-	a.inner.DismissToolFeedback(ctx, channel, chatID)
+func (a *channelManagerAdapter) DismissToolFeedback(ctx context.Context, channel, chatID string, outboundCtx *bus.InboundContext) {
+	a.inner.DismissToolFeedback(ctx, channel, chatID, outboundCtx)
 }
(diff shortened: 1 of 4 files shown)


lxowalle (commented):
I think this fix still misses topic-aware tool feedback tracking on Telegram.

At the new `DismissToolFeedback` call site in `pkg/agent/pipeline_execute.go`, the cleanup only receives the raw `ts.chatID`. However, tracked tool feedback cleanup currently resolves the tracking key through `trackedToolFeedbackMessageChatID(ch, chatID, outboundCtx)` in `pkg/channels/manager.go`, and Telegram overrides that resolution via `ToolFeedbackMessageChatID(...)` to use a topic-aware key such as `chatID/topicID`.

The existing tests in `pkg/channels/manager_test.go` already codify this behavior: dismissal is expected to target `-100123/42`, not just `-100123`.

With the new `ChannelManager.DismissToolFeedback(ctx, channel, chatID)` / `Manager.DismissToolFeedback(..., nil)` path, the cleanup loses `InboundContext`, so in Telegram forum-topic conversations it will look up the wrong tracker key and fail to clear the tracked feedback entry. That means the orphaned animation goroutine can still survive in exactly the scoped-chat case that depends on resolved tool-feedback keys.

lxowalle (approved):

PR #2719, opened 2026-04-29
feat(channels): add slack_webhook output-only channel

loafoe:
## Summary

- Add new `slack_webhook` channel for pushing messages to Slack via Incoming Webhooks
- Support Block Kit formatting with markdown-to-mrkdwn conversion
- Multiple webhook targets with fallback to "default", configurable username/icon per target
- Table rendering: narrow tables as formatted text, wide tables (>60 chars) as code blocks

## Test Plan

- [x] Unit tests for markdown converter (11 test cases)
- [x] Unit tests for table splitting and rendering
- [x] Unit tests for channel validation (empty webhooks, missing default, non-HTTPS)
- [x] Unit tests for Send with mock HTTP server
- [x] Unit tests for fallback to default target
- [x] Unit tests for error classification (4xx/5xx)

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -527,6 +527,18 @@ type MQTTSettings struct {
 	QoS         int          `json:"qos,omitempty"          yaml:"-"                  env:"PICOCLAW_CHANNELS_MQTT_QOS"`
 }
 
+// SlackWebhookSettings configures the output-only Slack webhook channel.
+type SlackWebhookSettings struct {
+	Webhooks map[string]SlackWebhookTarget `json:"webhooks" yaml:"webhooks,omitempty"`
+}
+
+// SlackWebhookTarget represents a single Slack Incoming Webhook destination.
+type SlackWebhookTarget struct {
+	WebhookURL SecureString `json:"webhook_url,omitzero" yaml:"webhook_url,omitempty"`
+	Username   string       `json:"username,omitempty"   yaml:"-"`
+	IconEmoji  string       `json:"icon_emoji,omitempty" yaml:"-"`
+}
+
 type HeartbeatConfig struct {
 	Enabled  bool `json:"enabled"  env:"PICOCLAW_HEARTBEAT_ENABLED"`
 	Interval int  `json:"interval" env:"PICOCLAW_HEARTBEAT_INTERVAL"` // minutes, min 5
--- a/pkg/channels/manager.go
+++ b/pkg/channels/manager.go
@@ -688,6 +688,8 @@ func (m *Manager) getChannelConfigAndEnabled(channelName string) (*config.Channe
 		return bc, true
 	case *config.TeamsWebhookSettings:
 		return bc, true
+	case *config.SlackWebhookSettings:
+		return bc, true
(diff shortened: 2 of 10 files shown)


loafoe:
@yinwm similar to the teams webhook channel. This oen is super interesting for pushing reports / summaries / etc to Slack if there is no need for bi-direcrional integration

loafoe:
@alexhoshina can we consider this one for merging?

pkg/config/config.go
```diff
@@ -512,6 +512,18 @@ type TeamsWebhookTarget struct {
...
 }
 
+// SlackWebhookSettings configures the output-only Slack webhook channel.
+type SlackWebhookSettings struct {
+	Webhooks map[string]SlackWebhookTarget `json:"webhooks" yaml:"webhooks,omitempty"`
+}
+
+// SlackWebhookTarget represents a single Slack Incoming Webhook destination.
+type SlackWebhookTarget struct {
+	WebhookURL SecureString `json:"webhook_url,omitzero" yaml:"webhook_url,omitempty"`
```
alexhoshina, line 537: When a slack_webhook channel is decoded, this SecureString is redacted to `[NOT_HERE]` during `Channel.MarshalJSON`, but `hiddenValues` has no slack_webhook case to add the real nested webhook URLs back into `toChannelHashes`. If only the webhook URL is rotated, `Manager.Reload` sees the same hash, does not recreate the channel, and the existing channel keeps posting to the old URL until a full restart.

pkg/config/config.go
```diff
@@ -512,6 +512,18 @@ type TeamsWebhookTarget struct {
...
+	Username   string       `json:"username,omitempty"   yaml:"username,omitempty"`
+	IconEmoji  string       `json:"icon_emoji,omitempty" yaml:"icon_emoji,omitempty"`
```
alexhoshina: Because `Channel.MarshalYAML` serializes the decoded settings into `.security.yml`, these YAML tags cause `username` and `icon_emoji` to be written to the security file even though they are not secrets. Since `loadSecurityConfig` merges `.security.yml` after the main config, a later edit to these fields in the main config can be overwritten by stale values from `.security.yml`; omit them from security serialization, e.g. with `yaml:"-"`.

alexhoshina:
@loafoe Please resolve the conflicts

pkg/config/config.go
loafoe: Fixed! Added the `slack_webhook` case to `hiddenValues()` in `manager_channel.go` - it now exposes the actual webhook URLs from SecureString for hash computation, following the same pattern as `teams_webhook`. This ensures config changes are detected and the channel is properly recreated on reload.

loafoe:
@alexhoshina resolved the conflicts and processed your feedback, thanks! The failing test seems to be a transient issue (it passes locally), but I have no way to retrigger in CI. Let me know if there's anything else that needs to be addressed.

alexhoshina (approved):

alexhoshina:
LGTM!

PR #2740, opened 2026-05-01
fix(deepseek): capture reasoning_content from streaming

cjkihl:
Two fixes for DeepSeek thinking-mode compatibility:

**1. Streaming parser drops `reasoning_content`**
The `Delta` struct in `parseStreamResponse` only parsed `content` and `tool_calls`. DeepSeek streaming deltas containing `reasoning_content` (thinking-mode tokens) were silently dropped.

**Fix:** Add `ReasoningContent` to the Delta struct, accumulate it alongside content, and include it in the returned `LLMResponse`.

**2. Non-tool turn `reasoning_content` stripped for DeepSeek**
`filterDeepSeekReasoningTurn` stripped `reasoning_content` from assistant messages that were not part of a tool-interaction turn. DeepSeek-V3/V4 Chat with Thinking Mode now requires `reasoning_content` to be echoed back for ALL turns in the message history.

**Fix:** Always preserve `reasoning_content` for DeepSeek. Remove the `!hasToolInteraction` guard and old comment referencing outdated API docs.

**Tests updated:**
- `TestProviderChat_DeepSeekOmitsReasoningContentForNonToolTurnHistory` → now asserts `reasoning_content` is preserved
- `TestProviderChat_DeepSeekDocsReplayRequirements` → now asserts plain turns preserve `reasoning_content`
- `TestProviderChat_HistoryCanonicalizationMatrix` deepseek case → turn1 now expects preserved `reasoning_content`

### DIFF
--- a/pkg/providers/openai_compat/provider.go
+++ b/pkg/providers/openai_compat/provider.go
@@ -419,6 +419,7 @@ func parseStreamResponse(
 	onChunk func(accumulated string),
 ) (*LLMResponse, error) {
 	var textContent strings.Builder
+	var reasoningContent strings.Builder
 	var finishReason string
 	var usage *UsageInfo
 
@@ -451,7 +452,8 @@ func parseStreamResponse(
 		var chunk struct {
 			Choices []struct {
 				Delta struct {
-					Content   string `json:"content"`
+					Content          string `json:"content"`
+					ReasoningContent string `json:"reasoning_content"`
 					ToolCalls []struct {
 						Index    int    `json:"index"`
 						ID       string `json:"id"`
@@ -486,6 +488,11 @@ func parseStreamResponse(
 			if onChunk != nil {
 				onChunk(textContent.String())
 			}
+		}
+
+		// Accumulate reasoning content (e.g. DeepSeek thinking-mode tokens)
+		if choice.Delta.ReasoningContent != "" {
+			reasoningContent.WriteString(choice.Delta.ReasoningContent)
 		}
 
 		// Accumulate tool call deltas
@@ -544,10 +551,11 @@ func parseStreamResponse(
 	}
 
 	return &LLMResponse{
-		Content:      textContent.String(),
-		ToolCalls:    toolCalls,
-		FinishReason: finishReason,
-		Usage:        usage,
+		Content:          textContent.String(),
+		ReasoningContent: reasoningContent.String(),
+		ToolCalls:        toolCalls,
+		FinishReason:     finishReason,
+		Usage:            usage,
 	}, nil
 }
 
--- a/pkg/providers/openai_compat/provider.go
+++ b/pkg/providers/openai_compat/provider.go
@@ -255,14 +255,6 @@ func filterDeepSeekReasoningMessages(messages []Message) []Message {
 }
 
 func filterDeepSeekReasoningTurn(messages []Message) []Message {
-	hasToolInteraction := false
-	for _, msg := range messages {
-		if msg.Role == "tool" || (msg.Role == "assistant" && len(msg.ToolCalls) > 0) {
-			hasToolInteraction = true
-			break
-		}
-	}
-
 	out := make([]Message, 0, len(messages))
 	for _, msg := range messages {
 		if messageutil.IsTransientAssistantThoughtMessage(msg) {
@@ -270,13 +262,10 @@ func filterDeepSeekReasoningTurn(messages []Message) []Message {
 		}
 
 		cloned := msg
-		// DeepSeek thinking-mode replay only requires reasoning_content for
-		// turns that participate in a tool interaction round. For plain
-		// assistant turns between two user messages, the docs say the API will
-		// ignore reasoning_content on replay, so we strip it here.
-		if cloned.Role == "assistant" && strings.TrimSpace(cloned.ReasoningContent) != "" && !hasToolInteraction {
-			cloned.ReasoningContent = ""
-		}
+		// DeepSeek thinking-mode requires reasoning_content to be echoed
+		// back for ALL assistant turns. If missing the API returns 400:
+		// "The reasoning_content in the thinking mode must be passed back
+		// to the API."
 		if assistantMessageEmpty(cloned) {
 			continue
 		}
--- a/pkg/providers/openai_compat/provider.go
+++ b/pkg/providers/openai_compat/provider.go
@@ -255,6 +255,14 @@ func filterDeepSeekReasoningMessages(messages []Message) []Message {
 }
 
 func filterDeepSeekReasoningTurn(messages []Message) []Message {
+	hasToolInteraction := false
+	for _, msg := range messages {
+		if msg.Role == "tool" || (msg.Role == "assistant" && len(msg.ToolCalls) > 0) {
+			hasToolInteraction = true
+			break
+		}
+	}
+
 	out := make([]Message, 0, len(messages))
 	for _, msg := range messages {
 		if messageutil.IsTransientAssistantThoughtMessage(msg) {
@@ -262,10 +270,13 @@ func filterDeepSeekReasoningTurn(messages []Message) []Message {
 		}
 
 		cloned := msg
-		// DeepSeek thinking-mode requires reasoning_content to be echoed
-		// back for ALL assistant turns. If missing the API returns 400:
-		// "The reasoning_content in the thinking mode must be passed back
-		// to the API."
+		// DeepSeek thinking-mode replay only requires reasoning_content for
+		// turns that participate in a tool interaction round. For plain
+		// assistant turns between two user messages, the docs say the API will
+		// ignore reasoning_content on replay, so we strip it here.
+		if cloned.Role == "assistant" && strings.TrimSpace(cloned.ReasoningContent) != "" && !hasToolInteraction {
+			cloned.ReasoningContent = ""
+		}
 		if assistantMessageEmpty(cloned) {
 			continue
 		}
--- a/pkg/providers/openai_compat/provider_test.go
+++ b/pkg/providers/openai_compat/provider_test.go
@@ -1195,6 +1195,57 @@ func TestProviderChatStream_CustomHeadersInjected(t *testing.T) {
 	}
 }
 
+func TestProviderChatStream_ParsesReasoningContent(t *testing.T) {
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "text/event-stream")
+		_, _ = w.Write([]byte(
+			"data: {\"choices\":[{\"delta\":{\"reasoning_content\":\"Let me \",\"content\":\"Checking \",\"tool_calls\":[{\"index\":0,\"id\":\"call_1\",\"function\":{\"name\":\"get_weather\",\"arguments\":\"{\\\"city\\\":\"}}]}}]}\n\n",
+		))
+		_, _ = w.Write([]byte(
+			"data: {\"choices\":[{\"delta\":{\"reasoning_content\":\"think step by step.\",\"content\":\"the weather\",\"tool_calls\":[{\"index\":0,\"function\":{\"arguments\":\"\\\"Hangzhou\\\"}\"}}]},\"finish_reason\":\"tool_calls\"}],\"usage\":{\"prompt_tokens\":10,\"completion_tokens\":6,\"total_tokens\":16}}\n\n",
+		))
+		_, _ = w.Write([]byte("data: [DONE]\n\n"))
+	}))
+	defer server.Close()
+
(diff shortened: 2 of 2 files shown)


pkg/providers/openai_compat/provider.go
```diff
@@ -255,28 +255,18 @@ func filterDeepSeekReasoningMessages(messages []Message) []Message {
...
-		// assistant turns between two user messages, the docs say the API will
-		// ignore reasoning_content on replay, so we strip it here.
-		if cloned.Role == "assistant" && strings.TrimSpace(cloned.ReasoningContent) != "" && !hasToolInteraction {
-			cloned.ReasoningContent = ""
-		}
+		// DeepSeek thinking-mode (V3/V4+ Chat with Thinking Mode) requires
+		// reasoning_content to be echoed back for ALL assistant turns in the
+		// message history. If it is missing, the API returns a 400 error:
+		// "The reasoning_content in the thinking mode must be passed back
+		// to the API." We always preserve it.
```
lc6464: This removes the tool-interaction distinction that was added to match DeepSeek's documented replay behavior.

For DeepSeek thinking mode, replaying `reasoning_content` is required for tool-interaction turns, but plain assistant turns between two user messages do not require it and the API ignores it if it is sent.

Preserving reasoning for every assistant turn is therefore not a compatibility fix; it regresses the documented contract.

Please keep the streaming `reasoning_content` parsing, but restore the `hasToolInteraction` gating here.

pkg/providers/openai_compat/provider_test.go
```diff
@@ -301,9 +301,14 @@ func TestProviderChat_DeepSeekOmitsReasoningContentForNonToolTurnHistory(t *test
...
+			"reasoning_content should be preserved for DeepSeek non-tool turns",
+		)
+	}
+	if assistantMsg["reasoning_content"] != "Let me think... 1+1=2" {
 		t.Fatalf(
-			"reasoning_content should be omitted for DeepSeek non-tool turns, got %v",
+			"reasoning_content = %v, want 'Let me think... 1+1=2'",
 			assistantMsg["reasoning_content"],
 		)
 	}
```
lc6464, line 309: This expectation is the opposite of the current DeepSeek docs.

A plain non-tool assistant turn should not require `reasoning_content` on replay.

Flipping this test bakes the docs regression into the suite.

pkg/providers/openai_compat/provider_test.go
```diff
@@ -492,11 +497,11 @@ func TestProviderChat_HistoryCanonicalizationMatrix(t *testing.T) {
...
-			t.Fatalf(
-				"turn1 reasoning_content should be stripped for DeepSeek non-tool turn, got %v",
-				msgs[1]["reasoning_content"],
-			)
+		if _, ok := msgs[1]["reasoning_content"]; !ok {
+			t.Fatalf("turn1 reasoning_content should be preserved for DeepSeek, but was missing")
+		}
+		if msgs[1]["reasoning_content"] != "plain thought" {
+			t.Fatalf("turn1 reasoning_content = %v, want 'plain thought'", msgs[1]["reasoning_content"])
 		}
```
lc6464, line 500: Same issue here: the matrix test currently protects the distinction between plain turns and tool-interaction turns.

Reversing the expectation for the plain DeepSeek turn removes an important regression guard.

pkg/providers/openai_compat/provider_test.go
```diff
@@ -602,11 +607,11 @@ func TestProviderChat_DeepSeekDocsReplayRequirements(t *testing.T) {
...
 	}
-	if _, exists := plainAssistant["reasoning_content"]; exists {
-		t.Fatalf(
-			"plain DeepSeek turn should omit reasoning_content on replay, got %v",
-			plainAssistant["reasoning_content"],
-		)
+	if _, exists := plainAssistant["reasoning_content"]; !exists {
+		t.Fatalf("plain DeepSeek turn should preserve reasoning_content, but was missing")
+	}
+	if plainAssistant["reasoning_content"] != "I know this from general knowledge." {
```
lc6464: This test was explicitly documenting the vendor rule that plain turns may omit `reasoning_content`, while tool-interaction turns must preserve it.

Reversing it changes the test from a docs lock to a docs violation.

lc6464 (changes requested):
This PR mixes one valid fix with a docs-level regression.

Capturing `reasoning_content` from streaming deltas is a real bug and should be fixed. However, the change in `filterDeepSeekReasoningTurn()` goes in the opposite direction of DeepSeek's documented replay rules.

According to the DeepSeek thinking-mode and multi-round chat docs, a plain assistant turn between two user messages does **not** need to replay `reasoning_content` when there was no tool interaction; if it is sent anyway, the API ignores it. Only turns that participate in a tool-interaction round must preserve and replay `reasoning_content` in later requests.

Because this PR removes the tool-interaction check and updates the tests to require replay for all assistant turns, it changes previously-correct behavior into behavior that contradicts the vendor docs.

Please keep the streaming parsing fix, but restore the existing non-tool-turn filtering logic and add a streaming-specific regression test instead of flipping the documented replay tests.

pkg/providers/openai_compat/provider.go
cjkihl: Thanks for feedback! I completely agree. I'll adress this.

lc6464:
@cjkihl Thank you for your attention to this issue!

You can refer to the implementation in my PR #2741. After you have resolved the review comments in this PR and signed the CLA, I will forward your current PR to the maintainers for further review.

After your PR is merged, I will close my PR or use it for other purposes.

You can either copy the diff of my PR directly or create your own fix—as long as it resolves the current issue correctly.

Since you were the first to discover this issue, the credit for this contribution should go to you.

cjkihl:
Thank you for your feedback! I have adressed the concerns, please review and let me know if you need any changes.  Found a related issue, for dedections of deepseek models from 3rd party providers, both are needed to run deepseek with for example Opencode Zen Go Subscription. https://github.com/sipeed/picoclaw/pull/2743

lc6464:
> Thank you for your feedback! I have adressed the concerns, please review and let me know if you need any changes. Found a related issue, for dedections of deepseek models from 3rd party providers, both are needed to run deepseek with for example Opencode Zen Go Subscription. #2743

You're right, that's a good point. However, I think that simply returning a thought message based on keyword matching `deepseek` might not be a good strategy. If it involves deepseek-v3 or even older versions, returning a thought message will not only not help, but may even be rejected by strict providers.

My personal suggestion is to retain the current logic of determining DeepSeek based solely on the Provider and API Base. Regardless of the actual provider used, `deepseek` can be used as the Provider value, and the API Base can be modified accordingly. In other words, I personally prefer to view DeepSeek as an independent protocol (although it largely reuses code from OpenAI-compatible APIs), similar to Anthropic/Google Gemini, rather than a standard OpenAI-compatible protocol.

Of course, you could consider explicitly checking whether it starts with `deepseek-v4-`, but in any case, you should consider whether your modifications might cause certain edge cases to not work properly—the current code, while it may require explicitly setting `deepseek` as the value of the Provider, at least guarantees that it can work with the OpenAI-compatible API when the provider does not support `reasoning_content` postbacks.

There does not appear to be any precedent in the current project for determining the protocol based on the model name.

lc6464 (approved):
The current code changes now look correct to me, and they are aligned with the minimal streaming `reasoning_content` fix I expected.

I personally approve this change and will forward it to maintainers with write access for the next step of review and merge handling.

One non-blocking note: the PR description should be updated to match the current implementation. It still describes the earlier behavior that preserved `reasoning_content` for all DeepSeek turns, while the current code no longer does that. Updating the PR body will help maintainers with write access evaluate the PR correctly without being misled by outdated context.

Please also note that the actual merge may happen only after the International Workers' Day holiday period in China.

cjkihl:
Additional commit: preserve reasoning_content for all DeepSeek turns

After deploying only the streaming capture fix (commit e0f5ccff), the 400 error persisted:
Error from provider (DeepSeek): The `reasoning_content` in the thinking mode must be passed back to the API.

Why the streaming fix alone wasn't enough:
The filterDeepSeekReasoningTurn() function was still stripping reasoning_content from assistant messages in non-tool interaction turns. DeepSeek's API behind this deployment (via opencode.ai/zen/go/v1) rejects requests where any assistant message is missing reasoning_content — not just tool-interaction turns as the docs suggest. The error occurs at iteration 1 when a session with prior assistant messages is loaded, confirming it's a replay requirement, not a streaming issue.

What changed in this commit:
Removed the hasToolInteraction guard and associated dead code from filterDeepSeekReasoningTurn(). reasoning_content is now always preserved for DeepSeek assistant messages, matching the API's actual (if undocumented) behavior.

Verification:
Tested end-to-end with a session that was previously hitting the 400 error on every message. After this fix + clearing the stale session (which had messages saved before the streaming capture fix), new conversations work without errors.

lc6464 (changes requested):
For the third-party deployment, it is suggested that create a new provider for it, instead of change the behaviors of DeepSeek official provider.

You can also apply `deepseek` model keyword matching in the specific provider.

For the `deepseek` provider, the better method is keeping behaviors as official documents.

lc6464:
My personal suggestion is that you clear past sessions and try again. The problem you are encountering may not necessarily come from the code itself, but more likely from past sessions where `reasoning_content` has been completely removed.

You can back up and remove the existing `.picoclaw` folder, then re-run `onboard` and start the conversation again; DeepSeek V4 will then be available.

Additionally, `ChatStream` is not currently included in the call chain for requesting large model APIs, so the main branch can already work well with DeepSeek V4—as long as it is not affected by historical sessions.

lc6464:
Alternatively, you can simply back up and empty the `sessions` folder, then execute `/clear` in the foreground and test again.

Through testing by **many** developers on tag v0.2.8 (6e1fab80e2f1da03605565e2dce20019d6be256f), DeepSeek V4 has been able to run well in v0.2.8—your PR for fixing ChatStream is valuable, but this method has not yet been included in the call chain.

cjkihl:
> For the third-party deployment, it is suggested that create a new provider for it, instead of change the behaviors of DeepSeek official provider.
> 
> You can also apply `deepseek` model keyword matching in the specific provider.
> 
> For the `deepseek` provider, the better method is keeping behaviors as official documents.

Thank you for the feedback. I've reverted the hasToolInteraction removal. PR now contains only the streaming reasoning_content parsing fix + the new streaming regression test. The official DeepSeek provider behavior is restored to match the documented replay rules.

For the third-party proxy issue (opencode.ai/zen/go/v1 requiring reasoning_content on all turns), I'll open a separate PR with a dedicated provider rather than modifying the DeepSeek official behavior

lc6464:
> Thank you for the feedback. I've reverted the hasToolInteraction removal. PR now contains only the streaming reasoning_content parsing fix + the new streaming regression test. The official DeepSeek provider behavior is restored to match the documented replay rules.
> 
> For the third-party proxy issue (opencode.ai/zen/go/v1 requiring reasoning_content on all turns), I'll open a separate PR with a dedicated provider rather than modifying the DeepSeek official behavior 

Thank you for your contribution!

The keyword matching in your closed PR is a good idea for a specific provider like OpenCode Zen Go. You can introduce it to your new provider.

I will review again tomorrow (UTC+8). The maintainers may review and merge this PR after 2026-05-05 if they approve it.

pkg/providers/openai_compat/provider.go
afjcjsbx, line 416: `parseStreamResponse` treats each `data:` line as a complete JSON payload and silently skips `json.Unmarshal` failures. SSE events are allowed to span multiple `data:` lines, so a provider, proxy, or future backend that folds one JSON event across lines will lose `content`, `reasoning_content`, tool-call fragments, `usage`, or `finish_reason` without surfacing an error. That can leave callers with truncated output and an incorrect fallback `FinishReason` of `stop`. Parse by SSE event boundary (blank line), join all `data:` lines, and unmarshal once; for malformed payloads, prefer returning an error or at least logging/marking the response as partial.

pkg/providers/openai_compat/provider.go
```diff
@@ -451,7 +452,8 @@ func parseStreamResponse(
 		var chunk struct {
 			Choices []struct {
 				Delta struct {
-					Content   string `json:"content"`
+					Content          string `json:"content"`
+					ReasoningContent string `json:"reasoning_content"`
```
afjcjsbx, line 456: The new stream logic only reads `delta.reasoning_content,` while the non-streaming parser already preserves `reasoning` and `reasoning_details`. That makes `Chat()` and `ChatStream()` behave differently for OpenAI-compatible backends that expose reasoning in another field or schema. The fix is to centralize reasoning extraction in a shared helper or expand the stream delta schema to cover all supported reasoning variants instead of baking a DeepSeek-specific field into the generic adapter.

PR #2743, opened 2026-05-01
fix(openai_compat): detect DeepSeek models behind non-deepseek proxies

cjkihl:
`isDeepSeekReasoningProvider()` only detected DeepSeek by `providerName == "deepseek"` or `apiBase` host `== "deepseek.com"`. Users proxying DeepSeek models through third-party endpoints (e.g. `opencode.ai`, `avian.io`, etc.) would not match either check, causing `prepareMessagesForRequest` to route into `stripReasoningMessages` which unconditionally blanks all `reasoning_content` — including for tool-interaction turns where DeepSeek requires it.

**Fix:** Add `isDeepSeekModel()` which checks if the model name contains `"deepseek"`. Pass the model through `prepareMessagesForRequest` → `isDeepSeekReasoningProvider` so the routing decision accounts for the model, not just the provider name and host.

**Changes:**
- `provider.go`: Add `model` parameter to `prepareMessagesForRequest` and `isDeepSeekReasoningProvider`. Add `isDeepSeekModel()` helper. Update call site in `buildRequestBody`.
- `provider_test.go`: Add `TestProviderChat_DeepSeekModelViaProxyPreservesReasoningForToolTurns` and `TestIsDeepSeekReasoningProvider_ModelDetection` (5 sub-cases).

**Detection matrix:**

| providerName | apiBase host | model name | Old result | New result |
|---|---|---|---|---|
| `"openai"` | `opencode.ai` | `deepseek-v4-flash` | ❌ false (stripped) | ✅ true (preserved) |
| `"deepseek"` | any | any | ✅ true | ✅ true |
| any | `deepseek.com` | any | ✅ true | ✅ true |
| `"openai"` | `openai.com` | `gpt-5.4` | ❌ false | ❌ false |

This does **not** change the existing `filterDeepSeekReasoningTurn` behavior for non-tool vs tool-interaction turns — those rules remain aligned with DeepSeek documentation.

### DIFF
--- a/pkg/providers/openai_compat/provider.go
+++ b/pkg/providers/openai_compat/provider.go
@@ -144,7 +144,7 @@ func (p *Provider) buildRequestBody(
 
 	requestBody := map[string]any{
 		"model":    model,
-		"messages": common.SerializeMessages(p.prepareMessagesForRequest(messages)),
+		"messages": common.SerializeMessages(p.prepareMessagesForRequest(messages, model)),
 	}
 
 	// When fallback uses a different provider (e.g. DeepSeek), that provider must not inject web_search_preview.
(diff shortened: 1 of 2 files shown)


cjkihl:
Please also see: https://github.com/sipeed/picoclaw/pull/2740 . Both are needed to run deepseek with a 3rd party provider. (In my case OpenCode Zen go)

cjkihl:
Closing this one. The better solutions is to use "deepseek" as the provider fileld for proxies. 

Example:

```
{
  "model_name": "deepseek-v4-flash-go",
  "provider": "deepseek",
  "model": "deepseek-v4-flash",
  "api_base": "https://opencode.ai/zen/go/v1",
  "api_keys": ["..."]
},
```

PR #2751, opened 2026-05-03
fix(pid): verify gateway identity before blocking startup on stale PID

### LINKED ISSUE #2720: [BUG] Singleton PID check doesn't verify process identity — stale PID causes crash loop
## Summary

## Environment

- Linux amd64, Go 1.25.9
- Running as systemd service

## Steps to Reproduce

## Observed Behavior

```
```

## Expected Behavior

## Suggested Fix

## Workaround

```bash
systemctl stop picoclaw
rm ~/.picoclaw/.picoclaw.pid
systemctl start picoclaw
```

mrigangha:
…ng gateway alive check

## 📝 Description

Fix stale PID file causing gateway startup failure when a foreign process 
reuses the PID of a previously crashed gateway instance. Added isGatewayAlive() 
health check to verify the PID belongs to an actual picoclaw gateway before 
blocking startup.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue

Fixes #2720

(https://github.com/sipeed/picoclaw/issues/2720)
The fix is suggested in the issue.

Fix stale PID file causing gateway startup failure when a foreign process 
reuses the PID of a previously crashed gateway instance. Added isGatewayAlive() 
health check to verify the PID belongs to an actual picoclaw gateway before 
blocking startup.

## 🧪 Test Environment
- **Hardware:** PC
- **OS:**  Ubuntu 22.04 

## ☑️ Checklist

### DIFF
--- a/pkg/pid/pidfile.go
+++ b/pkg/pid/pidfile.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net/http"
 	"os"
 	"path/filepath"
 	"sync"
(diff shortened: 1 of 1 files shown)


wj-xiao:
Thanks for working on this and for addressing the stale PID-file case. The health-check direction makes sense, but I think there are two details that need to be tightened before this is safe to merge.

1. `isGatewayAlive` currently hard-codes `localhost` and does not use the `host` stored in the PID file. If the existing gateway was started with a different bind/probe host, this could fail to detect a valid running gateway.

2. The check only verifies that `/health` returns HTTP 200. It does not decode the health response and compare the reported PID with `data.PID`, so an unrelated service on the recorded port could still be mistaken for the gateway.

A possible direction would be to build the health URL from the PID file’s host and port, decode the `/health` response, and only treat the existing instance as alive when the response PID matches the PID in the PID file. It would also be good to add regression tests for PID reuse, health PID mismatch, and non-`localhost` host handling.

PR #2752, opened 2026-05-03
feat: improve model configuration workflows

SiYue-ZO:
## 📝 Description

This PR improves the model configuration workflow in the web UI and backend API.

It adds upstream model fetching and saved model catalogs, introduces provider-aware validation and selection in add/edit forms, and adds a basic model connectivity test flow in the editor. It also tightens model identifier validation on the backend.

I'll post some pictures as a demonstration.
image
img_v3_0211b_19264354-f1e6-4464-82b5-60967e78c0cg

In fact, this update also enhanced other functions, and I will continue to improve it.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue

N/A

- **Reference URL:** N/A
- **Reasoning:**
  - Add backend endpoints for fetching upstream model lists, saving/retrieving cached catalogs, and testing model connectivity.
  - Unify provider metadata in a frontend registry so labels, icons, defaults, aliases, and common models stay consistent.
  - Improve add/edit flows with provider comboboxes, model validation, quick-fill options from common/catalog/fetched models, and JSON field validation.

## 🧪 Test Environment
- **Hardware:** PC
- **OS:** Linux
- **Model/Provider:** DeepSeek
- **Channels:** Web frontend / backend API

## 📸 Evidence (Optional)
<details>

Not included.

</details>

## ☑️ Checklist


SiYue-ZO:
@alexhoshina Please review

SiYue-ZO:
@afjcjsbx Please review

alexhoshina:
Sry, I'm not particularly familiar with React. Please ask @wj-xiao to review

SiYue-ZO:
> Sry, I'm not particularly familiar with React. Please ask @wj-xiao to review

ok thanks

web/frontend/src/components/models/catalog-dialog.tsx
```diff
@@ -0,0 +1,325 @@
...
+    try {
+      const modelsToAdd = entry.models.filter((m) => catalogSelected.has(m.id))
+      for (const model of modelsToAdd) {
+        const modelRef = entry.provider
+          ? `${entry.provider}/${model.id}`
+          : model.id
+        await addModel({
+          model_name: model.id,
+          provider: entry.provider || undefined,
+          model: modelRef,
```
wj-xiao: This builds the saved `model` as `${entry.provider}/${model.id}` while also sending `provider` separately. The backend treats an explicit `provider` as authoritative and preserves `model` as the provider-local model ID, so this can persist invalid IDs.

Example: importing an OpenAI catalog entry for `gpt-4o` saves `provider: "openai"` and `model: "openai/gpt-4o"`. That model ID is then sent to OpenAI as `openai/gpt-4o` instead of `gpt-4o`. For OpenRouter, an upstream ID like `openai/gpt-4o` would become `openrouter/openai/gpt-4o`.

Please save `model.id` unchanged when `provider` is set.

web/frontend/src/components/models/model-validation.ts
```diff
@@ -0,0 +1,115 @@
...
+  const slashIdx = trimmed.indexOf("/")
+  if (slashIdx === -1) {
+    // No provider prefix
+    if (selectedProvider) {
+      return {
+        level: "warning",
+        messageKey: "models.validation.useProvider",
+        messageParams: { provider: selectedProvider },
+        fix: `${selectedProvider}/${trimmed}`,
+      }
```
wj-xiao, line 66: When a provider is selected, the suggested fix prefixes the model ID with that provider, but the form already saves the provider in the separate `provider` field. This creates the same invalid double-encoding as the catalog import path.

Example: with provider `openai` selected, entering `gpt-4o` suggests `openai/gpt-4o`; saving then sends `provider: "openai"` and `model: "openai/gpt-4o"`, instead of the provider-local `model: "gpt-4o"`.

With an explicit provider selected, a plain upstream model ID should be considered valid, or the save path should strip only the selected provider prefix before sending it to the backend.

web/frontend/src/components/models/provider-registry.ts
```diff
@@ -0,0 +1,446 @@
...
+    priority: 100,
+    commonModels: ['gpt-4o', 'gpt-4o-mini', 'gpt-4-turbo', 'o1', 'o3-mini'],
+    aliases: ['gpt'],
+  },
+  {
+    key: 'anthropic',
+    label: 'Anthropic',
+    iconSlug: 'anthropic',
+    domain: 'anthropic.com',
+    defaultApiBase: 'https://api.anthropic.com',
```
wj-xiao: The Anthropic registry default is `https://api.anthropic.com`, but API-key Anthropic configs in the backend use `https://api.anthropic.com/v1` when no `api_base` is provided. Since the frontend auto-fills and saves this value, newly added Anthropic configs will bypass the backend default and call the wrong base URL.

Please either change the frontend default to `https://api.anthropic.com/v1` or avoid persisting the default API base so the backend can apply its known default.

wj-xiao:
Personal suggestion: from a user-flow perspective, I think there are a few places where the new model configuration workflow could be made a bit easier to understand and less error-prone.
Please just take these as reference points and feel free to decide whether any of them are worth including in this PR.

1. Keep the provider/model mental model consistent. Once a provider is selected, the model field could generally contain only the upstream model ID, without a provider prefix. This would make add/edit, fetched models, and catalog import follow the same rule.

2. Avoid auto-fetching into an expected failure state when an API key is required but missing. The fetch dialog could first show an API-key-required state, or allow the user to enter a temporary key inside the dialog.

3. Consider adding a review step for catalog imports from remote providers. Since the catalog only stores a masked key, directly importing models may create entries that are not usable until credentials are re-entered.

4. Make Test Connection able to test the current unsaved form values. Users usually want to adjust model/API base/API key, test that combination, and then save.

5. Consider making backend provider metadata and default API bases the source of truth for the frontend, so defaults do not drift between the UI and backend routing behavior.

web/frontend/src/components/models/catalog-dialog.tsx
SiYue-ZO: Thanks for catching this. Fixed: the model field now uses model.id directly instead of prepending ${entry.provider}/, since the provider is already sent as a separate field and ExtractProtocol treats it as authoritative.

web/frontend/src/components/models/model-validation.ts
SiYue-ZO: When selectedProvider is already set, a plain model ID like gpt-4o is provider-local and valid as-is. Changed the validation to return success with the parsed provider+model instead of suggesting a redundant prefix fix.

web/frontend/src/components/models/provider-registry.ts
SiYue-ZO: Frontend: defaultApiBase now includes /v1 suffix to match the backend default.
Backend: the "anthropic" generic HTTP provider path now uses common.NormalizeBaseURL (same as anthropic-messages), so even if a user manually enters https://api.anthropic.com without /v1, it will be normalized correctly.

SiYue-ZO:
Hi @wj-xiao, thanks a lot for the detailed review and thoughtful suggestions.

I’ve made a round of updates based on your comments, along with some additional refinements on my side to help improve the overall model configuration flow and make it more robust.

Could you please take another look when you have time? I’d really appreciate another review to see if these changes help move things in a better direction for the project.

Thanks again!

web/frontend/src/components/models/edit-model-sheet.tsx
```diff
@@ -280,267 +328,360 @@ export function EditModelSheet({
...
+        open={testOpen}
+        onClose={() => setTestOpen(false)}
+        inlineParams={{
+          provider: form.provider,
+          model: form.modelId,
+          apiBase: form.apiBase,
+          apiKey: form.apiKey,
+          authMethod: form.authMethod,
+        }}
+      />
```
wj-xiao, line 687: The edit-page connection test return false negatives for models that already have a saved API key.

Repro:
1. Save a remote model, e.g. OpenAI, with a valid API key.
2. Open that model in the edit sheet.
3. Do not type a new API key. The form correctly keeps `form.apiKey` empty because the real key is not exposed to the frontend.
4. Click "Test Connection".

`EditModelSheet` always passes `inlineParams`, so `TestModelDialog` calls `/api/models/test-inline` instead of `/api/models/{index}/test`. The inline request sends an empty `api_key`, so the backend treats the model as unconfigured/unreachable even though the saved config has a valid key.

A safer fix would be to have the inline test accept the saved model index and merge unsaved form fields with the stored secret/default API base server-side, or fall back to `/api/models/{index}/test` when the user has not entered a replacement key.

web/backend/api/models.go
```diff
@@ -614,3 +621,303 @@ func maskAPIKey(key string) string {
...
+		}
+		root = strings.TrimRight(root, "/")
+		fetchURL = root + "/api/tags"
+		return fetchOllamaModels(ctx, fetchURL)
+	default:
+		// OpenAI-compatible: /v1/models
+		fetchURL = apiBase + "/models"
+		return fetchOpenAICompatibleModels(ctx, fetchURL, apiKey)
+	}
+}
```
wj-xiao, line 879: The fetch-models endpoint is exposed for providers that are not actually handled by this implementation.

Repro:
1. Open Add Model.
2. Select a non OpenAI-compatible provider such as Gemini, Anthropic, or Azure.
3. Enter the required credential.
4. Click "Fetch Available Models".

The frontend allows the flow, but the backend only special-cases Ollama and treats every other provider as OpenAI-compatible by calling `GET {apiBase}/models` with `Authorization: Bearer ...`. That contract is not valid for several providers in the registry: Gemini, Anthropic, Azure, Bedrock, etc. use different list-model/deployment endpoints and/or different authentication headers/query parameters.

This will produce stable failures or misleading fetch errors for those providers. Either implement provider-specific fetch logic, or only enable this action for providers known to support the OpenAI-compatible `/models` endpoint plus Ollama.

SiYue-ZO:
@wj-xiao If you have time, please review again, thanks

web/frontend/src/components/models/edit-model-sheet.tsx
```diff
@@ -124,117 +148,145 @@ export function EditModelSheet({
...
-      setError(providerError)
-      return
-    }
     if (!form.modelId.trim()) {
       setError(t("models.add.errorRequired"))
       return
     }
+    if (modelValidation?.level === "error") return
+
+    let extraBody: Record<string, unknown> | undefined
```
wj-xiao, line 258: Editing a model can no longer clear `extra_body` or `custom_headers`.

In the edit form, empty textareas leave `extraBody` / `customHeaders` as `undefined`, and `JSON.stringify` omits those fields from the request. The backend update handler treats omitted fields as “preserve the existing value”; it only clears the stored value when it receives an explicit empty object `{}`.

This means a user can delete the JSON from the UI, save successfully, and still keep sending the old extra body / custom headers on future requests.

Please restore the previous contract by sending `{}` when the user clears these fields, or update the backend to distinguish “field omitted” from “field present but empty”.

web/frontend/src/components/models/provider-registry.ts
```diff
@@ -0,0 +1,466 @@
+/**
```
wj-xiao, line 1: The provider picker now relies on the frontend hard-coded `PROVIDERS` registry instead of the backend `provider_options` returned by `/api/models`.

That makes the UI diverge from the backend’s authoritative provider catalog. For example, backend-supported providers such as bedrock, anthropic-messages, qwen-us, coding-plan, coding-plan-anthropic, claude-cli, codex-cli, and elevenlabs are not present in the new frontend list. Backend policy fields like create_allowed, default_model_allowed, default_auth_method, and auth_method_locked are also no longer driving the Add/Edit forms.

The frontend registry is useful for presentation data such as labels, icons, aliases, and common models, but provider availability and behavioral rules should still come from `provider_options` so the UI stays aligned with what the backend actually supports.

web/frontend/src/components/models/provider-combobox.tsx
```diff
@@ -0,0 +1,195 @@
...
+            </span>
+          ) : (
+            <span className="text-muted-foreground">
+              {placeholder || t("models.combobox.selectProvider")}
+            </span>
+          )}
+          <IconChevronDown className="ml-2 size-4 shrink-0 opacity-50" />
+        </Button>
+      </PopoverTrigger>
+      <PopoverContent className="w-[--radix-popover-trigger-width] p-0">
```
wj-xiao: The provider combobox popover does not scroll correctly when used inside the Add/Edit model sheet.

`PopoverContent` is always rendered through a portal to `document.body`, while the sheet is a Radix Dialog/Sheet that locks scrolling outside the dialog content. As a result, the provider list is visually displayed above the sheet, but wheel events are not handled by the popover list; scrolling the mouse wheel over the provider popup scrolls the underlying sheet/drawer instead.

Please render the popover inside the sheet content, or otherwise make the popover content participate in the dialog’s allowed scroll area. At minimum, the command list should capture wheel scrolling so the provider list scrolls independently instead of moving the drawer.

wj-xiao:
Thank you for putting so much work into this PR. The model configuration flow is clearly becoming more capable.

It looks like the branch currently has merge conflicts with `main`. Could you please rebase or merge the latest `main` and resolve the conflicts when you have a chance? After that, we can continue reviewing the remaining issues more easily.

SiYue-ZO:
@wj-xiao I have resolved the above issue. Please review it again. Thank you. Thank you for your multiple reviews.

web/frontend/src/components/models/provider-combobox.tsx
```diff
@@ -0,0 +1,220 @@
...
+            </span>
+          ) : (
+            <span className="text-muted-foreground">
+              {placeholder || t("models.combobox.selectProvider")}
+            </span>
+          )}
+          <IconChevronDown className="ml-2 size-4 shrink-0 opacity-50" />
+        </Button>
+      </PopoverTrigger>
+      <PopoverContent className="w-[--radix-popover-trigger-width] p-0" container={containerRef?.current}>
```
wj-xiao, line 125: This reads `containerRef.current` during render, which fails the current frontend lint rules (`react-hooks/refs`). `pnpm lint` exits with an error here, so this PR cannot merge as-is. Please move the container resolution out of render, or pass a stable container value through state/effect.

web/backend/api/models.go
```diff
@@ -614,3 +633,324 @@ func maskAPIKey(key string) string {
...
+		APIBase:    strings.TrimSpace(req.APIBase),
+		AuthMethod: strings.TrimSpace(req.AuthMethod),
+	}
+	if req.APIKey != "" {
+		m.SetAPIKey(req.APIKey)
+	}
+
+	// When api_key is empty and model_index is provided, fall back to stored credentials.
+	// This lets the edit form test unsaved field changes while using the saved key.
+	if req.APIKey == "" && req.ModelIndex != nil {
```
wj-xiao, line 718: This fallback can reuse the saved API key with arbitrary unsaved request fields. In the edit flow, a caller can provide a new `api_base` while leaving `api_key` empty, and the backend will attach the stored key to that new endpoint during the probe. Please only reuse the stored key when the provider/API base match the saved model, or require the user to explicitly enter a key for changed endpoints.

web/backend/api/models.go
```diff
@@ -614,3 +633,324 @@ func maskAPIKey(key string) string {
...
+		result["error"] = "Endpoint unreachable"
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	json.NewEncoder(w).Encode(result)
+}
+
+// probeModelConnectivity performs a real network probe to verify model endpoint reachability.
+func probeModelConnectivity(m *config.ModelConfig) bool {
+	apiBase := modelProbeAPIBase(m)
```
wj-xiao, line 765: Inline tests for remote providers with an empty `api_base` will currently fail even when the provider has a known default base. `modelProbeAPIBase` only falls back to defaults for empty-key/local providers, so OpenAI/DeepSeek/etc. saved with only an API key get `apiBase == ""` and are reported unreachable. Please resolve the default provider API base before probing.

web/frontend/src/components/models/edit-model-sheet.tsx
```diff
@@ -280,267 +336,364 @@ export function EditModelSheet({
...
-                hint={t("models.field.toolSchemaTransformHint")}
-              >
-                <Input
-                  value={form.toolSchemaTransform}
-                  onChange={setField("toolSchemaTransform")}
-                  placeholder="google"
-                />
-              </Field>
+                <Field
+                  label={t("models.field.workspace")}
```
wj-xiao, line 549: This removes the previous `default_model_allowed` guard from the default-model toggle. Users can now select “set as default” for providers that the backend rejects as default chat models, causing the model add/update to succeed and the later `setDefaultModel` call to fail. Please restore the disabled state/hint based on backend `default_model_allowed`.

web/frontend/src/components/models/edit-model-sheet.tsx
```diff
@@ -280,267 +336,364 @@ export function EditModelSheet({
...
-              >
-                <Input
-                  value={form.requestTimeout}
-                  onChange={setField("requestTimeout")}
-                  placeholder="60"
-                  type="number"
-                  min={0}
-                />
-              </Field>
+              <SwitchCardField
```
wj-xiao, line 507: Same issue as the add flow: the default toggle no longer respects d`efault_model_allowed`. For an existing ASR-only or otherwise non-defaultable provider, saving can produce a partial success followed by a backend rejection from `setDefaultModel`. Please keep this toggle disabled for unsupported providers and clear/handle existing invalid defaults explicitly.

web/backend/api/model_catalog.go
```diff
@@ -0,0 +1,161 @@
...
+	return fmt.Sprintf("%s|%s|%x", provider, apiBase, hash[:6])
+}
+
+// maskAPIKeyValue masks an API key for display, keeping first 4 and last 4 chars.
+func maskAPIKeyValue(key string) string {
+	key = strings.TrimSpace(key)
+	if key == "" {
+		return ""
+	}
+	if len(key) <= 8 {
```
wj-xiao, line 57: For 9-12 character keys this mask reveals too much of the secret; for example a 9-character key exposes 8 characters and hides only one. Please reuse the existing `maskAPIKey` policy or otherwise guarantee a minimum hidden percentage for catalog display.

wj-xiao:
Thanks for continuing to iterate on this. The overall direction looks useful, but this PR now crosses several separate areas: provider metadata, add/edit model form behavior, backend fetch/catalog APIs, inline connectivity testing, credential handling, new UI primitives/dependencies, and i18n. Because these areas have different risk profiles, reviewing and stabilizing them all in one PR is becoming difficult.

Would you consider splitting this into a few smaller, feature-focused PRs?

1. **Provider selection and model form foundation**
   Provider registry/combobox, `provider_options` merging, add/edit form updates, model ID validation, and Anthropic default API base normalization. This one should preserve the existing backend-driven policies such as `create_allowed`, `default_model_allowed`, and `auth_method_locked`.

2. **Fetch Models and Saved Catalogs**
   `/api/models/fetch`, catalog persistence/list/delete, fetch dialog, and catalog dialog. This would be easier to review with focused backend tests around provider allow-listing, catalog masking, default API base behavior, and fetch error handling.

3. **Test Connection**
   Saved/inline model test endpoints and the test dialog. This is worth reviewing separately because it involves network probes and reuse of stored credentials. The credential reuse boundary and default `api_base` fallback should be covered by tests here.

A 3-PR split should keep the work reasonably grouped by user-facing feature without making the review too fragmented. I think this would make it much easier to get each part merged confidently.

SiYue-ZO:
> Thanks for continuing to iterate on this. The overall direction looks useful, but this PR now crosses several separate areas: provider metadata, add/edit model form behavior, backend fetch/catalog APIs, inline connectivity testing, credential handling, new UI primitives/dependencies, and i18n. Because these areas have different risk profiles, reviewing and stabilizing them all in one PR is becoming difficult.
> 
> Would you consider splitting this into a few smaller, feature-focused PRs?
> 
>     1. **Provider selection and model form foundation**
>        Provider registry/combobox, `provider_options` merging, add/edit form updates, model ID validation, and Anthropic default API base normalization. This one should preserve the existing backend-driven policies such as `create_allowed`, `default_model_allowed`, and `auth_method_locked`.
> 
>     2. **Fetch Models and Saved Catalogs**
>        `/api/models/fetch`, catalog persistence/list/delete, fetch dialog, and catalog dialog. This would be easier to review with focused backend tests around provider allow-listing, catalog masking, default API base behavior, and fetch error handling.
> 
>     3. **Test Connection**
>        Saved/inline model test endpoints and the test dialog. This is worth reviewing separately because it involves network probes and reuse of stored credentials. The credential reuse boundary and default `api_base` fallback should be covered by tests here.
> 
> 
> A 3-PR split should keep the work reasonably grouped by user-facing feature without making the review too fragmented. I think this would make it much easier to get each part merged confidently.

Thanks for reviewing. I now realize that this PR should indeed be split. Please let me resubmit the PR after resolving the issue.

SiYue-ZO:
@wj-xiao I've split this PR into 3 smaller PRs as requested:

1. **[#2831](https://github.com/sipeed/picoclaw/pull/2831) — Provider Selection & Model Form Foundation**
   Provider registry, combobox, model validation, CRUD endpoints, add/edit sheets

2. **[#2832](https://github.com/sipeed/picoclaw/pull/2832) — Fetch Models & Saved Catalogs**
   `/api/models/fetch`, catalog endpoints, FetchModelsDialog, CatalogDialog

3. **[#2833](https://github.com/sipeed/picoclaw/pull/2833) — Test Connection**
   `/api/models/{index}/test`, `/api/models/test-inline`, TestModelDialog

PR1 is the base. PR2 and PR3 are independent of each other and can be reviewed in parallel after PR1 merges. All 12 review issues from the previous rounds are addressed in the split.

This PR can be closed in favor of the three new ones.

wj-xiao:
The first PR has been merged. Could you please resolve the conflicts in the remaining two PRs?

SiYue-ZO:
> The first PR has been merged. Could you please resolve the conflicts in the remaining two PRs?

Okay, I'm dealing with it

PR #2758, opened 2026-05-04
fix(telegram): media group album handling

bogdanovich:
## Summary

- buffer Telegram media group album updates and process them as one inbound message
- preserve album captions/text by selecting the caption/text message as the primary context
- attach album media in Telegram message order using the same per-message collector as normal single messages
- make the media group idle delay configurable via `media_group_delay_ms`
- guard stale timer callbacks with a generation token and flush pending groups during shutdown
- protect Telegram `chatIDs` writes because media group timer callbacks process messages asynchronously

## Why

Telegram sends albums as separate updates with the same `media_group_id`, and the caption is usually attached to only one of those updates. Processing each update independently can lose the combined album context and caption.

This mirrors the OpenClaw Telegram handling pattern: configurable short debounce, sort by message id, choose a caption/text primary message, then process the album once. The timer generation guard prevents an old debounce callback from flushing a newer batch after `Timer.Stop()` loses the race with an already-running callback.

## Testing

- `go test ./pkg/channels/telegram`
- `go test -race ./pkg/channels/telegram`

### DIFF
--- a/pkg/channels/telegram/telegram.go
+++ b/pkg/channels/telegram/telegram.go
@@ -1060,7 +1060,10 @@ func (c *TelegramChannel) collectTelegramMessageParts(
 	if msg.Voice != nil {
 		voicePath := c.downloadFile(ctx, msg.Voice.FileID, ".ogg")
 		if voicePath != "" {
-			parts.mediaPaths = append(parts.mediaPaths, storeMedia(voicePath, indexedMediaFilename("voice", ".ogg", index, total)))
+			parts.mediaPaths = append(
+				parts.mediaPaths,
+				storeMedia(voicePath, indexedMediaFilename("voice", ".ogg", index, total)),
+			)
 			parts.content = append(parts.content, "[voice]")
 		}
 	}
--- a/pkg/channels/telegram/telegram.go
+++ b/pkg/channels/telegram/telegram.go
@@ -11,6 +11,7 @@ import (
 	"net/url"
 	"os"
 	"regexp"
+	"slices"
 	"strconv"
 	"strings"
 	"sync"
@@ -43,20 +44,38 @@ var (
 	reInlineCode = regexp.MustCompile("`([^`]+)`")
 )
 
+const defaultMediaGroupDelay = 500 * time.Millisecond
+
 type TelegramChannel struct {
 	*channels.BaseChannel
-	bot      *telego.Bot
-	bh       *th.BotHandler
-	bc       *config.Channel
-	chatIDs  map[string]int64
-	ctx      context.Context
-	cancel   context.CancelFunc
-	tgCfg    *config.TelegramSettings
-	progress *channels.ToolFeedbackAnimator
+	bot       *telego.Bot
+	bh        *th.BotHandler
+	bc        *config.Channel
+	chatIDsMu sync.Mutex
+	chatIDs   map[string]int64
+	ctx       context.Context
+	cancel    context.CancelFunc
+	tgCfg     *config.TelegramSettings
+	progress  *channels.ToolFeedbackAnimator
 
 	registerFunc      func(context.Context, []commands.Definition) error
 	commandRegDelayFn func(int) time.Duration
 	commandRegCancel  context.CancelFunc
+
+	mediaGroupMu    sync.Mutex
+	mediaGroups     map[string]*telegramMediaGroup
+	mediaGroupDelay time.Duration
+}
+
+type telegramMediaGroup struct {
+	messages   []*telego.Message
+	timer      *time.Timer
+	generation uint64
+}
+
+type telegramMessageParts struct {
+	content    []string
+	mediaPaths []string
 }
 
 func NewTelegramChannel(
@@ -112,9 +131,19 @@ func NewTelegramChannel(
 		bc:          bc,
 		chatIDs:     make(map[string]int64),
 		tgCfg:       telegramCfg,
+
+		mediaGroups:     make(map[string]*telegramMediaGroup),
+		mediaGroupDelay: telegramMediaGroupDelay(telegramCfg),
 	}
 	ch.progress = channels.NewToolFeedbackAnimator(ch.EditMessage)
 	return ch, nil
+}
+
+func telegramMediaGroupDelay(telegramCfg *config.TelegramSettings) time.Duration {
+	if telegramCfg != nil && telegramCfg.MediaGroupDelayMS > 0 {
+		return time.Duration(telegramCfg.MediaGroupDelayMS) * time.Millisecond
+	}
+	return defaultMediaGroupDelay
 }
 
 func (c *TelegramChannel) Start(ctx context.Context) error {
@@ -167,6 +196,7 @@ func (c *TelegramChannel) Stop(ctx context.Context) error {
 	if c.bh != nil {
 		_ = c.bh.StopWithContext(ctx)
 	}
+	c.flushPendingMediaGroups(ctx)
 
 	// Cancel our context (stops long polling)
 	if c.cancel != nil {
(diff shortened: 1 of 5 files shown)


pkg/channels/telegram/telegram.go
```diff
@@ -178,6 +191,14 @@ func (c *TelegramChannel) Stop(ctx context.Context) error {
 	if c.commandRegCancel != nil {
 		c.commandRegCancel()
 	}
+	c.mediaGroupMu.Lock()
```
afjcjsbx: `Stop` stops each timer and deletes the pending mediaGroups entry without flushing it first. Any album received during the 500ms buffer window has already been acknowledged by the Telegram handler, so a normal restart/deploy will silently lose user input permanently. Drain or synchronously flush pending groups before shutdown returns.

pkg/channels/telegram/telegram.go
```diff
@@ -713,6 +734,103 @@ func (c *TelegramChannel) SendMedia(ctx context.Context, msg bus.OutboundMediaMe
...
+		group = &telegramMediaGroup{}
+		c.mediaGroups[key] = group
+	}
+	group.messages = append(group.messages, &msgCopy)
+	if group.timer != nil {
+		group.timer.Stop()
+	}
+	delay := c.mediaGroupDelay
+	if delay <= 0 {
+		delay = 500 * time.Millisecond
```
afjcjsbx: This code infers album completeness from a hard-coded idle timeout. Telegram does not guarantee that all parts of a media group arrive within 500ms, so jitter or backpressure will flush early and turn one logical album into multiple inbound messages. If you keep timeout-based assembly, it needs a stronger design, configurability, and tests for staggered delivery.

afjcjsbx (changes requested):

pkg/channels/telegram/telegram.go
```diff
@@ -713,6 +736,122 @@ func (c *TelegramChannel) SendMedia(ctx context.Context, msg bus.OutboundMediaMe
 }
 
 func (c *TelegramChannel) handleMessage(ctx context.Context, message *telego.Message) error {
+	if message != nil && strings.TrimSpace(message.MediaGroupID) != "" {
+		return c.bufferMediaGroupMessage(ctx, message)
+	}
+	return c.handleMessages(ctx, []*telego.Message{message})
+}
+
+func (c *TelegramChannel) bufferMediaGroupMessage(ctx context.Context, message *telego.Message) error {
```
afjcjsbx, line 752: in `bufferMediaGroupMessage` timer-reset logic, calls `group.timer.Stop()` and ignores the result. That is incorrect. In Go, `Stop() == false` means the old callback may already be running, so a stale timer can flush the album immediately after a new item is appended. Failure mode: albums split unpredictably around the debounce boundary. Action: add a generation/version token per batch and only let the latest generation flush.

pkg/channels/telegram/telegram.go
```diff
@@ -713,6 +736,122 @@ func (c *TelegramChannel) SendMedia(ctx context.Context, msg bus.OutboundMediaMe
...
+	}
+	if err := c.handleMessages(ctx, messages); err != nil {
+		logger.ErrorCF("telegram", "Failed to handle media group", map[string]any{
+			"key":   key,
+			"error": err.Error(),
+		})
+	}
+}
+
+func (c *TelegramChannel) handleMessages(ctx context.Context, messages []*telego.Message) error {
```
afjcjsbx, line 857: in `handleMessage` -> `bufferMediaGroupMessage / handleMessages` split, forks inbound parsing logic instead of reusing it cleanly. Only photo handling was lifted into the multi-message loop; voice/audio/document remain singleton logic on one "representative" message. That is logic fragmentation. Failure mode: grouped non-photo attachments are silently dropped, and this path will drift from normal single-message handling. Action: extract one per-message collector helper and aggregate its outputs across the group. If the feature is intentionally photo-only, enforce that explicitly and document it as photo-only.

pkg/channels/telegram/telegram.go
```diff
@@ -57,6 +60,15 @@ type TelegramChannel struct {
...
 	commandRegCancel  context.CancelFunc
+
+	mediaGroupMu    sync.Mutex
+	mediaGroups     map[string]*telegramMediaGroup
+	mediaGroupDelay time.Duration
+}
+
+type telegramMediaGroup struct {
+	messages []*telego.Message
+	timer    *time.Timer
```
afjcjsbx: the timer callback now runs `handleMessages` asynchronously, but that path still writes `c.chatIDs[platformID] = chatID` on a plain map. You introduced a concurrent execution path with no synchronization. Failure mode: two timer flushes, or a timer flush racing a live update, can crash the process with concurrent map writes. Action: protect chatIDs with a mutex or replace it with sync.Map, then audit every field touched from the timer goroutine.

afjcjsbx (changes requested):

pkg/channels/telegram/telegram.go
bogdanovich: Fixed in the latest push. I added a `generation` token to each pending media group batch and pass that generation into the timer callback. `flushMediaGroup` now ignores stale callbacks when the stored generation no longer matches, so a callback that races with `Timer.Stop()` cannot flush a newer batch. Added `TestFlushMediaGroupIgnoresStaleTimerGeneration` for this case.

pkg/channels/telegram/telegram.go
bogdanovich: Fixed. I extracted `collectTelegramMessageParts` and now `handleMessages` calls it for every message in the group, aggregating text/captions plus photo, voice, audio, and document media paths. This keeps grouped message handling on the same per-message collection path instead of keeping a photo-only fork.

pkg/channels/telegram/telegram.go
bogdanovich: Fixed. I added `chatIDsMu` around writes to `chatIDs`. I also re-ran `go test -race ./pkg/channels/telegram` after the media-group timer changes; it passes locally.

bogdanovich:
@afjcjsbx I addressed the review feedback in the latest push and resolved the threads.

Summary:
- added a media-group generation token so stale debounce callbacks cannot flush newer batches
- extracted per-message media collection and aggregate it across grouped updates
- guarded `chatIDs` writes with a mutex
- kept pending group shutdown flush synchronous

Validation:
- `go test ./pkg/channels/telegram`
- `go test -race ./pkg/channels/telegram`

Could you take another look when you have a chance?

bogdanovich:
Rebased this PR on the current `upstream/main` and reran the Telegram album tests after the latest feedback.

Validation rerun after rebase:
- `go test ./pkg/channels/telegram -count=1`
- `go test -race ./pkg/channels/telegram -count=1`

The branch still includes the earlier fixes for the review points:
- synchronous flush of pending media groups during shutdown
- configurable debounce window (`media_group_delay_ms`)
- generation token to ignore stale timer callbacks
- shared per-message collector aggregated across the whole media group
- synchronized `chatIDs` writes for the async timer path

Could you take another look when you have a chance?

bogdanovich:
I pushed a follow-up lint fix: `5aba20e8 fix(telegram): wrap long voice media append`.\n\nCurrent CI interpretation:\n- `Tests` were already passing\n- `Linter` was failing only because `golines` wanted one long append call wrapped\n- `Security Check` is not caused by this PR logic; it is reporting Go 1.25.9 standard-library vulnerabilities in the shared workflow environment (fixed in Go 1.25.10), which affects repository-wide analysis rather than this Telegram album patch specifically

bogdanovich:
@afjcjsbx Could you please re-review this?

bogdanovich:
Rebased on current `upstream/main` and reran telegram-channel validation:

- `go test ./pkg/channels/telegram -count=1`
- `go test -race ./pkg/channels/telegram -count=1`

The earlier review threads on this PR look resolved/outdated from my side, so this should be ready for another pass.

afjcjsbx (approved):
LGTM!

bogdanovich:
@afjcjsbx Would you merge it as well, please?

PR #2759, opened 2026-05-04
fix(seahorse): retrieval tools to current session

bogdanovich:
## Summary
- scope short_grep to the active tool session by default
- keep all_conversations=true as the explicit cross-session search escape hatch
- scope short_expand to the active session so guessed message IDs cannot expand other conversations

## Tests
- go test ./pkg/seahorse
- go test ./pkg/agent ./pkg/tools ./pkg/tools/shared ./pkg/seahorse

### DIFF
--- a/pkg/seahorse/short_retrieval.go
+++ b/pkg/seahorse/short_retrieval.go
@@ -44,6 +44,7 @@ type GrepInput struct {
 	Pattern          string     `json:"pattern"`
 	Scope            string     `json:"scope,omitempty"` // "both" (default), "summary", or "message"
 	Role             string     `json:"role,omitempty"`  // "user", "assistant", or "" (all)
+	ConversationID   int64      `json:"conversationId,omitempty"`
 	AllConversations bool       `json:"allConversations,omitempty"`
 	Since            *time.Time `json:"since,omitempty"`
 	Before           *time.Time `json:"before,omitempty"`
@@ -120,6 +121,7 @@ func (r *RetrievalEngine) Grep(ctx context.Context, input GrepInput) (*GrepResul
 		Pattern:          input.Pattern,
 		Mode:             mode,
 		Role:             input.Role,
+		ConversationID:   input.ConversationID,
 		AllConversations: input.AllConversations,
 		Since:            since,
 		Before:           input.Before,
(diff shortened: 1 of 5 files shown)


pkg/seahorse/short_retrieval.go
```diff
@@ -193,8 +195,43 @@ func (r *RetrievalEngine) Grep(ctx context.Context, input GrepInput) (*GrepResul
 	return result, nil
 }
 
+// ConversationIDForSession returns the storage conversation ID for a session key.
+func (r *RetrievalEngine) ConversationIDForSession(ctx context.Context, sessionKey string) (int64, error) {
```
afjcjsbx: ConversationIDForSession returns `0, nil` for an empty or unknown session, and the store only scopes searches when `ConversationID > 0`. Failure case: brand-new session, missing tool session metadata, or lookup miss -> `short_grep` searches every conversation anyway. That is a cross-conversation data leak. Action: change the contract to `id, found, err`; if `!all_conversations && !found`, return empty results or an explicit error. Add tests for empty and unknown session keys.

pkg/seahorse/tool_expand.go
```diff
@@ -82,7 +82,12 @@ func (t *ExpandTool) Execute(ctx context.Context, args map[string]any) *tools.To
 		}
 	}
 
-	result, err := t.engine.ExpandMessages(ctx, messageIDs)
+	conversationID, err := t.engine.ConversationIDForSession(ctx, tools.ToolSessionKey(ctx))
+	if err != nil {
+		return tools.ErrorResult("Expand failed: resolve current conversation: " + err.Error())
+	}
+
+	result, err := t.engine.ExpandMessagesScoped(ctx, messageIDs, conversationID, false)
```
afjcjsbx: `ExpandMessagesScoped(..., false)` hard-codes current-conversation behavior even when the caller intentionally used `short_grep` with `all_conversations: true`. Failure case: grep returns a valid ID from another conversation, expand silently drops it, and the user gets a misleading partial result. Action: add `all_conversations` or `{message_id, conversation_id}` input to `short_expand`, and report rejected IDs instead of silently discarding them.

afjcjsbx (changes requested):

bogdanovich:
I rechecked the two requested points on this branch and rebased it on current `upstream/main`.

What is now on the branch:
- `ConversationIDForSession` returns `(id, found, err)`
- `short_grep` no longer falls back to cross-conversation search when the current session is empty/unknown; when `all_conversations` is false and no scoped conversation is found, it returns an empty scoped result with a hint instead
- `short_expand` accepts `all_conversations` and reports `rejectedMessageIds` when scoped expansion filters out IDs from other conversations
- there is an explicit unknown-session test for expand, plus the scoped/all-conversations expand tests

Validation rerun after rebase:
- `go test ./pkg/seahorse -run 'Test(ExpandTool|Retrieval).*' -count=1`

Please take another look when convenient.

bogdanovich:
Rebased on current `upstream/main` and reran focused validation:

- `go test ./pkg/seahorse ./pkg/agent ./pkg/tools ./pkg/tools/shared -count=1`

The earlier review threads on this PR look resolved/outdated from my side, so this should be ready for another pass.

PR #2770, opened 2026-05-05
Add MCP section to config web UI

Gabrielsv01:
## 📝 Description

Add a dedicated MCP section to the web Config page so users can manage MCP enable/discovery settings and MCP servers directly from the UI.

This removes the need to edit raw config for common MCP setup tasks and fixes server removal persistence by sending null entries in patch payloads.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue

N/A

- **Reference URL:** N/A
- **Reasoning:**
  - Add MCP controls to Config UI (enabled/discovery/ttl/max_search_results/use_bm25/use_regex).
  - Add MCP server editor for stdio/http modes with validation.
  - Serialize MCP servers into tools.mcp.servers patch payload.
  - Preserve delete semantics by sending removed servers as null in merge patch payload.

## 🧪 Test Environment
- **Hardware:** PC
- **OS:** macOS
- **Model/Provider:** N/A (UI/config feature)
- **Channels:** N/A (launcher web config page)

## 📸 Evidence (Optional)
<details>

Frontend build succeeded after changes (vite build completed successfully).

</details>

## ☑️ Checklist


web/frontend/src/components/config/config-page.tsx
```diff
@@ -214,10 +266,139 @@ export function ConfigPage() {
...
+
+        const removedServerEntries = Array.from(baselineServerNames)
+          .filter((name) => !currentServerNames.has(name))
+          .map((name) => [name, null] as const)
+
+        const upsertServerEntries = normalizedServers.map((server) => {
+          const deferredPatch =
+            server.deferredOverride === null
+              ? {}
+              : { deferred: server.deferredOverride }
```
wj-xiao: When an existing MCP server is changed back to “inherit” discovery mode, `server.deferredOverride` becomes `null`, but this builds an empty patch `{}` instead of sending `{ deferred: null }`.

Since the backend applies JSON Merge Patch semantics, omitting `deferred` keeps the previous `deferred: true` or `deferred: false` value in the saved config. The UI will appear to save successfully, but the server will not actually return to inherited discovery behavior.

web/frontend/src/components/config/config-page.tsx
```diff
@@ -214,10 +266,139 @@ export function ConfigPage() {
...
+              min: 0,
+            })
+          : Number.parseInt(EMPTY_FORM.mcpDiscoveryTTL, 10)
+        const mcpDiscoveryMaxSearchResults = mcpDiscoveryValidationEnabled
+          ? parseIntField(
+              form.mcpDiscoveryMaxSearchResults,
+              "MCP discovery max search results",
+              { min: 0 },
+            )
+          : Number.parseInt(EMPTY_FORM.mcpDiscoveryMaxSearchResults, 10)
```
wj-xiao: The problematic flow is:

1. The user has custom MCP discovery values saved, for example `ttl: 30` or `max_search_results: 20`.
2. The user disables MCP or MCP discovery, which hides these inputs in the UI.
3. The user saves any config change.
4. This code replaces the hidden discovery values with `EMPTY_FORM` defaults and still sends them in the patch.

Step 4 is the issue. Saving unrelated changes while discovery is hidden overwrites the user’s existing discovery settings, even though the user did not edit those fields.

web/frontend/src/components/config/config-page.tsx
```diff
@@ -214,10 +266,144 @@ export function ConfigPage() {
...
+            {
+              ...deferredPatch,
+              enabled: server.enabled,
+              type: "stdio",
+              command: server.command,
+              args: parseMultilineList(server.argsText),
+              env: parseJSONObjectField(
+                server.envText,
+                `MCP server ${server.name} env`,
+              ),
```
wj-xiao: The `headers: null` / `env: null` cases are fine for clearing the whole field when switching transports. The remaining issue is deleting individual keys while the server stays on the same transport type.

For example, if the existing config has `headers: { "Authorization": "old", "X-Test": "1" }` and the user removes `Authorization` in the UI, this code sends only `headers: { "X-Test": "1" }`. Since the backend applies JSON Merge Patch recursively, that patch merges into the existing object and does not remove omitted keys, so `Authorization` remains in the saved config.

To persist key deletion, the patch should include removed map entries as `null`, e.g. `headers: { "Authorization": null, "X-Test": "1" }` and similarly for `env`, or use another replace strategy for these nested map fields.

web/frontend/src/i18n/locales/en.json
```diff
@@ -701,9 +701,40 @@
       "allowed_cidrs": "Allowed Network CIDRs",
       "allowed_cidrs_hint": "Only clients from these CIDR ranges can access the service. One per line or comma-separated. Leave empty to allow all.",
       "allowed_cidrs_placeholder": "192.168.1.0/24\n10.0.0.0/8",
+      "mcp_section_hint": "Configure MCP servers without editing config.json manually.",
+      "mcp_enabled": "Enable MCP",
+      "mcp_enabled_hint": "Turn MCP server integration on or off.",
+      "mcp_discovery_enabled": "Enable MCP Discovery",
+      "mcp_discovery_enabled_hint": "Allow MCP discovery tools to search registered MCP servers.",
+      "mcp_discovery_ttl": "Discovery Cache TTL (minutes)",
+      "mcp_discovery_ttl_hint": "How long MCP discovery results stay cached.",
```
wj-xiao: The English copy describes MCP discovery TTL as a cache duration in minutes, but the backend uses `tools.mcp.discovery.ttl` as the temporary unlock TTL for discovered hidden tools. Search results call `PromoteTools(..., ttl)`, and that TTL is decremented by `TickTTL()` during the tool-execution pipeline; it is not time-based. Please update the label/hint to avoid implying minutes or cache duration, e.g. “Discovered tool unlock TTL” and “How many tool-execution TTL ticks discovered tools remain available after search.”

web/frontend/src/components/config/config-page.tsx
```diff
@@ -214,10 +281,171 @@ export function ConfigPage() {
...
+          baseline.mcpServers
+            .map((server) => ({
+              ...server,
+              name: server.name.trim(),
+            }))
+            .filter((server) => server.name !== "")
+            .map((server) => [server.name, server] as const),
+        )
+
+        const upsertServerEntries = normalizedServers.map((server) => {
```
wj-xiao: `handleSave` validates and serializes every MCP server whenever the config form is dirty, even when MCP is disabled. Since the MCP server editor is hidden when `mcpEnabled` is false, an existing disabled or incomplete MCP server can prevent users from saving unrelated config changes, and they cannot fix the invalid server from this page.

Please either keep the server editor accessible when MCP is disabled, or only require `command` / `url` for servers that are actually enabled and relevant to the active MCP config.

wj-xiao (approved):

PR #2788, opened 2026-05-06
feat(session): add per-message created_at timestamps

### LINKED ISSUE #2787: [BUG] Session messages lack individual timestamps — all messages share session.updated time
## Problem

The Session API (`GET /api/sessions/{id}`) returns a list of messages without any per-message timestamps. The frontend has no choice but to use the session-level `updated` timestamp for all messages, making message times inaccurate.

For example, the actual timeline might be:
- 23:01 User sends "Hello"
- 23:01 Assistant replies with greeting
- 23:48 User asks "What time is it?"
- 23:48 Assistant replies with the time

But in the API response, every message shows the same timestamp (the session's `updated` time).

## Root Cause

1. **`providers.Message` struct has no timestamp field** — `pkg/providers/protocoltypes/types.go`
2. **Each JSONL line stores only the Message itself** without any creation time
3. **The API response includes no time info per message** — `web/backend/api/session.go`
4. The frontend can only fall back to `session.updated` for message timing

## Relevant Code

- `pkg/providers/protocoltypes/types.go` — Message struct definition
- `pkg/memory/jsonl.go` — JSONL storage (addMsg, etc.)
- `web/backend/api/session.go` — Session API (handleGetSession)
- `pkg/session/session_store.go` — SessionStore interface

## Suggested Direction

Add a `created_at` timestamp to each message. Options include:

1. Adding `CreatedAt` to the `Message` struct directly (simplest)
2. Introducing a `StoredMessage` wrapper to separate storage metadata from the business object
3. Handling backward compatibility for existing JSONL lines missing this field

## Environment

- Version: latest
- Components: session API, JSONL store

LiusCraft:
## Background

Session API (`GET /api/sessions/{id}`) returns messages without individual timestamps. The frontend has to use the session-level `updated` time for all messages, which is inaccurate when messages span different times.

## Changes

image

### Backend

- **`pkg/providers/protocoltypes/types.go`** — Added `CreatedAt time.Time` field to `Message` struct with `json:"created_at,omitempty"` for backward compatibility
- **`pkg/memory/jsonl.go`** — Set `CreatedAt = time.Now()` in `addMsg` when the field is zero (after transient message check), so every stored message gets a timestamp
- **`web/backend/api/session.go`** — Expose `created_at` in `sessionChatMessage` API response; pass through from `providers.Message` to all transcript message types (user, assistant, thought, tool_calls, visible tool messages); fallback to session `updated` time for legacy messages with zero `CreatedAt`
- **`pkg/memory/jsonl_test.go`** — Added `TestAddMessage_SetsCreatedAt` verifying timestamps are set on write

### Frontend

- **`src/api/sessions.ts`** — Added `created_at?: string` to `SessionDetail.messages` type
- **`src/features/chat/history.ts`** — Use `message.created_at` when available (`message.created_at ?? detail.updated`)
- **`src/components/chat/user-message.tsx`** — Added `timestamp` prop and displays formatted time below message bubble
- **`src/components/chat/assistant-message.tsx`** — Show timestamp in collapsed block header (thought, tool_calls) — previously only shown for normal assistant messages
- **`src/components/chat/chat-page.tsx`** — Pass `timestamp` to `UserMessage`

## Behavior Impact

- New messages written after this change will have accurate per-message `created_at` timestamps in API responses
- Legacy messages (existing JSONL lines without `created_at`) will use the session `updated` time as fallback — no more `0001-01-01T00:00:00Z`
- All message types now display timestamps: user, assistant, thought, tool_calls
- No change to LLM provider communication (`Message` is never directly serialized to provider APIs)

## Risks and Rollback

- **Risk:** Legacy JSONL lines lack the field — handled by `omitempty`, `IsZero()` guard, and backend fallback to `session.updated`
- **Rollback:** Revert the commits, old behavior is fully preserved

## Verification

- `go test ./pkg/memory/` — 44/44 PASS
- `go test ./web/backend/api/` — all PASS
- `go build ./pkg/providers/protocoltypes/ ./pkg/memory/ ./web/backend/api/` — all pass
- Frontend build — `pnpm build:backend` passes

## Notes

Fixes #2787

### DIFF
--- a/pkg/memory/jsonl.go
+++ b/pkg/memory/jsonl.go
@@ -561,6 +561,12 @@ func (s *JSONLStore) addMsg(sessionKey string, msg providers.Message) error {
 	l.Lock()
 	defer l.Unlock()
 
+	now := time.Now()
+
+	if msg.CreatedAt == nil {
+		msg.CreatedAt = &now
+	}
+
 	// Append the message as a single JSON line.
 	line, err := json.Marshal(msg)
 	if err != nil {
@@ -598,7 +604,6 @@ func (s *JSONLStore) addMsg(sessionKey string, msg providers.Message) error {
 	if err != nil {
 		return err
 	}
-	now := time.Now()
 	if meta.Count == 0 && meta.CreatedAt.IsZero() {
 		meta.CreatedAt = now
 	}
@@ -725,6 +730,12 @@ func (s *JSONLStore) SetHistory(
 	meta.Skip = 0
 	meta.Count = len(history)
 	meta.UpdatedAt = now
+
+	for i := range history {
+		if history[i].CreatedAt == nil {
+			history[i].CreatedAt = &now
+		}
+	}
 
 	// Write meta BEFORE rewriting the JSONL file. If we crash between
 	// the two writes, meta has Skip=0 and the old file is still intact,
--- a/pkg/providers/protocoltypes/types.go
+++ b/pkg/providers/protocoltypes/types.go
@@ -1,5 +1,7 @@
 package protocoltypes
 
+import "time"
+
 type ToolCall struct {
 	ID               string         `json:"id"`
 	Type             string         `json:"type,omitempty"`
@@ -81,6 +83,7 @@ type Attachment struct {
 type Message struct {
 	Role             string         `json:"role"`
 	Content          string         `json:"content"`
+	CreatedAt        *time.Time     `json:"created_at,omitempty"`
 	Media            []string       `json:"media,omitempty"`
 	Attachments      []Attachment   `json:"attachments,omitempty"`
 	ReasoningContent string         `json:"reasoning_content,omitempty"`
--- a/pkg/providers/protocoltypes/types.go
+++ b/pkg/providers/protocoltypes/types.go
@@ -1,5 +1,7 @@
 package protocoltypes
 
+import "time"
+
 type ToolCall struct {
 	ID               string         `json:"id"`
 	Type             string         `json:"type,omitempty"`
@@ -87,6 +89,7 @@ type Message struct {
 	Role             string         `json:"role"`
 	Content          string         `json:"content"`
 	ModelName        string         `json:"model_name,omitempty"`
+	CreatedAt        *time.Time     `json:"created_at,omitempty"`
 	Media            []string       `json:"media,omitempty"`
 	Attachments      []Attachment   `json:"attachments,omitempty"`
 	ReasoningContent string         `json:"reasoning_content,omitempty"`
--- a/pkg/agent/agent_test.go
+++ b/pkg/agent/agent_test.go
@@ -370,6 +370,8 @@ func TestProcessMessage_BtwCommandRunsWithoutPersistingHistory(t *testing.T) {
 	defaultAgent.Sessions.SetHistory(sessionKey, initialHistory)
 	defaultAgent.Sessions.SetSummary(sessionKey, "The team decided to keep state request-scoped.")
 
+	initialHistory = defaultAgent.Sessions.GetHistory(sessionKey)
+
 	response, err := al.processMessage(context.Background(), msg)
 	if err != nil {
 		t.Fatalf("processMessage() error = %v", err)
@@ -487,6 +489,8 @@ func TestProcessMessage_BtwCommandUsesIsolatedProvider(t *testing.T) {
 		{Role: "assistant", Content: "Right, keep it request-scoped."},
 	}
 	defaultAgent.Sessions.SetHistory(mainSessionKey, initialHistory)
+
+	initialHistory = defaultAgent.Sessions.GetHistory(mainSessionKey)
 
 	// Process a /btw command
 	response, err := al.processMessage(context.Background(), bus.InboundMessage{
--- a/pkg/agent/agent_test.go
+++ b/pkg/agent/agent_test.go
@@ -1062,6 +1062,8 @@ func TestProcessMessage_BtwCommandRunsWithoutPersistingHistory(t *testing.T) {
 	defaultAgent.Sessions.SetHistory(sessionKey, initialHistory)
 	defaultAgent.Sessions.SetSummary(sessionKey, "The team decided to keep state request-scoped.")
 
+	initialHistory = defaultAgent.Sessions.GetHistory(sessionKey)
+
 	response, err := al.processMessage(context.Background(), msg)
 	if err != nil {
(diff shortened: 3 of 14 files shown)


pkg/memory/jsonl.go
```diff
@@ -556,6 +556,9 @@ func (s *JSONLStore) addMsg(sessionKey string, msg providers.Message) error {
 	if messageutil.IsTransientAssistantThoughtMessage(msg) {
 		return nil
 	}
+	if msg.CreatedAt.IsZero() {
+		msg.CreatedAt = time.Now()
```
afjcjsbx: It's not optimal to assigns `time.Now()` before acquiring the per-session lock. Two concurrent writers can therefore persist messages in an order that does not match their timestamps, which breaks chronology and any merge/sort logic that trusts `created_at`. Capture one `now` only after the lock is held, then use it for both `msg.CreatedAt` and `meta.UpdatedAt`.

pkg/providers/protocoltypes/types.go
```diff
@@ -81,6 +83,7 @@ type Attachment struct {
 type Message struct {
 	Role             string         `json:"role"`
 	Content          string         `json:"content"`
+	CreatedAt        time.Time      `json:"created_at,omitempty"`
```
afjcjsbx: Using omitempty on a `time.Time` field does not omit zero values with Go's `encoding/json`. Any uninitialized message written through `SessionManager` or `rewriteJSONL` will be serialized as `0001-01-01T00:00:00Z`, which is garbage data masquerading as a real timestamp. Use `*time.Time` for an optional field, or guarantee population on every marshal path and add raw-JSON tests to prove zero timestamps are never written.

afjcjsbx (changes requested):

LiusCraft:
Thanks for the review! I've addressed all the feedback — @afjcjsbx could you take another look when you get a chance?

pkg/memory/jsonl.go
```diff
@@ -561,6 +561,11 @@ func (s *JSONLStore) addMsg(sessionKey string, msg providers.Message) error {
 	l.Lock()
 	defer l.Unlock()
 
+	now := time.Now()
+	if msg.CreatedAt == nil {
+		msg.CreatedAt = &now
```
alexhoshina, line 567: When using the JSONL backend, this timestamp is added only to the local copy passed into `addMsg`; callers such as `turnState.recordPersistedMessage` still keep the same message with `CreatedAt == nil`. After a mid-turn restore-point refresh (for example context-window retry compression), `refreshRestorePointFromSession` reads the persisted messages back with non-nil `CreatedAt` and compares them to the recorded messages with `reflect.DeepEqual`, so the current turn tail no longer matches and gets included in the rollback snapshot; aborting that turn can then leave partially persisted current-turn messages in history.

LiusCraft: @alexhoshina  The CreatedAt normalization (setting to nil on both sides before DeepEqual) in refreshRestorePointFromSession fixes the specific regression this PR introduced. However, after deeper analysis, there is a pre-existing issue with the same tail-matching logic: fields tagged with `json:"-"` — such as `ToolCall.Name`, `ToolCall.Arguments`, and `Message.PromptLayer` — are also lost during the JSONL roundtrip, so `reflect.DeepEqual` would still mismatch when comparing an in-memory persisted message (which has these fields populated) with a deserialized history message (which does not).

This means the tail-matching heuristic is fundamentally fragile — it breaks on any field that differs between the in-memory and serialized representations, not just CreatedAt.

I think the proper long-term fix is to introduce a unique MessageID so we can use set-subtraction instead of content-based DeepEqual matching. For now, the CreatedAt nil normalization is sufficient to fix this PR's regression. The json:"-" issue is pre-existing and unrelated to this PR — I'll file a separate PR to address it.

LiusCraft:
Thanks for the reviews! Both issues from the initial review have been addressed in commit `409206f0`:

1. **Concurrency**: `time.Now()` is now called **after** acquiring the per-session lock — two concurrent writers will get consistent ordering.
2. **`omitempty`**: Changed `CreatedAt` from `time.Time` to `*time.Time`, so zero values are properly omitted during JSON marshaling.

Regarding the **DeepEqual tail-matching regression** raised by @alexhoshina — `CreatedAt` is now normalized to `nil` on both sides before `reflect.DeepEqual`, which resolves the mismatch. The pre-existing issue with `json:"-"" fields is unrelated to this PR and will be handled separately.

@afjcjsbx @alexhoshina could you take another look when you get a chance? The blocking `CHANGES_REQUESTED` is still pending. Thanks!

afjcjsbx:
The main problem here is not the timestamp itself, but its ownership.

With this patch `CreatedAt` is not guaranteed by the persistence layer: it is populated manually at various runtime call sites (`pipeline_*`, `prompt_turn`, `steering`), while `AddFullMessage` continues to accept and persist messages with `CreatedAt == nil`.

This makes the invariant fragile: it only takes a future `AddFullMessage` without a timestamp to revert to writing history partially without `created_at`, at which point the backend and frontend "fix" it differently (`handleGetSession` uses `sess.Updated`, the frontend fallbacks to `detail.updated`). It is dangerous because history merge uses timestamp in signature/ordering, so we can introduce duplicate or inconsistent reorder after reload/hydration.

In my opinion the correct fix is to move the responsibility to the persistence boundary:
- normalize `CreatedAt` in the store when it is missing
- do it at least in `JSONLStore.addMsg`
- ideally also in `SetHistory`
- ideally also align the `SessionManager` fallback, so the behavior does not depend on the active backend

Call sites should set `CreatedAt` only when they want to preserve a real event time; they should not be the ones to guarantee the persistence invariant.

I would also ask for specific tests on `AddFullMessage`, `SetHistory`, and API transcripts, because currently new coverage only tests `AddMessage`, which is not the main path used by the runtime.

alexhoshina:
Please resolve the conflicts♥️

LiusCraft:
> Please resolve the conflicts♥️

@alexhoshina Resolved conflicts, PTAL

alexhoshina (approved):
lgtm

alexhoshina:
> lgtm

CC: @afjcjsbx

afjcjsbx (approved):
It might be useful to have:
- explicit test that checks created_at in the output of GET /api/sessions/{id}.
- frontend test on history merge with mixed legacy + new sessions, where some messages have created_at and some do not.

But nothing blocking, they can be added with a second PR

LGTM!

PR #2531, opened 2026-04-15
feat(tools): add delegate tool for cross-agent task handoff

is-Xiaoen:
## Summary

Implements the `delegate` tool — **PR C** from the Phase 2 plan (#2148).

`delegate(agent_id, task)` hands off a task to a specific named agent and waits synchronously for the result. The target agent runs with its own workspace, model, tools, and system prompt, satisfying D3.

### How it differs from existing tools

| | subagent | spawn | **delegate** |
|---|---|---|---|
| Mode | sync | async | sync |
| Target | generic (caller's model) | optional agent_id | **required** agent_id |
| Agent context | caller's | caller's | **target's** |
| Use case | internal helper | background task | cross-agent handoff |

## Changes

**SubTurn layer** (`pkg/agent/subturn.go`)
- `TargetAgentID` field on `SubTurnConfig` — when set, `spawnSubTurn` resolves the target `AgentInstance` from the registry instead of shallow-copying the caller
- `Model` validation relaxed: empty is accepted when `TargetAgentID` provides it implicitly
- Existing callers unaffected (field defaults to empty string → old behavior)

**Delegate tool** (`pkg/tools/delegate.go`)
- Params: `agent_id` (required), `task` (required)
- `agent_id` normalized via `routing.NormalizeAgentID` before any logic — prevents case/whitespace variants from bypassing self-check while still resolving to the same agent
- Self-delegation explicitly blocked
- Permission via `subagents.allow_agents` (D4)
- Spawner errors preserve underlying error via `.WithError(err)`
- Result attributed: `[Response from agent "X"]\n...`

**Wiring** (`pkg/agent/loop.go`)
- Auto-registered when `len(agents) > 1` — single-agent unaffected
- Independent of `subagent` enable — delegate uses SubTurn directly, not SubagentManager

**Tests** (17 new tests total)
- 13 tool-level: success, validation, permission, self-delegation, normalization bypass, nil result, spawner error
- 2 subturn-level with `modelRecordingProvider`: asserts the provider actually received `model-beta` when alpha delegates to beta; plus not-found and empty-model-accepted
- 2 wiring: single-agent has no delegate tool; multi-agent has delegate on both agents

## Scope

This PR covers delegation only (PR C + minimal wiring). It does not touch:
- Agent discovery / system prompt injection (PR A/B — #2158)
- Async agent messaging (Layer 3)
- Token budget sharing across delegation chains

The implementation does not depend on #2158, but will rebase cleanly once it lands.

## Test plan

[...]

PR #2790, opened 2026-05-07
fix(agents): route spawn tool to target agent

bogdanovich:
## Summary

- pass the `spawn` tool's optional `agent_id` through to sub-turn execution
- run the child turn with the requested agent's model, tools, workspace, skills, and injected files
- return an explicit error when the requested target agent does not exist
- mark background `spawn` sub-turns as critical so they can continue after the parent turn finishes

## Problem

The `spawn` tool already exposes an `agent_id` parameter and validates it with the configured allowlist, but the actual sub-turn execution ignored the target agent. The child turn was created from the parent/default agent instead, so `spawn(agent_id: "...")` could acknowledge the target while still running with the wrong agent configuration.

That breaks multi-agent setups where agents intentionally have different tools, workspaces, skills, or models.

## Fix

This PR adds `AgentID` to the tool/agent sub-turn config path and resolves the target agent inside `spawnSubTurn`. If no `agent_id` is provided, behavior remains unchanged: the child turn uses the parent agent and falls back to the default agent.

## Validation

```bash
go test ./pkg/tools ./pkg/agent -run 'TestSpawnTool|TestSubTurn|TestSpawn|SubTurn'
go test ./pkg/tools ./pkg/agent
```

### DIFF
--- a/pkg/agent/subturn.go
+++ b/pkg/agent/subturn.go
@@ -103,6 +103,7 @@ type subTurnRuntimeConfig struct {
 //	// Result also available in parent's pendingResults channel
 //	// Parent turn will poll and process it in a later iteration
 type SubTurnConfig struct {
+	AgentID      string
 	Model        string
 	Tools        []tools.Tool
 	SystemPrompt string
@@ -224,6 +225,7 @@ func (s *AgentLoopSpawner) SpawnSubTurn(
 
 	// Convert tools.SubTurnConfig to agent.SubTurnConfig
 	agentCfg := SubTurnConfig{
+		AgentID:            cfg.AgentID,
 		Model:              cfg.Model,
 		Tools:              cfg.Tools,
 		SystemPrompt:       cfg.SystemPrompt,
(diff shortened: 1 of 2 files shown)


afjcjsbx:
Hello! Thanks for the PR. In the meantime, #2531 (https://github.com/sipeed/picoclaw/pull/2531) was merged, your pull request now has merge conflicts. Could you please perform a rebase to resolve them? 🙏

bogdanovich:
I updated this PR branch on top of the latest `upstream/main` and resolved the subturn conflict so the target-agent routing change applies cleanly again.

Validation rerun on the rebased branch:
- `go test ./pkg/tools -run TestSpawn -count=1`
- `go test ./pkg/agent -run SubTurn -count=1`

The branch should no longer be in a conflicting state.

bogdanovich:
Closing as superseded by upstream changes. Current `upstream/main` already has `TargetAgentID` in `SubTurnConfig`, and both `spawn` and `delegate` now route through that field. The remaining rebased diff would only add compatibility logic around the older `AgentID` path, which no longer looks like the right upstream shape. If needed, the follow-up should be a small regression-test PR against current main rather than reviving this branch.

PR #2793, opened 2026-05-07
fix(tools): hidden tools promoted in the parent registry rather than the subagent's registry

bogdanovich:
## Summary

Fix deferred tool discovery inside cloned tool registries, especially subagent turns.

`ToolRegistry.Clone()` previously shallow-copied every tool entry. Discovery tools such as `BM25SearchTool` and `RegexSearchTool` store a pointer to the registry they search and promote tools in. After cloning a registry for a subagent, the cloned search tool still pointed at the parent registry.

That meant a subagent could call `tool_search_tool_bm25`, get a successful "promoted tools" result, but the hidden tools were promoted in the parent registry rather than the subagent's cloned registry. The next subagent LLM iteration still could not call the discovered tools.

## Changes

- Recreate BM25 and regex discovery tools during `ToolRegistry.Clone()` so they point at the cloned registry.
- Add a regression test that verifies searching in a clone promotes hidden tools only in that clone, not in the parent registry.

## Validation

- `go test ./pkg/tools -run 'TestToolRegistry_Clone_RebindsDiscoveryToolsToClone|TestToolRegistry_Clone' -count=1`
- `go test ./pkg/tools ./pkg/agent`

### DIFF
--- a/pkg/tools/registry.go
+++ b/pkg/tools/registry.go
@@ -30,6 +30,10 @@ type mediaStoreAware interface {
 	SetMediaStore(store media.MediaStore)
 }
 
+type registryCloneAware interface {
+	CloneForRegistry(registry *ToolRegistry) Tool
+}
+
 func NewToolRegistry() *ToolRegistry {
 	return &ToolRegistry{
 		tools: make(map[string]*ToolEntry),
@@ -413,11 +417,8 @@ func (r *ToolRegistry) Clone() *ToolRegistry {
 	}
 	for name, entry := range r.tools {
 		tool := entry.Tool
-		switch t := entry.Tool.(type) {
-		case *RegexSearchTool:
-			tool = NewRegexSearchTool(clone, t.ttl, t.maxSearchResults)
-		case *BM25SearchTool:
-			tool = NewBM25SearchTool(clone, t.ttl, t.maxSearchResults)
+		if aware, ok := entry.Tool.(registryCloneAware); ok {
+			tool = aware.CloneForRegistry(clone)
 		}
 		clone.tools[name] = &ToolEntry{
 			Tool:   tool,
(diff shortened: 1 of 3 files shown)


pkg/tools/registry.go
afjcjsbx, line 411: `ToolRegistry.Clone` now hard-codes `RegexSearchTool` and `BM25SearchTool` construction even though the registry-bound state already belongs to those tool types in `pkg/tools/search_tool.go` and `pkg/tools/search_tool.go.` This is the wrong abstraction boundary. The next registry-bound tool, or the next field added to either search tool, will silently clone wrong and either mutate the parent registry again or behave differently in subagents.
Try to introduce an optional tool-owned contract such as `CloneForRegistry(*ToolRegistry)` Tool or `RebindRegistry(*ToolRegistry)` Tool; make Clone dispatch through that interface instead of a concrete-type switch.

afjcjsbx (changes requested):

bogdanovich:
Addressed the review point.

I removed the concrete-type switch from `ToolRegistry.Clone()` and replaced it with a tool-owned optional contract: discovery tools now implement `CloneForRegistry(*ToolRegistry) Tool`, and registry cloning dispatches through that interface. That keeps the registry generic and makes future registry-bound tool types responsible for cloning their own internal state.

Also rebased the branch on current `upstream/main` and reran:
- `go test ./pkg/tools -run 'TestToolRegistry_Clone_RebindsDiscoveryToolsToClone|TestToolRegistry_Clone' -count=1`

Please take another look when convenient.

bogdanovich:
Closing this as superseded by upstream. The requested architectural shape is now already present on `upstream/main`: `ToolRegistry.Clone()` dispatches through a tool-owned `CloneForRegistry(*ToolRegistry)` contract, and the discovery-tool clone regression coverage is already in tree. This branch is no longer adding distinct value beyond what upstream already has.

PR #2813, opened 2026-05-07
fix(pid): (updated) verify gateway identity before blocking startup on stale PID

### LINKED ISSUE #2720: [BUG] Singleton PID check doesn't verify process identity — stale PID causes crash loop

mrigangha:
## 📝 Description

The singleton check in WritePidFile only verified that a process with the
recorded PID existed, but did not verify it was actually a picoclaw gateway.
When a gateway crashed without cleaning up its PID file and the OS reused
that PID for an unrelated process (e.g. systemd-resolved), the gateway would
refuse to start and enter a crash loop under systemd.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue

Fixes #2720 

- **Reference URL:**
- **Reasoning:**

## 🧪 Test Environment
- **Hardware:** PC
- **OS:** Ubuntu 22.04 

## ☑️ Checklist

### DIFF
--- a/pkg/pid/pidfile.go
+++ b/pkg/pid/pidfile.go
@@ -6,11 +6,13 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net"
 	"net/http"
 	"os"
 	"path/filepath"
 	"sync"
 	"time"
+	"strconv"
 
 	"github.com/sipeed/picoclaw/pkg/config"
 	"github.com/sipeed/picoclaw/pkg/logger"
@@ -50,7 +52,7 @@ func generateToken() string {
 // then verifies the reported PID matches expectedPID to confirm the process
 // is actually a picoclaw gateway and not a foreign service on the same port.
 func isGatewayAlive(host string, port int, expectedPID int) bool {
-	url := fmt.Sprintf("http://%s:%d/health", host, port)
+	url := "http://" + net.JoinHostPort(host, strconv.Itoa(port)) + "/health"
 	client := &http.Client{Timeout: 2 * time.Second}
 	resp, err := client.Get(url)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


weissfl:
Hey @mrigangha, any updates on this PR? The stale bot is about to close both the PR and the issue. Would be great to get this merged — the bug is still reproducing.

mrigangha:
Hi, this PR is ready for review from my side. The reported bug is still reproducible, so it would be great if someone could review and merge this when available. Posting an update to prevent the stale bot from auto-closing it. Thanks!

mrigangha:
Update just so GitHub bot doesn't mark it as stale.

mrigangha:
An update to prevent it fromgoing stale.

afjcjsbx:
Hi @mrigangha, thanks for the PR:

### 1. Missing Host Normalization in `isGatewayAlive`

`isGatewayAlive` currently uses the host from the pidfile exactly as it is, without normalizing it into a valid probe host (`pidfile.go`).

* **The Problem:** Throughout the rest of the codebase, before querying the `/health` endpoint, the value is passed through `gatewayProbeHost`. This ensures that un-probeable bind hosts (like `0.0.0.0`, `::`, or legacy hosts) are correctly transformed into a concrete loopback address (`gateway.go`, `gateway_host.go`).
* **The Impact:** As it stands, a pidfile written by an older version that saved the raw bind host could cause a perfectly healthy, active gateway to be misclassified as "stale." This will break the singleton check during restarts or upgrades.

> **Recommendation:** Pass the pidfile host through `gatewayProbeHost` before checking if the gateway is alive to maintain consistency and backward compatibility.

### 2. Ineffective Test Coverage (False Positives)

Two of the newly added tests do not actually exercise the conditional logic branch modified by this PR, creating a false sense of test coverage.

* **`TestWritePidFileHealthPIDMismatch` (`pidfile_test.go`):** The pidfile here is created using `PID: os.Getpid()`. However, `WritePidFile` completely skips the "existing process" validation check when `data.PID == os.Getpid()` (`pidfile.go`).
* **`TestWritePidFileForeignPIDReuse` (`pidfile_test.go`):** This test uses `PID: 1`, which is explicitly excluded from the newly introduced condition in `pidfile.go`.

Out of the three tests added in this PR, only the one using `os.Getppid()` is actually verifying your new logic. Please refactor the other two tests so they properly hit and test the target code paths.

PR #2819, opened 2026-05-07
feat(cli): Add non-destructive /reset command for session history

bogdanovich:
## Summary
- add a non-destructive `/reset` command that starts a fresh session without deleting stored history
- persist per-route session overrides in workspace state so future messages in the same conversation continue in the new session
- add `/reset clear` to remove the override and return to the default routed session
- add tests for reset behavior and state persistence

## Why this is needed
`/clear` currently clears persisted session history, and with the Seahorse context manager that also clears the session's SQLite conversation state. That is much more destructive than many users expect from a conversational "reset" action.

This change adds a softer operation:
- `/reset` starts a fresh session for future messages in the current conversation
- existing history is preserved
- `/reset clear` returns the conversation to the original routed session

That gives users a safe way to start over without losing previous context permanently.

## Behavior notes
- `/clear` remains destructive and unchanged
- `/reset` only affects the current routed conversation/session boundary
- explicit session keys still take precedence over reset overrides

## Verification
- `go test ./pkg/state ./pkg/commands`
- `go test ./pkg/agent -run 'TestProcessMessage_Reset' -count=1`

### DIFF
--- a/pkg/agent/agent_command.go
+++ b/pkg/agent/agent_command.go
@@ -335,6 +335,28 @@ func (al *AgentLoop) buildCommandsRuntime(
 			return al.contextManager.Clear(ctx, opts.SessionKey)
 		}
 
+		rt.ResetSession = func(clear bool) (string, error) {
+			if opts == nil {
+				return "", fmt.Errorf("process options not available")
+			}
+			routeSessionKey := strings.TrimSpace(opts.Dispatch.RouteSessionKey)
+			if routeSessionKey == "" {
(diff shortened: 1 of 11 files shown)


bogdanovich:
Closing in favor of #2820. PR #2819 demonstrated a workable non-destructive reset via session-key override, but that approach is better treated as a workaround than the final design because it behaves more like branching than a true fresh-session reset and introduces confusing command semantics such as .

PR #2823, opened 2026-05-08
fix(agent): dismiss tool feedback when outbound is skipped

bogdanovich:
## Summary
- dismiss tracked tool feedback when the final outbound is skipped because the  tool already replied to the same chat
- add a regression test covering  in that path

## Testing
- go test ./pkg/agent -run 'TestPublishResponseIfNeeded_DismissesToolFeedbackWhenMessageToolAlreadySent|TestProcessMessage_PublishesToolFeedbackWhenEnabled|TestNewAgentLoop' -count=1

### DIFF
--- a/pkg/agent/agent_outbound.go
+++ b/pkg/agent/agent_outbound.go
@@ -56,6 +56,16 @@ func (al *AgentLoop) PublishResponseIfNeeded(ctx context.Context, channel, chatI
 	}
 
 	if alreadySentToSameChat {
+		if al.channelManager != nil && channel != "" && chatID != "" {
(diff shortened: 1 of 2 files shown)


bogdanovich:
Follow-up summary:

- dismiss tracked tool feedback when the final outbound is skipped because the `message` tool already replied to the same chat
- add a regression test covering `PublishResponseIfNeeded` in that path

Tested with:
- `go test ./pkg/agent -run 'TestPublishResponseIfNeeded_DismissesToolFeedbackWhenMessageToolAlreadySent|TestProcessMessage_PublishesToolFeedbackWhenEnabled|TestNewAgentLoop' -count=1`

afjcjsbx (approved):
LGTM!

PR #2826, opened 2026-05-08
fix: resolve relative paths correctly in exec tool safety guard

### LINKED ISSUE #2749: [BUG] Bash evaluates relative path as absolute path
##  Environment & Tools
- **PicoClaw Version:** `a94ba82`
- **Go Version:** go1.25.9
- **AI Model & Provider:** GLM-4.7-Flash on Llama.cpp
- **Operating System:** Ubuntu 24.04 (mainly Docker, reproducible on host)
- **Channels:** Discord, OneBot

## 📸 Steps to Reproduce
1. Add a skill or similar file to the workspace; must be a subdirectory, e.g. `workspace/skill/...`
2. Restrict model access to workspace
3. Prompt the model to access the file using bash commands (eg. grep, wc, cat, etc.) on the *relative* path
  - Note, in trying to reproduce this programmatically, I found that the `read_file` tool is unaffected
4. Prompt the model to try again with the absolute path.

## ❌ Actual Behavior
The relative path is evaluated incorrectly: the command `cat skills/whoami/SKILL.md` identifies the path as `/whoami/SKILL.md`, starting at root, not the relative path from the workspace.

The safety guard correctly blocks the evaluated path (as `/home/picoclaw/.picoclaw/workspace/../../../../whoami/SKILL.md`), while allowing the absolute path `/home/picoclaw/.picoclaw/workspace/**skills**/whoami/SKILL.md`.

## ✅ Expected Behavior
Safety guard should allow the model to access the file in both cases, as both of these specs should evaluate to the same absolute path.

## 💬 Additional Context
While technically possible for a valid command to be generated, this is difficult to ensure in practice without heavy steering or allowing the model a copious number of turns. This will very quickly burn through tokens and/or compute time.

I first encountered this issue when working with a script-heavy skill, which used relative paths for the constituent python and resource files in the example snippets of the skilldef. I had tried to coerce the model into pointing to the absolute path in the system config, but the results were flaky at best.

yuxuan-7814:
## 📝 Description

Fix issue #2749: Bash evaluates relative path as absolute path

When the exec tool (bash) restricts access to the workspace, relative paths in commands were incorrectly resolved using the process's current working directory instead of the command's intended working directory. This caused legitimate relative paths (e.g., `skills/file.md`) to be evaluated as absolute paths from the process root, triggering false-positive safety blocks.

The fix modifies the `guardCommand` method in `pkg/tools/shell.go` to correctly resolve paths:
- For absolute paths: use as-is (cleaned)
- For relative paths: join with the command's working directory (`cwd`) before cleaning and validation

This ensures that relative paths are resolved relative to the intended working directory, not the process's global cwd.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation
- [ ] 🤖 Fully tool-generated
- [ ] 🛠️ Mixed (tool-assisted)
- [x] 👨‍💻 Human-written

## 🔗 Related Issue

Fixes #2749

- **Reference URL:**
- **Reasoning:**
  The `absolutePathPattern` regex matches paths in commands. The guard validates each matched path to ensure it stays within the workspace boundary. Previously, `filepath.Abs(raw)` was used for all matched paths, which is incorrect for relative paths because it uses the process's current working directory. The fix distinguishes absolute and relative paths, resolving them correctly against the command's cwd.

## 🧪 Test Environment
- **Hardware:** x86_64
- **OS:** Linux (Ubuntu 24.04)
- **Model/Provider:** Not applicable (tooling fix)
- **Channels:** All (exec tool is channel-agnostic)

## 📸 Evidence (Optional)
<details>

Manual verification: With workspace restriction enabled, running `exec` with a relative path like `cat skills/test.md` now correctly resolves to the workspace subdirectory and is allowed, while attempts to traverse outside (`cat ../../etc/passwd`) remain blocked.

</details>

## ☑️ Checklist

### DIFF
--- a/pkg/tools/shell.go
+++ b/pkg/tools/shell.go
@@ -1094,9 +1094,15 @@ func (t *ExecTool) guardCommand(command, cwd string) string {
 				}
 			}
 
-			p, err := filepath.Abs(raw)
-			if err != nil {
-				continue
+			// Resolve the path correctly based on cwd:
+			// - Absolute paths: use as-is (cleaned)
+			// - Relative paths: join with cwd to resolve correctly
+			var p string
+			if filepath.IsAbs(raw) {
+				p = filepath.Clean(raw)
+			} else {
+				p = filepath.Join(cwdPath, raw)
+				p = filepath.Clean(p)
 			}
 
 			if safePaths[p] {


afjcjsbx (changes requested):
Hi, the patch doesn't really fix relative paths. In [pkg/tools/shell.go] `absolutePathPattern` only matches absolute paths, so the new `else` branch in [pkg/tools/shell.go] is actually dead code.
Forward you are duplicating validation logic instead of reusing the shared validator in [pkg/tools/fs/filesystem.go]. No new tests on a security-sensitive change.

Chris-dash-T4:
The regex in `absolutePathPattern` needs to be updated, +1 on previous comment re: adding tests

PR #2828, opened 2026-05-08
fix(agent): transcribe queued voice follow-ups

bogdanovich:
## Summary
- transcribe queued voice follow-up messages before they are converted into steering messages
- preserve normal voice-note behavior when a second inbound message arrives during an already active turn
- add a regression test covering queued voice follow-ups

## Problem
When a user sent a second Telegram voice note while the first turn was still running, the follow-up did not go through the normal inbound `processMessage()` path.

The active-turn branch in `AgentLoop.Run()` converted the inbound message directly into a steering `providers.Message` and enqueued it. That bypassed `transcribeAudioInMessage()`, so queued voice follow-ups could reach the model as raw `[voice]` media annotations instead of `[voice: ...]` text.

In practice this made follow-up additions easy to lose. A common example was:
- first voice: what I ate
- second voice while the first turn is still running: and also two pieces of bread

The second voice note could be ignored because it was never transcribed before steering injection.

## Fix
Before enqueueing a follow-up into the steering queue, run the same audio transcription step used for normal inbound messages:
- transcribe queued audio media refs
- replace `[voice]` with `[voice: ...]`
- preserve transcription placeholder behavior for queued follow-ups too

This keeps queued voice notes behaviorally consistent with ordinary turn starts.

## Validation
- `go test ./pkg/agent -run 'TestAgentLoop_Run_(AutoContinuesLateSteeringMessage|QueuedVoiceMessageIsTranscribedBeforeSteering)' -count=1`
- `go test ./pkg/agent -run 'TestProcessMessage_MessageTool|TestNewAgentLoop' -count=1`

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -182,6 +182,8 @@ func (al *AgentLoop) Run(ctx context.Context) error {
 					continue
 				}
 
+				msg = al.prepareInboundMessageForAgent(ctx, msg)
+
 				// Another turn is already active (or reserved) for this session — enqueue
 				if err := al.enqueueSteeringMessage(sessionKey, agentID, providers.Message{
 					Role:    "user",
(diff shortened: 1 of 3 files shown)


afjcjsbx (approved):
this is an annoying picoclaw bug! thanks for the PR, LGTM!

PR #2830, opened 2026-05-09
fix(spawn): add explicit async delivery policy with configurable spawn routing

### LINKED ISSUE #2829: [Proposal] explicit async result delivery policy for async tool results
## Problem

When a parent agent uses `spawn` to launch a subagent that produces a user-facing terminal result, PicoClaw can also re-inject that same async completion back into the parent as a synthetic follow-up.

This can cause an unnecessary second parent turn that:
- reinterprets the subagent result
- adds an extra explanatory layer
- duplicates or distorts the original user-facing output

This is a real runtime delivery problem.

Further runtime testing showed that this is only one layer of the broader orchestration problem:
- async completion routing is one concern
- parent/child ownership of final user-facing output is a separate concern

This issue should stay focused on the async delivery primitive for spawned async results.

## Design goal

Introduce an explicit async result delivery policy in runtime metadata so async completion routing is intentional instead of always defaulting to both user-facing output and parent follow-up.

## Proposed delivery modes

- `user_only`
  - publish to the user/channel
  - do not re-inject into the parent agent

- `parent_only`
  - deliver internally to the parent agent
  - do not publish directly as a terminal user-facing result

- `user_and_parent`
  - publish to the user
  - also deliver internally to the parent

## Scope of this issue

This issue is specifically about async completion routing for `spawn` and similar async tool results.

It is **not** intended to fully solve:
- `delegate` composition behavior
- mixed ownership between parent and child
- duplicate user-facing text caused by both child and parent speaking about the same result
- visibility of already-delivered side effects such as media/file sends

## Proposed semantics

Longer-term tool semantics should become explicit:

- `delegate`
  - likely default semantic: parent-owned composition
- `spawn`
  - likely default semantic: autonomous completion

But the MVP here remains:
- add explicit async delivery support in runtime
- apply a sensible default to `spawn`
- allow explicit override where needed

## Acceptance criteria

1. A parent agent spawns a subagent for a long-running workflow
2. The subagent returns a terminal user-facing async result
3. The user sees exactly one terminal result from that spawned workflow
4. The parent does not start a second interpretation turn from that same async completion unless parent delivery was explicitly requested

## Follow-up work

Separate follow-up work should address:
- parent/child ownership semantics
- `delegate` default behavior
- mixed user-facing ownership
- structured reporting of already-delivered side effects

bogdanovich:
Closes #2829

## Summary

Introduce an explicit async result delivery policy for async tool results, keep `spawn` default behavior aligned with autonomous background work, and expose an optional `delivery_mode` override on `spawn` for partial orchestration cases.

This started from a real runtime problem:

- a parent agent spawns a subagent for long-running work
- the subagent already produces a terminal user-facing result
- PicoClaw also re-injects that same async completion back into the parent
- the parent starts a second interpretation turn and may duplicate, distort, or over-explain the child result

Runtime testing confirmed that this problem is real.

That testing also showed that hardcoding all `spawn` completions to `user_only` is too narrow:
- some spawned workflows are truly autonomous and should remain `user_only`
- other spawned workflows intentionally need parent continuation
- so the runtime primitive should exist, and `spawn` should expose an explicit caller override

## What changed

- add `AsyncDelivery` to `ToolResult`
- add delivery modes:
  - `user_only`
  - `parent_only`
  - `user_and_parent`
- update the async callback path to decide separately whether to:
  - publish the result to the user
  - queue a parent follow-up event
- keep runtime default for unset async delivery as current behavior (`user_and_parent`)
- make `spawn` default to `user_only`
- add optional `delivery_mode` argument to `spawn` so callers can explicitly request:
  - `user_only`
  - `parent_only`
  - `user_and_parent`

## Why this shape

This keeps delivery semantics in runtime metadata instead of trying to infer them from:
- message text
- channel formatting
- agent naming conventions
- parent-agent prompt heuristics

It also separates:
- low-level async completion routing
- higher-level orchestration decisions made by the caller

## Scope

This PR is about async delivery semantics and `spawn` routing control.

It does **not** fully solve the broader parent/child ownership problem, including:
- duplicate user-facing text in `delegate` flows
- mixed ownership where child both messages the user and returns the same semantic result to the parent
- parent visibility into child side effects like already-sent media/files

Those need follow-up ownership/routing semantics above this runtime primitive.

## Semantics after this change

- `spawn`
  - default: `user_only`
  - can be overridden per call with `delivery_mode`

This matches the observed split between:
- autonomous background specialists
- partial stage work that intentionally needs parent continuation

Longer-term, `delegate` vs `spawn` ownership semantics should be made more explicit, but that is outside this PR.

### Autonomous spawned workflow
A parent agent spawns a browser-oriented subagent for a long-running site workflow.
If the browser subagent finishes with a terminal user-facing result, that result should be delivered once to the user. It should not automatically trigger a second parent turn unless the tool explicitly requests parent delivery.

### Partial spawned workflow
A parent agent spawns a specialist for stage work that should continue in the parent.
For that case, the caller can explicitly use:
- `delivery_mode="parent_only"`
- or `delivery_mode="user_and_parent"`

instead of relying on hardcoded `user_only`.

## Tests

Ran focused tests:

```bash
go test ./pkg/tools -run 'TestAsyncResult|TestToolResultWithAsyncDelivery|TestSpawnTool_ExecuteAsync_MarksCallbackResultUserOnly|TestSpawnTool_SpawnStatusSeesSpawnedTask' -count=1
go test ./pkg/agent -run 'TestAgentLoop_AsyncFollowUpEmitsRuntimeEvent|TestAgentLoop_AsyncToolUserOnly_DoesNotEmitFollowUpQueued' -count=1
```

### DIFF
--- a/pkg/agent/eventbus_test.go
+++ b/pkg/agent/eventbus_test.go
@@ -699,6 +699,7 @@ type asyncFollowUpTool struct {
 	name          string
 	followUpText  string
 	completionSig chan struct{}
+	deliveryMode  tools.AsyncDeliveryMode
 }
 
 func (t *asyncFollowUpTool) Name() string {
@@ -726,7 +727,11 @@ func (t *asyncFollowUpTool) ExecuteAsync(
 	cb tools.AsyncCallback,
 ) *tools.ToolResult {
 	go func() {
-		cb(ctx, &tools.ToolResult{ForLLM: t.followUpText})
+		res := &tools.ToolResult{ForLLM: t.followUpText}
+		if t.deliveryMode != "" {
+			res.WithAsyncDelivery(t.deliveryMode)
+		}
+		cb(ctx, res)
 		if t.completionSig != nil {
 			close(t.completionSig)
 		}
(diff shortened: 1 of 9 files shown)


afjcjsbx (changes requested):
`pkg/tools/spawn.go` introduces a second async-delivery mechanism instead of reusing the existing subturn delivery contract. The new `SetSpawner` wrapper routes completion through the manager callback, but it also spawns the child subturn with `Async: true`. Existing code already treats `Async=true` as "deliver into pendingResults" in `pkg/agent/subturn.go` and verifies that contract in pkg/agent/subturn_test.go.

This fragments result routing and creates duplicate/contradictory delivery paths. A supposedly `user_only` result can still enter parent LLM context through pendingResults.

Make one component own delivery. For spawn, either run the subturn with `Async: false` and let `SubagentManager` callback deliver the result, or remove callback routing and let the parent consume pendingResults. Do not keep both.

bogdanovich:
Rebased on current `upstream/main` and reworked the spawn path to match the review direction.

What changed:
- `spawn` now has a single delivery owner: `SubagentManager` callback delivery
- the child subturn is spawned with `Async: false` inside the manager goroutine
- this avoids the previous split between callback delivery and parent `pendingResults` delivery
- the explicit async delivery policy remains in the async callback path, and `spawn` marks its callback result as `user_only`

Validation rerun:
- `go test ./pkg/tools -run 'TestAsyncResult|TestToolResultWithAsyncDelivery|TestSpawnTool_ExecuteAsync_MarksCallbackResultUserOnly|TestSpawnTool_SpawnStatusSeesSpawnedTask' -count=1`\n- `go test ./pkg/agent -run 'TestAgentLoop_AsyncFollowUpEmitsRuntimeEvent|TestAgentLoop_AsyncToolUserOnly_DoesNotEmitFollowUpQueued' -count=1`\n- `golangci-lint run ./pkg/tools ./pkg/agent`

bogdanovich:
Context update after more runtime testing:

I still think the async delivery primitive in this PR is correct, but I no longer think hardcoding all `spawn` completions to `user_only` is broad enough.

What testing confirmed:
- the original duplicate parent follow-up problem is real
- explicit async delivery metadata is the right runtime layer for solving it

What additional testing showed:
- some spawned workflows are autonomous and should stay `user_only`
- other spawned workflows intentionally need parent continuation
- so `spawn` should expose an explicit `delivery_mode` override instead of treating `user_only` as universal behavior

I’ve updated the framing accordingly:
- runtime primitive stays
- `spawn` default remains `user_only`
- callers can explicitly request `parent_only` or `user_and_parent` when orchestration needs it

This still does not fully solve parent/child ownership semantics in general, especially in `delegate` flows. That should be treated as a separate follow-up concern.

PR #2832, opened 2026-05-09
feat(web,api): fetch models and saved catalog support

SiYue-ZO:
## Summary (Part 2 of 3)

Split from [#2752](https://github.com/sipeed/picoclaw/pull/2752) per maintainer request.

Depends on [#2831](https://github.com/sipeed/picoclaw/pull/2831) (PR1).

**Backend:**
- `POST /api/models/fetch` — fetch available models from upstream providers
- `GET /api/models/catalog` — list saved model catalogs
- `DELETE /api/models/catalog/{id}` — delete a saved catalog
- Provider allow-listing (only OpenAI-compatible + Ollama providers)
- Auto-save fetched models to catalog with masked API key storage

**Frontend:**
- `FetchModelsDialog` — browse, filter, and select upstream models
- `CatalogDialog` — view, expand, filter, and import from saved catalogs
- Replace PR1 dynamic imports with static imports for fetch/catalog
- Fetch/Test buttons enabled (no longer disabled)

### DIFF
--- a/web/backend/api/model_catalog.go
+++ b/web/backend/api/model_catalog.go
@@ -1,0 +1,161 @@
+package api
+
+import (
+	"crypto/sha256"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/sipeed/picoclaw/pkg/config"
+	"github.com/sipeed/picoclaw/pkg/fileutil"
+)
+
+// CatalogModel represents a single model entry in a saved catalog.
+type CatalogModel struct {
+	ID      string         `json:"id"`
+	OwnedBy string         `json:"owned_by,omitempty"`
+	Extra   map[string]any `json:"extra,omitempty"`
+}
+
+// CatalogEntry is a saved list of upstream models fetched for a specific provider+key combination.
+type CatalogEntry struct {
+	ID         string         `json:"id"`
+	Provider   string         `json:"provider"`
+	APIBase    string         `json:"api_base"`
+	APIKeyMask string         `json:"api_key_mask"`
+	Models     []CatalogModel `json:"models"`
+	FetchedAt  string         `json:"fetched_at"`
+}
+
+// CatalogStore holds all saved model catalogs.
+type CatalogStore struct {
+	Entries map[string]*CatalogEntry `json:"entries"`
+}
+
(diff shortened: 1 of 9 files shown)


SiYue-ZO:
@wj-xiao  I have resolved the conflict; please review.

web/backend/api/models.go
```diff
@@ -614,3 +631,172 @@ func maskAPIKey(key string) string {
...
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("upstream returned status %d", resp.StatusCode)
+	}
+
+	var parsed struct {
```
wj-xiao: `fetchOpenAICompatibleModels` only decodes the OpenAI `{ "data": [...] }` envelope, but the existing CLI model fetch path already supports both `{data:[...]}` and bare-array responses because some OpenAI-compatible servers return `[...]` directly.

With the current implementation, those providers will fail through the web API with a decode error/502 even though the CLI supports them. Please align this parser with `cmd/picoclaw/internal/model/online.go` and add tests for envelope, bare-array, and unrecognized response shapes before merging.

web/backend/api/model_catalog.go
```diff
@@ -0,0 +1,161 @@
...
+
+// generateCatalogKey creates a deterministic key for a provider+base+key combination.
+func generateCatalogKey(provider, apiBase, apiKey string) string {
+	provider = strings.ToLower(strings.TrimSpace(provider))
+	apiBase = strings.TrimRight(strings.TrimSpace(apiBase), "/")
+	hash := sha256.Sum256([]byte(apiKey))
+	return fmt.Sprintf("%s|%s|%x", provider, apiBase, hash[:6])
+}
+
+// maskAPIKeyValue masks an API key for display, keeping first 4 and last 4 chars.
```
wj-xiao: `maskAPIKeyValue` leaks too much of short API keys. For 9-12 character keys it exposes the first 4 and last 4 characters, meaning a 9-character key only has one character masked. The existing `maskAPIKey` helper in `models.go` already uses a safer policy for short keys. Please reuse that helper or match its masking behavior, and add tests covering 9-12 character keys before merging.

wj-xiao (approved):

PR #2836, opened 2026-05-09
fix(powershell):  windows security enhancement, sec deny  powershell encoding bypass   via iex inje…

sky5454:
…ction.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

- **Reference URL:**
- **Reasoning:**

## 🧪 Test Environment
- **Hardware:** 
- **OS:** 
- **Model/Provider:** 
- **Channels:** 

## 📸 Evidence (Optional)
<details>

</details>

## ☑️ Checklist

### DIFF
--- a/pkg/tools/shell.go
+++ b/pkg/tools/shell.go
@@ -95,14 +95,25 @@ var (
 		regexp.MustCompile(`\bssh\b.*@`),
 		regexp.MustCompile(`\beval\b`),
 		regexp.MustCompile(`\bsource\s+.*\.sh\b`),
-		// PowerShell encoding bypass: [Text.Encoding] used to construct command strings.
-		regexp.MustCompile(`\[text\.encoding\]`),
-		// PowerShell -EncodedCommand flag (base64-encoded command).
-		regexp.MustCompile(`-encodedcommand`),
+	}
+
+	// windowsDenyPatterns contains PowerShell-specific deny patterns that only
+	// apply on Windows, where commands are executed via powershell -Command.
+	windowsDenyPatterns = []*regexp.Regexp{
+		// [Text.Encoding] used to construct command strings at runtime.
+		// Matches [Text.Encoding] and [System.Text.Encoding] variants.
+		regexp.MustCompile(`\[(?:\w+\.)?text\.encoding\]`),
+		// PowerShell -EncodedCommand flag (base64-encoded command) and short forms -e, -ec, -enc.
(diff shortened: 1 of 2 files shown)


lxowalle (commented):
Thanks for working on this. The new deny rules do block the exact examples covered by the new test, but I do not think the PowerShell bypass is fully closed yet.

`pkg/tools/shell.go` only matches the full `-EncodedCommand` form. PowerShell also accepts short forms such as `-e` and `-ec`, so commands like `pwsh -e <base64>` or `pwsh -ec <base64>` would still bypass this check.

The new decoding patterns are also very literal:

- `\[text\.encoding\]` does not match `[System.Text.Encoding]`
- `\.getstring\(\[byte\[\]` only matches one no-whitespace form

Because of that, equivalent commands with normal spacing, or commands that store the byte array in a variable before calling `GetString`, can still avoid the guard.

Could we broaden the detection and add a few more tests for these variants? For example:

- `pwsh -e ...`
- `pwsh -ec ...`
- `[System.Text.Encoding]::ASCII.GetString ([byte[]](...))`

lxowalle (commented):
Thanks for the follow-up fix. I looked at the latest commit again, and I think there are still a couple of things that need to be fixed before this can merge.

The main issue is the new `-EncodedCommand` pattern:

```go
regexp.MustCompile(` -e[cn]\b`)
```

This does not actually match all the forms mentioned in the commit message. It matches `-ec` and `-en`, but it still misses:

- `-EncodedCommand`
- `-enc`
- `-e`

So commands like these can still get past the guard:

```powershell
pwsh -e <base64>
pwsh -enc <base64>
powershell -EncodedCommand <base64>
```

There is also a test/implementation mismatch. The new PowerShell and Windows path tests run on every platform, but the implementation only enables those deny patterns and env/tilde expansion when `runtime.GOOS == "windows"`. Because of that, `go test ./pkg/tools` fails on Linux. The GitHub Tests check is failing for the same reason.

One smaller test issue: `TestShellTool_SymlinkBypassPrevented` currently does not really assert anything. If the command is not blocked, the test just enters an empty `if` block and still passes, so it would not catch a regression.

Could you update the encoded-command regex to cover the real PowerShell forms, and either make the Windows-specific tests Windows-only or adjust the implementation/tests so they agree across platforms?

lxowalle (approved):
LGTM

Issue #2841, opened 2026-05-09
[Feature] synthesize steering-chain final replies from user-facing outcomes


bogdanovich:
## Summary

After steering-heavy turns, PicoClaw can execute multiple user-visible actions but still return only the last natural-language reply.

This leaves a UX gap even when the underlying actions were all completed correctly.

## Example

A tracking-style agent receives:
- `log 100 pushups`
- then, while still working: `and 100 situps`
- then: `and 10 pullups`

The agent may successfully record all three actions, but the final visible reply can end up reporting only the last one.

## Problem

Today the final user-facing reply is still largely transcript-shaped:
- the runtime remembers the latest assistant response,
- but it does not keep a compact list of user-facing outcomes across the whole steering chain,
- so the final reply is weak when several actions complete over multiple follow-ups.

## Proposed direction

Add an optional steering-turn synthesis mode that:
- collects only explicit user-facing outcomes during the turn
  - `toolResult.ForUser`
  - direct assistant replies from no-tool iterations
- skips internal transcript plumbing and `ForLLM`
- on steering-heavy turns with multiple outcomes, runs one final narrow LLM formatting pass over that compact action log

## Why this shape

This keeps the core logic:
- tool-agnostic
- grounded in explicit user-facing outcomes
- free of per-tool operation contracts
- free of text denylist heuristics for internal steps

## Scope

This is intentionally smaller than a full operation ledger or rollback model.
It is about better final reporting for multi-step turns, not transaction semantics.

bogdanovich:
Closing this feature proposal for now.

The original idea here was a detached action-log summarization pass. After local testing, that shape is not quite right:
- it is too detached from the main agent context
- it does not reliably preserve the full accumulated intent across a steering chain
- it can produce a final reply that reflects only the latest follow-up instead of the whole resolved request chain

We are reworking the design locally toward a same-agent final no-tools render pass instead. That should preserve the existing agent prompt/style/context while still avoiding another tool loop.

I’ll reopen with a revised proposal once that design is validated.

PR #2842, opened 2026-05-09
[Feature] synthesize steering-chain final replies from action log

bogdanovich:
## Summary

Add an optional steering-turn final synthesis mode that builds the terminal reply from a compact log of explicit user-facing outcomes.

Related to #2841.

## Problem

After steering-heavy turns, PicoClaw can complete multiple user-visible actions but still respond with only the last assistant message.

That works poorly for workflows like activity or calorie tracking, where the underlying actions may all succeed but the visible reply reports only the last item from the chain.

## Design

This PR adds a narrow, optional synthesis path:
- collect only explicit user-facing outcomes during the turn
  - `toolResult.ForUser`
  - direct assistant replies from no-tool iterations
- ignore internal transcript plumbing and `ForLLM`
- when steering was observed and at least two user-facing outcomes were recorded, run one final LLM formatting pass over that compact action log

The mode is gated behind:
- `agents.defaults.final_action_summary_mode = "llm"`

## Why this shape

This keeps the feature:
- tool-agnostic
- grounded in explicit user-facing outcomes
- free of per-tool operation schemas
- free of transcript-text heuristics for internal steps

It is intentionally smaller than a full operation ledger or rollback model.

## Example

A tracking-style agent receives:
- `log 100 pushups`
- then: `and 100 situps`
- then: `and 10 pullups`

If the tool executions succeed across the steering chain, the final reply can now be synthesized from all recorded user-facing outcomes instead of only the last response.

## Tests

Added coverage for:
- steering-heavy turns that complete one action, accept steering, complete another action, and synthesize one final reply from both outcomes
- config/default loading for the new optional mode

### DIFF
--- a/pkg/agent/action_summary.go
+++ b/pkg/agent/action_summary.go
@@ -1,0 +1,157 @@
+package agent
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/sipeed/picoclaw/pkg/logger"
+	"github.com/sipeed/picoclaw/pkg/providers"
+)
+
+type TurnActionRecord struct {
+	Source string `json:"source"`
+	Tool   string `json:"tool,omitempty"`
+	Text   string `json:"text"`
+	Error  bool   `json:"error,omitempty"`
+}
+
+func appendTurnActionRecord(
+	records []TurnActionRecord,
+	source, tool, text string,
+	isError bool,
+) []TurnActionRecord {
+	text = strings.TrimSpace(text)
+	if text == "" {
+		return records
+	}
+	rec := TurnActionRecord{
+		Source: source,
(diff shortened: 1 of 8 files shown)


bogdanovich:
Closing this POC PR for now.

After local testing, the detached action-log summarizer turned out to be the wrong abstraction boundary:
- it runs outside the main agent context
- it does not inherit the full agent prompt/skills/style
- it can over-focus on the last follow-up even when earlier requests in the same chain were already resolved

We are replacing this with a different direction locally: a same-agent final no-tools render pass that preserves the full accumulated turn context while still avoiding another tool loop.

So this branch is being closed rather than iterated further. If that replacement works well, I will reopen with a revised issue/PR built around that design.

PR #2853, opened 2026-05-11
feat(pico): add ChatStream support for real-time token streaming

loafoe:
## Summary

This PR adds full ChatStream support to the pico channel, enabling real-time token streaming to connected WebSocket clients. Key changes:

- **Streamer tracking in turnState** - acquire and release streamer per conversation turn
- **ChatStream integration** - use provider's `ChatStream` method when a streamer is available, falling back to `Chat` when not
- **Streaming lifecycle management** - properly wire streamer acquisition/release through the agent loop
- **Config support** - add `Streaming` field to PicoSettings to enable/disable per deployment
- **Duplicate publish prevention** - skip bus publish when response was already streamed to avoid double-sending

## Why this matters

Currently, pico channel responses are only sent after the full LLM response is complete. With ChatStream support, tokens stream in real-time as the LLM generates them, providing a much more responsive user experience - especially for longer responses or when the model is "thinking."

This builds on the existing streaming infrastructure and extends it to the pico WebSocket channel.

## Web UI compatibility

The built-in web UI works seamlessly with this change - no frontend modifications needed. The UI already handles `message.update` events via the existing protocol handler. When streaming is enabled:

1. `BeginStream` creates a placeholder message via `message.create`
2. Tokens stream via throttled `message.update` events (100ms or 20 chars minimum)
3. `Finalize` sends the complete content as a final `message.update`

## Configuration

Streaming is **opt-in** and disabled by default. To enable:

**YAML:**
```yaml
channel_list:
  pico:
    streaming: true
```

**JSON:**
```json
"channel_list": {
  "pico": {
    "streaming": true
  }
}
```

When disabled (default), the system falls back to the existing placeholder mechanism.

## Test plan

- [x] Unit test scaffold for streaming integration
- [x] Manual testing with pico WebSocket client
- [x] Verify streaming works with Bedrock provider
- [x] Verify fallback to non-streaming works when streamer unavailable

### DIFF
--- a/pkg/agent/adapters/messagebus.go
+++ b/pkg/agent/adapters/messagebus.go
@@ -34,3 +34,7 @@ func (a *messageBusAdapter) PublishOutboundMedia(ctx context.Context, msg bus.Ou
 func (a *messageBusAdapter) InboundChan() <-chan bus.InboundMessage {
 	return a.inner.InboundChan()
 }
+
+func (a *messageBusAdapter) GetStreamer(ctx context.Context, channel, chatID string) (bus.Streamer, bool) {
+	return a.inner.GetStreamer(ctx, channel, chatID)
+}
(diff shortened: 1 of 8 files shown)


loafoe:
superseded by #2892

PR #2856, opened 2026-05-11
feat(message): support media attachments and Telegram rich delivery

### LINKED ISSUE #2855: [Feature] Extend message tool to support media attachments and channel-aware rich outbound delivery
## Summary

The current `message` tool is text-only, while rich outbound delivery already exists lower in the bus/channel stack. This leads to awkward workflows where agents have to split media and text into separate sends, even for channels that natively support `text + media` in one logical message.

This feature request is to evolve `message` into the main outbound message primitive for:

- text-only messages
- media-only messages
- text + media messages

with channel-aware rendering and safe fallbacks.

## Motivation

A few concrete cases:

- Telegram can send a photo/video/document with caption, but the current tool surface does not expose that cleanly.
- Telegram can also send multiple images as media groups, but today the agent has to degrade into multiple separate sends.
- Slack, Feishu, Matrix, Discord, etc. do not all support captions in the same way, but they still need a sound fallback where message text is never silently lost.

OpenClaw already leans toward a single outbound payload model (`text + mediaUrl/mediaUrls`) instead of separate low-level send primitives. PicoClaw can move in the same direction while keeping channel-specific rendering in adapters.

## Proposed direction

Extend `message` so it accepts:

- `content` (optional)
- `media[]` (optional)

with the rule that at least one of them must be present.

Example shape:

```json
{
  "content": "Recipe translation",
  "media": [
    {
      "path": "/tmp/video.mp4",
      "type": "video",
      "filename": "recipe.mp4"
    }
  ]
}
```

### Channel behavior

- Telegram:
  - `content <= caption limit`: use caption on first media item
  - long content: send full text first via normal text path, then send media
  - multiple images: use media groups, chunked by 10
- Slack / Feishu / other channels without clean native caption semantics:
  - keep media delivery
  - make sure `content` survives via a text fallback instead of being silently dropped

## Scope for first iteration

A first implementation is ready on a branch and includes:

- richer `message` tool schema and execution path
- Telegram support for:
  - single media + caption
  - multiple images as media groups
  - long-content policy (`text first`, then media)
- simple fallback semantics for Slack and Feishu so text is preserved

## Why this should be one tool

This keeps the public tool surface smaller and cleaner:

- no `send_text`
- no separate `send_media`
- no Telegram-specific `send_album`

Instead, one semantic outbound message tool maps to channel capabilities in adapters.

## Follow-ups

Likely follow-up work after the first iteration:

- richer fallback semantics for additional channels
- tests for each adapter path
- docs/examples for `message + media`
- possibly de-emphasize `send_file` as the primary user-facing delivery primitive over time

bogdanovich:
Closes #2855

## Summary

This is the first iteration of richer outbound delivery for the existing `message` tool.

Instead of keeping `message` text-only and forcing agents to combine it with lower-level file delivery primitives, this change lets `message` carry a single semantic outbound payload with:

- text only
- media only
- text + media

The design is intentionally close to OpenClaw's `text + mediaUrl/mediaUrls` model, while keeping channel-specific rendering inside adapters.

### `message` tool

`message` now accepts:

- `content` (optional)
- `media[]` (optional)

with the rule that at least one of them must be present.

Each `media[]` item accepts:

- `path`
- optional `type`
- optional `filename`

When media is present:

- local files are validated against the existing workspace/path/media-store rules
- media refs are registered through the `MediaStore`
- `content` is attached to the first media part as the logical caption/body text

### Telegram

Telegram now gets the richest implementation in this first pass:

- single media + caption
- multiple images as `sendMediaGroup`
- groups chunked by 10 images
- long-content policy:
  - if content is within Telegram caption limits, keep it as caption on the first media item
  - if content exceeds the caption limit, send the full text first through the normal text path, then send media without captions

This avoids partial caption splitting and keeps long content easier to copy.

### Fallback semantics for other channels

For channels in this PR where native caption semantics are weaker, the rule is:

- media still sends
- message text must not be silently dropped

This first pass adds simple fallback behavior for:

- Slack
- Feishu

They now send the media and then send the first non-empty caption as a plain text follow-up.

## Why extend `message` instead of adding a new tool

This keeps the public tool surface smaller and cleaner:

- no `send_text`
- no separate `send_media`
- no Telegram-specific album tool

The semantic primitive is still just: send a message to the user.
The channel decides how to render `content + media[]`.

## Tests

Added/updated tests for:

- `pkg/tools/integration/message_test.go`
- `pkg/channels/telegram/telegram_test.go`
- `pkg/agent/agent_test.go`

Verified locally with:

- `go test ./pkg/tools/integration -count=1`
- `go test ./pkg/channels/telegram ./pkg/channels/slack ./pkg/channels/feishu -count=1`
- `go test ./pkg/agent -run 'TestPublishResponseIfNeeded_DismissesToolFeedbackWhenMessageToolAlreadySent' -count=1`
- `make build`

## Follow-ups

This PR intentionally stops at a clean first iteration.
Follow-up work can expand:

- richer rendering/fallback semantics for more channels
- additional adapter-level tests
- docs/examples for `message + media`
- eventual guidance around when `send_file` is still the right primitive vs. `message`

### DIFF
--- a/docs/guides/configuration.md
+++ b/docs/guides/configuration.md
@@ -400,6 +400,7 @@ Even with `restrict_to_workspace: false`, the `exec` tool blocks these dangerous
 |------------|------|---------|-------------|
 | `tools.allow_read_paths` | string[] | `[]` | Additional paths allowed for reading outside workspace |
 | `tools.allow_write_paths` | string[] | `[]` | Additional paths allowed for writing outside workspace |
+| `tools.message.media_enabled` | bool | `false` | Allows the `message` tool to attach local media files by path. This is separate from `tools.send_file.enabled`; enable it only when unified text/media/caption delivery is intended. |
 
 ### Read File Mode
 
--- a/pkg/agent/agent_init.go
+++ b/pkg/agent/agent_init.go
@@ -161,9 +161,16 @@ func registerSharedTools(
 		// Message tool
 		if cfg.Tools.IsToolEnabled("message") {
 			messageTool := tools.NewMessageTool()
+			messageTool.ConfigureLocalMedia(
+				agent.Workspace,
+				cfg.Agents.Defaults.RestrictToWorkspace,
+				cfg.Agents.Defaults.GetMaxMediaSize(),
+				allowReadPaths,
+			)
 			messageTool.SetSendCallback(func(
 				ctx context.Context,
 				channel, chatID, content, replyToMessageID string,
+				mediaParts []bus.MediaPart,
 			) error {
 				pubCtx, pubCancel := context.WithTimeout(context.Background(), 5*time.Second)
 				defer pubCancel()
@@ -173,6 +180,15 @@ func registerSharedTools(
 					tools.ToolSessionKey(ctx),
 					tools.ToolSessionScope(ctx),
 				)
+				if len(mediaParts) > 0 {
+					return msgBus.PublishOutboundMedia(pubCtx, bus.OutboundMediaMessage{
+						Context:    outboundCtx,
+						AgentID:    outboundAgentID,
+						SessionKey: outboundSessionKey,
+						Scope:      outboundScope,
+						Parts:      mediaParts,
+					})
+				}
 				return msgBus.PublishOutbound(pubCtx, bus.OutboundMessage{
 					Context:          outboundCtx,
 					AgentID:          outboundAgentID,
--- a/pkg/agent/agent_init.go
+++ b/pkg/agent/agent_init.go
@@ -161,19 +161,19 @@ func registerSharedTools(
 		// Message tool
 		if cfg.Tools.IsToolEnabled("message") {
 			messageTool := tools.NewMessageTool()
-			messageTool.ConfigureLocalMedia(
-				agent.Workspace,
-				cfg.Agents.Defaults.RestrictToWorkspace,
-				cfg.Agents.Defaults.GetMaxMediaSize(),
-				allowReadPaths,
-			)
+			if cfg.Tools.Message.MediaEnabled {
+				messageTool.ConfigureLocalMedia(
+					agent.Workspace,
+					cfg.Agents.Defaults.RestrictToWorkspace,
+					cfg.Agents.Defaults.GetMaxMediaSize(),
+					allowReadPaths,
+				)
+			}
 			messageTool.SetSendCallback(func(
 				ctx context.Context,
 				channel, chatID, content, replyToMessageID string,
 				mediaParts []bus.MediaPart,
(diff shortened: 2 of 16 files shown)


afjcjsbx:
This PR expands message until it is, in fact, a second `send_file`, even when `send_file` is disabled. In `message.go` the tool exposes `media[].path`, and it reads and registers local files in the `MediaStore`. However, this capability is enabled simply when message is enabled in `agent_init.go`, regardless of the separate flag for `send_file` in `agent_init.go`. If today a deployment keeps message on but `send_file` off to prevent file exfiltration, with this PR that permission boundary disappears.

afjcjsbx:
Attachments sent via message no longer have the synchronous and reliable path that `send_file` uses, so real delivery errors never return to the model. The new callback immediately publishes on the bus in `agent_init.go`, while the robust media path goes from `channelManager.SendMedia(...)` in `pipeline_execute.go` and `manager.go`. Result: if the channel does not support media, the upload fails or retries run out, the tool has already returned success and `PublishResponseIfNeeded` also suppresses the final response because it sees that message "has already sent" in `agent_outbound.go`. The user may then receive neither attachment nor textual fallback.

bogdanovich:
Addressed the review concerns in the PR branch.

Changes pushed:
- Added an explicit capability gate for local media attachments on the message tool: tools.message.media_enabled.
- message remains text-capable with tools.message.enabled=true, but media[].path is only exposed/accepted when media_enabled is true.
- This keeps the boundary independent from send_file, so deployments can disable send_file while still explicitly allowing unified media+caption sends through message.
- Changed message media delivery to use channelManager.SendMedia(...) when a channel manager is available, instead of returning success immediately after publishing to the bus.
- The message tool now records a send only after the callback succeeds; failed media delivery returns a tool error and does not suppress the final fallback response.

Validation:
- go test ./pkg/tools/integration ./pkg/config -count=1
- go test ./pkg/agent -run 'Test.*Message|Test.*Outbound|Test.*Media' -count=1

I also mirrored the same nested config shape into fork/main as tools.message.media_enabled.

afjcjsbx:
Just a quick observation: if the caption fallback fails after the uploads, the manager retries and duplicates the attachments.

In `slack.go` and `feishu_64.go`, the caption text is only sent after the files have already been successfully uploaded. If that second step fails, `SendMedia` returns an error, and `manager.go` retries the entire operation, re-sending media that has already been delivered. Here, we should either treat the caption posting as a best-effort operation or avoid a complete retry after a partial success.

Other than that, everything looks good! This can be addressed in a follow-up PR.

LGTM

afjcjsbx (approved):

PR #2892, opened 2026-05-18
Support streaming

lxowalle:
### Summary

- Adds configuration-driven provider streaming with a dual opt-in model: model entries and channels must both explicitly enable streaming.
- Implements generic agent streaming eligibility and delivery, with Pico using existing `message.create` / `message.update` messages for real incremental output.
- Keeps streaming omitted by default in generated/example config, preserves Telegram streaming environment variable compatibility, and updates multilingual documentation.

### What Changed

- Added `model_list[].streaming.enabled` as a model-level opt-in.
- Added generic channel `settings.streaming` support for streaming-capable channels, including delivery controls such as throttling and minimum content growth.
- Implemented the agent streaming path without Pico-specific branching or hardcoded channel allowlists.
- Added Pico streaming delivery that creates one assistant message and updates that same message as provider chunks arrive.
- Suppressed duplicate final answers after a streamed final response has already been delivered.
- Preserved non-streaming fallback behavior when any streaming eligibility gate fails.
- Updated Web model add/edit forms and channel configuration UI to expose streaming settings.
- Updated backend model/channel/gateway APIs so streaming config changes are surfaced and restart signatures reflect active streaming changes.
- Added OpenSpec change documentation and multilingual user documentation for streaming setup.

### Behavior Notes

- Streaming is disabled unless both the selected model entry and the current channel settings enable it.
- If streaming is unconfigured, config serialization omits the `streaming` parameter instead of writing disabled defaults.
- If provider streaming fails before the first chunk, the agent falls back to the normal `Chat()` path.
- If provider streaming fails after at least one chunk, the error is returned and the normal non-streaming path is not replayed.
- Provider `ChatStream()` final responses remain authoritative for tool calls, usage, hooks, history, and final content.
- 
## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

- **Reference URL:**
- **Reasoning:**

## 🧪 Test Environment
- **Hardware:** 
- **OS:** 
- **Model/Provider:** 
- **Channels:** 

## 📸 Evidence (Optional)
<details>

</details>

## ☑️ Checklist


pkg/channels/manager.go
```diff
@@ -555,27 +608,90 @@ func (m *Manager) GetStreamer(ctx context.Context, channelName, chatID string) (
...
+	})
+	return found
+}
+
 // finalizeHookStreamer wraps a Streamer to run a hook on Finalize.
 type finalizeHookStreamer struct {
 	Streamer
-	onFinalize func(context.Context)
+	onFinalize  func(context.Context, string)
+	clearMarker func()
```
imguoguo, line 822: 🔴 **Critical:** this wrapper doesn't forward `ReasoningStreamer`, so Pico reasoning stream is silently dropped in production.

The struct embeds `Streamer` and manually forwards `FinalizeWithContext` and `ClearFinalizedStreamMarker`, but omits `UpdateReasoning` and `FinalizeReasoning`. When `pipeline_streaming.go` type-asserts the wrapped streamer to `bus.ReasoningStreamer`, it gets `false` and silently discards all reasoning chunks.

Existing tests pass because they either test `picoStreamer` directly (bypassing the manager wrapper) or use mocks that implement `ReasoningStreamer`.

Fix — add two forwarding methods:

```go
func (s *finalizeHookStreamer) UpdateReasoning(ctx context.Context, content string) error {
    if rs, ok := s.Streamer.(bus.ReasoningStreamer); ok {
        return rs.UpdateReasoning(ctx, content)
    }
    return nil
}

func (s *finalizeHookStreamer) FinalizeReasoning(ctx context.Context, content string) error {
    if rs, ok := s.Streamer.(bus.ReasoningStreamer); ok {
        return rs.FinalizeReasoning(ctx, content)
    }
    return nil
}
```

pkg/providers/openai_compat/provider.go
```diff
@@ -380,6 +381,28 @@ func (p *Provider) ChatStream(
...
+		options,
+		func(chunk StreamChunk) {
+			if onChunk != nil && strings.TrimSpace(chunk.Content) != "" {
+				onChunk(chunk.Content)
+			}
+		},
+	)
+}
+
+func (p *Provider) ChatStreamEvents(
```
imguoguo, line 402: ⚠️ **Warning:** the streaming HTTP client (used internally by this function) drops the request timeout but adds no idle/read timeout either.

Removing `http.Client.Timeout` for SSE is correct — it would kill valid long-running streams. But the agent's streaming context normally carries no deadline, so a provider that connects but stalls sending data can hang the turn indefinitely.

Context cancellation (user interrupt, hard abort, shutdown) still works as a safety net, and real providers rarely stall silently, so this is a hardening item rather than a functional break.

Suggested fix: add a per-read idle timeout (e.g. `context.WithTimeout` refreshed on each SSE line, or a custom `net.Conn` wrapper with `SetReadDeadline`). Not a merge blocker.

imguoguo (commented):
## Review of PR #2892

Clean architecture — dual opt-in, fallback semantics (before/after visible output), and streaming suppression are all well thought out. One real bug to fix before merge; other items are hardening or nits.

### Critical

Pico reasoning stream is completely broken through the production path. `finalizeHookStreamer` wraps the inner streamer but doesn't forward `UpdateReasoning` / `FinalizeReasoning`, so the type assertion in `pipeline_streaming.go` silently drops all reasoning chunks. Two forwarding methods fix it — see inline comment.

### Other observations (not blocking)

- `streamingConfigFromDecodedSettings` uses `reflect.FieldByName("Streaming")` — fragile compared to the type-switch pattern used in `validateChannelStreamingConfig` in the same package. Consider aligning on one pattern.
- `streamActiveForChat` does a `sync.Map.Range` scan on every non-auxiliary preSend. Fine at current scale; worth a second index if concurrent stream count grows.
- `pipeline_finalize.go` streaming conditional is a 4-clause AND/OR nest — a one-line comment explaining intent would help the next reader.

### Verdict

`COMMENT`. The reasoning-stream bug is a scoped fix (two methods). Happy to re-review once it lands.

imguoguo (approved):

PR #2900, opened 2026-05-20
fix: add CSRF protection, path traversal validation, and security headers

yangwenjie1231:
fix: add CSRF protection, path traversal validation, and security headers to web backend

Security fixes for the web UI launcher backend:

1. handleDeleteSkill path boundary validation:
   - Add filepath.Clean + isWithinDir prefix check against workspace dir
   - Resolve symlinks via filepath.EvalSymlinks before final check
   - Return 400 if deletion target escapes workspace directory

2. CSRF middleware:
   - Check X-Requested-With header on POST/PUT/DELETE requests
   - Return 403 for requests missing the header
   - Frontend automatically adds X-Requested-With: XMLHttpRequest on mutating requests

3. Security response headers:
   - X-Content-Type-Options: nosniff (prevent MIME sniffing)
   - X-Frame-Options: DENY (prevent clickjacking)

### DIFF
--- a/web/backend/middleware/middleware.go
+++ b/web/backend/middleware/middleware.go
@@ -79,3 +79,28 @@ func Recoverer(next http.Handler) http.Handler {
 		next.ServeHTTP(w, r)
 	})
 }
+
+// CSRF checks for the X-Requested-With header on mutating requests (POST, PUT,
+// DELETE). Browsers block cross-origin requests that include custom headers,
+// so this simple check mitigates CSRF attacks. AJAX calls from the frontend
+// must include the header explicitly.
+func CSRF(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Method == http.MethodPost || r.Method == http.MethodPut || r.Method == http.MethodDelete {
+			if r.Header.Get("X-Requested-With") == "" {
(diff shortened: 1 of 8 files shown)


web/backend/middleware/middleware.go
```diff
@@ -79,3 +79,28 @@ func Recoverer(next http.Handler) http.Handler {
...
 	})
 }
+
+// CSRF checks for the X-Requested-With header on mutating requests (POST, PUT,
+// DELETE). Browsers block cross-origin requests that include custom headers,
+// so this simple check mitigates CSRF attacks. AJAX calls from the frontend
+// must include the header explicitly.
+func CSRF(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Method == http.MethodPost || r.Method == http.MethodPut || r.Method == http.MethodDelete {
```
afjcjsbx, line 89: The new CSRF logic only treats `POST/PUT/DELETE` as mutating. `PATCH /api/config` is a real write endpoint in this backend, and the config UI already uses it via patchAppConfig(). As written, that route remains completely outside the new protection, so the PR’s "CSRF protection for mutating requests" claim is incomplete.

afjcjsbx (changes requested):

PR #2904, opened 2026-05-20
Fix agent loop reload and panic cleanup stability

SiYue-ZO:
## 📝 Description

It fixes three related problems in `pkg/agent`:

1. `ReloadProviderAndConfig` no longer creates a detached goroutine just to build a new registry. The reload path now uses a synchronous `defer/recover` flow, which removes the possibility of leaving behind a blocked goroutine when the caller context is canceled.
2. Worker panic cleanup now reliably releases the session turn state. Previously, if a panic happened after the placeholder turn state had already been replaced by a real turn state, the panic recovery path only logged the failure and could leave the session permanently blocked. The new cleanup path unifies session release so panic, stop, and semaphore-wait cancellation paths all release the session deterministically.
3. Provider hot-reload cleanup no longer uses a fixed 100ms grace period. Instead, reload waits for tracked in-flight requests to drain, up to a 30s grace period, before closing the previous stateful provider. This avoids aggressively terminating requests that are still actively running during reload.

In addition to the production changes, this PR adds focused regression tests that cover:

- worker panic recovery releasing the session so the next message can start a new turn
- provider reload waiting for an in-flight request before closing the old provider
- context cancellation behavior while waiting for active requests
- panic handling during registry creation in the reload path

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue

Addresses the first issue in `improvement-report.md`:
"Agent Loop resource leaks and stability risks"

- **Reference URL:**
  - Internal report: `improvement-report.md`
- **Reasoning:**
  - The original implementation had a real risk of session-level deadlock after worker panic, unnecessary goroutine lifetime during provider reload, and an unrealistically short provider shutdown grace period for live LLM requests. The goal of this change is to make agent turn cleanup deterministic and make provider reload significantly safer under real traffic.

## 🧪 Test Environment
- **Hardware:** PC
- **OS:** Linux
- **Model/Provider:** test doubles / mock providers in `pkg/agent` tests
- **Channels:** agent unit tests, no external chat channel dependency

## 📸 Evidence (Optional)
<details>

Validated locally:

- `go test ./pkg/agent -run 'Test(ReloadProviderAndConfigRefreshesRuntimeEventLogger|ReloadProviderAndConfigWaitsForInFlightRequestsBeforeClosingOldProvider|WaitForActiveRequestsHonorsContextCancellation|ReloadProviderAndConfigReturnsCanceledErrorWhenRegistryCreationPanics|RunWorkerPanicReleasesSessionTurnState)'`
- `make build`

Notes about environment limitations:

- `make lint` could not be executed in the current environment because `golangci-lint` was not installed in `PATH`.
- `go test ./pkg/...` hit unrelated sandbox limitations in tests that use `httptest.NewServer` because local listener sockets were not permitted in this environment.

</details>

## ☑️ Checklist

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -117,6 +117,7 @@ const (
 	handledToolResponseSummary = "Requested output delivered via tool attachment."
 	sessionKeyAgentPrefix      = "agent:"
 	pendingTurnPrefix          = "pending-"
+	providerReloadGracePeriod  = 30 * time.Second
 	metadataKeyMessageKind     = "message_kind"
 	metadataKeyToolCalls       = "tool_calls"
 	metadataKeyOutboundKind    = "outbound_kind"
@@ -208,6 +209,7 @@ func (al *AgentLoop) Run(ctx context.Context) error {
 			// slot. The goroutine is spawned immediately so the main loop keeps
 			// draining the inbound channel. The goroutine blocks on the semaphore.
 			go func(m bus.InboundMessage) {
+				var releaseSession bool
 				// Acquire semaphore slot (blocks if at capacity)
 				select {
 				case al.workerSem <- struct{}{}:
@@ -215,7 +217,7 @@ func (al *AgentLoop) Run(ctx context.Context) error {
 				case <-ctx.Done():
 					// Context canceled while waiting for a slot — clean up the
 					// placeholder to prevent session-level deadlock.
-					al.activeTurnStates.Delete(sessionKey)
+					al.releaseSessionTurnState(sessionKey, nil)
 					return
 				}
 
@@ -224,16 +226,21 @@ func (al *AgentLoop) Run(ctx context.Context) error {
 				// completes normally, clearActiveTurn deletes the real turnState and
 				// this becomes a no-op (the key is already gone).
 				defer func() {
+					if releaseSession {
+						al.releaseSessionTurnState(sessionKey, nil)
+						return
+					}
 					if actual, ok := al.activeTurnStates.Load(sessionKey); ok {
 						if ts, ok := actual.(*turnState); ok && strings.HasPrefix(ts.turnID, pendingTurnPrefix) {
 							// Placeholder still present — runTurn never replaced it.
-							al.activeTurnStates.Delete(sessionKey)
+							al.releaseSessionTurnState(sessionKey, ts)
 						}
 					}
 				}()
 
 				defer func() {
 					if r := recover(); r != nil {
+						releaseSession = true
 						logger.RecoverPanicNoExit(r)
 						logger.ErrorCF("agent", "Worker goroutine panicked",
 							map[string]any{
@@ -251,7 +258,7 @@ func (al *AgentLoop) Run(ctx context.Context) error {
 				}
 
 				if al.takePendingStop(sessionKey) {
-					al.activeTurnStates.Delete(sessionKey)
+					al.releaseSessionTurnState(sessionKey, nil)
 					target := &continuationTarget{
 						SessionKey: sessionKey,
 						Channel:    m.Channel,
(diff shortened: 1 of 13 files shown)


afjcjsbx:
#### `pkg/channels/discord/voice.go` & `pkg/bus/bus.go`

**Issue: Context timeout mismatch masks backpressure handling and drops metrics/events**

The only runtime caller of `PublishAudioChunk` passes a `context.WithTimeout` set to `100ms`. However, the new bus drop budget defined in `pkg/bus/bus.go` is set to `150ms`.

Because of this mismatch, whenever the audio queue is full, `ctx.Done()` will always trigger *before* the `ErrBusBackpressure` / `bus.message.dropped` branch can ever be reached. In production, this means drop counters, health stats, and the newly introduced event will never be updated or fired; the system will exclusively report a `context deadline exceeded` error. The new tests use `context.Background()`, which is why this configuration discrepancy went unnoticed.


#### `pkg/bus/bus.go`, `pkg/agent/agent.go` & `pkg/channels/base.go`

**Issue: Dead timeout logic causes unbounded blocking on non-dropping streams**

For all streams configured with `dropOnFull: false`, the internal `publish()` function returns from the first `select` block and never reaches the timer-based fallback logic. As a result, `defaultCriticalPublishTimeout` is effectively dead configuration, despite being explicitly passed by `PublishInbound`, `PublishOutbound`, `PublishOutboundMedia`, and `PublishVoiceControl`.

Consequently, inbound, outbound, and voice streams remain unboundedly blocking, meaning the original failure mode has not actually been resolved. Real production call sites within `pkg/agent/agent.go` and `pkg/channels/base.go` can still hang indefinitely until their own parent context is cancelled.

SiYue-ZO:
> #### `pkg/channels/discord/voice.go` & `pkg/bus/bus.go`
> **Issue: Context timeout mismatch masks backpressure handling and drops metrics/events**
> 
> The only runtime caller of `PublishAudioChunk` passes a `context.WithTimeout` set to `100ms`. However, the new bus drop budget defined in `pkg/bus/bus.go` is set to `150ms`.
> 
> Because of this mismatch, whenever the audio queue is full, `ctx.Done()` will always trigger _before_ the `ErrBusBackpressure` / `bus.message.dropped` branch can ever be reached. In production, this means drop counters, health stats, and the newly introduced event will never be updated or fired; the system will exclusively report a `context deadline exceeded` error. The new tests use `context.Background()`, which is why this configuration discrepancy went unnoticed.
> 
> #### `pkg/bus/bus.go`, `pkg/agent/agent.go` & `pkg/channels/base.go`
> **Issue: Dead timeout logic causes unbounded blocking on non-dropping streams**
> 
> For all streams configured with `dropOnFull: false`, the internal `publish()` function returns from the first `select` block and never reaches the timer-based fallback logic. As a result, `defaultCriticalPublishTimeout` is effectively dead configuration, despite being explicitly passed by `PublishInbound`, `PublishOutbound`, `PublishOutboundMedia`, and `PublishVoiceControl`.
> 
> Consequently, inbound, outbound, and voice streams remain unboundedly blocking, meaning the original failure mode has not actually been resolved. Real production call sites within `pkg/agent/agent.go` and `pkg/channels/base.go` can still hang indefinitely until their own parent context is cancelled.

Could you please take a look at the latest commit? Thank you.

afjcjsbx:
- `pkg/agent/agent_utils.go:` with `pkg/agent/pipeline_llm.go` and `pkg/agent/context_legacy.go`: `waitForActiveRequests` starts a background `Wait()` on `al.activeRequests` while the same `WaitGroup` can still receive new `Add(1)` calls from fresh LLM work. That is not a safe `sync.WaitGroup` pattern: once the counter hits zero, a new request arriving before `Wait` fully returns can panic the process with `sync: WaitGroup is reused before previous Wait has returned` / `Add called concurrently with Wait`. Since reload does not fully quiesce the agent loop before calling this helper, this race is live under concurrent traffic.

- `pkg/agent/agent.go:` and `pkg/agent/agent.go`, together with `pkg/agent/turn_coord.go`: the new panic cleanup can delete the wrong session state. If `runTurn` panics after registering a real turn, its own `defer al.clearActiveTurn(ts)` runs first during unwinding. After that, the worker-level recover path sets `releaseSession=true`, and the outer cleanup does `releaseSessionTurnState(sessionKey, nil)` unconditionally. A new message for the same session can claim the slot in between those two cleanups, and this unconditional delete will wipe the new placeholder/turn, reopening duplicate-turn or stuck-session behavior.

SiYue-ZO:
@afjcjsbx can you review again? Thank you!

afjcjsbx (approved):
Thanks @SiYue-ZO, LGTM!

PR #2906, opened 2026-05-20
Fix message bus backpressure handling and health visibility

SiYue-ZO:
## 📝 Description

The change focuses on two parts of the runtime:

1. `pkg/bus`
2. `pkg/health` / gateway health wiring

In `pkg/bus`, publish operations now use bounded waiting instead of unbounded blocking when a queue is saturated. The bus records per-stream drop statistics, emits a dedicated runtime event when a message is dropped due to backpressure, and exposes queue depth/capacity snapshots for observability.

In `pkg/health`, checks are now evaluated dynamically at request time instead of being frozen at registration time. That allows the gateway to expose live message-bus state through the existing health endpoints rather than a stale startup snapshot.

Concretely, this PR adds:

- bounded publish wait policies for all bus streams
- a shorter drop timeout for audio chunks, which are naturally lossy and should not pile up indefinitely
- `ErrBusBackpressure` to distinguish bus saturation from caller cancellation or shutdown
- per-stream queue stats and drop counters via `MessageBus.Stats()`
- a new `bus.message.dropped` runtime event kind with queue depth/capacity metadata
- a `MessageBus.HealthCheck()` summary registered into the gateway health server
- regression tests for dropped-message events, bus stats, and dynamic health-check evaluation

This keeps the existing bus API intact while making overload behavior explicit and observable.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue

No tracked GitHub issue. This addresses item 3 ("Message Bus lacks backpressure handling") from the local improvement report.

- **Reference URL:**
  - Local report: `improvement-report.md`
  - Relevant code: `pkg/bus/bus.go`, `pkg/health/server.go`, `pkg/gateway/gateway.go`
- **Reasoning:**
  - Fixed-size buffered channels are not enough on their own. When consumers slow down, publishers can block for arbitrarily long periods, which is especially risky in fan-in/fan-out runtime paths.
  - The runtime needed an explicit overload policy so saturation becomes bounded and diagnosable.
  - Health checks also needed to be evaluated dynamically; otherwise queue depth reporting would remain stale and not reflect actual runtime pressure.

## 🧪 Test Environment
- **Hardware:** PC
- **OS:** Linux
- **Model/Provider:** N/A
- **Channels:** N/A

### DIFF
--- a/pkg/bus/bus.go
+++ b/pkg/bus/bus.go
@@ -314,6 +314,12 @@ func (mb *MessageBus) Stats() MessageBusStats {
 
 func (mb *MessageBus) HealthCheck() (bool, string) {
 	stats := mb.Stats()
+	totalDropped := stats.Inbound.DroppedTotal +
+		stats.Outbound.DroppedTotal +
+		stats.OutboundMedia.DroppedTotal +
+		stats.AudioChunks.DroppedTotal +
+		stats.VoiceControls.DroppedTotal
+
 	return true, fmt.Sprintf(
 		"in=%d/%d out=%d/%d media=%d/%d audio=%d/%d voice=%d/%d dropped=%d",
 		stats.Inbound.Depth, stats.Inbound.Capacity,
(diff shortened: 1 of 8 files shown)


afjcjsbx:
The new backpressure turns temporary saturation into permanent message loss even on critical streams. In `pkg/bus/bus.go` the publish now discards the message when the timer expires, and then all major publishers trigger it with `dropOnFull: true` in `pkg/bus/bus.go.` The problem is that `ErrBusBackpressure` is not handled by any production caller outside `pkg/bus` in the tests: in practice, under full queue we can lose user inbound, follow-up and final outbound instead of applying true backpressure or retry. This is a semantic regression from the previous behavior, which let the caller's ctx decide how long to wait.

PR #2907, opened 2026-05-20
Fix JSONL store metadata drift after crash

SiYue-ZO:
## 📝 Description

This PR fixes a crash-consistency gap in the JSONL-backed memory store.

`pkg/memory/jsonl.go` appends new data to the session `.jsonl` file first and updates `.meta.json` afterwards. If the process crashes between those two writes, the JSONL file may already contain the new message while the metadata still reflects the old count or skip state. In practice this means:

- `GetHistory` can read the appended message from disk, but the metadata drift is left behind
- later operations such as `TruncateHistory` or `Compact` may keep working from stale metadata
- repeated crashes can accumulate incorrect `Count` / `Skip` values and eventually lead to incorrect truncation decisions or silent history loss

This change makes the metadata layer self-healing without changing the on-disk format:

- `NewJSONLStore` now performs a startup repair pass over existing `.meta.json` files
- `GetSessionMeta`, `GetHistory`, `TruncateHistory`, and `Compact` now reconcile metadata with the real JSONL file before proceeding
- stale `Count` values are repaired to match the real raw JSONL line count
- out-of-range `Skip` values are clamped back to the actual file size
- missing session keys and missing timestamps on non-empty sessions are repaired during reconciliation
- focused tests were added for stale count recovery, startup recovery, and `Skip` past EOF recovery
- a design note was added to document the fix scope, tradeoffs, and the follow-up WAL direction

This is intentionally a compatibility-preserving fix. It improves recovery behavior for the existing JSONL + metadata design and reduces the risk of permanent metadata drift, while leaving room for a future WAL-based implementation if stricter atomicity is needed.

## 🗣️ Type of Change
- [x] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue

Addresses the 4th item in `improvement-report.md`: `JSONL Store 的崩溃一致性问题`.

- **Reference URL:** Repository-local analysis from `improvement-report.md` and the current `pkg/memory/jsonl.go` implementation
- **Reasoning:** The current write order in `addMsg` prioritizes not losing appended JSONL lines, but it leaves a crash window where `.meta.json` can permanently lag behind the real file contents. The lowest-risk fix within the current architecture is to automatically reconcile metadata on startup and on critical read/write paths, so the store converges back to a correct state without introducing a new storage format.

## 🧪 Test Environment
- **Hardware:** PC
- **OS:** Linux
- **Model/Provider:** N/A
- **Channels:** N/A

## 📸 Evidence (Optional)
<details>

Validated locally with:

```bash
go test ./pkg/memory
go test ./pkg/memory ./pkg/fileutil
CGO_ENABLED=0 go build -tags goolm,stdjson ./cmd/picoclaw
```

Notes:

- `go test ./...` was also attempted during validation.
- The remaining failure on that command was environment-specific and unrelated to this PR: the machine did not have `olm/olm.h`, which is required by `maunium.net/go/mautrix/crypto/libolm`.

</details>

## ☑️ Checklist

### DIFF
--- a/docs/design/jsonl-crash-consistency-fix-plan.zh.md
+++ b/docs/design/jsonl-crash-consistency-fix-plan.zh.md
@@ -1,0 +1,80 @@
+# JSONL Store 崩溃一致性修复方案
+
+## 1. 问题定义
+
+`pkg/memory/jsonl.go` 的 `addMsg` 先追加 `.jsonl` 数据行，再更新 `.meta.json`。
+
+当进程在两步之间崩溃时，会出现：
+
+- `.jsonl` 已包含新消息
+- `meta.Count` 仍是旧值
+- `GetHistory` 能读到新消息，但不会修复 `meta.Count`
+- `readMessages` 只按 `meta.Skip` 读取，不校验元数据是否与真实文件一致
+
+这会导致 metadata 长期漂移，并在后续截断、压缩、重启恢复中放大为静默丢消息风险。
+
+## 2. 本次目标
+
+- 在不引入 WAL 的前提下，消除 `Count/Skip` 长期漂移
+- 把“崩溃后永久不一致”收敛为“首次访问自动自愈”
+- 保持现有 JSONL 存储格式兼容，不做破坏性迁移
+- 用单元测试覆盖启动恢复、读取恢复、越界 `Skip` 恢复
+
+## 3. 已实施改动
+
+- `NewJSONLStore` 增加启动期 metadata repair：
+  - 扫描现有 `.meta.json`
+  - 对比真实 JSONL 行数与 `meta.Count`
+  - 当 `Skip > Count` 或 `Count` 漂移时自动回写修复
+- 增加统一修复入口 `repairSessionMetaLocked`
+  - 使用真实 JSONL 原始非空行数作为 `Count`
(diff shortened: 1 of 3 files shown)


afjcjsbx:
Hey, I noticed something: PRs #2904, #2905, #2906, and #2907 all modify the same core files (pkg/agent/agent.go, pkg/bus/bus.go, pkg/providers/fallback.go, etc.) within a 2-hour window. #2905 and #2906 even share exactly the same 14 files. It looks like these changes are being stacked sequentially on top of each other rather than being properly separated. Would it make sense to rebase each one on the merged result of the previous to avoid duplication?

SiYue-ZO:
> Hey, I noticed something: PRs #2904, #2905, #2906, and #2907 all modify the same core files (pkg/agent/agent.go, pkg/bus/bus.go, pkg/providers/fallback.go, etc.) within a 2-hour window. #2905 and #2906 even share exactly the same 14 files. It looks like these changes are being stacked sequentially on top of each other rather than being properly separated. Would it make sense to rebase each one on the merged result of the previous to avoid duplication?

Thanks for the reminder that has been processed. I cleaned up the branch boundaries and force-pushed all four PR branches. They are now separated as independent PRs on top of `main`, without carrying commits from the others.

PR #2940, opened 2026-05-25
fix(providers): omit temperature for claude-opus-4-7

### LINKED ISSUE #2939: claude-opus-4-7 fails with 'temperature is deprecated for this model'
### Summary

Calls to `claude-opus-4-7` (configured via `provider: anthropic`) consistently fail with HTTP 400:

```
{"error":{"code":"invalid_request_error","message":"temperature is deprecated for this model.","type":"invalid_request_error"}}
```

The claude-opus-4-7 family no longer accepts the `temperature` parameter — sending it (even at the picoclaw default `0.7`) returns a 400 before the model is invoked. There is no per-model config knob to suppress the field today, so every request to this model fails.

### Reproduce

1. Add a model entry:
   ```json
   {
     "model_name": "claude-opus-4.7",
     "provider": "anthropic",
     "model": "claude-opus-4-7",
     "api_base": "https://api.anthropic.com/v1"
   }
   ```
2. Set a valid `sk-ant-…` Anthropic API key for that entry (no `auth_method`).
3. Send any chat message → 400 from Anthropic with the message above.

### Where temperature is injected

- `pkg/agent/instance.go:178` — defaults `temperature` to `0.7` when `agents.defaults.temperature` is unset.
- `pkg/providers/openai_compat/provider.go:178-184` — always writes `requestBody["temperature"]` for the chat-completions path that the `anthropic` protocol routes through.
- `pkg/providers/anthropic_messages/provider.go:176-179` — same unconditional write for the native Messages API.

A workaround exists (set `thinking_level: "off"` on the model entry — only works in some cases) but the root cause is the unconditional `temperature` send.

### Suggested fix

Skip the field for models that deprecate it, mirroring the existing kimi-k2 special case in `openai_compat`. PR follows.

### Environment

- picoclaw v0.2.7 (`git_commit` 71c877a6, built 2026-04-22)
- Docker image `sipeed/picoclaw:launcher`
- Anthropic API

LegendAlessandro-Liguori:
## Summary

- `claude-opus-4-7` deprecates the `temperature` parameter — sending it returns HTTP 400 `invalid_request_error` "temperature is deprecated for this model" and the request never reaches the model.
- Picoclaw unconditionally writes `temperature` into the request body in both the chat-completions path (`openai_compat`, used by the `anthropic` protocol) and the native Messages API path (`anthropic_messages`), so every call to this model fails.
- This patch adds `common.AnthropicModelOmitsTemperature(model)` and uses it in both providers to skip the field for affected models. Mirrors the existing kimi-k2 special case in `openai_compat`.

Fixes #2939.

## Test plan

- [x] `go test ./pkg/providers/...` — all green locally (golang:1.25 container).
- [x] New unit tests `TestAnthropicModelOmitsTemperature` cover bare/dated/cased/whitespace variants and confirm sonnet/haiku/non-anthropic models are unaffected.
- [ ] Manual: with the patch applied, `claude-opus-4-7` returns a normal completion against `https://api.anthropic.com/v1` instead of the 400.

## Notes

- Match is a prefix on `claude-opus-4-7` so dated variants (e.g. `claude-opus-4-7-20260601`) are covered without further changes.
- Defaults are unchanged for every other model — sonnet 4.6, haiku 4.5, opus 4.5 and earlier all still receive temperature as before.

### DIFF
--- a/pkg/providers/anthropic_messages/provider.go
+++ b/pkg/providers/anthropic_messages/provider.go
@@ -173,8 +173,9 @@ func buildRequestBody(
 		"messages":   []any{},
 	}
 
-	// Set temperature from options
-	if temp, ok := common.AsFloat(options["temperature"]); ok {
+	// Set temperature from options, unless the target model deprecates it.
+	if temp, ok := common.AsFloat(options["temperature"]); ok &&
+		!common.ModelOmitsTemperature(model) {
 		result["temperature"] = temp
 	}
 
--- a/pkg/providers/common/anthropic_common.go
+++ b/pkg/providers/common/anthropic_common.go
@@ -2,6 +2,33 @@ package common
 
 import "strings"
 
+// ModelOmitsTemperature reports whether an LLM model rejects requests that
+// include the temperature parameter.
+//
+// Affected models:
+//   - Anthropic claude-opus-4-7 family: returns HTTP 400 invalid_request_error
+//     "temperature is deprecated for this model".
+//   - OpenAI gpt-5.5 family: returns HTTP 400
+//     "Unsupported value: 'temperature' does not support … with this model.
+//     Only the default (1) value is supported." Setting temperature=1 works,
+//     but omitting it is simpler and matches the documented default.
+//
+// Checks are prefix matches so dated variants (e.g. claude-opus-4-7-20260601,
+// gpt-5.5-20260601) are covered without further code changes.
+func ModelOmitsTemperature(model string) bool {
+	lower := strings.ToLower(strings.TrimSpace(model))
+	if lower == "" {
+		return false
+	}
+	switch {
+	case strings.HasPrefix(lower, "claude-opus-4-7"):
+		return true
+	case strings.HasPrefix(lower, "gpt-5.5"):
+		return true
+	}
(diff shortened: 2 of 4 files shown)


LegendAlessandro-Liguori:
Scope extended — same root cause turned out to apply to OpenAI `gpt-5.5` as well:

> Unsupported value: 'temperature' does not support 0.7 with this model. Only the default (1) value is supported.

Pushed an amended commit that renames the helper to `common.ModelOmitsTemperature` and adds a `gpt-5.5` prefix match alongside `claude-opus-4-7`. Tests extended to cover both families plus negative cases (`gpt-5.4`, `gpt-5.4-mini`, sonnet 4.6, haiku 4.5, opus 4.5).

Couldn't update the PR title/body via `gh pr edit` (`projectCards` GraphQL deprecation error) — happy to retitle via the web UI if a maintainer prefers, e.g.

> fix(providers): omit temperature for claude-opus-4-7 and gpt-5.5

afjcjsbx (approved):
LGTM!

afjcjsbx:
@LegendAlessandro-Liguori pls accept cla

PR #2956, opened 2026-05-27
fix: preserve channel enabled state when merging security.yml

### LINKED ISSUE #2742: [BUG] gateway starts with no channels in v0.2.8
##  Environment & Tools
- **PicoClaw Version:** 0.2.8
- **Go Version:** go1.25.9
- **AI Model & Provider:** openrouter/tencent/hy3-preview:free
- **Operating System:** Ubuntu 22.04
- **Channels:** Telegram

## 📸 Steps to Reproduce
1. config.json 
"telegram": {
      "enabled": true,
      "type": "telegram",
      "allow_from": [""]
 },
2. .security.yml
telegram:
    token: "my-bot-token" 
3. picoclaw gateway
## ❌ Actual Behavior
📦 Agent Status:
  • Tools: 16 loaded
  • Skills: 8/8 available
✓ Cron service started
✓ Heartbeat service started
⚠ Warning: No channels enabled
09:38:49 WRN channels ../home/runner/work/picoclaw/picoclaw/pkg/channels/manager.go:743 > No channels enabled
✓ Health endpoints available at http://localhost:18790/health, /ready and /reload (POST)
✓ Gateway started on [::1]:18790
✓ Gateway started on 127.0.0.1:18790
## ✅ Expected Behavior
telegram chat with bot working

yuxuan-7814:
## 📝 Description

Fix an issue where channels configured as `enabled: true` in `config.json` were being disabled after loading `.security.yml`. When users added credentials (e.g., `telegram.token`) to security.yml without explicitly setting `enabled: true`, the merge process would overwrite the enabled state from the main config.

This fix preserves the original `enabled` state when the YAML node in security.yml does not explicitly contain an `enabled` field.

## 🗣️ Type of Change

## 🤖 AI Code Generation

## 🔗 Related Issue

Closes #2742

## 📚 Technical Context
- **Reference URL:** https://github.com/sipeed/picoclaw/issues/2742
- **Reasoning:** The `Channel.UnmarshalYAML` method was not checking if the `enabled` field was explicitly present in the YAML before overwriting the existing value. This adds a check using `isYAMLFieldPresent` to preserve the original state.

## 🧪 Test Environment
- **Hardware:** PC / Server
- **OS:** Linux / Ubuntu 22.04
- **Model/Provider:** N/A

## ☑️ Checklist

### DIFF
--- a/pkg/config/config_channel.go
+++ b/pkg/config/config_channel.go
@@ -3,7 +3,9 @@ package config
 import (
 	"encoding/json"
 	"fmt"
+	"os"
 	"reflect"
+	"strconv"
 	"strings"
 
 	"github.com/caarlos0/env/v11"
@@ -33,6 +35,8 @@ const (
 	ChannelWhatsApp       = "whatsapp"
 	ChannelWhatsAppNative = "whatsapp_native"
 	ChannelTeamsWebHook   = "teams_webhook"
+	ChannelMQTT           = "mqtt"
+	ChannelSlackWebHook   = "slack_webhook"
 )
 
 func initChannel() {
@@ -237,6 +241,7 @@ func (b Channel) MarshalJSON() ([]byte, error) {
 		if err != nil {
 			return nil, err
 		}
+		raw = preserveExplicitDisabledStreaming(raw, b.Settings)
 		settings = raw
 	} else {
 		settings = b.Settings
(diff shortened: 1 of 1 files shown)


afjcjsbx:
Hi @yuxuan-7814 thanks for the PR but this doesn't actually resolve #2742. The reproduction case for the issue uses a flat format in `.security.yml`:

```yaml
telegram:
  token: ...
```

However, `UnmarshalYAML` still only decodes the nested `settings:` block inside `Channel.Settings`. I reproduced this flow on the PR's merge-base: after the merge, `Enabled` remains `true`, but `TelegramSettings.Token` stays empty. Because of this, the logic in `pkg/channels/manager.go` discards the Telegram channel anyway, meaning the final result is still "No channels enabled".

Additionally, the new logic using `isYAMLFieldPresent` implies that an explicit `enabled:` field in `.security.yml` should override the previous value. However, this cannot happen because `Channel.Enabled` is still marked as `yaml:"-"`, meaning `value.Decode(&a)` will never read that field. In practice, `enabled: false` inside `.security.yml` continues to be ignored for existing channels, so the implementation does not match the PR description.

PR #2965, opened 2026-05-29
fix(tools): stop workspace guard misreading scheme-less URLs

### LINKED ISSUE #1042: [BUG]exec工具的guardCommand方法问题
## Quick Summary

当设置`restrict_to_workspace`为`true`时，`exec`工具的`guardCommand`方法会判断命令的相对路径是否合法，但是判断方式有点简单粗暴。

举个例子，查询北京的天气，天气技能会执行`curl -s "wttr.in/Beijing?T"`这个命令，但方法的正则match出来的结果会得到`../../../../Beijing?T`这样的相对路径，导致`Command blocked by safety guard (path outside working dir)`，问题是这个命令压根不涉及到任何路径。

maxmilian:
## Description

When `restrict_to_workspace` is enabled, the `exec` tool's `guardCommand`
extracted every `/...`-led run from the command and treated each as an absolute
path. A **scheme-less URL** — which `curl` accepts without an explicit `http://`
prefix, e.g. `curl -s "wttr.in/Beijing?T"` — had its `/Beijing?T` component
matched, resolved with `filepath.Abs`, and `filepath.Rel(cwd, …)` turned it into
`../../Beijing?T`, tripping the `..` "path outside working dir" check. Reporters
note this makes **most curl-based skills unusable**.

## Fix

Skip a `/path` match **only** when the run of characters immediately before the
slash parses as a hostname: host-legal bytes `[A-Za-z0-9.-]`, not starting with
`-`, containing a `.`, not starting/ending with `.`.

Because a skipped match is glued to its host with **no separator**, the shell
treats the whole word as a single URL/relative token — never a standalone
absolute path. The skip therefore cannot whitelist a real escape:

| Command | Left token | Skipped? | Result |
|---|---|---|---|
| `curl wttr.in/Beijing` | `wttr.in` | ✅ host | allowed |
| `curl 192.168.1.1/status` | `192.168.1.1` | ✅ host | allowed |
| `tar -xC/etc/cron.d …` | `-xC` (starts `-`) | ❌ | **blocked** |
| `cat $x/etc/passwd` | `x` (no dot) | ❌ | **blocked** |
| `java -cp app.jar:/etc/passwd` | empty (walk stops at `:`) | ❌ | **blocked** |
| `scp host:/etc/passwd` | empty (stops at `:`) | ❌ | **blocked** |
| `cat /etc/passwd` | empty (stops at space) | ❌ | **blocked** |

`..` traversal is still caught by the existing pre-loop check; the existing
web-scheme (`//host` after `http:`/`https:`/…) skip is unchanged.

## Type of Change

## AI Code Generation
- [x] 🤖 Fully AI-generated (AI-written, human-reviewed + validated)

The fix approach was human-selected and the security boundary was hardened
across three independent adversarial review passes — two of which found and
closed sandbox-escape classes (flag-glued `-C` paths, and `:` path-lists) before
this PR was opened.

## Related Issue
Fixes #1042

## Technical Context
- Root cause: `pkg/tools/shell.go` `absolutePathPattern` (`/[^\s"']+`) matches a
  URL's host-relative path component; the prior exemption only covered `//` after
  a web scheme, not scheme-less hosts.
- New helper `hostLikeBeforeSlash` gates the skip on a hostname check.
- `:` is deliberately excluded from the host-character walk so classpath /
  remote-spec path-lists (`java -cp a.jar:/abs`, `scp host:/abs`) are not mistaken
  for `host:port` and stay blocked.

## Test Environment
- OS: macOS (Darwin) · Go 1.25
- `go test ./pkg/tools/`: new `TestShellTool_SchemelessURLGuard` passes (5
  scheme-less URLs allowed + 5 escape regressions blocked); existing
  `TestShellTool_URLsNotBlocked` / `URLBypassPrevented` / `PathTraversalVariants`
  / `RestrictToWorkspace` / `SafePaths` all pass.
- `TestShellTool_DevNullAllowed` and `TestShellTool_FileURISandboxing` fail
  locally **only** due to the macOS `/tmp`→`/private/tmp` symlink (pre-existing on
  `main`, unrelated to this change); they pass on Linux CI.

## Checklist
- [x] I have read and tested the AI-generated code
- [x] gofmt-clean and `go vet` passes
- [x] Added test coverage (both directions, incl. escape regressions)
- [x] No new shell-injection / path-traversal surface

### DIFF
--- a/pkg/tools/shell.go
+++ b/pkg/tools/shell.go
@@ -1155,6 +1155,21 @@ func (t *ExecTool) guardCommand(command, cwd string) string {
 				}
 			}
 
+			// Skip a "/path" match that is the path component of a scheme-less
+			// URL (e.g. "wttr.in/Beijing" — curl accepts URLs without an explicit
+			// http:// prefix), which is a URL token, not an absolute-path argument
+			// (issue #1042). We skip ONLY when the run of characters immediately
+			// before the "/" parses as a hostname/authority. This deliberately
+			// does NOT skip flag-glued absolute paths like "tar -xC/etc/cron.d"
+			// (preceding token "-xC" starts with "-") or "cat $x/etc/passwd"
+			// (token "x" has no dot), so genuine sandbox escapes stay blocked.
+			// A skipped match is always glued to its host with no separator, so
+			// the shell treats it as one token (a URL or relative path), never as
+			// a standalone absolute path. ".." traversal is caught earlier.
+			if hostLikeBeforeSlash(cmd, loc[0]) {
+				continue
+			}
+
 			p, err := filepath.Abs(raw)
 			if err != nil {
 				continue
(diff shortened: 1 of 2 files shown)


afjcjsbx:
Hi @maxmilian , thanks for the PR, 

`pkg/tools/shell.go` and `pkg/tools/shell.go` introduce a new sandbox bypass for dotted relative paths. `hostLikeBeforeSlash()` treats any `foo.bar/...` token as URL-like and skips the existing `filepath.Abs` / `EvalSymlinks` / `Rel` checks entirely.

In practice, in a workspace that contains a versioned symlink like `foo.bar -> /`, running `cat foo.bar/etc/hosts` now bypasses the guard and reads outside the workspace, whereas it was blocked prior to this PR.

We need to either properly identify a true URL token-instead of just matching any prefix with a dot-or re-run the validation on the full token before triggering the `continue` statement.

afjcjsbx:
`hostLikeBeforeSlash()` now treats any `foo.bar/...` token as a scheme-less URL and skips the path check before `EvalSymlinks()`. That means a relative path inside the workspace can evade the guard if it contains a dot and resolves through a symlink outside the workspace.

Repro:
1. Create `safe.example -> /tmp/secret` inside the workspace
2. Run `cat safe.example/secret.txt`

On `main` this is blocked by the workspace guard; on this PR it succeeds and reads the file, because `safe.example/secret.txt` is classified as a URL-like token.

Could we tighten the heuristic and add a regression test for a dotted symlink/path case? Right now this looks like a security regression, not just an edge case.

maxmilian:
Closing — superseded by #3001 (merged as `b86ab718`), which fixes the same scheme-less URL workspace-guard misread (`wttr.in/Beijing`) using `looksLikeDomain` + `localPathExists`.

Importantly, #3001's local-path-exists guard is exactly the protection @afjcjsbx correctly flagged my `hostLikeBeforeSlash` approach was missing — without it, a dotted relative path could slip past `EvalSymlinks()` (a sandbox bypass). The merged approach is the safer one. Thanks for the review! 🙏

PR #2982, opened 2026-06-01
fix(bedrock): drop temperature for models that deprecate it (Opus 4.8)

loafoe:
## Problem

Bumping the model to Claude Opus 4.8 on AWS Bedrock causes every LLM call to fail:

```
LLM call failed after retries: bedrock converse: operation error Bedrock Runtime: Converse,
https response error StatusCode: 400, ValidationException: The model returned the following errors:
temperature is deprecated for this model.
```

Opus 4.8 no longer accepts the `temperature` inference parameter, so any config that sets a temperature (or a non-nil default) breaks the provider.

## Fix

`buildConverseParams` now receives the model id and omits `temperature` from the `InferenceConfiguration` for models that deprecate it. A new `modelDeprecatesTemperature` helper matches `claude-opus-4-8*` via a case-insensitive substring check, which covers both bare model ids (`claude-opus-4-8-…`) and region-prefixed inference profiles (`us.anthropic.claude-opus-4-8-…`). When the parameter is dropped, it logs a line for visibility. `max_tokens` and all other models are unaffected.

## Tests

- `TestModelDeprecatesTemperature` — table test across Opus 4.8 (bare/profile/mixed-case), Opus 4.7, Sonnet, empty.
- `TestBuildConverseParams_DropsTemperatureForDeprecatedModel` — temperature omitted, max_tokens preserved.
- `TestBuildConverseParams_KeepsTemperatureForSupportedModel` — temperature preserved for Opus 4.7.

All pass with `go test -tags bedrock ./pkg/providers/bedrock/`.

### DIFF
--- a/pkg/providers/bedrock/provider_bedrock.go
+++ b/pkg/providers/bedrock/provider_bedrock.go
@@ -143,7 +143,22 @@ type converseParams struct {
 	toolConfig     *types.ToolConfiguration
 }
 
-func buildConverseParams(messages []Message, tools []ToolDefinition, options map[string]any) converseParams {
+// modelDeprecatesTemperature reports whether the given Bedrock model rejects the
+// temperature inference parameter. Newer Claude models (Opus 4.8 and later) treat
+// temperature as deprecated and return a ValidationException if it is supplied:
+//
+//	ValidationException: The model returned the following errors:
+//	temperature is deprecated for this model.
+//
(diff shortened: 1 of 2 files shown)


loafoe:
@afjcjsbx  this PR addresses Opus 4.8 compatiblity for the bedrock provider

afjcjsbx (approved):

PR #2985, opened 2026-06-02
fix(context): show both summarize and compress thresholds in /context

### LINKED ISSUE #2968: [BUG] /context always show Compress at: 76800 tokens
##  Environment & Tools
- **Go Version:** 1.25.10
- **Channels:** Web

## 📸 Steps to Reproduce
1. Set provider&model to MiniMax-M2.7-highspeed, max_tokens: 128000, context_window: 204800, summarize_message_threashold: 200, summarize_token_percent: 80
3. Talk with agent in WebUI
4. Send /contenxt command

## ❌ Actual Behavior
You will see:
```
Context usage
Messages: 36
Used: ~38807 / 204800 tokens (18%)
Compress at: 76800 tokens
Compression progress: 50%
Remaining: ~37993 tokens
```

## ✅ Expected Behavior
I think 80% of 204800 should be:
```
Compress at: 163910 tokens
Remaining: ~165993 tokens
```

## 💬 Additional Context
My setting:
```
      "image_model": "MiniMax-M2.7",
      "max_tokens": 128000,
      "context_window": 204800,
      "max_tool_iterations": 50,
      "summarize_message_threshold": 200,
      "summarize_token_percent": 80,
      "steering_mode": "one-at-a-time"
```

I found that whatever I set summarize_token_percent to what, like 50, 75, 80, Compress at is always 76800 tokens, I can't understand what is going on.

chengzhichao-xydt:
## Description

The /context command previously showed only the hard budget compression threshold (contextWindow - maxTokens), which confused users who expected to see the soft summarization trigger from summarize_token_percent. This was reported in #2968.

This PR adds SummarizeAtTokens alongside the existing CompressAtTokens so that both thresholds are visible:

- **Compress at**: contextWindow - maxTokens ??hard budget, triggers proactive compression when exceeded (existing behavior)
- **Summarize at**: contextWindow * summarizeTokenPercent / 100 ??soft trigger, matches maybeSummarize threshold (new)

## Type of Change

## AI Code Generation

## Related Issue

Fixes #2968

## Technical Context

- **Reference URL:** https://github.com/sipeed/picoclaw/issues/2968
- **Reasoning:** The compressAt value is calculated as contextWindow - maxTokens in computeContextUsage(), which is the hard prompt-budget compression point. Users with high max_tokens see a very low Compress at value that does not reflect their summarize_token_percent setting. The fix adds a separate Summarize at line that shows the soft summarization trigger (contextWindow * summarizeTokenPercent / 100), matching the existing maybeSummarize logic in context_legacy.go.

### Changes

- pkg/bus/types.go: Added SummarizeAtTokens field to ContextUsage`n- pkg/agent/context_usage.go: Compute SummarizeAtTokens in computeContextUsage()`n- pkg/commands/runtime.go: Added SummarizeAtTokens to ContextStats`n- pkg/commands/cmd_context.go: Show both thresholds in /context output
- pkg/agent/agent_command.go: Pass SummarizeAtTokens through
- pkg/channels/pico/pico.go: Serialize summarize_at_tokens in WebSocket payload
- web/frontend/src/store/chat.ts: Added summarize_at_tokens to TypeScript interface
- web/frontend/src/features/chat/protocol.ts: Parse summarize_at_tokens from JSON
- web/frontend/src/components/chat/context-usage-ring.tsx: Display both values in popover

## Test Environment

- **OS:** Windows (local verification)
- **Verification:** go build ./pkg/bus/... ./pkg/agent/... ./pkg/commands/... ./pkg/channels/pico/... all pass; go test ./pkg/agent/... and go test ./pkg/channels/pico/... all pass (9 changed files, 59 insertions, 26 deletions)


afjcjsbx:
Hi @chengzhichao-xydt ! Thanks for the PR. Here are a few critical points that need to be addressed before this can be merged:

### 1. Logic Misalignment in Summarization Threshold

The new `Summarize at` threshold does not match the actual runtime condition that triggers summarization.

* In `context_usage.go`, it is exposed as `contextWindow * SummarizeTokenPercent`.
* However, the engine determines whether to trigger summarization by looking solely at history tokens and the message count, rather than the total `UsedTokens` (as seen in `context_legacy.go`).
* Since `/context` displays this threshold right next to `UsedTokens`—which includes the system prompt, summary, and tool definitions (handled in `cmd_context.go`)—the UI might incorrectly indicate to the user that the threshold has been exceeded when the actual trigger hasn't been met yet.

> **Recommendation:** Please align the displayed metric with the actual logic used by the engine to avoid confusing UX.

### 2. Hardcoded Text in Frontend (i18n Regression)

Hardcoded English text was introduced inside a component that is already localized.

* In `context-usage-ring.tsx`, the labels `Compress at` and `Summarize at` are hardcoded as plain strings, whereas the rest of the UI relies on `i18n`.
* This will cause an immediate visual regression for users on non-English locales.

> **Recommendation:** Refactor these strings to utilize the standard `i18n` localization system.

### 3. Missing Test Coverage

There is currently no test coverage for the newly exposed field.

* The existing tests for the `context_usage` payload in `pico_test.go` continue to verify only the older fields.
* There is no dedicated test for the new output of `/context`. If `summarize_at_tokens` stops being properly serialized or breaks in the future, the CI pipeline would currently pass without catching it.

> **Recommendation:** Add explicit test cases in `pico_test.go` to assert the correct serialization and presence of the new field.

chengzhichao-xydt:
Addressed all three review points:

1. **Threshold alignment**: Added code comment clarifying that summarize_at_tokens matches maybeSummarize's formula (history-token based), while UsedTokens in /context includes system prompt + summary + tools - explaining the difference to avoid UX confusion.

2. **i18n regression**: Replaced hardcoded Compress at / Summarize at strings with 	(\chat.contextCompressAt\) / 	(\chat.contextSummarizeAt\), added translations to all 5 locale files (en, zh, pt-br, cs, bn-in).

3. **Test coverage**: Added SummarizeAtTokens: 60 to TestBeginStream_FinalizeIncludesContextUsage and assert its presence in the serialized payload.

afjcjsbx:
#### Token metrics misalignment (`pkg/commands/cmd_context.go`, `context-usage-ring.tsx`, `context_usage.go`, `context_legacy.go`)

The alignment issue isn't quite fixed yet. We are still displaying `Summarize at: ...` right next to `Used: ...`, but these two metrics are apples and oranges:

* **`UsedTokens`** aggregates everything (history + system prompt + summary + tool definitions).
* **`maybeSummarize`** evaluates the threshold **strictly against history tokens**.

While the new code comment acknowledges this mismatch, the user-facing output from `/context` and the UI tooltip will still mislead users. They will see "you are past the summarize threshold" even when the engine won't actually trigger a summary.

* **Suggestion:** We either need to expose a dedicated `history-only` token counter to the UI, or explicitly relabel the threshold so users know it only tracks history.


#### Workflow artifacts leaked into PR (`.codebuddy/github-contribute/*`)

The PR includes five files under `.codebuddy/github-contribute/` that look like local/generated state from an automation or contribution workflow (tracking active branches, open PRs, reviewers), rather than actual product code.

While they won't break the build, leaving them in introduces permanent noise and will cause useless commit churn on future runs.

chengzhichao-xydt:
Addressed the remaining feedback:

1. **Token metrics alignment**: Added `HistoryTokens` field to `ContextUsage`/`ContextStats`, now shown as a dedicated `History:` line in `/context` output and the Web UI tooltip. This makes it clear that `Summarize at` is compared against history-only tokens (not total `UsedTokens`), resolving the apples-to-oranges UX concern.

2. **State files**: Removed the `.codebuddy/github-contribute/` files accidentally included in the PR.

Changes span 10 files: `bus/types.go`, `context_usage.go`, `commands/runtime.go`, `cmd_context.go`, `agent_command.go`, `pico.go`, `pico_test.go`, `context-usage-ring.tsx`, `store/chat.ts`, `features/chat/protocol.ts`.

afjcjsbx (approved):

PR #2986, opened 2026-06-02
fix(tools): add Stop() to SessionManager to prevent goroutine leak

chengzhichao-xydt:
## Description

The SessionManager's background cleanup goroutine previously had no shutdown mechanism. Each call to NewSessionManager() started a ticker goroutine that ran indefinitely, causing goroutine leaks when multiple SessionManagers were created (e.g. in tests).

This commit adds a Stop() method that cleanly shuts down the background cleanup goroutine via a channel. Stop() is safe to call multiple times (idempotent). The SessionManager remains usable after Stop() - only the cleanup goroutine is terminated.

All existing tests now call 	.Cleanup(sm.Stop) to ensure goroutines are cleaned up after each test.

## Type of Change

- [x] Bug fix (non-breaking change)

## Test Environment

- **OS:** Windows
- **Verification:** go build ./pkg/tools/... and go test ./pkg/tools/ -run TestSessionManager all pass (3 files, +29/-6)

### DIFF
--- a/pkg/tools/session.go
+++ b/pkg/tools/session.go
@@ -171,23 +171,42 @@ func (s *ProcessSession) ToSessionInfo() SessionInfo {
 type SessionManager struct {
 	mu       sync.RWMutex
 	sessions map[string]*ProcessSession
+	stopCh   chan struct{}
 }
 
 func NewSessionManager() *SessionManager {
 	sm := &SessionManager{
 		sessions: make(map[string]*ProcessSession),
+		stopCh:   make(chan struct{}),
 	}
 
 	// Start cleaner goroutine - runs every 5 minutes, cleans up sessions done for >30 minutes
 	go func() {
 		ticker := time.NewTicker(5 * time.Minute)
 		defer ticker.Stop()
-		for range ticker.C {
-			sm.cleanupOldSessions()
+		for {
+			select {
+			case <-sm.stopCh:
+				return
+			case <-ticker.C:
+				sm.cleanupOldSessions()
+			}
 		}
 	}()
 
 	return sm
+}
+
+// Stop shuts down the background cleanup goroutine. Safe to call multiple times.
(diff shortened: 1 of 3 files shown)


afjcjsbx:
LGTM! The changes look great and are ready to go as soon as the linter errors are fixed.

I did notice one minor edge-case issue regarding concurrency, but it absolutely **doesn't need to block this PR**. We can merge this now and tackle this fix in a follow-up PR or tracking issue.

### Stop() is not fully thread-safe (`pkg/tools/session.go`, line 203)

The `Stop()` method is documented/intended to be "safe to call multiple times," but the current pattern isn't entirely safe under concurrent conditions:

```go
select { 
case <-sm.stopCh: 
default: 
    close(sm.stopCh) 
}

```

* This approach is only idempotent if the calls are strictly serial (one after the other). If two goroutines call `Stop()` at the exact same time, both can pass the `case <-sm.stopCh:` check and enter the `default` block simultaneously. This will cause the second goroutine to panic by trying to close an already closed channel.
* To fully guarantee the contract of being safe for multiple concurrent calls, we should protect this logic using a `sync.Once` or a standard mutex lock.

As mentioned, this is a minor edge case and shouldn't hold up this delivery. Feel free to merge this whenever you're ready, and we can address this thread-safety nit in a separate, quick PR later!

chengzhichao-xydt:
Thanks for the review! I've addressed the feedback:

- **Stop() concurrency**: Replaced the select/default pattern with sync.Once to guarantee exactly-once close semantics, matching the documented contract of being safe for concurrent use.

  sync.Once is already imported (part of the standard sync package), so no new dependencies.

afjcjsbx:
thanks! could you fix lint pls? 🙏

chengzhichao-xydt:
Fixed lint (gofmt applied).

afjcjsbx (approved):
LGTM!

PR #2987, opened 2026-06-02
fix(channels): exclude tool_calls from auxiliary message filtering

### LINKED ISSUE #2958: tool_calls messages dropped during consecutive requests via pico channel
## Bug Description

When making consecutive requests that invoke tools via the pico channel, subsequent `tool_calls` messages are not delivered to the UI. Only the first request's tool_calls appear.

## Steps to Reproduce

1. Connect to picoclaw via the pico WebSocket channel
2. Send a message that triggers a tool call (e.g., "whoami")
3. Wait for the response
4. Immediately send another message that triggers a tool call
5. Observe that the second tool_calls message does not appear in the UI

## Expected Behavior

All tool_calls messages should be delivered to the UI, regardless of streaming state.

## Actual Behavior

The second (and subsequent) tool_calls messages are silently dropped when:
- There is an active stream from a previous request
- There is a tombstone from a recently finished stream (within 30 seconds)

## Root Cause

Introduced in #2892. The `preSend()` function filters "auxiliary messages" (messages with `message_kind` set) when streaming is active or has recently finished. Since `tool_calls` messages have `message_kind=tool_calls`, they are incorrectly filtered out.

The intent was to drop stale feedback/thoughts, but `tool_calls` are not stale content - they represent new tool invocations that must be delivered.

## Affected Code

`pkg/channels/manager.go`:
```go
if isAuxiliaryMessage {
    if _, loaded := m.streamActive.Load(streamKey); loaded {
        return nil, true  // tool_calls incorrectly dropped here
    }
    if m.streamAuxiliaryTombstoneActive(streamKey) {
        return nil, true  // tool_calls incorrectly dropped here
    }
}
```

## Fix

PR #2957 excludes `tool_calls` from auxiliary message filtering.

## Environment

- picoclaw version: v0.2.9 and later (after #2892 merged)
- Channel: pico

yuxuan-7814:
## Summary

This PR fixes issue #2958 where tool_calls messages were incorrectly dropped during active streaming sessions.

## Root Cause

The preSend() function in pkg/channels/manager.go was filtering out all auxiliary messages when a stream was active or had recently finished. Since tool_calls messages have message_kind=tool_calls, they were incorrectly filtered out.

## Changes

- Modified the auxiliary message filter to exclude tool_calls type messages
- Added comment explaining the rationale

## Related Issue

Fixes #2958

## Testing

- Code change is minimal and focused on the root cause
- The fix ensures tool_calls are delivered even during active streaming

### DIFF
--- a/pkg/channels/manager.go
+++ b/pkg/channels/manager.go
@@ -23,6 +23,7 @@ import (
 	"github.com/sipeed/picoclaw/pkg/bus"
 	"github.com/sipeed/picoclaw/pkg/config"
 	"github.com/sipeed/picoclaw/pkg/constants"
+	runtimeevents "github.com/sipeed/picoclaw/pkg/events"
 	"github.com/sipeed/picoclaw/pkg/health"
 	"github.com/sipeed/picoclaw/pkg/logger"
 	"github.com/sipeed/picoclaw/pkg/media"
@@ -40,6 +41,8 @@ const (
 	janitorInterval = 10 * time.Second
 	typingStopTTL   = 5 * time.Minute
 	placeholderTTL  = 10 * time.Minute
+
+	streamAuxiliaryTombstoneTTL = 30 * time.Second
 )
 
 // typingEntry wraps a typing stop function with a creation timestamp for TTL eviction.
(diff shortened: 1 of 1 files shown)


afjcjsbx:
Hi @yuxuan-7814 , thanks for the PR, please fix conflicts 🙏

PR #2988, opened 2026-06-02
fix(agent): use summarize_token_percent config for context compression

### LINKED ISSUE #2968: [BUG] /context always show Compress at: 76800 tokens

yuxuan-7814:
## Summary

This PR fixes issue #2968 where the `/context` command always showed 'Compress at: 76800 tokens' regardless of the `summarize_token_percent` configuration setting.

## Root Cause

The `computeContextUsage()` function in `pkg/agent/context_usage.go` was using the full `effectiveWindow` as the `compressAt` threshold, ignoring the `summarize_token_percent` config value.

## Changes

- Modified `compressAt` calculation to use `effectiveWindow * summarizeTokenPercent / 100`
- Added default fallback of 75% when config value is not set

## Related Issue

Fixes #2968

## Testing

- Verified the fix uses the configured percentage value
- Default fallback of 75% maintains backward compatibility

### DIFF
--- a/pkg/agent/context_usage.go
+++ b/pkg/agent/context_usage.go
@@ -57,9 +57,15 @@ func computeContextUsage(agent *AgentInstance, sessionKey string) *bus.ContextUs
 		effectiveWindow = contextWindow
 	}
 
-	// compressAt = effectiveWindow: aligns with isOverContextBudget's
-	// proactive trigger (msgTokens + toolTokens + maxTokens > contextWindow).
-	compressAt := effectiveWindow
+	// compressAt = effectiveWindow * summarizeTokenPercent / 100
+	// This ensures compression triggers at the configured percentage of the context window.
+	// Fixes issue #2968 where CompressAtTokens was always showing 76800 (default effectiveWindow)
+	// regardless of the summarize_token_percent configuration.
+	summarizeTokenPercent := agent.SummarizeTokenPercent
+	if summarizeTokenPercent <= 0 {
+		summarizeTokenPercent = 75 // default fallback
+	}
+	compressAt := effectiveWindow * summarizeTokenPercent / 100
 
 	usedPercent := 0
 	if compressAt > 0 {


afjcjsbx:
Thanks for putting this PR together to tackle #2968! While the intent is spot on, there are a couple of critical logic gaps regarding how the context threshold is calculated and how it impacts the rest of the application. As it stands, the PR doesn't fully resolve the linked issue and introduces some semantic misalignments. 

## Incorrect Calculation Logic (Doesn't fix #2968)

The new formula introduced in `context_usage.go` doesn't implement the expected behavior outlined in issue #2968.

* If we plug in the numbers from the issue (`context_window=204800`, `max_tokens=128000`, `summarize_token_percent=80`), this PR computes $(204800 - 128000) \times 80\% = 61440$. However, the expected target is $204800 \times 80\% = 163840$.
* The actual, real-world summarization threshold remains tightly coupled to `context_window * summarize_token_percent / 100` over in `context_legacy.go` (line 85). Because this legacy logic hasn't been updated to match, the PR unfortunately leaves the reported bug unresolved.

### Semantic Mismatch & UI/API Desync

This change alters the definition of `CompressAtTokens` without actually updating the backend logic that executes the context compression.

* `CompressAtTokens` is explicitly documented as the *"threshold that triggers compression"* (`types.go`), is rendered literally as *"Compress at"* in the `/context` command (`cmd_context.go`), and acts as the denominator for the progress ring in the Web UI (`context-usage-ring.tsx`).
* Proactive compression still actually triggers based on `used + max_tokens > context_window` in `context_budget.go`. By changing the meaning of this token count here, the UI and API will end up reporting "100%" or "remaining=0" way before or way after the actual compression event kicks in. We definitely want to avoid giving the user a mini heart attack with a premature 100% warning.

### Open Questions & Required Tests

* **Missing Regression Tests:** I don't see any unit tests covering `computeContextUsage` in this change.
* **Action Item:** Please add at least one test case using the exact scenario from the issue (`204800` / `128000` / `80`). If we had that test coverage in place, the `61440` result would have flagged the build immediately! Let's get that added to prevent future regressions.

PR #2990, opened 2026-06-02
fix(web): read full session history for Web UI display

### LINKED ISSUE #2796: [BUG]历史记录中，一次对话有多次用户消息，只能查看到最后一条用户消息，先前的都看不到
##  Environment & Tools
g4d3070e8

## 📸 Steps to Reproduce
1. 新建对话，对话中多次发送用户消息。
2. 新建其他对话，沟通后，再从历史记录中查看先前的对话
3. 先前对话多次消息中只能查看到最后一条用户消息，先前的都看不到

## ❌ Actual Behavior
先前对话多次消息中只能查看到最后一条用户消息，先前的都看不到
## ✅ Expected Behavior
消息压缩应该是针对大模型的，对用户显示的历史消息应该完整

yuxuan-7814:
## Summary

This PR fixes issue #2796 where users could only see the last user message in a conversation with multiple user messages when viewing session history.

## Root Cause

The `readJSONLSession()` function in `web/backend/api/session.go` was passing `meta.Skip` to `readSessionMessages()`, which skipped messages that had been compressed for context management. This is correct for Agent internal use (where compressed messages are replaced by summaries), but incorrect for Web UI history display.

## Changes

- Modified `readJSONLSession()` to always read all messages (skip = 0) for Web UI display
- Added comment explaining the rationale

## Related Issue

Fixes #2796

## Testing

- Users can now see their complete conversation history in the Web UI
- Agent internal context compression behavior is unchanged

### DIFF
--- a/web/backend/api/session.go
+++ b/web/backend/api/session.go
@@ -15,6 +15,7 @@ import (
 	"github.com/sipeed/picoclaw/pkg/config"
 	"github.com/sipeed/picoclaw/pkg/memory"
 	"github.com/sipeed/picoclaw/pkg/providers"
+	"github.com/sipeed/picoclaw/pkg/providers/messageutil"
 	"github.com/sipeed/picoclaw/pkg/session"
 	"github.com/sipeed/picoclaw/pkg/utils"
 )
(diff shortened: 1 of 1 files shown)


afjcjsbx:
Thanks for working on a fix for #2796! Restoring the full history for the session detail view is definitely needed, but changing the shared reader logic here introduces a severe performance regression for the session list endpoint.

We need to decouple these two paths before this can be merged. Here is the breakdown:

### Performance & Memory Regression on Session List

The shared reader in `web/backend/api/session.go` (line 182) is used by both the **session detail** and **session list** endpoints.

* By changing this to `skip = 0`, the list endpoint loop (line 837) will now fully unmarshal *every single raw JSONL transcript for every session* before pagination is even applied further down (line 867).
* This effectively changes the history dropdown behavior from "read active tail" to "read full archived history of all sessions." On workspaces with numerous compressed chats, this will cause massive latency and memory spikes.
* Since issue #2796 specifically targets restoring full history in the session detail view, the safer and cleaner approach is to create a dedicated full-history path strictly for `GET /api/sessions/{id}`, while preserving the original `meta.Skip` behavior for the list path.

### Testing & Validation

* We currently lack a targeted test case where `meta.Skip > 0` (which is exactly the path broken by this issue). Please add a test covering this scenario to ensure the list pagination behavior remains protected and we don't regress on it in the future.

PR #2768, opened 2026-05-04
fix(agent): retry transient LLM HTTP errors

afjcjsbx:
## 📝 Description

Fixes LLM request retry handling for transient provider HTTP errors. Previously, OpenRouter/OpenAI-compatible `Status: 500` responses could fail the agent turn immediately when no model fallback candidate was available. This PR reuses the existing provider error classifier in the main LLM retry loop so transient HTTP/server errors are retried according to the configured retry policy.

## 🗣️ Type of Change
- [x] 🐞 Bug fix (non-breaking change which fixes an issue)
- [ ] ✨ New feature (non-breaking change which adds functionality)
- [ ] 📖 Documentation update
- [ ] ⚡ Code refactoring (no functional changes, no api changes)

## 🤖 AI Code Generation
- [ ] 🤖 Fully AI-generated (100% AI, 0% Human)
- [x] 🛠️ Mostly AI-generated (AI draft, Human verified/modified)
- [ ] 👨‍💻 Mostly Human-written (Human lead, AI assisted or none)

## 🔗 Related Issue

Fixes #629

## 📚 Technical Context (Skip for Docs)
- **Reference URL:** https://github.com/sipeed/picoclaw/issues/629
- **Reasoning:** The provider error classifier already maps transient HTTP statuses such as `500`, `502`, and `503` to retryable failover reasons. The main LLM retry loop now uses that classifier too, so single-provider runs can retry transient provider failures instead of dropping the task.

## 🧪 Test Environment
- **Hardware:** 
- **OS:**
- **Model/Provider:** 
- **Channels:** 

## 📸 Evidence (Optional)
<details>
<summary>Click to view Logs/Screenshots</summary>

</details>

## ☑️ Checklist
- [x] My code/docs follow the style of this project.
- [x] I have performed a self-review of my own changes.
- [ ] I have updated the documentation accordingly.

PR #2991, opened 2026-06-02
fix(agent): retry transient LLM HTTP errors using provider error classifier

chengzhichao-xydt:
## Description

Fixes LLM request retry handling for transient provider HTTP errors. Previously, OpenRouter/OpenAI-compatible Status: 500 responses could fail the agent turn immediately when no model fallback candidate was available.

This PR replaces the separate timeout/network retry branches with a unified 	ransientLLMRetryReason() helper that:

1. Uses providers.ClassifyError() to detect server_error (HTTP >=500), timeout, network, and rate_limit errors
2. Falls back to the existing string-based detection for errors not classified by the provider

A regression test (TestPipeline_CallLLM_HTTP5xxRetry) verifies that HTTP 500 errors are retried and recover successfully.

This is a clean rebase of the approach originally proposed in #2768 by @afjcjsbx.

## Type of Change

- [x] Bug fix (non-breaking change)

## Test Environment

- **OS:** Windows
- **Verification:** go build ./pkg/agent/... and go test ./pkg/agent/ -run TestPipeline_CallLLM_HTTP5xxRetry pass (2 files, +85/-54)

### DIFF
--- a/pkg/agent/pipeline_llm.go
+++ b/pkg/agent/pipeline_llm.go
@@ -280,22 +280,8 @@ func (p *Pipeline) CallLLM(
 		}
 
 		errMsg := strings.ToLower(err.Error())
-		isTimeoutError := errors.Is(err, context.DeadlineExceeded) ||
-			strings.Contains(errMsg, "deadline exceeded") ||
-			strings.Contains(errMsg, "client.timeout") ||
-			strings.Contains(errMsg, "timed out") ||
-			strings.Contains(errMsg, "timeout exceeded")
-
-		isNetworkError := !isTimeoutError && (strings.Contains(errMsg, "connection reset") ||
-			strings.Contains(errMsg, "connection refused") ||
-			strings.Contains(errMsg, "broken pipe") ||
-			strings.Contains(errMsg, "no such host") ||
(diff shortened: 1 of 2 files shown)


afjcjsbx:
LGTM! The overall implementation looks solid. I have spotted two minor edge cases, but they don't need to block this merge, we can easily track and address them in a follow-up PR.

### 1. 5xx Error Coverage Gap

The new helper function relies on the error classifier, but the internal status classification currently only treats a specific subset of 5xx codes as transient (such as 500, 502, 503, 521-524, 529).

As a result, an error like an HTTP 504 (Gateway Timeout) will still fail immediately without a retry, even though the PR description mentions handling all HTTP >= 500. The newly added test only validates HTTP 500, which is why this gap is currently hidden.

### 2. Context Limit Regression Edge Case

Calling the error classification before handling the context limit introduces a small regression. Because status extraction takes precedence over context overflow patterns, an error message containing text like *"maximum context length is 500 tokens"* might be misclassified as a transient HTTP 500 error instead of a context overflow.

If this happens, the loop will execute a backoff and retry the exact same request, instead of compressing the history and recovering as expected.

afjcjsbx (approved):
LGTM!

PR #2992, opened 2026-06-02
fix(session): skip main-session alias during history promotion

### LINKED ISSUE #2972: [BUG] After upgrade to v0.2.9, Web UI message chaos, every new session always attached some old message history.
##  Environment & Tools
- **Go Version:** 1.25.10
- **Channels:** Web

## 📸 Steps to Reproduce
1. Open Web UI and start new chat session.
2. Send something.
3. After responsed, refresh web page, and open session you just create before.

## ❌ Actual Behavior
The title is some old title, and there are a lot of old messages (look like from cron) above the message you just sent.

## ✅ Expected Behavior
Title is new session title like you sent, and you only seen the message you just send in this chat session.

## 💬 Additional Context

https://github.com/user-attachments/assets/ed286bc0-34b1-43f4-9e50-136d00f1aa5c

chengzhichao-xydt:
## Description

Fixes #2972: after upgrading to v0.2.9, new Web UI sessions would get old messages attached because PromoteAliasHistory copied the migrated main session content into every new session.

The root cause is that agent:main:main is always the first alias. When PromoteAliasHistory finds this alias has content (from migration) and the new session is empty, it promotes the old messages into the new session.

This commit adds isMainSessionAlias() to detect and skip the main session alias during promotion.

## Type of Change

- [x] Bug fix

## Related Issue

Fixes #2972

### DIFF
--- a/pkg/memory/jsonl.go
+++ b/pkg/memory/jsonl.go
@@ -4,6 +4,8 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"hash/fnv"
@@ -227,6 +229,11 @@ func (s *JSONLStore) UpsertSessionMeta(
 
 // PromoteAliasHistory atomically promotes the first non-empty alias session
 // into the canonical session when the canonical session is still empty.
+//
+// Main-session aliases (e.g. "agent:main:main" or its opaque form) are
+// skipped during promotion.  The main session is a shared global fallback
+// and promoting its history into individual sessions would attach stale
+// messages to every new Web UI session (issue #2972).
 func (s *JSONLStore) PromoteAliasHistory(
 	_ context.Context,
 	sessionKey string,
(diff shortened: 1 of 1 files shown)


afjcjsbx:
`pkg/memory/jsonl.go`: It classifies any legacy alias that starts with `agent:` and ends with `:main` as a main-session alias. However, this also catches perfectly valid aliases like `agent:<agent>:direct:main` or `agent:<agent>:slack:channel:main`, which can be generated by `BuildLegacyDirectAliases` and `BuildLegacyPeerAlias` (`pkg/session/key.go`).

In those cases, `PromoteAliasHistory` would stop promoting the legacy history, and after `ResolveSessionKey`, that history would effectively remain hidden behind the canonical session. The check here must recognize only the exact format `agent:<agent>:main`, rather than "any alias that ends with `:main`.

chengzhichao-xydt:
Good catch! Fixed: now uses strings.SplitN to only match exactly 3 colon-separated parts (agent:X:main), not longer forms like agent:X:direct:main.

afjcjsbx:
**Heads-up**: please rebase on `main` — the previous Go version (`1.25.10`) was affected by **GO-2026-5039** (`net/textproto`: arbitrary inputs in errors without escaping), which was likely causing the security check to fail.

This has now been fixed in `main` by bumping Go to `1.25.11` (#2997). A rebase should clear the security check. 🦞

afjcjsbx (approved):
LGTM!

PR #2996, opened 2026-06-03
fix(tools): handle json.Marshal errors in exec tool responses

chengzhichao-xydt:
Replace 7 instances of ignored json.Marshal errors in pkg/tools/shell.go with proper error handling. Previously, if marshaling an ExecResponse failed, a nil byte slice would be silently converted to an empty string in the LLM response. Now each site returns ErrorResult with the marshal error message.

This is a defensive improvement: while ExecResponse fields are simple (string, int) and unlikely to fail today, future additions of complex types could silently break.

### DIFF
--- a/pkg/tools/shell.go
+++ b/pkg/tools/shell.go
@@ -671,7 +671,10 @@ func (t *ExecTool) runBackground(ctx context.Context, command, cwd string, ptyEn
 		SessionID: sessionID,
 		Status:    "running",
 	}
-	data, _ := json.Marshal(resp)
+	data, err := json.Marshal(resp)
+	if err != nil {
+		return ErrorResult(err.Error())
+	}
 	return &ToolResult{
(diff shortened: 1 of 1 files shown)


afjcjsbx:
**Heads-up**: please rebase on `main` — the previous Go version (`1.25.10`) was affected by **GO-2026-5039** (`net/textproto`: arbitrary inputs in errors without escaping), which was likely causing the security check to fail.

This has now been fixed in `main` by bumping Go to `1.25.11` (#2997). A rebase should clear the security check. 🦞

afjcjsbx:
Hi @chengzhichao-xydt, thanks for the PR, could you fix lint please? 🙏

afjcjsbx (dismissed):
LGTM!

afjcjsbx (approved):

PR #3000, opened 2026-06-04
fix(pid): verify process identity in singleton PID check

### LINKED ISSUE #2720: [BUG] Singleton PID check doesn't verify process identity — stale PID causes crash loop

chengzhichao-xydt:
## Description

The singleton PID check in isProcessRunning() only verified whether a process with the stored PID existed (via Signal(0) on Unix / OpenProcess on Windows), without confirming the process was actually a picoclaw gateway.

When a gateway was terminated with SIGKILL (kill -9) or crashed without cleaning its PID file, the PID could be reused by an unrelated process such as systemd-resolved. The next gateway start would then see a live process at that PID and refuse to start with:
`

gateway is already running (PID: 151, version: 0.2.7)
`

Under systemd this results in an infinite crash-restart loop.

## Fix

Add isPicoclawProcess() identity verification to both platform implementations:

- **Unix**: Reads /proc/<pid>/comm and checks for picoclaw in the process name
- **Windows**: Uses QueryFullProcessImageNameW to get the executable path and checks for picoclaw`n
When a PID is alive but the process is NOT picoclaw, the PID file is treated as stale and normal startup proceeds. If identity verification is unavailable (e.g., /proc unreadable, API call fails), the code falls back to trusting the liveness check as before.

## Type of Change

- [x] Bug fix

## AI Code Generation

## Related Issue

Fixes #2720

## Test Environment

- OS: Windows (local verification)
- Verification: go build ./pkg/pid/... and go vet ./pkg/pid/... pass. Logic review: the new isPicoclawProcess() returns 	rue (pass-through) when identity data is unavailable, ensuring no regression on systems without /proc or with restricted process access.

### DIFF
--- a/pkg/pid/pidfile.go
+++ b/pkg/pid/pidfile.go
@@ -64,7 +64,14 @@ func WritePidFile(homePath, host string, port int) (*PidFileData, error) {
 			// pass the isProcessRunning check, blocking new gateway starts.
 			// Treat recorded PID 1 as always stale.
 			if data.PID != 1 && isProcessRunning(data.PID) {
-				return nil, fmt.Errorf("gateway is already running (PID: %d, version: %s)", data.PID, data.Version)
+				// Verify the process is actually a picoclaw instance.
+				// If the PID was reused by an unrelated process
+				// (e.g. systemd-resolved after a kill -9), treat
+				// the PID file as stale and proceed with startup.
+				if isPicoclawProcess(data.PID) {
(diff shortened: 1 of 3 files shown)


afjcjsbx (approved):
LGTM

PR #3001, opened 2026-06-04
fix(tools): allow scheme-less URLs in workspace guard

### LINKED ISSUE #1042: [BUG]exec工具的guardCommand方法问题

chengzhichao-xydt:
## Description

When 
estrict_to_workspace is enabled, commands like curl wttr.in/Beijing are incorrectly blocked by the workspace guard. The bsolutePathPattern regex matches /Beijing as an absolute path. Since wttr.in is not a recognized web scheme (http:, https:, tp:, etc.), the path goes through workspace sandbox validation instead of being treated as a URL path.

This breaks many skills that rely on curl with bare (scheme-less) domain names, which curl natively supports.

## Fix

Add detection for domain-like tokens immediately preceding /path matches:

1. **looksLikeDomain(s)** ??checks that the token contains a dot, starts with alphanumeric, and does NOT end with a common file extension (.py, .go, .exe, etc.)
2. **localPathExists(cwd, token)** ??verifies the token does NOT exist as a local filesystem entry relative to the command's working directory

When both conditions are met, the /path match is treated as part of a scheme-less URL and skipped during workspace sandbox validation.

### Security: symlink bypass prevention

The localPathExists guard directly addresses the security concern raised in the review of PR #2965: if oo.bar exists as a local symlink or directory, the path still undergoes full workspace validation, preventing symlink-to-external-directory escapes.

`go
if looksLikeDomain(token); !localPathExists(cwd, token) {
    continue  // treat as URL, skip workspace validation
}
`

### Behavior table

| Command | Token before / | Domain-like? | Exists locally? | Result |
|---------|-------------------|--------------|-----------------|--------|
| curl wttr.in/Beijing | wttr.in | Yes | No | Allowed |
| cat foo.bar/secret (symlink) | oo.bar | Yes | **Yes** | Blocked (validated) |
| 	ar -xC/etc/cron.d | -xC | No | ??| Blocked |
| cat /etc/passwd | (empty) | No | ??| Blocked |
| python3 /tmp/script.py | python3 | No | ??| Blocked |
| scp host:/etc/passwd | (colon boundary) | ??| ??| Blocked |

## Type of Change

- [x] Bug fix

## AI Code Generation

## Related Issue

Fixes #1042

Related: PR #2965 (earlier attempt with security concerns)

## Test Environment

- OS: Windows
- Verification: go build ./pkg/tools/..., go vet ./pkg/tools/... pass

### DIFF
--- a/pkg/tools/shell.go
+++ b/pkg/tools/shell.go
@@ -1156,6 +1156,27 @@ func (t *ExecTool) guardCommand(command, cwd string) string {
 				}
 			}
 
+			// Skip scheme-less URL paths like "wttr.in/Beijing".
+			// When a /path is immediately preceded by a token that looks
+			// like a domain name and that token does NOT exist as a local
+			// filesystem entry, treat the path as part of a URL and skip
+			// workspace sandbox validation.
+			//
+			// The local-path-exists guard prevents symlink bypass: if
+			// "foo.bar" exists as a local symlink or directory, the path
+			// still undergoes full workspace validation (see #2965).
+			if loc[0] > 0 && raw[0] == '/' {
+				// Find the token immediately before the "/".
+				j := loc[0] - 1
+				for j >= 0 && !isShellTokenBoundary(cmd[j]) {
+					j--
+				}
+				token := cmd[j+1 : loc[0]]
+				if looksLikeDomain(token) && !localPathExists(cwd, token) {
+					continue
+				}
+			}
+
 			p, err := filepath.Abs(raw)
 			if err != nil {
 				continue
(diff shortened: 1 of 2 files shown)


afjcjsbx:
Hi @chengzhichao-xydt , thanks for the PR, looks good, would you also add some unit tests? 🙏

chengzhichao-xydt:
Added unit tests for scheme-less URL detection in TestShellTool_SchemelessURLDetection. The new test covers all 7 recognized web schemes individually, plus multi-URL commands.

afjcjsbx (approved):
LGTM

PR #3010, opened 2026-06-05
fix(channels): add ok checks for type assertions in toChannelHashes

chengzhichao-xydt:
## Problem

Two unchecked type assertions in `toChannelHashes` could panic when channel config values had unexpected types from `json.Unmarshal`:

1. `value["enabled"].(bool)` panics if the `enabled` key is missing, null, or not a boolean
2. `vv.(map[string]string)` panics when `json.Unmarshal` produces `map[string]any` (its default map type) instead of `map[string]string` - this can happen when `teams_webhook` is enabled with webhooks configured

## Fix

Add safe `ok`-checked type assertions:
- Line 20: `if enabled, ok := value["enabled"].(bool); !ok; !enabled`
- Line 97: Handle both `map[string]string` and `map[string]any` (JSON default) types for webhooks

## Changes

- `pkg/channels/manager_channel.go`: +10/-2

## Verification

- `go build ./pkg/channels/...` passes

### DIFF
--- a/pkg/channels/manager_channel.go
+++ b/pkg/channels/manager_channel.go
@@ -17,7 +17,7 @@ func toChannelHashes(cfg *config.Config) map[string]string {
 	_ = json.Unmarshal(marshal, &channelConfig)
 
 	for key, value := range channelConfig {
-		if !value["enabled"].(bool) {
+		if enabled, ok := value["enabled"].(bool); !ok || !enabled {
 			continue
 		}
 		hiddenValues(key, value, ch.Get(key))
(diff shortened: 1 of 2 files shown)


afjcjsbx:
Great fix, thanks @chengzhichao-xydt! 

One suggestion: could you add tests covering the edge cases that were previously panicking? Specifically:

  1. Missing enabled key when value["enabled"] doesn't exist in the settings
  2. enabled is not a bool e.g., "enabled": "yes" (string instead of bool)
  3. teams_webhook with configured webhooks, the most likely real-world scenario that triggered the panic

Adding these would prevent regressions and verify the fix works correctly under all conditions.

Other than that, LGTM.

afjcjsbx (approved):

PR #3011, opened 2026-06-05
fix(agent): add ok check for LoadAndDelete type assertion

chengzhichao-xydt:
## Problem

`UnsubscribeEvents` uses `sync.Map.LoadAndDelete` which returns `any`, then directly asserts the value as `legacyEventSubscription` without checking `ok`. If an unexpected type were ever stored under an event ID, this would panic.

## Fix

Add a safe `ok`-checked type assertion:
```go
sub, ok := value.(legacyEventSubscription)
if !ok {
    return
}
```

## Changes

- `pkg/agent/legacy_events.go`: +4/-1

## Verification

### DIFF
--- a/pkg/agent/legacy_events.go
+++ b/pkg/agent/legacy_events.go
@@ -2,11 +2,13 @@ package agent
 
 import (
 	"context"
+	"fmt"
 	"sync"
 	"sync/atomic"
 
 	"github.com/sipeed/picoclaw/pkg/bus"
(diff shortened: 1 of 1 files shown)


afjcjsbx:
LGTM. Clean, surgical defensive fix with zero runtime overhead on the happy path.

Optional suggestion:
I'd consider logging a warning if the type assertion fails. Right now it swallows the failure completely silently; making a mismatched type observable would save someone a debugging headache if things drift in the future.

afjcjsbx (approved):

PR #3016, opened 2026-06-06
fix: cancel old dispatchTask on reload and guard nil ts.agent

chengzhichao-xydt:
## Problem

Two defensive fixes:

1. **Goroutine leak in `Manager.Reload()`**: When channels are reloaded (config change), `m.dispatchTask` is overwritten without calling `cancel()` on the previous context. The old dispatch goroutines (`dispatchOutbound` + `dispatchOutboundMedia`) keep running indefinitely. `StopAll()` already handles this correctly at line 1325.

2. **Potential nil dereference in `configuredStreamingEligible`**: `ts.agent.ID` is accessed in debug logs at lines 268/278 without a nil guard on `ts.agent`, even though `ts` itself is already nil-checked.

## Fix

1. Cancel old `dispatchTask` before overwriting, matching the pattern in `StopAll()`.
2. Add `ts.agent == nil` guard to the existing checks.

## Changes

- `pkg/channels/manager.go`: +4
- `pkg/agent/pipeline_streaming.go`: +1/-1

## Verification

- `go build ./pkg/channels/...` passes

### DIFF
--- a/pkg/agent/pipeline_streaming.go
+++ b/pkg/agent/pipeline_streaming.go
@@ -264,13 +264,16 @@ func (p *Pipeline) configuredStreamingEligible(ts *turnState, exec *turnExecutio
 		return false
 	}
 	if ts.agent == nil || strings.TrimSpace(ts.channel) == "" || strings.TrimSpace(ts.chatID) == "" {
-		logger.DebugCF("agent", "configured streaming not used", map[string]any{
-			"agent_id": ts.agent.ID,
+		logFields := map[string]any{
 			"channel":  ts.channel,
 			"chat_id":  ts.chatID,
 			"model":    exec.activeModel,
 			"reason":   "missing_channel_context",
-		})
+		}
+		if ts.agent != nil {
+			logFields["agent_id"] = ts.agent.ID
+		}
+		logger.DebugCF("agent", "configured streaming not used", logFields)
 		return false
 	}
 	if !ts.opts.SendResponse && !ts.opts.AllowInterimPicoPublish {
--- a/pkg/agent/pipeline_streaming.go
+++ b/pkg/agent/pipeline_streaming.go
@@ -265,10 +265,10 @@ func (p *Pipeline) configuredStreamingEligible(ts *turnState, exec *turnExecutio
 	}
 	if ts.agent == nil || strings.TrimSpace(ts.channel) == "" || strings.TrimSpace(ts.chatID) == "" {
 		logFields := map[string]any{
-			"channel":  ts.channel,
-			"chat_id":  ts.chatID,
-			"model":    exec.activeModel,
-			"reason":   "missing_channel_context",
+			"channel": ts.channel,
+			"chat_id": ts.chatID,
+			"model":   exec.activeModel,
+			"reason":  "missing_channel_context",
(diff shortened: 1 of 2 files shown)


afjcjsbx:
Hi @chengzhichao-xydt, 

#### `pkg/channels/manager.go` & `pkg/channels/manager_channel.go`

**Issue: Critical bug in `Reload()` lifecycle causing goroutine starvation and service teardown**

During `Reload()`, the previous `dispatchTask` context is cancelled, but the function fails to restart the background goroutines governed by that specific context (`dispatchOutbound`, `dispatchOutboundMedia`, `runTTLJanitor`, as well as the existing workers initialized around line 1260).

Because `compareChannels` (in `pkg/channels/manager_channel.go`) filters out and skips unchanged channels, a reload triggered with no real configuration changes—or containing unchanged channels—effectively shuts down the active workers and exits without spinning up new ones. From this point forward, outbound events will stop being consumed entirely or will deadlock the system once the queues fill up.

Additionally, if `toChannelConfig` or `initChannels` fails *after* the context cancellation has occurred, the botched reload still tears down the currently active service, leaving the system in a broken state with no fallback.

#### `pkg/agent/pipeline_streaming.go`

**Issue: Broken nil-guard resulting in a guaranteed panic**

The promised nil-guard around line 266 does not actually prevent a crash. Even though `ts.agent == nil` was added to the conditional check, the logging statement immediately following it (line 268) still attempts to read `ts.agent.ID`.

If `ts.agent` is indeed `nil`, a nil pointer dereference panic will trigger right there, completely bypassing the intended `return false`. You need to ensure `ts.agent.ID` is not evaluated when `ts.agent` is nil.

chengzhichao-xydt:
Thanks for the thorough review @afjcjsbx! I've addressed both issues:

**pipeline_streaming.go:** Fixed the nil-guard - 	s.agent.ID is no longer unconditionally accessed in the log map when 	s.agent == nil.

**manager.go Reload():** Moved the dispatchTask cancel/reset to after initChannels succeeds, and skip it entirely when len(added) == 0. This avoids cancelling the old service without replacement workers on no-change reloads or intermediate failures.

afjcjsbx:
hI @chengzhichao-xydt, 

In `pkg/channels/manager.go`, `Reload()` cancels `m.dispatchTask`. However, this context currently governs not only the new workers, but also all existing workers, `dispatchOutbound`, `dispatchOutboundMedia`, and `runTTLJanitor` initialized during `StartAll()`.

Right after this cancellation, `Reload()` only recreates workers for channels present in the `added` list. Unchanged channels are completely skipped because `compareChannels()` (in `pkg/channels/manager_channel.go`) filters them out, and the core background dispatchers/janitor are never restarted.

As a result, upon the very first reload containing at least one added or modified channel, outbound events will stop being consumed entirely, and even unchanged channels will lose their active workers. In practice, this fix replaces the resource leak with a total breakdown of the active service.

chengzhichao-xydt:
Thanks for the thorough review. You are right - the dispatchTask context is shared across all workers and core goroutines, and canceling it without restarting everything properly is a fundamental flaw that requires a deeper refactor beyond this fix. Closing.

PR #3017, opened 2026-06-06
fix(agent): close base64 encoder on io.Copy error path

chengzhichao-xydt:
## Problem

In `encodeMediaFile`, when `io.Copy` to the base64 encoder fails, `encoder.Close()` is skipped. The `base64.Encoder` buffers internal state (padding) that is never flushed, leaving an incomplete output in the buffer.

## Fix

Always call `encoder.Close()` regardless of `io.Copy` result. Check both errors separately and return on either failure.

## Changes

- `pkg/agent/agent_media.go`: +11/-3

## Verification

### DIFF
--- a/pkg/agent/agent_media.go
+++ b/pkg/agent/agent_media.go
@@ -161,14 +161,22 @@ func encodeImageToDataURL(localPath, mime string, info os.FileInfo, maxSize int)
 	buf.WriteString(prefix)
 
 	encoder := base64.NewEncoder(base64.StdEncoding, &buf)
-	if _, err := io.Copy(encoder, f); err != nil {
+	_, copyErr := io.Copy(encoder, f)
+	closeErr := encoder.Close()
+	if copyErr != nil {
 		logger.WarnCF("agent", "Failed to encode media file", map[string]any{
 			"path":  localPath,
-			"error": err.Error(),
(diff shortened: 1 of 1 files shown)


afjcjsbx:
Hi @chengzhichao-xydt , the PR doesn’t appear to fix an observable bug in the current implementation. On the only path it changes, `encodeImageToDataURL` already returns `""` immediately when `io.Copy` fails, so any partially buffered base64 output is discarded and never reaches callers. In addition, `encoder` is writing into a `bytes.Buffer`, so `encoder.Close()` cannot produce a new write error in practice. That makes the new `closeErr` branch effectively dead code here, and the change ends up being behaviorally a no-op rather than a real fix.

chengzhichao-xydt:
Thanks for the review @afjcjsbx. You're right - on the only path it changes, encodeImageToDataURL already returns "" immediately on io.Copy failure, and encoder.Close() against a ytes.Buffer can't produce an error in practice. The change is dead code; closing.

chengzhichao-xydt:
Thanks for the review @afjcjsbx. You are right - on the only path it changes, encodeImageToDataURL already returns empty string immediately on io.Copy failure, and encoder.Close() against a bytes.Buffer cannot produce an error in practice. The change is dead code; closing.

PR #3018, opened 2026-06-06
fix: add ok checks for type assertions and handle os.Getwd error

chengzhichao-xydt:
## Problem

Three defensive fixes for unchecked type assertions and error returns:

1. **LINE channel** (`line.go:478`): `sync.Map.LoadAndDelete` result is directly type-asserted as `replyTokenEntry` without `ok` check.

2. **Evolution store** (`store.go:558`): `sync.Map.LoadOrStore` result directly type-asserted as `*sync.Mutex` without `ok` check.

3. **Context builder** (`context.go:105`): `os.Getwd()` error silently discarded with `_`. On failure, `wd` is `""` and `builtinSkillsDir` becomes `"/skills"`, causing confusing downstream errors.

## Fix

1. Add `ok` check for `replyTokenEntry` assertion, return error on mismatch.
2. Add `ok` check for `*sync.Mutex` assertion, fallback to new mutex (should never happen).
3. Check `os.Getwd()` error, fallback to global config dir.

## Changes

- `pkg/channels/line/line.go`: +3
- `pkg/evolution/store.go`: +4/-1
- `pkg/agent/context.go`: +6/-2

## Verification

- `go build ./pkg/channels/line/... ./pkg/evolution/... ./pkg/agent/...` passes

### DIFF
--- a/pkg/identity/identity.go
+++ b/pkg/identity/identity.go
@@ -51,19 +51,35 @@ func MatchAllowed(sender bus.SenderInfo, allowed string) bool {
 		if !isNumeric(platform) {
 			candidate := BuildCanonicalID(platform, id)
 			if candidate != "" && sender.CanonicalID != "" {
-				return strings.EqualFold(sender.CanonicalID, candidate)
+				if strings.EqualFold(sender.CanonicalID, candidate) {
+					return true
+				}
 			}
 			// If sender has no canonical ID, try matching platform + platformID
-			return strings.EqualFold(platform, sender.Platform) &&
-				sender.PlatformID == id
+			if strings.EqualFold(platform, sender.Platform) &&
+				sender.PlatformID == id {
+				return true
+			}
 		}
+		// Canonical match failed but we still parsed a colon separator.
+		// The input could be a Matrix-style ID ("@user:domain") where the
+		// colon is part of the ID, not a canonical separator. Fall through
+		// to let the remaining match strategies (PlatformID, Username) try.
+		// For Matrix, sender.PlatformID and sender.Username are both the
+		// full MXID ("@alice:example.com"), so we preserve the leading "@"
+		// in the legacy path.
 	}
 
 	// Keep track of explicit username format
 	isAtUsername := strings.HasPrefix(allowed, "@")
 
-	// Strip leading "@" for username matching
-	trimmed := strings.TrimPrefix(allowed, "@")
+	// Strip leading "@" for username matching, but only when the
+	// remaining string does not contain a colon (otherwise the "@"
+	// is part of a Matrix-style user ID like "@alice:example.com").
+	trimmed := allowed
+	if isAtUsername && !strings.Contains(allowed, ":") {
+		trimmed = strings.TrimPrefix(allowed, "@")
+	}
 
 	// Split compound "id|username" format
 	allowedID := trimmed
--- a/pkg/identity/identity_test.go
+++ b/pkg/identity/identity_test.go
@@ -223,6 +223,40 @@ func TestMatchAllowed(t *testing.T) {
 			allowed: "  123456  ",
 			want:    true,
 		},
+		// Matrix user IDs with colon (regression test for #3044)
+		{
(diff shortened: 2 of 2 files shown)


afjcjsbx:
Hi @chengzhichao-xydt, 

The new fallback in `pkg/evolution/store.go` breaks the per-path lock invariant. If `storeFileLocks` ever contains a bad value, the PR now creates a fresh local `sync.Mutex` and continues. Two callers hitting the same `path` can then each lock different mutexes and write the same store file concurrently, which risks corrupting the JSONL/profile files. For a defensive branch, we need to repair/replace the map entry atomically or fail fast; continuing with a detached mutex is not safe.

`pkg/channels/line/line.go` still has the same unchecked `sync.Map` pattern one line above the new fix: `quoteTokens.LoadAndDelete` is followed by `qt.(string)` without an `ok` check. So in the same “bad map contents” scenario this PR is trying to harden against, `Send` can still panic before it reaches the new `replyTokens` guard.

chengzhichao-xydt:
Thanks for the review @afjcjsbx! Both issues fixed:

**store.go:** The fallback path now atomically calls storeFileLocks.Store(path, mu) so all callers share the same mutex, preserving the per-path lock invariant.

**line.go:** Added ok check for the quoteTokens.LoadAndDelete type assertion on the line right above.

afjcjsbx:
Hi @chengzhichao-xydt, thank you for the fix, there are two leftovers:

#### `pkg/evolution/store.go`

**Issue: Defensive branch does not atomically repair the "one mutex per path" invariant**

In `lockStoreFile()`, the defensive branch fails to fix the "one mutex per path" invariant atomically. If `storeFileLocks[path]` contains a corrupted or invalid value and two goroutines concurrently enter `lockStoreFile(path)`, both can retrieve the same wrong value from `LoadOrStore`.

Consequently, both goroutines will instantiate two separate `*sync.Mutex` objects, execute `Store(path, mu)`, and end up locking completely distinct mutexes. This allows concurrent writes to the exact same store file, failing to eliminate the risk of JSONL/profile corruption. This needs to be handled via a loop with a CAS (CompareAndSwap + retry) mechanism or a fail-fast strategy.

#### `pkg/channels/line/line.go`

**Issue: Hardening logic introduces avoidable message loss on invalid cache types**

If `replyTokens` contains a value of the wrong type, the new logic prematurely returns `ErrSendFailed` before even attempting the `PushMessageWithHttpInfo` fallback.

However, the push operation does not rely on the reply token. In a "corrupt cache but valid ChatID" scenario, the message is still entirely deliverable. This hardening change introduces an avoidable message drop; it would be much better to log the type mismatch issue as a warning and proceed directly to the push fallback.

afjcjsbx:
#### `pkg/agent/context.go`

**Issue: Fallback logic introduces a regression in skill lookup and alters tier semantics**

The fallback to `getGlobalConfigDir()` alters the semantic behavior of the builtin tier instead of simply handling the error. If `Getwd()` fails, `builtinSkillsDir` becomes identical to `globalSkillsDir`. Consequently, the builtin skills effectively disappear from the lookup pool, and the global path ends up being scanned twice.

Furthermore, the core premise of this PR is flawed: `filepath.Join("", "skills")` evaluates to `skills`, not `/skills`. Therefore, this proposed fix introduces a regression into the skill lookup mechanism precisely within the error handling path.

#### `pkg/evolution/store.go`

**Issue: Lock recovery is non-atomic, allowing concurrent access to the critical section**

The lock recovery mechanism is not truly atomic. If two concurrent goroutines both detect a corrupted map entry, a race condition occurs: the first goroutine can execute `Store` with a new mutex A and lock it, while the second goroutine can immediately follow up, execute `Store` with a completely different mutex B, and lock it instead.

At that point, both callers will enter the critical section for the exact same file simultaneously. This behavior was verified with a standalone test script that yielded two distinct mutex addresses. Therefore, the inline comment stating *"Replace atomically and retry so all callers share one mutex"* is factually incorrect.

PR #3019, opened 2026-06-06
fix: type-switch capture, nil guard, check LastInsertId errors

chengzhichao-xydt:
## Problem

Three defensive fixes:

1. **WhatsApp channel** (`whatsapp_native.go:272`): `switch evt.(type)` type-switch followed by redundant unchecked `evt.(*events.Message)` assertion. Should use type-switch capture pattern `switch v := evt.(type)`.

2. **Config** (`config.go:196`): `FilterSensitiveData` has no nil receiver guard. Called on nil `*Config` the method panics.

3. **Seahorse store** (`store.go:59,196,285`): `result.LastInsertId()` error discarded with `_` in 3 locations. On unexpected failure, `id` is 0 which could cause subtle bugs (ID 0 indistinguishable from zero-value).

## Fix

1. Use `switch v := evt.(type)` to capture typed value.
2. Add `if c == nil { return content }` guard.
3. Check and return `LastInsertId()` errors with wrapped error messages.

## Changes

- `pkg/channels/whatsapp_native/whatsapp_native.go`: +2/-2
- `pkg/config/config.go`: +3
- `pkg/seahorse/store.go`: +12/-3

## Verification

- `go build ./pkg/channels/whatsapp_native/... ./pkg/config/... ./pkg/seahorse/...` passes

### DIFF
--- a/pkg/channels/whatsapp_native/whatsapp_native.go
+++ b/pkg/channels/whatsapp_native/whatsapp_native.go
@@ -269,9 +269,9 @@ func (c *WhatsAppNativeChannel) Stop(ctx context.Context) error {
 }
 
 func (c *WhatsAppNativeChannel) eventHandler(evt any) {
-	switch evt.(type) {
+	switch v := evt.(type) {
 	case *events.Message:
-		c.handleIncoming(evt.(*events.Message))
+		c.handleIncoming(v)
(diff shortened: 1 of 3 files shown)


afjcjsbx (approved):
LGTM

PR #3036, opened 2026-06-07
fix(config): use canonical Anthropic default model ID

### LINKED ISSUE #2941: Default config seeds claude-sonnet-4.6 with invalid model ID (uses dots, Anthropic API requires hyphens)
### Summary

A fresh launcher install generates a default `model_list` entry for Anthropic that the API rejects with HTTP 404 the first time it's used:

```
{"error":{"code":"not_found_error","message":"model: claude-sonnet-4.6 was not found. Did you mean claude-sonnet-4-6?","type":"invalid_request_error"}}
```

The default seeder writes the user-facing alias `claude-sonnet-4.6` (with dots) into the `Model` field — but `Model` is the canonical ID sent verbatim to Anthropic, which uses hyphens (`claude-sonnet-4-6`). `ModelName` and `Model` should be different values for this entry.

### Reproduce

1. Spin up a fresh launcher (e.g. `docker run sipeed/picoclaw:launcher` with an empty data volume).
2. In the UI, configure an Anthropic API key on the seeded `claude-sonnet-4.6` entry. **Do not** set `auth_method`.
3. Send any message → 404 with the message above.

The HTTP-API path (`provider: anthropic` + API key → routes through `openai_compat`) doesn't normalize the dotted form before forwarding it.

### Root cause

`pkg/config/defaults.go:88-93`:

```go
{
    ModelName: "claude-sonnet-4.6",
    Provider:  "anthropic",
    Model:     "claude-sonnet-4.6",   // ← should be "claude-sonnet-4-6"
    APIBase:   "https://api.anthropic.com/v1",
},
```

`Model` should hold the canonical Anthropic ID — `claude-sonnet-4-6`. Compare with the metadata in `pkg/providers/provider_metadata.go:51`, which lists the correct hyphenated form as a common model.

### Why some installs aren't seeing this

The OAuth-based Anthropic provider at `pkg/providers/anthropic/provider.go:217-219` *does* normalize the dotted form to hyphens before calling the SDK:

```go
// Normalize model ID: Anthropic API uses hyphens (claude-sonnet-4-6),
// but config may use dots (claude-sonnet-4.6).
apiModel := strings.ReplaceAll(model, ".", "-")
```

So installs that use `auth_method: "oauth"` / `"token"` never hit the bug. The HTTP-API path (`pkg/providers/openai_compat`) and the native Messages path (`pkg/providers/anthropic_messages`) both forward `model` verbatim.

That gives two reasonable fixes — applying either or both:

1. **One-line fix** — change `defaults.go:91` to `Model: "claude-sonnet-4-6"`. Smallest blast radius; matches Anthropic's canonical naming.
2. **Defense in depth** — apply the same normalization that `anthropic/provider.go:219` does in `openai_compat` and `anthropic_messages` when the protocol is Anthropic. Saves any user from this footgun if they paste a dotted ID elsewhere.

Happy to send a PR for (1); (2) is a slightly larger judgment call that I'd leave to maintainers.

### Environment

- picoclaw `sipeed/picoclaw:launcher` (rebuilt from `main` at `ab6d3946`, today)
- Anthropic API
- Reproduced on a freshly-created instance with an empty data volume

SutraHsing:
## Summary

- keep the seeded Anthropic ModelName as the user-facing claude-sonnet-4.6 alias
- change the actual default Model sent to Anthropic from claude-sonnet-4.6 to claude-sonnet-4-6
- add regression coverage that first-party Anthropic default models and Anthropic common models do not use dotted Claude API IDs

Fixes #2941.

## Reference

Anthropic documents Claude API model IDs as hyphenated formats such as claude-{name}-{major}-{minor} for the 4.6 generation and later, while dotted prefixes are platform-specific IDs such as Bedrock:
https://platform.claude.com/docs/en/about-claude/models/model-ids-and-versions

## Test plan

- go test ./pkg/config/...
- go test ./pkg/providers/...

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -907,6 +907,29 @@ func TestDefaultConfig_WorkspacePath(t *testing.T) {
 	}
 }
 
+// TestDefaultConfig_AnthropicModelsUseClaudeAPIIDs verifies that first-party
+// Anthropic defaults use Claude API model IDs, not dotted display names or
+// Bedrock-style provider prefixes. See:
+// https://platform.claude.com/docs/en/about-claude/models/model-ids-and-versions
+func TestDefaultConfig_AnthropicModelsUseClaudeAPIIDs(t *testing.T) {
+	cfg := DefaultConfig()
+
+	checked := 0
+	for _, model := range cfg.ModelList {
+		if model.Provider != "anthropic" {
+			continue
+		}
+		checked++
+		if strings.Contains(model.Model, ".") {
+			t.Fatalf("Anthropic default model %q uses dotted ID %q", model.ModelName, model.Model)
+		}
+	}
+
+	if checked == 0 {
+		t.Fatal("DefaultConfig() missing Anthropic models")
+	}
+}
+
 // TestDefaultConfig_MaxTokens verifies max tokens has default value
 func TestDefaultConfig_MaxTokens(t *testing.T) {
 	cfg := DefaultConfig()
--- a/pkg/config/defaults.go
+++ b/pkg/config/defaults.go
@@ -88,7 +88,7 @@ func DefaultConfig() *Config {
 			{
 				ModelName: "claude-sonnet-4.6",
 				Provider:  "anthropic",
-				Model:     "claude-sonnet-4.6",
+				Model:     "claude-sonnet-4-6",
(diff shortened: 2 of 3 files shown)


afjcjsbx (approved):
LGTM!

PR #3043, opened 2026-06-07
fix: check strconv.Atoi and json.Unmarshal errors

chengzhichao-xydt:
## Problem

Two places silently discard errors that should not be ignored:

1. **`pkg/seahorse/short_retrieval.go:25`** — `strconv.Atoi` result discarded with `_`. While a regex guarantees numeric input today, any future regex change could silently return a zero value instead of reporting the parse failure.

2. **`pkg/config/gateway.go:91`** — `json.Unmarshal` error silently discarded with `_`. If the gateway config file contains malformed JSON, the user gets default values with no indication that their config was ignored.

## Fix

1. Check `Atoi` error and return a clear error message.
2. Log a warning when gateway config JSON cannot be parsed, including the file path and parse error.

## Changes

- `pkg/seahorse/short_retrieval.go`: +3/-1
- `pkg/config/gateway.go`: +5/-1

## Verification

- `go build ./pkg/seahorse/... ./pkg/config/...` passes

### DIFF
--- a/pkg/config/gateway.go
+++ b/pkg/config/gateway.go
@@ -88,7 +88,12 @@ func ResolveGatewayLogLevel(path string) string {
 
 	data, err := os.ReadFile(path)
 	if err == nil {
-		_ = json.Unmarshal(data, &cfg)
+		if err := json.Unmarshal(data, &cfg); err != nil {
+			logger.WarnCF("config", "failed to parse gateway config, using defaults", map[string]any{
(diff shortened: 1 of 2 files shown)


afjcjsbx:
Hi @chengzhichao-xydt, thank you for the PR, could you fix conflicts pls? 🙏

afjcjsbx (approved):

PR #3045, opened 2026-06-07
fix(identity): allow_from fallthrough for Matrix user IDs with colon

### LINKED ISSUE #3044: Bug: allow_from fails for Matrix user IDs containing colon
## Summary

`allow_from` does not work when using standard Matrix user ID format (`@localpart:domain`). Messages are silently rejected even though the user ID in the config matches the sender exactly.

## Steps to Reproduce

1. Configure a Matrix channel with `allow_from` using the documented format:
   ```json
   "matrix": {
     "enabled": true,
     "type": "matrix",
     "allow_from": ["@alice:example.com"],
     "settings": {
       "homeserver": "https://matrix.org",
       "user_id": "@bot:example.com",
       "access_token": "..."
     }
   }
   ```
2. Send a message to the bot from `@alice:example.com`.
3. The message is silently rejected with log: `"Message rejected by allowlist"`.

## Expected Behavior

The message should be accepted since the sender matches the `allow_from` entry.

## Root Cause

`identity.MatchAllowed` (`pkg/identity/identity.go:41`) calls `ParseCanonicalID` on the allow-list entry before attempting other match strategies. `ParseCanonicalID` splits on the **first** colon:

```go
// ParseCanonicalID("@alice:example.com")
// -> platform = "@alice", id = "example.com", ok = true
```

Since `"@alice"` is not numeric, the function enters the canonical-match branch and builds a candidate:

```go
candidate := BuildCanonicalID("@alice", "example.com")
// -> "@alice:example.com"
```

This candidate is compared against `sender.CanonicalID` which is `"matrix:@alice:example.com"` — they don't match, so the function returns `false` **without ever trying** the PlatformID or Username match paths.

The core issue: Matrix user IDs always contain a colon in `@localpart:domain` format, which `ParseCanonicalID` misinterprets as the canonical `"platform:id"` separator.

## Workaround

Use the full canonical format in `allow_from`:

```json
"allow_from": ["matrix:@alice:example.com"]
```

This works because `ParseCanonicalID("matrix:@alice:example.com")` correctly splits into `platform="matrix"`, `id="@alice:example.com"`, which then matches `sender.CanonicalID`.

## Suggested Fix

`ParseCanonicalID` (or the caller in `MatchAllowed`) should only treat the colon as a canonical separator when the `platform` part is a recognized platform name (e.g., `matrix`, `telegram`, `discord`, etc.), not an arbitrary string like `@alice`. Alternatively, `MatchAllowed` should fall through to the remaining match strategies when the canonical match fails, rather than returning early.

## Affected Files

- `pkg/identity/identity.go` — `ParseCanonicalID`, `MatchAllowed`
- `pkg/channels/README.md:798` — documents the non-working format `"@user:example.com"`

## Environment

- PicoClaw built from latest `main`
- Matrix channel with Synapse homeserver

chengzhichao-xydt:
## Problem

`allow_from` silently rejects messages from Matrix users when using the standard Matrix user ID format (`@alice:example.com`). Fixes #3044.

**Root cause**: `ParseCanonicalID` splits on the first colon, so `@alice:example.com` is parsed as `platform="@alice"`, `id="example.com"`. Since `@alice` is not numeric, the code enters the canonical-match branch and builds `"@alice:example.com"` as the candidate — which doesn't match `sender.CanonicalID` (`"matrix:@alice:example.com"`). The function then returned `false` immediately, never trying the remaining PlatformID or Username match strategies.

## Fix

When the canonical match fails but a colon was present, fall through to the remaining match strategies (PlatformID, Username) instead of returning early. This allows Matrix-style IDs to still match via the legacy paths.

## Changes

- `pkg/identity/identity.go`: refactor early returns into fallthrough

## Verification

- `go build ./pkg/identity/...` passes
- `go test ./pkg/identity/...` passes (all existing tests green)

### DIFF
--- a/pkg/identity/identity.go
+++ b/pkg/identity/identity.go
@@ -48,22 +48,38 @@ func MatchAllowed(sender bus.SenderInfo, allowed string) bool {
 	if platform, id, ok := ParseCanonicalID(allowed); ok {
 		// Only treat as canonical if the platform portion looks like a known platform name
 		// (not a pure-numeric string, which could be a compound ID)
-		if !isNumeric(platform) {
+		if !isNumeric(platform) && !strings.HasPrefix(platform, "@") {
 			candidate := BuildCanonicalID(platform, id)
 			if candidate != "" && sender.CanonicalID != "" {
-				return strings.EqualFold(sender.CanonicalID, candidate)
+				if strings.EqualFold(sender.CanonicalID, candidate) {
+					return true
+				}
 			}
 			// If sender has no canonical ID, try matching platform + platformID
-			return strings.EqualFold(platform, sender.Platform) &&
-				sender.PlatformID == id
+			if strings.EqualFold(platform, sender.Platform) &&
+				sender.PlatformID == id {
+				return true
+			}
 		}
+		// Canonical match failed but we still parsed a colon separator.
+		// The input could be a Matrix-style ID ("@user:domain") where the
+		// colon is part of the ID, not a canonical separator. Fall through
+		// to let the remaining match strategies (PlatformID, Username) try.
+		// For Matrix, sender.PlatformID and sender.Username are both the
+		// full MXID ("@alice:example.com"), so we preserve the leading "@"
+		// in the legacy path.
 	}
 
 	// Keep track of explicit username format
 	isAtUsername := strings.HasPrefix(allowed, "@")
 
-	// Strip leading "@" for username matching
-	trimmed := strings.TrimPrefix(allowed, "@")
+	// Strip leading "@" for username matching, but only when the
+	// remaining string does not contain a colon (otherwise the "@"
+	// is part of a Matrix-style user ID like "@alice:example.com").
+	trimmed := allowed
+	if isAtUsername && !strings.Contains(allowed, ":") {
+		trimmed = strings.TrimPrefix(allowed, "@")
+	}
 
 	// Split compound "id|username" format
 	allowedID := trimmed
--- a/pkg/identity/identity.go
+++ b/pkg/identity/identity.go
@@ -60,14 +60,19 @@ func MatchAllowed(sender bus.SenderInfo, allowed string) bool {
 				sender.PlatformID == id {
 				return true
 			}
+			// This was a valid canonical "platform:id" entry that
+			// didn't match. Return false immediately — do NOT fall
+			// through to legacy matching. Otherwise an entry like
+			// "discord:98765432" could be satisfied by a sender with
+			// Platform="matrix" and PlatformID="discord:98765432".
+			return false
 		}
-		// Canonical match failed but we still parsed a colon separator.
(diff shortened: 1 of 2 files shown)


afjcjsbx:
Hi @chengzhichao-xydt, thanks for digging into this, I think the underlying bug is real, but I don't think this patch fully fixes it yet.

For Matrix, `sender.PlatformID` and `sender.Username` are both set to the full MXID, e.g. `@alice:example.com`. In `MatchAllowed`, the fallback path strips the leading `@` before doing the legacy checks, so after the canonical match fails it still compares `alice:example.com` against `@alice:example.com`, which will not match.

So unless I'm missing something, `allow_from: ["@alice:example.com"]` still looks broken even with this fallthrough change.

I think we need one of these instead:
- only treat `platform:id` as canonical when `platform` is a recognized platform name
- or preserve the leading `@` for Matrix-style IDs in the fallback path

Could you also add a regression test for the Matrix case in `pkg/identity/identity_test.go`? The current tests stay green because this scenario is not covered.

afjcjsbx:
**pkg/identity/identity.go** The new fallthrough overly broadens the match for canonical `platform:id` values. Previously, if `allowed == "discord:98765432"` and the canonical match failed, the function returned `false`. With this PR, however, it continues into the legacy checks and could authorize a sender from another platform if their `PlatformID` matches the entire canonical string. Concrete example: a sender with `Platform="matrix"` and `PlatformID="discord:98765432"` now matches `allow_from: ["discord:98765432"]`. Since `allow_from` is an authorization check, this is a semantic regression: the fallthrough should be restricted solely to the case of raw MXIDs like `alice:example.com` or to explicitly recognized platform names.

**pkg/identity/identity_test.go** One of the new tests does not exercise the bug fixed by the PR. The case `allowed: "matrix:@alice:example.com"` was already the documented workaround and worked previously as well, so it doesn't guard the control flow change introduced here. What is missing, instead, is a test to prevent the regression mentioned above: a canonical mismatch must not fall back to legacy matches simply because `sender.PlatformID` contains the `platform:id` string.

chengzhichao-xydt:
Thanks for the detailed review. Addressed in dfc66d0:

1. **Fallthrough scope** — When a canonical `platform:id` entry is recognized as such (platform is not numeric and doesn't start with `@`) and fails to match, we now `return false` immediately. Only Matrix-style IDs (`@user:domain`) and numeric compound IDs fall through to legacy matching. This prevents the authorization widening scenario you described.

2. **Regression test** — Added `TestMatchAllowed/canonical_mismatch_does_not_fallthrough_to_PlatformID`: a sender with Platform=matrix and PlatformID=discord:98765432 does NOT match `allow_from: ["discord:98765432"]`.

chengzhichao-xydt:
Closing this PR due to prolonged inactivity (18 days without maintainer review). The fix remains available on the branch if needed. If there is renewed interest, please feel free to reopen or reference this work.

Thanks!

PR #3047, opened 2026-06-07
fix(web): restore full JSONL history for session detail

### LINKED ISSUE #2796: [BUG]历史记录中，一次对话有多次用户消息，只能查看到最后一条用户消息，先前的都看不到

SutraHsing:
## Summary
- Add a detail-only JSONL session reader that ignores meta.Skip so GET /api/sessions/{id} can show archived messages still present in the JSONL file.
- Keep the session list endpoint on the existing active-tail reader so list pagination does not unmarshal full archived transcripts.
- Add meta.Skip > 0 regression coverage for both detail and list behavior.

Fixes #2796

## Test Plan
- go test -count=1 ./web/backend/api

### DIFF
--- a/web/backend/api/session.go
+++ b/web/backend/api/session.go
@@ -170,6 +170,19 @@ func (h *Handler) readSessionMessages(path string, skip int) ([]providers.Messag
 }
 
 func (h *Handler) readJSONLSession(dir, sessionKey string) (sessionFile, error) {
+	return h.readJSONLSessionWithHistory(dir, sessionKey, false)
+}
+
+// readFullJSONLSession is for detail views; list views should keep using
+// readJSONLSession so archived JSONL lines remain skipped.
+func (h *Handler) readFullJSONLSession(dir, sessionKey string) (sessionFile, error) {
+	return h.readJSONLSessionWithHistory(dir, sessionKey, true)
+}
+
+func (h *Handler) readJSONLSessionWithHistory(
+	dir, sessionKey string,
+	fullHistory bool,
+) (sessionFile, error) {
 	base := filepath.Join(dir, sanitizeSessionKey(sessionKey))
 	jsonlPath := base + ".jsonl"
 	metaPath := base + ".meta.json"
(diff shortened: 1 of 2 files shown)


afjcjsbx:
Hi @SutraHsing, thanks for the PR,

#### `web/backend/api/session.go`, `pkg/agent/context_legacy.go` & `pkg/session/jsonl_backend.go`

**Issue: Proposed fix only works prior to file compaction; missing integration test coverage**

The proposed change only resolves the edge case where historical log rows are still present in the `.jsonl` file by zeroing out `meta.Skip`. However, the actual production summarization pipeline executes `TruncateHistory()` followed by `Save()`, and `Save()` physically compacts and overwrites the file.

Once compaction occurs, those older messages cease to exist entirely. As a result, the `GET /api/sessions/{id}` endpoint will continue to return only the active queue. Furthermore, the newly added test in `session_test.go` does not call `Save()`, completely missing the specific execution path that triggers this production bug.

#### `web/backend/api/session_test.go` (Frontend impact: `session-history-menu.tsx`)

**Issue: User-visible data discrepancy between session list and detail endpoints**

The updated test intentionally hardcodes an assertion that the session list response remains strictly derived from the active queue. However, the frontend UI renders the `message_count` directly within the historical session menu.

Following this PR, a session might display a badge indicating "2 messages" in the history list, but clicking to open it will reveal 4 messages in the detail view. Similarly, the session title and preview snippet could end up lagging or referring to a later conversation turn. This introduces a jarring, user-visible state inconsistency between the list and detail views.

PR #3053, opened 2026-06-08
fix(evolution): add ok check for LoadOrStore type assertion in lockStoreFile

chengzhichao-xydt:
## Problem
`lockStoreFile` in `pkg/evolution/store.go` uses an unchecked type assertion on `sync.Map.LoadOrStore` result. A mismatched value causes a panic.

## Fix
Add `ok` check. On failure, atomically replace with a new `*sync.Mutex`, preserving the per-path lock invariant.

## Verification
- `go build ./pkg/evolution/...` passes

## Related
Split from #3018.

### DIFF
--- a/pkg/evolution/store.go
+++ b/pkg/evolution/store.go
@@ -554,19 +554,20 @@ func isInvalidJSON(err error) bool {
 }
 
 func lockStoreFile(path string) func() {
-	actual, _ := storeFileLocks.LoadOrStore(path, &sync.Mutex{})
-	mu, ok := actual.(*sync.Mutex)
-	if !ok {
-		// Delete the corrupted entry so that the next LoadOrStore
-		// atomically creates a fresh mutex. Multiple goroutines may
-		// race to Delete, but LoadOrStore ensures they all converge
-		// on the same mutex before entering the critical section.
-		storeFileLocks.Delete(path)
-		actual, _ = storeFileLocks.LoadOrStore(path, &sync.Mutex{})
-		mu = actual.(*sync.Mutex)
+	for {
+		actual, _ := storeFileLocks.LoadOrStore(path, &sync.Mutex{})
+		mu, ok := actual.(*sync.Mutex)
+		if !ok || mu == nil {
+			// Corrupted entry (wrong type or nil *sync.Mutex).
+			// Atomically swap in a fresh mutex via CompareAndSwap.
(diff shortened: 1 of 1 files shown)


afjcjsbx:
Hi @chengzhichao-xydt, thank you for the PR,

The new `!ok` branch in `lockStoreFile` does not restore the per-path lock atomically. If two goroutines observe the corrupted value together, both can create a new `*sync.Mutex`, do Store one over the other, and then lock different mutexes. At that point the critical section is no longer serialized and concurrent writes are reopened on the same file, i.e., the very invariant that PR says to preserve. Here you need a retry with `CompareAndSwap` or an equivalent mechanism that converges all callers on the same mutex before `Lock()`.

afjcjsbx:
**pkg/evolution/store.go:** The `!ok` branch is still non-atomic. If two goroutines concurrently read the same corrupted value, the first one might perform `Delete`, `LoadOrStore`, and lock the `m1` mutex; immediately after, the second one might execute its own `Delete` (erasing `m1` itself), create `m2`, and lock that one too. At that point, both enter the critical section using different mutexes, thereby reopening concurrent writes on the same file. Therefore, the fix doesn't truly preserve the "one mutex per path" invariant; a repair using CAS/retry or a fail-fast approach is needed here.

**pkg/evolution/store.go** The hardening is incomplete for corrupted values that are of type `*sync.Mutex` but are `nil`. In that case, the type assertion passes (`ok == true`), but `mu.Lock()` will panic anyway. Furthermore, the second assertion on `actual.(*sync.Mutex)` in the recovery path is once again unchecked.

chengzhichao-xydt:
Thanks for the thorough review. Addressed in bbdf746:

1. **Atomicity** — Replaced Delete+LoadOrStore with a `CompareAndSwap` retry loop. When a corrupted entry is detected, CAS atomically swaps in a fresh `*sync.Mutex`. If CAS fails (another goroutine got there first), the loop retries to pick up the valid entry.

2. **Nil `*sync.Mutex`** — Added `|| mu == nil` check. A nil `*sync.Mutex` now enters the same CAS recovery path instead of panicking on `Lock()`.

3. **Unchecked second assertion** — Eliminated entirely by the retry loop; there is no second unchecked assertion.

afjcjsbx (approved):

PR #3063, opened 2026-06-08
feat: add deltachat gateway

trufae:
## 🗣️ Type of Change
- [x] 📖 Documentation update

## 🤖 AI Code Generation

- **Reference URL:**
- **Reasoning:**

## 🧪 Test Environment
- **Hardware:** Jetson Thor
- **OS:** Ubuntu 24.04
- **Model/Provider:** Nemotron3
- **Channels:** DeltaChat

## ☑️ Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -494,6 +494,7 @@ Talk to your PicoClaw through 19+ messaging platforms:
 | **QQ** | Easy (AppID + AppSecret) | WebSocket | [Guide](docs/channels/qq/README.md) |
 | **Slack** | Easy (bot + app token) | Socket Mode | [Guide](docs/channels/slack/README.md) |
 | **Matrix** | Medium (homeserver + token) | Sync API | [Guide](docs/channels/matrix/README.md) |
+| **Delta Chat** | Easy (account script or email/password) | JSON-RPC (email/E2EE) | [Guide](docs/channels/deltachat/README.md) |
 | **DingTalk** | Medium (client credentials) | Stream | [Guide](docs/channels/dingtalk/README.md) |
 | **Feishu / Lark** | Medium (App ID + Secret) | WebSocket/SDK | [Guide](docs/channels/feishu/README.md) |
 | **LINE** | Medium (credentials + webhook) | Webhook | [Guide](docs/channels/line/README.md) |
--- a/docs/channels/deltachat/README.md
+++ b/docs/channels/deltachat/README.md
@@ -1,0 +1,148 @@
+> Back to [README](../../../README.md)
+
+# Delta Chat Channel
+
+PicoClaw can run as a Delta Chat bot by launching a local
+`deltachat-rpc-server` process and talking to it over JSON-RPC. The RPC server
+handles the email account, IMAP/SMTP connection, message store, and encryption
+keys.
+
+## Install
+
+Install the Delta Chat RPC server. If `deltachat-rpc-server` is on `PATH`,
+PicoClaw can find it automatically; otherwise set `rpc_server_path` to the
+exact binary path.
+
+```bash
+pip install deltachat-rpc-server
+which deltachat-rpc-server
+```
+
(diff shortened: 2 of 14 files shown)


trufae:
CLA accepted but the job is still not running

trufae:
If you need some help in testing it let me know and i will update the readme or provide some guidelines. Imho deltachat is very interesting bwcause its distributed, secure and easy to setup, there’s no other channel that can match those requirements and i find it very handy to use from any OS from bots to cli or gui apps.

let me know if i need to change anything in the pr.

trufae:
cleaned up the code and rebased. im using it in daily basis and its quite stable

trufae:
updated to support voice messages and attachments

trufae:
only missing features to be 1:1 with telegram (which is the reference channel) would be to handle message reactions and messages editted, which are probably far from essential and hardly handled by the agent. let me know if there's anything related to style that blocks merging this

trufae:
Theres nothing more to do. The pr works perfectly and its featurefull im just waiting for feedback or merge

afjcjsbx:
Hi @trufae, thank you for the PR,

### 1. Premature Message Marking (`handler.go`)

`markseen_msgs` is currently called before the `group_trigger` filter and before dispatching to `HandleInboundContext`.
* **The Bug:** If the local publish fails, the message is already marked as "seen" and will never be reprocessed.
* **The Side Effect:** In groups with `mention_only` or `prefix` enabled, the bot will send read receipts even for messages it ultimately decides to ignore.
* **Fix:** We should defer marking messages as seen until after they have been successfully filtered and dispatched. 

### 2. Configuration Cache Invalidation (`deltachat.go`)

Channel settings are only applied during the very first bootstrap. Since `configureAccount()` runs exclusively when `is_configured == false`, any subsequent changes to runtime configuration (e.g., `password`, `display_name`, or IMAP/SMTP overrides) are completely ignored unless the `data_dir` is manually wiped.
* **Fix:** We need a mechanism to re-apply or update the account configuration when settings change, rather than relying on a one-time bootstrap flag. 

### 3. Naive Mention Detection (`handler.go`)

Mention detection currently relies on a simple `strings.Contains` check against the `display_name`.
* **The Issue:** If the bot has a short or common name (e.g., "bot", "ai"), it will false-trigger on regular words (like "ro**bot**" or "m**ai**l"), effectively bypassing the `group_trigger.mention_only` restriction.
* **Fix:** We should boundary-check the mention (e.g., using a regex with word boundaries `\b`, checking for an `@` prefix, or splitting by whitespace) to ensure we are matching the exact handle.

trufae:
Solved (1) and (2), but for (3) i think it will be better to have an api shared across all the gateways instead. i can refactor this but for this pr i dont want to touch code outside the teelgram gateway. pr updated now

trufae:
I have redesigned the way account creation and auth works and updated the documentation, because it's now more secure and easy:

* If no email address is defined in the configuration picoclaw will show a list of default servers to create an account
* when the email address contains no username (`@foo.com`) it will create the account and tell you the full email address (deltachat accounts are randomized)
* the password is stored by the jsonrpc, so there's no need to dump it or store it in plain in the config file
* when the emial address is specified in the config file it will just show the QR and the invite link to start chatting

trufae:
oh, also i added the `avatar_image` directive in the deltachat channel, when defined, on startup it will use this image in the profile picture of the configured account. this way there's no need to use a separate client to define or change it. The avatar is not transfered witohut key exchange that happens after the first message

afjcjsbx (approved):
LGTM!

trufae:
I was working on a cleanup, simplifying the implementation and improving the documentation! But happy to work on this in a separate pr!

Thanks for merging! I found some other issues in the memory management and i have some questions about some logic that can be probably unified across the channels

PR #3067, opened 2026-06-09
fix: add DmScope field to SessionConfig to persist dm_scope setting

SiYue-ZO:
## Problem

The "Session Scope"（运行时会话隔离范围）setting on the config page can be modified in the UI but **cannot be saved**. After saving and reloading the page, it always reverts to the default `per-channel-peer`.

## Root Cause

The frontend sends `dm_scope` as part of the `session` config, but the backend `SessionConfig` struct lacked the corresponding `DmScope` field. Go's `encoding/json` silently discards unknown fields during unmarshal, so the value was lost on every PATCH request.

Additionally, the custom `MarshalJSON` only emits the `session` block when `Dimensions` or `IdentityLinks` are non-empty, so even if `dm_scope` were stored, it would not appear in GET responses.

## Fix

- Add `DmScope string` field with json tag `dm_scope` to `SessionConfig` struct
- Update `MarshalJSON` condition to include session when `DmScope` is set

Only one file changed: `pkg/config/config.go` (2 insertions, 1 deletion).

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -257,7 +257,7 @@ func (c *Config) MarshalJSON() ([]byte, error) {
 		Alias: (*Alias)(c),
 	}
 
-	if len(c.Session.Dimensions) > 0 || len(c.Session.IdentityLinks) > 0 {
+	if len(c.Session.Dimensions) > 0 || len(c.Session.IdentityLinks) > 0 || c.Session.DmScope != "" {
 		sessionCfg := c.Session
 		aux.Session = &sessionCfg
 	}
@@ -349,6 +349,7 @@ type DispatchSelector struct {
 type SessionConfig struct {
 	Dimensions    []string            `json:"dimensions,omitempty"`
 	IdentityLinks map[string][]string `json:"identity_links,omitempty"`
+	DmScope       string              `json:"dm_scope,omitempty"`
 }
 
 // RoutingConfig controls the intelligent model routing feature.
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -352,6 +352,26 @@ type SessionConfig struct {
 	DmScope       string              `json:"dm_scope,omitempty"`
 }
 
+// ApplyDmScope translates the user-facing dm_scope value into the internal
+// dimensions array that the routing layer consumes. It is a no-op when
+// DmScope is empty or when Dimensions is already set (explicit Dimensions
+// take precedence over the derived value).
+func (s *SessionConfig) ApplyDmScope() {
+	if s.DmScope == "" || len(s.Dimensions) > 0 {
+		return
+	}
+	switch s.DmScope {
+	case "per-channel-peer":
+		s.Dimensions = []string{"chat", "sender"}
+	case "per-channel":
+		s.Dimensions = []string{"chat"}
+	case "per-peer":
+		s.Dimensions = []string{"sender"}
+	case "global":
+		s.Dimensions = nil
+	}
+}
+
 // RoutingConfig controls the intelligent model routing feature.
 // When enabled, each incoming message is scored against structural features
 // (message length, code blocks, tool call history, conversation depth, attachments).
@@ -1476,6 +1496,8 @@ func LoadConfig(path string) (*Config, error) {
 		homePath := GetHome()
 		cfg.Agents.Defaults.Workspace = filepath.Join(homePath, pkg.WorkspaceName)
 	}
+
+	cfg.Session.ApplyDmScope()
 
 	return cfg, nil
 }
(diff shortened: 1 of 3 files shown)


SiYue-ZO:
## Update: dm_scope is now wired into runtime behavior

The initial commit only fixed the **storage** problem (`dm_scope` was silently dropped during JSON unmarshal). This follow-up commit fixes the **runtime** problem — `dm_scope` was a dead config field that never affected actual session isolation.

### What was missing

The routing layer reads `SessionConfig.Dimensions` (e.g. `["chat", "sender"]`) to construct session keys, but the frontend only sends the user-friendly `dm_scope` value (e.g. `"per-channel-peer"`). No code ever translated one into the other.

### What this commit adds

1. **`SessionConfig.ApplyDmScope()`** — maps `dm_scope` → `Dimensions`:
   | `dm_scope` | `Dimensions` |
   |---|---|
   | `per-channel-peer` | `["chat", "sender"]` |
   | `per-channel` | `["chat"]` |
   | `per-peer` | `["sender"]` |
   | `global` | `[]` |

   Explicit `Dimensions` take precedence when already set.

2. **Called in `LoadConfig` post-processing** — covers startup and config reload.

3. **Called in PATCH and PUT API handlers** — covers frontend save flow.

4. **Table-driven tests** — all 4 scope values + precedence + no-op cases.

afjcjsbx:
Hi @SiYue-ZO, thank you for the PR,

**PATCH /api/config leaves session.dimensions stale**
Changing the "Session Scope" via the dashboard still fails to update runtime behavior. In `web/backend/api/config.go:147`, the merge starts from `config.LoadConfig()`, and this PR adds `ApplyDmScope()` in `pkg/config/config.go:1479`. This means the base merge object already contains the dimensions derived from the *old* `dm_scope`. After a patch like `{"session":{"dm_scope":"per-channel"}}`, `newCfg.Session.Dimensions` remains populated. As a result, `ApplyDmScope()` exits early (`pkg/config/config.go:349`), causing the saved file to end up with a mismatched configuration (e.g., `dm_scope: "per-channel"` but `dimensions: ["chat","sender"]`). Since routing relies on `Dimensions` (`pkg/routing/route.go:101`), the UI will display the new scope while the runtime continues to use the old one.

**Inconsistent scope mapping for legacy/fresh configs in UI**
Existing or fresh configurations still reload with a UI-displayed scope that doesn't match reality. `handleGetConfig()` directly serializes the output of `LoadConfig()` (`web/backend/api/config.go:36`), but `ApplyDmScope()` only handles the `dm_scope -> dimensions` conversion; it lacks the inverse mapping when only `dimensions` are present. Therefore, for legacy configs—or even `DefaultConfig()` which initializes with `["chat"]` (`pkg/config/defaults.go:47`)—the API response might completely omit `session.dm_scope`. The frontend form then falls back to its default (`per-channel-peer` in `web/frontend/src/components/config/form-model.ts:376`). Consequently, the page can display an incorrect scope until the configuration is explicitly rewritten.

SiYue-ZO:
> Hi @SiYue-ZO, thank you for the PR,
> 
> **PATCH /api/config leaves session.dimensions stale** Changing the "Session Scope" via the dashboard still fails to update runtime behavior. In `web/backend/api/config.go:147`, the merge starts from `config.LoadConfig()`, and this PR adds `ApplyDmScope()` in `pkg/config/config.go:1479`. This means the base merge object already contains the dimensions derived from the _old_ `dm_scope`. After a patch like `{"session":{"dm_scope":"per-channel"}}`, `newCfg.Session.Dimensions` remains populated. As a result, `ApplyDmScope()` exits early (`pkg/config/config.go:349`), causing the saved file to end up with a mismatched configuration (e.g., `dm_scope: "per-channel"` but `dimensions: ["chat","sender"]`). Since routing relies on `Dimensions` (`pkg/routing/route.go:101`), the UI will display the new scope while the runtime continues to use the old one.
> 
> **Inconsistent scope mapping for legacy/fresh configs in UI** Existing or fresh configurations still reload with a UI-displayed scope that doesn't match reality. `handleGetConfig()` directly serializes the output of `LoadConfig()` (`web/backend/api/config.go:36`), but `ApplyDmScope()` only handles the `dm_scope -> dimensions` conversion; it lacks the inverse mapping when only `dimensions` are present. Therefore, for legacy configs—or even `DefaultConfig()` which initializes with `["chat"]` (`pkg/config/defaults.go:47`)—the API response might completely omit `session.dm_scope`. The frontend form then falls back to its default (`per-channel-peer` in `web/frontend/src/components/config/form-model.ts:376`). Consequently, the page can display an incorrect scope until the configuration is explicitly rewritten.

maybe you can review again?thanks!

afjcjsbx (approved):
LGTM!

PR #3087, opened 2026-06-09
fix(tools): allow workspace relative exec paths

jp39:
## 📝 Description

Fixes a false positive in the exec safety guard when `restrict_to_workspace` is enabled. Relative workspace paths containing `/`, such as `skills/calendar-query/scripts/query_calendar.py`, were incorrectly treated as absolute path fragments
and blocked as outside the working directory.

This PR resolves relative path tokens against the exec working directory before applying workspace-boundary checks, while preserving blocking for real absolute paths, path traversal, `file://` escapes, and unsafe outside-workspace references.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

- **Reference URL:**
- **Reasoning:** The exec guard’s `absolutePathPattern` matched `/...` substrings inside relative paths. The fix classifies slash matches before validation and resolves relative path tokens under the exec `cwd`, avoiding false blocks without
weakening absolute path or traversal checks.

## 🧪 Test Environment
- **Hardware:** PC
- **OS:** Linux
- **Model/Provider:** Codex
- **Channels:** CLI

## 📸 Evidence (Optional)
<details>

```bash
GOCACHE=/tmp/picoclaw-go-build go test ./pkg/tools -count=1
ok  	github.com/sipeed/picoclaw/pkg/tools	11.622s

GOCACHE=/tmp/picoclaw-go-build go test ./pkg/config -count=1
ok  	github.com/sipeed/picoclaw/pkg/config	1.097s
</details>

## ☑️ Checklist

### DIFF
--- a/pkg/tools/shell.go
+++ b/pkg/tools/shell.go
@@ -1201,7 +1201,7 @@ func (t *ExecTool) guardCommand(command, cwd string) string {
 				}
 			}
 
-			p, err := filepath.Abs(raw)
+			p, err := commandPathAbs(commandPathTextFromMatch(cmd, loc[0], loc[1]), cwdPath)
 			if err != nil {
 				continue
 			}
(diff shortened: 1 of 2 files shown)


afjcjsbx:
Hi @jp39, thanks for you PR, 

- Path validation is checking the entire flag-value token instead of extracting just the path. For example, with `--config=link/passwd`, `EvalSymlinks()` ends up evaluating `<cwd>/--config=link/passwd` rather than the actual path `<cwd>/link/passwd`. If `link` is a symlink pointing outside the workspace (e.g., to `/etc`), the check completely misses the escape. This reopens the exact path traversal bypass the guard was meant to prevent.


- `isUnixAbsolutePathMatchStart()` incorrectly restricts "absolute paths attached to a flag" to prefixes of exactly length 2 (`len(prefix) == 2`), missing longer valid options. For instance, I verified locally that `clang -isystem/usr/include` is correctly parsed by clang as `-isystem /usr/include`. However, with this patch, `/usr/include` is misclassified as a relative token (`-isystem/usr/include`), resolved against `cwd`, and successfully bypasses the out-of-workspace absolute path check.

jp39:
Thanks for catching this. You were right: the previous patch validated the whole option token in cases like --config=link/passwd, which could miss a symlink escape, and the len(prefix) == 2 check was too narrow for longer attached options like -isystem/usr/include.

I’ve amended the commit so the guard now:

- extracts and validates the value portion for --flag=rel/path forms;
- keeps ambiguous attached option paths like -isystem/usr/include conservative by validating the slash-starting path;
- still allows normal relative workspace paths such as skills/calendar-query/scripts/query_calendar.py.

I also added regression coverage for both cases you pointed out:

- --config=link/passwd where link is a symlink outside the workspace;
- longer attached absolute option paths such as -isystem/usr/include.

Tests pass locally.

afjcjsbx (approved):
LGTM!

PR #3091, opened 2026-06-10
fix(openai_compat): add ok check for native_search type assertion

chengzhichao-xydt:
## Problem
`pkg/providers/openai_compat/provider.go:157` discards the `ok` from a type assertion on `options["native_search"]`. A non-bool value silently becomes `false`, which can disable native search without any diagnostic.

## Fix
Capture `ok` and explicitly ignore it with `_ = ok`, consistent with the `maxTokens` pattern on line 164.

## Verification
- `go build ./pkg/providers/openai_compat/...` passes

### DIFF
--- a/pkg/providers/openai_compat/provider.go
+++ b/pkg/providers/openai_compat/provider.go
@@ -155,7 +155,14 @@ func (p *Provider) buildRequestBody(
 
 	// When fallback uses a different provider (e.g. DeepSeek), that provider must not inject web_search_preview.
 	nativeSearch, ok := options["native_search"].(bool)
-	_ = ok
+	if !ok {
+		// If the option is present but not a bool, log a warning and
+		// treat it as false — web_search_preview must not be injected
+		// when the caller cannot express a well-typed intent.
+		if _, present := options["native_search"]; present {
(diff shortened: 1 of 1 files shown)


afjcjsbx:
Thanks for the patch. I don't think this actually fixes the issue described in the PR.

Before:
`nativeSearch, _ := options["native_search"].(bool)`

After:
`nativeSearch, ok := options["native_search"].(bool)`
`_ = ok`

Because `ok` is still ignored, the behavior is unchanged: if `options["native_search"]` is not a `bool`, `nativeSearch` still silently becomes `false` with no diagnostic.

So as written, this looks like a no-op rather than a functional fix. If we want to address the problem, we probably need to explicitly handle the `!ok` case (for example by logging, returning an error, or adding a helper with clear coercion/validation semantics), and add a test that covers the invalid-type case.

chengzhichao-xydt:
Thanks for the review. You're right — the previous patch was a no-op.

Updated in 409cc05: when the type assertion fails and the option key is present, we now log a warning with the unexpected type. The behavior stays conservative (nativeSearch defaults to false), but the caller gets a diagnostic instead of silent failure.

When the option is absent entirely, there's nothing to warn about — nativeSearch simply stays false, which is the intended default.

afjcjsbx (approved):

PR #3115, opened 2026-06-12
Fix inline data URL media extraction for generic tool output

jp39:
## 📝 Description

This PR fixes a session-history corruption bug where PicoClaw treated `data:image/...;base64,...` strings found inside plain text tool output as real media attachments.

Generic tools like `read_file` and `exec` can legitimately return source code, logs, shell output, HTML/CSS/JSON, markdown, or test fixtures containing data URL literals. Those strings are text, not produced media. Previously, normalization globally scanned tool output, extracted matching inline data URLs, stored them in the media store, and persisted `media://...` refs on the tool message.

That could poison the session history. A later LLM turn would replay the persisted media ref, and the provider could reject the request with:

The image data you provided does not represent a valid image

### How to reproduce the original bug

1. Create or read a source file containing a data URL literal whose payload is not actually an image:

```
package demo

const badFixture = "data:image/png;base64,aGVsbG8="
```

`aGVsbG8=` decodes to `hello`, which is valid base64 but not valid PNG bytes.

2. Ask PicoClaw to read that file using a generic text tool such as `read_file`.
3. Inspect the persisted session JSONL.

Buggy behavior looked like this:

```
{
  "role": "tool",
  "name": "read_file",
  "content": "... data:image/png;base64,aGVsbG8= ...",
  "media": [
    "media://..."
  ]
}
```

This is wrong because `read_file` returned source text, not an image.

4. Send another message in the same conversation.

The later request may fail because history replay resolves and sends the bad `media://...` image to the model API.

### Fix summary

Inline data URL media extraction is now explicit/opt-in via `InlineMediaProducer`. Generic textual tools no longer create media attachments from data URL literals. Opt-in inline image payloads are validated before storage, so invalid, empty, unsupported, truncated, or magic-byte-mismatched payloads are rejected.

## 🗣️ Type of Change

- [ ] 📖 Documentation update

## 🤖 AI Code Generation

- Reference URL:
- Reasoning: Validation alone is not enough because even a valid image data URL inside source code is still source text, not an intentional media artifact. The fix gates extraction behind explicit tool opt-in and keeps existing explicit media-producing paths working.

## 🧪 Test Environment

- Hardware: PC
- OS: Linux
- Model/Provider: N/A
- Channels: N/A

## 📸 Evidence (Optional)
<details>

ok  	github.com/sipeed/picoclaw/pkg/tools	10.478s
ok  	github.com/sipeed/picoclaw/pkg/tools	(cached)
ok  	github.com/sipeed/picoclaw/pkg/tools/fs	(cached)
ok  	github.com/sipeed/picoclaw/pkg/tools/hardware	(cached)
ok  	github.com/sipeed/picoclaw/pkg/tools/integration	(cached)
ok  	github.com/sipeed/picoclaw/pkg/tools/shared	(cached)
ok  	github.com/sipeed/picoclaw/pkg/agent	0.021s
</details>

## ☑️ Checklist


jp39:
Ping. This is IMHO a pretty serious usability bug as it can make entire sessions unusable.

jp39:
Ping

afjcjsbx:
Hi, thanks you for the PR, 

generic tools continue to mangle "large" data URLs instead of preserving them as text. `sanitizeToolLLMContent` still calls `sanitizeInlineDataURLsForLLM(..., false)`, which removes any data URL with a `payload >= 1024` even when `AllowInlineMediaExtraction` is false. Therefore, a `read_file` on HTML/CSS/fixtures with a realistic inline asset still results in something like `const fixture = ""` plus the omission note, and the LLM no longer sees the actual source code. The new tests only cover small payloads in `pkg/tools/registry_test.go`, so this case remains uncovered.

jp39:
Pushed a follow-up fix for the large data URL case.

Changes:

- Generic text output now preserves inline `data:*;base64,...` URLs instead of stripping them.
- Large data URLs in source/text output no longer get mistaken for standalone large base64 payloads.
- MCP text output now follows the same preservation behavior.
- Explicit inline-media tools still use the opt-in extraction path and still validate image payloads before registering media.

Added regression coverage for:

- Generic `read_file` output containing a data URL with a payload `>= 1024` characters.
- MCP text output containing a large data URL.

Validation:

```bash
go test ./pkg/tools ./pkg/tools/integration
go test ./pkg/tools/...
```

afjcjsbx:
The opt-in introduced in `registry.go` (line 334) effectively disables inline media extraction for all production tools.
Currently, `normalizeToolResult` only extracts `data:` URLs if the tool implements `InlineMediaProducer`. 
However, the only implementation present in the codebase seems to be the test mock in `registry_test.go`.
Critical production tools like `MCPTool` do not implement this interface. As a result, an MCP server returning inline images via markdown or data URLs will no longer produce `media://` attachments—instead, it will either leak raw base64 data into the text or leave it completely unprocessed.
This introduces a regression compared to the previous global behavior.  Did we break this intentionally, or should `MCPTool` (and potentially others) implement `InlineMediaProducer`?

jp39:
Pushed a follow-up for the MCP inline-media regression.

Changes:

- `MCPTool` now implements `InlineMediaProducer`, so MCP text content that intentionally contains inline image data URLs still goes through media extraction.
- Generic text tools remain non-opt-in, so source/log/fixture data URL literals are still preserved as text.
- Added a registry-level regression test covering a real MCP tool returning a markdown data URL and producing a `media://` attachment.

Validation:

```bash
go test ./pkg/tools ./pkg/tools/integration
go test ./pkg/tools/...
```

jp39:
Ping

jp39:
Ping

afjcjsbx:
I've reviewed the changes and spotted a couple of issues that still need to be addressed:

* **Incomplete fix for history corruption (`pkg/tools/integration/mcp_tool.go` and `pkg/tools/registry.go`)**
Currently, `MCPTool.ProducesInlineMedia()` returns `true` for *every* MCP tool. Because of this, the registry still extracts `data:...;base64,...` from any MCP text result. While `MCPTool.Execute()` now preserves the literal, `ToolRegistry.ExecuteWithContext()` immediately strips it back out and persists a `media://...` ref instead. This reintroduces the history-corruption bug for generic MCP tools (like filesystem/shell/log readers).
*Note:* The new `TestMCPTool_Execute_LargeDataURLTextIsPreserved` test misses this regression because it only covers the direct `Execute()` path rather than the actual registry path.
* **Insufficient WebP validation (`pkg/tools/normalization.go`)**
The new WebP validation relies only on a header check. Unlike the branches for PNG, JPEG, and GIF, the `image/webp` branch never actually decodes or parses the payload. This means a truncated or header-only "fake" WebP can successfully pass `validateInlineImagePayload`, get stored as `image/webp`, and cause a failure later when the provider tries to replay it as an image. We need to fully parse the WebP payload to fulfill the PR's goal of rejecting invalid/truncated inline payloads.

jp39:
Updated the branch to address this with a per-MCP-server opt-in instead of enabling inline media extraction for all MCP text output.

The new config option is:

```json
"inline_media_extraction": true
```

under an individual MCP server config. It defaults to false, so generic MCP text output like source files, logs, and fixtures keeps data:*;base64,... literals as text. Servers that intentionally return inline media in text output, such as screenshot tools, can opt in and have valid inline media extracted into attachments.

jp39:
Ping

PR #3118, opened 2026-06-12
Add remote Pico WebSocket mode to picoclaw agent

jp39:
## 📝 Description

This PR adds an optional remote mode to the existing `picoclaw agent`
command.

Local behavior is unchanged:

```bash
picoclaw agent
picoclaw agent -m "hello"
```

New remote behavior:

```
picoclaw agent --remote ws://localhost:18790/pico/ws
picoclaw agent --remote ws://localhost:18790/pico/ws --token "$PICO_TOKEN"
picoclaw agent --remote ws://localhost:18790/pico/ws -s test-session -m "hello"
```

When `--remote` is set, the CLI connects to a remote Pico WebSocket
endpoint instead of starting the local agent loop.

The remote client:

- Adds or replaces `session_id` in the WebSocket URL.
- Sends `message.send` with the same session ID.
- Supports `--token` and falls back to `PICO_TOKEN`.
- Supports one-shot mode with `-m`.
- Supports interactive mode with the lobster prompt.
- Handles Ctrl-C and EOF cleanly.
- Renders Pico events such as:
    - `message.create`
    - `message.update`
    - `message.delete`
    - `typing.start`
    - `typing.stop`
    - `media.create`
    - `error`
    - `pong`
    - unknown event types
- Generates a fresh `cli:<uuid>` session when `--session` is not passed.

The PR also adds simple client metadata so the agent can know that a
turn came from the remote CLI over WebSocket. The remote client sends
metadata like:

client_kind=remote_cli
client_name=picoclaw agent --remote
transport=websocket

The Pico server records this on inbound messages. The server sets
`transport=websocket` itself, so the client cannot spoof the transport.

## 🗣️ Type of Change

- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue

N/A

- Reference URL: N/A
- Reasoning:

The Pico WebSocket server expects clients to connect to:

/pico/ws?session_id=<session>

The same session ID must also be used in outgoing `message.send`
events. If the URL session and message session differ, the remote
response can be broadcast to a session that has no connected client.

When --session is omitted in remote mode, the CLI generates a fresh
cli:<uuid> session. This avoids accidentally sharing one remote
conversation across separate CLI runs.

Local mode is unchanged and still uses the existing cli:default
session.

Example:

```bash
picoclaw agent --remote ws://localhost:18790/pico/ws
```

connects with a generated session ID, for example:

ws://localhost:18790/pico/ws?session_id=cli%3A550e8400-e29b-41d4-a716-446655440000

while this uses the explicit session:

```
picoclaw agent --remote ws://localhost:18790/pico/ws -s test-session -m "hello"
```

ws://localhost:18790/pico/ws?session_id=test-session

And sends a Pico message with:

type=message.send
session_id=test-session

## 🧪 Test Environment

- Hardware: Development PC, plus remote PicoClaw gateway host
- OS: Linux
- Model/Provider: Existing configured PicoClaw provider
- Channels: Pico WebSocket, local CLI agent

## 📸 Evidence (Optional)
<details>

Focused test run:

go test ./cmd/picoclaw/internal/agent ./pkg/channels/pico ./pkg/agent

Result:

ok   github.com/sipeed/picoclaw/cmd/picoclaw/internal/agent
ok   github.com/sipeed/picoclaw/pkg/channels/pico
ok   github.com/sipeed/picoclaw/pkg/agent

Example remote one-shot command:

```
picoclaw agent \
  --remote ws://localhost:18790/pico/ws \
  --token "$PICO_TOKEN" \
  --session test-session \
  --message "hello"
```

Example remote interactive command:

```
picoclaw agent \
  --remote ws://localhost:18790/pico/ws \
  --token "$PICO_TOKEN"
```
</details>

## ☑️ Checklist

### DIFF
--- a/cmd/picoclaw/internal/agent/command.go
+++ b/cmd/picoclaw/internal/agent/command.go
@@ -20,12 +20,16 @@ func NewAgentCommand() *cobra.Command {
 		Args:  cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, _ []string) error {
 			if remoteURL != "" {
+				remoteSessionKey := sessionKey
+				if !cmd.Flags().Changed("session") {
+					remoteSessionKey = ""
+				}
 				return remoteAgentCmd(
 					cmd.Context(),
 					remoteURL,
 					token,
 					message,
-					sessionKey,
+					remoteSessionKey,
 					cmd.InOrStdin(),
 					cmd.OutOrStdout(),
 				)
--- a/cmd/picoclaw/internal/agent/command.go
+++ b/cmd/picoclaw/internal/agent/command.go
@@ -10,6 +10,8 @@ func NewAgentCommand() *cobra.Command {
 		sessionKey string
 		model      string
 		debug      bool
+		remoteURL  string
+		token      string
 	)
 
 	cmd := &cobra.Command{
@@ -17,6 +19,17 @@ func NewAgentCommand() *cobra.Command {
 		Short: "Interact with the agent directly",
 		Args:  cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, _ []string) error {
+			if remoteURL != "" {
+				return remoteAgentCmd(
+					cmd.Context(),
+					remoteURL,
+					token,
+					message,
+					sessionKey,
+					cmd.InOrStdin(),
+					cmd.OutOrStdout(),
+				)
+			}
 			return agentCmd(message, sessionKey, model, debug)
 		},
 	}
@@ -25,6 +38,8 @@ func NewAgentCommand() *cobra.Command {
 	cmd.Flags().StringVarP(&message, "message", "m", "", "Send a single message (non-interactive mode)")
 	cmd.Flags().StringVarP(&sessionKey, "session", "s", "cli:default", "Session key")
 	cmd.Flags().StringVarP(&model, "model", "", "", "Model to use")
+	cmd.Flags().StringVar(&remoteURL, "remote", "", "Connect to a remote Pico WebSocket URL")
+	cmd.Flags().StringVar(&token, "token", "", "Bearer token for remote Pico auth")
 
 	return cmd
 }
--- a/cmd/picoclaw/internal/agent/command.go
+++ b/cmd/picoclaw/internal/agent/command.go
@@ -42,7 +42,12 @@ func NewAgentCommand() *cobra.Command {
 	cmd.Flags().StringVarP(&message, "message", "m", "", "Send a single message (non-interactive mode)")
 	cmd.Flags().StringVarP(&sessionKey, "session", "s", "cli:default", "Session key")
 	cmd.Flags().StringVarP(&model, "model", "", "", "Model to use")
-	cmd.Flags().StringVar(&remoteURL, "remote", "", "Connect to a remote Pico WebSocket URL")
+	cmd.Flags().StringVar(
+		&remoteURL,
+		"remote",
+		"",
+		"Connect to a remote Pico WebSocket URL, e.g. ws://host:18790/pico/ws",
+	)
 	cmd.Flags().StringVar(&token, "token", "", "Bearer token for remote Pico auth")
 
 	return cmd
--- a/cmd/picoclaw/internal/agent/command_test.go
+++ b/cmd/picoclaw/internal/agent/command_test.go
@@ -30,4 +30,6 @@ func TestNewAgentCommand(t *testing.T) {
 	assert.NotNil(t, cmd.Flags().Lookup("message"))
 	assert.NotNil(t, cmd.Flags().Lookup("session"))
 	assert.NotNil(t, cmd.Flags().Lookup("model"))
+	assert.NotNil(t, cmd.Flags().Lookup("remote"))
+	assert.NotNil(t, cmd.Flags().Lookup("token"))
 }
--- a/cmd/picoclaw/internal/agent/remote.go
+++ b/cmd/picoclaw/internal/agent/remote.go
@@ -1,0 +1,695 @@
+package agent
+
+import (
+	"bufio"
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"mime"
+	"net/http"
+	"net/url"
+	"os"
+	"os/signal"
+	"path/filepath"
+	"strings"
+	"sync"
+	"syscall"
+	"time"
+
+	"github.com/ergochat/readline"
+	"github.com/google/uuid"
+	"github.com/gorilla/websocket"
+
+	"github.com/sipeed/picoclaw/cmd/picoclaw/internal"
+	"github.com/sipeed/picoclaw/pkg/channels/pico"
+)
+
+const (
+	defaultRemoteSessionID    = "cli:default"
+	remoteOneShotFirstTimeout = 30 * time.Second
+	remoteOneShotIdleTimeout  = 1500 * time.Millisecond
+)
+
+var errRemoteReadlineUnavailable = errors.New("remote readline unavailable")
+
+type remoteClient struct {
+	conn *websocket.Conn
+	out  *lockedWriter
+	mu   sync.Mutex
+}
+
+type remoteEventResult struct {
+	Displayed bool
+	Err       error
+}
+
+type remoteInputResult struct {
+	Line string
+	Err  error
+}
+
+type lockedWriter struct {
+	mu            sync.Mutex
+	w             io.Writer
+	refreshPrompt func()
+	typingVisible bool
+}
+
+func (w *lockedWriter) Printf(format string, args ...any) {
+	w.mu.Lock()
+	defer w.mu.Unlock()
+	w.prepareAsyncOutputLocked()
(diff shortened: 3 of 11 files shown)


SiYue-ZO:
Overall a high-quality PR 👏 — clean architecture, zero invasion of the local path, thorough test coverage, and an excellent PR description.

---

### 🔴 Suggested Fixes

**1. Concurrency risk between `readLoop` and `Close()`**

In `readLoop`, the mutex is held only to obtain a reference to `conn`, then released before calling `conn.ReadJSON()`. Meanwhile, `Close()` may concurrently set `c.conn` to nil and call `conn.Close()` on the underlying connection. gorilla/websocket's `Conn` does not guarantee safe concurrent reads/writes, which could lead to a panic.

```go
// In readLoop:
c.mu.Lock()
conn := c.conn
c.mu.Unlock()
// ← Here, Close() may concurrently call conn.Close()
var msg pico.PicoMessage
if err := conn.ReadJSON(&msg); err != nil { // concurrent read on a closed conn
```

**Suggestion**: Use `sync.Once` or a `done` channel to coordinate shutdown, ensuring `ReadJSON` and `Close` never operate on the same underlying conn concurrently. For example:

```go
type remoteClient struct {
    conn      *websocket.Conn
    out       *lockedWriter
    mu        sync.Mutex
    closed    chan struct{}
    closeOnce sync.Once
}

func (c *remoteClient) Close(code int) {
    c.closeOnce.Do(func() {
        close(c.closed)
        c.mu.Lock()
        defer c.mu.Unlock()
        // ... write close message and close conn
    })
}
```

In `readLoop`, `select` on `c.closed` to avoid continuing to read after the connection is closed.

---

### 🟡 Improvement Suggestions

**2. Add WebSocket heartbeat/keepalive**

The code handles `TypePong` events but never actively sends pings. If there is a reverse proxy or load balancer in between (nginx defaults to a 60s proxy timeout), long-idle interactive connections may be silently terminated. Consider adding periodic pings, or use `conn.SetReadDeadline` + pong handler for automatic deadline renewal.

**3. One-shot idle timeout (1.5s) may be too short**

`remoteOneShotIdleTimeout = 1500ms` — for slow models or long streaming responses, a 1.5s silence threshold may prematurely end the response. Consider increasing to 3–5s, or allow adjustment via an environment variable.

**4. Scanner fallback path has no test coverage**

`runScannerInteractive` is the fallback when readline is unavailable, but it currently has no dedicated test. Consider adding a test case that passes a plain `io.Reader` to cover this code path.

**5. Should `sender_id` / `sender_name` be extracted server-side?**

`buildRemoteMessageSend` sets `"sender_id": "picoclaw-cli"` and `"sender_name": "PicoClaw CLI"`, but `copyPicoClientMetadata` only copies `client_kind` and `client_name`. Are these fields consumed downstream? If not, consider adding a comment explaining their purpose or removing them to avoid confusion.

---

### 🟢 Minor Suggestions

**6. Flag help documentation** — The `--remote` description could include a short URL example (e.g., `ws://host:port/pico/ws`) so users can see the expected format directly from `--help`.

**7. Terminal compatibility** — The `\033[1M` (Delete Line) escape sequence used in `clearTypingLineLocked` may not be supported on all terminal emulators. Consider adding a comment noting the expected behavior.

---

### ✅ Highlights

- Local agent path is completely unaffected — excellent backward compatibility
- `cmd.Flags().Changed("session")` to distinguish explicit args from defaults is a clever design
- Server-side enforcement of `transport=websocket` prevents client spoofing — good security practice
- Tests cover URL construction, auth headers, one-shot interaction, session generation, and event rendering

---

**Conclusion**: The only critical concern is **#1 — the `readLoop` concurrency safety issue**, which is the only risk that could cause a runtime panic. Recommend merging after that fix.

jp39:
Thanks, addressed concern #1 in `c0a1593fabd6`.

I changed the remote client lifecycle to make shutdown explicit:

- `Close()` is now guarded by `sync.Once`, so it is idempotent.
- The client has a `closed` channel that records shutdown state.
- `readLoop` checks that channel and treats read errors after close as
  expected shutdown, not real remote errors.
- `Send()` now checks the closed state and serializes normal writes with
  a dedicated write mutex.

One important detail: I did not try to prevent `Close()` and `ReadJSON()`
from overlapping. That overlap is intentional and safe for Gorilla
WebSocket. The Gorilla concurrency docs say that `Close` and
`WriteControl` may be called concurrently with other connection methods:
https://pkg.go.dev/github.com/gorilla/websocket#hdr-Concurrency

Closing the WebSocket from another goroutine is also the normal way to
unblock a goroutine currently waiting in `ReadJSON()`. If we held a mutex
around `ReadJSON()` and made `Close()` wait for it, Ctrl-C shutdown could
hang forever on an idle connection.

So the fix here is not to serialize `Close()` behind the read. Instead,
it makes shutdown state explicit and makes the expected close path quiet
and repeatable.

I also added `TestRemoteClientCloseUnblocksReadLoop`, which verifies that:

- `Close()` makes `readLoop` exit.
- Calling `Close()` more than once is safe.
- `readLoop` does not report a normal local close as a remote error.
- Sending after close returns a closed connection error.

Validation:

```bash
go test -race -run TestRemoteClientCloseUnblocksReadLoop -timeout=20s ./cmd/picoclaw/internal/agent
go test ./cmd/picoclaw/internal/agent ./pkg/channels/pico ./pkg/agent
```

jp39:
> 4. Scanner fallback path has no test coverage
> 
> runScannerInteractive is the fallback when readline is unavailable, but it currently has no dedicated test. Consider adding a test case that passes a plain io.Reader to cover this code path.

Good point. I addressed this by removing the scanner fallback instead
of adding tests for it.

Commit: `0fd0c806e8e2`

The fallback duplicated the interactive REPL loop with different prompt
and terminal behavior. That means we would have had to maintain and test
two interactive paths, even though users should normally use the readline
path for interactive mode.

The behavior is now simpler:

- `picoclaw agent --remote ...` uses the readline interactive client.
- `picoclaw agent --remote ... -m "hello"` remains the non-interactive /
  scripted path.
- If readline cannot initialize, the command returns a clear error and
  points users to `-m` for one-shot mode.

This removes the untested fallback path rather than adding coverage for
code that did not provide much practical value.

jp39:
Pushed the follow-up commits.

Changes:

- Relaxed one-shot idle timeout from `1.5s` to `5s`, with coverage for a
  delayed second response chunk.
- Removed unused `sender_id` / `sender_name` from remote CLI payloads.
  The server owns trusted sender identity.
- Added an example `--remote` URL in help text:
  `ws://host:18790/pico/ws`
- Added a short comment for the terminal escape sequence that clears the
  transient `[typing]` line.

I left WebSocket heartbeat out for now. I agree it is useful for long-idle
interactive sessions behind proxies, but it is a larger behavior change
with its own choices around ping interval, pong deadline, close handling,
and tests. I think it is better as a separate follow-up PR.

jp39:
Ping

afjcjsbx:
Hi, thank you for the PR,

#### Remote one-shot truncation during "visible silence"

The remote one-shot can truncate a valid response after 5 seconds of "visible silence" and still exit successfully.

* In `remote.go`, the idle timeout is hardcoded to 5s, and it is only reset when `evt.Displayed` is true.
* However, `typing.start/stop` and `thought` payloads are explicitly excluded from refreshing the timeout in `remote.go`.

Consequently, a normal turn that emits an initial chunk and then spends >5s in tool execution or reasoning before the final chunk will be closed prematurely with exit code 0. The newly added test in `remote_test.go` only covers a 2-second gap. I reproduced this issue with a temporary test (first chunk, 6s pause, second chunk), and the command only printed the first one.

#### Remote error events do not cause the one-shot to fail

Remote error events fail to trigger a non-zero exit code in the one-shot execution.

* In `remote.go`, `pico.TypeError` is treated as `Displayed=true`.
* Then, further down in `remote.go`, any subsequent close/read error is considered a success if `seenResponse` is already true.

As a result, if the server rejects the message and only sends an error, the CLI prints `error[...]` but terminates with status 0. This breaks scripts and is inconsistent with the local `picoclaw agent -m` behavior. I confirmed this as well with a temporary test on the PR worktree.

jp39:
Pushed a follow-up fix for the two one-shot issues.

Changes:

- Increased the remote one-shot idle window from `5s` to `30s`.
- One-shot mode now refreshes its timeout on any remote event activity, not only visible output. This keeps it alive through typing/thought/tool-execution gaps before the next visible chunk.
- Remote Pico `error` events now make one-shot mode return an error instead of exiting successfully after printing the error.

Added regression coverage for:

- A response with a 6s gap between visible chunks.
- A server `type=error` event in one-shot mode.

Validation:

```bash
go test ./cmd/picoclaw/internal/agent
go test -run 'TestRemoteOneShot(WaitsThroughVisibleSilence|ReturnsErrorForRemoteErrorEvent|WebSocket)' -timeout=20s ./cmd/picoclaw/internal/agent
go test ./cmd/picoclaw/internal/agent ./pkg/channels/pico
```

jp39:
Ping

jp39:
Ping

PR #3120, opened 2026-06-14
feat(config): add RegisterChannelSettings hook for out-of-tree channels

carlosprados:
Makes PicoClaw extensible for **out-of-tree channels** — channels that live in a third-party module and register via the already-public `channels.RegisterFactory`, without forking PicoClaw.

## Why

The factory side is public (`channels.RegisterFactory`), but the **config side is not**: `InitChannelList` (called from `LoadConfig`) rejects any channel whose `type` is not in the **private** `channelSettingsFactory` map:

```go
if !isValidChannelType(bc.Type) {
    return fmt.Errorf("channel %q has unknown type %q", name, bc.Type)
}
```

So today a custom channel is impossible without editing PicoClaw. This PR closes that asymmetry: register the factory **and** its settings prototype, both from your own `init()`.

## What

One small, additive export (map access now guarded by a `sync.RWMutex`):

```go
func RegisterChannelSettings(channelType string, prototype any)
```

Usage from an out-of-tree channel package:

```go
func init() {
    config.RegisterChannelSettings("my_channel", MyChannelSettings{})
    channels.RegisterFactory("my_channel", newMyChannel)
}
```

## Why it is safe / low-effort to accept

- **Purely additive** — no existing channel or code path changes; built-in channels keep using the same map.
- **Tested** — unit tests cover the hook and the full `InitChannelList -> GetDecoded` path that previously failed with "unknown type".
- **Tiny surface** — one exported function + a mutex.
- `go test ./pkg/config/` passes; `go vet ./pkg/config/` clean.

Happy to adjust the API shape or naming if you prefer a different convention. Thanks for PicoClaw — we are building on it as a library and this is the one seam we are missing.

### DIFF
--- a/pkg/config/config_channel.go
+++ b/pkg/config/config_channel.go
@@ -7,6 +7,7 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
+	"sync"
 
 	"github.com/caarlos0/env/v11"
 	"gopkg.in/yaml.v3"
@@ -655,6 +656,8 @@ func filterSecureFields(r RawNode, secureFields map[string]struct{}) RawNode {
 // channelSettingsFactory maps channel type to a zero-value prototype of the
 // corresponding Settings struct. InitChannelList uses reflect.New to create
 // fresh instances, avoiding repeated closure boilerplate.
+var channelSettingsMu sync.RWMutex
+
 var channelSettingsFactory = map[string]any{
 	ChannelPico:           (PicoSettings{}),
 	ChannelPicoClient:     (PicoClientSettings{}),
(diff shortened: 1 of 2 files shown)


afjcjsbx:
Hi @carlosprados, thank you for the PR, pls accept the CLA

carlosprados:
Thanks @afjcjsbx — CLA is signed and the check is green now. Whenever you have a moment, I'd appreciate a review.

The change is purely additive (one exported function + a mutex), built-in channels are untouched, and tests cover the full `InitChannelList -> GetDecoded` path that previously failed with "unknown type". Happy to adjust the API shape or naming if you prefer a different convention.

afjcjsbx (approved):

PR #3122, opened 2026-06-14
fix(evolution): capture Close() error on write file in appendJSONLRecords

chengzhichao-xydt:
`defer f.Close()` on a file opened with `O_WRONLY|O_APPEND` silently discards the `Close()` error. On write-only files, `Close()` can surface delayed write failures (e.g. disk full, NFS errors).

Changed to a named defer that captures the close error when the function would otherwise return nil.

**Verification:** `go build ./pkg/evolution/...` and `go test ./pkg/evolution/...` pass.


afjcjsbx (approved):
LGTM!

PR #3132, opened 2026-06-15
fix: add panic recovery to core-path goroutines

SiYue-ZO:
## Summary

Add defer-recover protection to goroutines on critical execution paths to prevent a single panicked goroutine from crashing the entire process.

## Motivation

Unprotected goroutines that panic will terminate the whole process. This PR wraps goroutines on core paths — tool execution, event dispatch, HTTP serving, and shell process management — with defer-recover blocks so that a panic in any one of them is caught, logged with a full stack trace, and does not bring down the gateway.

## Changes

| File | Location | What was added |
|------|----------|----------------|
| pkg/tools/toolloop.go | L166 | Recover for parallel tool execution goroutine — sets ErrorResult so the LLM gets a meaningful error instead of nil |
| pkg/channels/manager.go | L1281, L1294 | Recover for HTTP server goroutines (×2) |
| pkg/channels/manager.go | L1945 | Recover for channel registration deferred goroutine |
| pkg/events/subscription.go | L228 | Recover for concurrent event dispatch goroutine |
| pkg/events/subscription.go | L255 | Recover for timeout-handler goroutine — sends fallback handlerResult{panicked: true} to prevent caller deadlock |
| pkg/events/subscription.go | L304 | Recover for watchContext goroutine |
| pkg/tools/shell.go | L418 | Recover for cmd.Wait goroutine — sends error to done channel to unblock caller |
| pkg/tools/shell.go | L584, L606 | Recover for PTY cmd.Wait + PTY read goroutines — sets session.Status = "error" on panic to avoid stale "running" state |
| pkg/tools/shell.go | L645 | Recover for pipe read goroutine |

> pkg/agent/agent.go:218 already has a well-implemented recover mechanism (with session cleanup logic) and was left unchanged.

## Design Decisions

- Deadlock prevention: Goroutines that send on a channel (shell.done, subscription.done) send a fallback value inside recover so the receiver never blocks forever.
- State consistency: PTY cmd.Wait recover sets session.Status = "error" so the session does not get stuck in "running".
- Tool loop: toolloop recover sets ErrorResult so the LLM receives a meaningful error response rather than a nil result.
- Logging style: subscription goroutines use log.Printf (consistent with the existing invokeHandler code in that file); all other packages use the project logger ErrorCF with full stack traces.

## Testing

- [x] go build passes
- [x] go vet passes
- [x] golangci-lint run passes on all changed packages
- [ ] Manual verification: multi-channel concurrent message handling

### DIFF
--- a/pkg/channels/manager.go
+++ b/pkg/channels/manager.go
@@ -13,6 +13,7 @@ import (
 	"math"
 	"net"
 	"net/http"
+	"runtime/debug"
 	"sort"
 	"strings"
 	"sync"
@@ -1279,6 +1280,16 @@ func (m *Manager) StartAll(ctx context.Context) error {
 			for _, listener := range m.httpListeners {
 				ln := listener
 				go func() {
+					defer func() {
+						if r := recover(); r != nil {
+							logger.ErrorCF("channels", "HTTP server goroutine panic recovered",
+								map[string]any{
+									"addr":  ln.Addr().String(),
+									"panic": fmt.Sprintf("%v", r),
+									"stack": string(debug.Stack()),
+								})
+						}
+					}()
 					logger.InfoCF("channels", "Shared HTTP server listening", map[string]any{
 						"addr": ln.Addr().String(),
 					})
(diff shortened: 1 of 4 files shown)


afjcjsbx (approved):
LGTM!

PR #3156, opened 2026-06-22
feat(pico): emit per-turn LLM token usage on finalized message

loafoe:
## Summary

Emit real per-turn LLM token usage on the finalized assistant message over the Pico channel, so downstream consumers can track input/output token consumption per conversation. Input and output tokens are carried separately (they are billed at different rates); `total_tokens` is a convenience sum.

The token counts are the real provider-reported values already captured by the agent loop (`turnState.GetLastUsage()` → `providers.UsageInfo`), sourced from the provider response (verified for the Bedrock Converse provider's streaming and non-streaming paths). Previously these never surfaced on the Pico channel, which only forwarded context-window occupancy (`context_usage`).

## Changes

- `pkg/channels/pico/protocol.go`: add `PayloadKeyUsage = "usage"`.
- `pkg/channels/pico/pico.go`: add `setTurnUsagePayload` helper (mirrors `setContextUsagePayload`), `SetTurnUsage` on `picoStreamer`, and emit the `usage` block on the finalized message. A test-only `broadcastFn` seam allows capturing broadcasts without a live WebSocket.
- `pkg/agent/pipeline_streaming.go`: wire `ts.GetLastUsage()` into the streamer via `SetTurnUsage` before finalize, mirroring the existing `SetModelName` interface-assertion pattern.

Wire shape on the finalized assistant `message.create`/`message.update`:

```json
"usage": { "input_tokens": 1234, "output_tokens": 567, "total_tokens": 1801 }
```

## Compatibility

Fully additive. The `usage` key is only emitted when a count is > 0, and is omitted entirely otherwise — older clients ignore the unknown key, and the change does not alter any existing payload. Usage is emitted exactly once per turn (on the finalized message); interim throttled updates carry no usage.

## Testing

- `TestSetTurnUsagePayload` — helper populates/omits the block correctly.
- `TestStreamerEmitsUsageOnFinalize` — finalized payload carries the usage block.
- Full `pkg/channels/pico` and `pkg/agent` suites pass.

### DIFF
--- a/pkg/agent/pipeline_llm.go
+++ b/pkg/agent/pipeline_llm.go
@@ -495,11 +495,16 @@ func (p *Pipeline) CallLLM(
 		}
 	}
 
-	// Save finishReason to turnState for SubTurn truncation detection
-	if innerTS := turnStateFromContext(ctx); innerTS != nil {
-		innerTS.SetLastFinishReason(exec.response.FinishReason)
+	// Save finishReason and usage on the turn state. Use ts directly (the
+	// authoritative turn state for this call) rather than a context lookup:
+	// the raw ctx passed to CallLLM is not seeded with turnState (only turnCtx
+	// is), so turnStateFromContext(ctx) returns nil here and silently dropped
+	// both the finish reason and the per-turn token usage. ts is also exactly
+	// what the streaming publisher reads via GetLastUsage at finalize.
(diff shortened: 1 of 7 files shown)


afjcjsbx (approved):
LGTM!

PR #3163, opened 2026-06-23
feat(bedrock): leverage Converse prompt caching via cache points

loafoe:
## Summary

AWS Bedrock's Converse API supports [prompt caching](https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html) via explicit **cache points** in `system`, `tools`, and `messages` — everything up to a cache point is cached as a prefix (reads bill at ~0.1× input, writes at ~1.25×).

The agent runtime already produces the breakpoint metadata (`Message.SystemParts` carrying `CacheControl{Type:"ephemeral"}`), and the Anthropic provider already consumes it — but the Bedrock adapter discarded it entirely. This PR has the Bedrock adapter honor that existing convention, behind a model-capability guard.

## Changes

- **System prompt caching** — `convertMessages` now reads structured `SystemParts`: each part becomes its own `SystemContentBlockMemberText`, and a cache point is inserted **after the last `ephemeral` part**. The stable prefix (identity, tool docs, skills, memory) is cached; volatile trailing parts (time, session, summary) stay outside the cached region. If no part is `ephemeral`, no cache point is emitted — so per-request dynamic context never anchors the cache.
- **Tool-schema caching** — a cache point is appended after the tool list (large, identical every turn). Omitted when no tools survive name validation.
- **Cache-usage reporting** — new additive `CacheReadTokens` / `CacheWriteTokens` fields on `UsageInfo`, populated from Bedrock's `CacheReadInputTokens` / `CacheWriteInputTokens` in both `Converse` and `ConverseStream`. Other providers leave them zero.

## Capability guard

`modelSupportsCaching` gates everything to cache-enabled Claude families on Bedrock (3.5 v2, 3.7, 4.x). Unsupported and non-Claude models keep the original behavior with zero cache points, avoiding `ValidationException`. Max 2 cache points per request (system + tools), well under Bedrock's limit of 4. Caching is fully opt-in by model and changes nothing for existing configurations.

## Testing

`go test -tags bedrock ./pkg/providers/bedrock/` — existing suite plus 11 new tests covering the capability guard, system/tool cache-point placement, the no-ephemeral / disabled / empty cases, and cache-usage parsing for both streaming and non-streaming paths.

### DIFF
--- a/pkg/providers/bedrock/provider_bedrock.go
+++ b/pkg/providers/bedrock/provider_bedrock.go
@@ -26,6 +26,7 @@ import (
 	"github.com/aws/aws-sdk-go-v2/service/bedrockruntime/document"
 	"github.com/aws/aws-sdk-go-v2/service/bedrockruntime/types"
 
+	"github.com/sipeed/picoclaw/pkg/logger"
 	"github.com/sipeed/picoclaw/pkg/providers/common"
 	"github.com/sipeed/picoclaw/pkg/providers/protocoltypes"
 )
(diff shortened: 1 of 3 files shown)


loafoe:
@afjcjsbx please consider this one for merging, the original Bedrock support did not support prompt caching even though picoclaw already has all of the plumbing in place. This PR wires up the bedrock cache point API resulting in significant cost saving!

loafoe:
@afjcjsbx rebased this PR so all tests pass

loafoe:
@afjcjsbx rebased against latest upstream, anything blocking this from being merged?

afjcjsbx:
Hi, there is the Security Check step that is failing, I created a PR to fix it: https://github.com/sipeed/picoclaw/pull/3248 but no one gives me the approval to merge it, I am not sure but it would seem that this project is no longer maintained

PR #3165, opened 2026-06-24
fix(openai_compat): recover Seed XML tool calls

### LINKED ISSUE #3153: [BUG] Volcengine Doubao Seed tool calls occasionally leak as <seed:tool_call> text
## Quick Summary

When using PicoClaw v0.2.8 with Volcengine Coding Plan and `doubao-seed-2.0-pro`, tool calls are sometimes returned to the user as raw `<seed:tool_call>` text instead of being executed.

Example output:

```xml
<seed:tool_call>
  <function name="exec">
    <parameter name="action" string="true">run</parameter>
    <parameter name="command" string="true">ls -la /root/.picoclaw/workspace/library/skills/rss_filter/temp/虎嗅网/</parameter>
  </function>
</seed:tool_call>
```

This started happening recently while I was still running PicoClaw v0.2.8. The timing was after the v0.2.9 release, but I am not claiming that v0.2.9 caused it.

## Environment & Tools

* PicoClaw Version: v0.2.8
* Installation method: prebuilt release binary
* Go Version: N/A
* AI Model & Provider: Doubao Seed 2.0 Pro via Volcengine Coding Plan
* API Base: `https://ark.cn-beijing.volces.com/api/coding/v3`
* Operating System: `Debian 12`
* Architecture: `x86_64`
* Channels: `Feishu `

Relevant model configuration:

```json
{
  "model_name": "Doubao-Seed-2.0-pro",
  "provider": "",
  "model": "volcengine/Doubao-Seed-2.0-pro",
  "api_base": "https://ark.cn-beijing.volces.com/api/coding/v3",
  "enabled": true
}
```

No API key is included above.

## Steps to Reproduce

1. Configure PicoClaw v0.2.8 to use Volcengine Coding Plan with `doubao-seed-2.0-pro`.
2. Start the PicoClaw gateway.
3. Run a skill that requires multiple tool calls. In my case, this occurs while running the `rss_filter` skill.
4. Let the agent invoke the `exec` tool, for example to list a temporary RSS directory.
5. Occasionally, the assistant sends the raw `<seed:tool_call>` XML to the chat instead of executing the tool.

The problem appears intermittently and may be more likely in a long or tool-heavy conversation.

## ❌ Actual Behavior

Raw Seed tool-call markup is displayed in the chat:

```xml
<seed:tool_call>
  <function name="exec">
    <parameter name="action" string="true">run</parameter>
    <parameter name="command" string="true">ls -la ...</parameter>
  </function>
</seed:tool_call>
```

The intended tool call is not executed in that turn.

Clearing the conversation or restarting the gateway may temporarily help, but the problem can occur again.

## ✅ Expected Behavior

PicoClaw should receive and process the tool call as a structured tool call, execute the `exec` tool, and never expose internal `<seed:tool_call>` markup to the user.

If the provider unexpectedly places Seed tool-call markup inside `message.content`, PicoClaw could ideally detect it, prevent it from being displayed, and either parse it safely or return a clear provider compatibility error.

## Additional Context

I tested the same Volcengine endpoint directly with a non-streaming OpenAI-compatible request containing an `exec` tool definition.

The endpoint returned a standard structured tool call:

```json
{
  "finish_reason": "tool_calls",
  "content": "",
  "tool_calls": [
    {
      "function": {
        "arguments": "{\"action\": \"execute\", \"command\": \"pwd\"}",
        "name": "exec"
      },
      "id": "call_mpnvitb4x5bdskxa7v2wdtdu",
      "type": "function"
    }
  ],
  "error": null
}
```

This indicates that the endpoint is capable of returning standard OpenAI-compatible `tool_calls`.

The direct test used a simplified tool schema, so the returned `"action": "execute"` is not relevant to the original issue. In the actual leaked XML, the action was correctly generated as `"run"`.

Possible areas to investigate:

* Whether tools are omitted from some agent iterations.
* Handling of tool calls in long or tool-heavy conversations.
* Conversation history containing earlier raw `<seed:tool_call>` messages.
* Streaming or response parsing behavior.
* Fallback handling when Seed tool-call markup appears in `message.content`.

Please let me know which debug logs would be useful, and I can try to collect them.

Alix-007:
## Summary
- recover Volcengine Doubao Seed `<seed:tool_call>` XML blocks from OpenAI-compatible response content as structured tool calls
- strip recovered Seed tool-call XML from user-visible content
- suppress leaked Seed XML from streaming chunks while preserving final tool-call extraction

## Verification
- `go test ./pkg/providers/openai_compat -run 'Seed|ToolCalls'`
- `go test ./pkg/providers/openai_compat`
- `git diff --check`

Fixes #3153

### DIFF
--- a/pkg/providers/openai_compat/provider.go
+++ b/pkg/providers/openai_compat/provider.go
@@ -631,6 +631,7 @@ func parseStreamResponse(
 	var reasoningDetails []ReasoningDetail
 	var finishReason string
 	var usage *UsageInfo
+	seedToolCallStreamingContent := false
 
 	// Tool call assembly: OpenAI streams tool calls as incremental deltas
 	type toolAccum struct {
@@ -703,7 +704,14 @@ func parseStreamResponse(
 		if choice.Delta.Content != "" {
 			textContent.WriteString(choice.Delta.Content)
 			if onChunk != nil {
-				onChunk(StreamChunk{Content: textContent.String()})
+				if visible, seedStarted := stripSeedToolCallStart(textContent.String()); !seedStarted {
+					onChunk(StreamChunk{Content: visible})
+				} else if !seedToolCallStreamingContent && strings.TrimSpace(visible) != "" {
+					onChunk(StreamChunk{Content: visible})
+					seedToolCallStreamingContent = true
+				} else {
+					seedToolCallStreamingContent = true
+				}
 			}
 		}
 
@@ -892,6 +900,14 @@ func parseSeedToolCallsFromContent(content string) ([]ToolCall, string) {
 	}
 
 	return toolCalls, strings.TrimSpace(sb.String())
+}
+
+func stripSeedToolCallStart(content string) (string, bool) {
+	start := strings.Index(content, "<seed")
+	if start < 0 {
+		return content, false
+	}
+	return strings.TrimSpace(content[:start]), true
 }
 
 // xmlSeedFunction mirrors the inner XML structure of a <seed:tool_call> block.
--- a/pkg/providers/openai_compat/provider.go
+++ b/pkg/providers/openai_compat/provider.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"encoding/xml"
 	"fmt"
 	"io"
 	"maps"
@@ -160,9 +161,11 @@ func (p *Provider) buildRequestBody(
 		// treat it as false — web_search_preview must not be injected
 		// when the caller cannot express a well-typed intent.
 		if _, present := options["native_search"]; present {
-			log.Printf(
-				"[openai_compat] native_search option has unexpected type %T, ignoring",
-				options["native_search"],
+			logger.WarnCF("provider.openai_compat",
+				"native_search option has unexpected type, ignoring",
+				map[string]any{
+					"type": fmt.Sprintf("%T", options["native_search"]),
+				},
 			)
 		}
 	}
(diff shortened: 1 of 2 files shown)


afjcjsbx:
Thanks for the PR! There are two critical edge cases regarding the stripping logic and content cleanup that we need to address:

* **Non-robust SSE Stream Stripping (`pkg/providers/openai_compat/provider.go`):** The streaming stripping logic currently relies on finding the first `<seed` in the accumulated buffer and stops emitting content from that point onward. This approach is prone to bugs at SSE delta boundaries:
1. If the opening tag is split across two deltas (e.g., `<se` followed by `ed:tool_call>`), the first chunk will leak partial XML, and the subsequent one will re-emit the prefix.
2. If there is visible text *after* `</seed:tool_call>`, that suffix will never be streamed because `seedToolCallStreamingContent` does not reset, causing `stripSeedToolCallStart` to always return only the prefix before the first match.
*Action:* We need to update the streaming parser to properly handle chunk boundaries and reset its state. Let's also expand the test suite, as the current tests only cover the "tool call at the very end" scenario.


* **Content Leakage in Hybrid Cases (`pkg/providers/openai_compat/provider.go`):**
`extractSeedToolCalls` early-returns as soon as `resp.ToolCalls` is already populated. In hybrid scenarios—where the provider sends both standard `tool_calls` and leaked XML inside `message.content`—the user will still see the raw `<seed:tool_call>...</seed:tool_call>` block in the final content.
*Action:* This contradicts the goal of consistently cleaning up visible content and diverges from the streaming path behavior. The cleanup of `Content` should run unconditionally; only the merging/overriding of `ToolCalls` should be conditional.

PR #3180, opened 2026-06-26
fix(cli): skip tool calls with invalid arguments

Alix-007:
## Summary
- skip CLI-emitted tool calls when `function.arguments` is not valid JSON
- keep valid tool calls from the same response instead of dropping the whole batch
- add regression coverage for invalid and mixed valid/invalid CLI tool-call arguments

Without this check, malformed CLI tool-call arguments were parsed as an empty argument map, which could let an invalid tool call proceed with missing inputs.

## Test plan
- `go test ./pkg/providers/cli`

### DIFF
--- a/pkg/providers/cli/claude_cli_provider_test.go
+++ b/pkg/providers/cli/claude_cli_provider_test.go
@@ -789,6 +789,32 @@ func TestExtractToolCalls_InvalidJSON(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


afjcjsbx:
Hi, thanks for the PR, 

it might be helpful to add a log when the call to the tool is skipped

yofabr (approved):

PR #3191, opened 2026-06-27
chore: remove duplicate build/ entry in .gitignore

chengzhichao-xydt:
## Description

Remove duplicate `build/` entry in `.gitignore`. The pattern appears on both line 4 and line 21.

## Type of Change
- [x] Chore (config cleanup)


afjcjsbx:
Hi, it seems that the PR also touches files not strictly related to the description.

PR #3200, opened 2026-07-01
feat(models): add configurable default fallback chain

lc6464:
## 📝 Description

Add a configurable default fallback chain for models in the web UI and persist it through the backend API.

This change introduces a dedicated default-chain workflow on the models page so users can set the default model, add fallback models, reorder the chain, and save the full draft in one write. Unsaved chain edits are rebased across model changes, model and chain writes are coordinated, and duplicate aliases are validated as a group before users can select them.

The backend now exposes default-chain read/write endpoints, validates alias and raw model references consistently, keeps default-chain references in sync when models are renamed or deleted, normalizes legacy references without changing their provider semantics, and treats fallback order changes as gateway-restart-sensitive configuration updates. Runtime fallback resolution preserves the exact model configuration and provider identity across duplicate aliases and model-list reordering, while keeping model switching, media routing, native search, and provider cleanup consistent with the active candidate chain.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 📚 Technical Context
- **Reference URL:** N/A
- **Reasoning:** The default model fallback chain already existed in config, but it could only be managed through JSON. This change makes the chain manageable from the frontend while preserving unsaved user edits, enforcing the same alias/raw-reference rules across UI, API, and runtime, and keeping candidate provider identity stable. It also ensures fallback ordering affects gateway restart detection and keeps default-chain references aligned with model add/rename/delete flows.

## 🧪 Test Environment
- **OS:** Ubuntu Server 24
- **Model/Provider:** -
- **Channels:** -

## 📸 Evidence
<details>

image

</details>

## ☑️ Checklist


imguoguo (commented):
Verified against PR head `bb5eb226` in an isolated worktree: `go vet ./web/backend/api/` is clean and the 10 new/related backend tests pass.

## Overview

This adds UI + API for managing `agents.defaults.model_name` and `agents.defaults.model_fallbacks` as one "default chain":

- **Backend** — `GET`/`PUT /api/models/default-chain`; a shared validator (`validateDefaultChainModelSet` / `validateDefaultModelChain`) now backs both the new endpoint and the existing `POST /api/models/default`; rename/delete of a model syncs default-chain references; `Handler.configMu` serializes the four model write handlers; `computeModelStreamingSignatures` now encodes chain *position* so reordering triggers restart-required.
- **Frontend** — draft/baseline state on the models page with a sticky save/reset bar, a reorder dialog, and per-card "add to fallback chain".

The design is sound and consolidating the validator is a genuine improvement — the old `handleSetDefaultModel` had three separate loops over `ModelList`. Test coverage for the new backend paths is good (rename, delete, duplicate alias, unsupported provider, restart detection).

---

## Blocking issues

### 1. `POST /api/models/default` bypasses the new invariant and persists a state the new endpoint rejects

`handleUpdateDefaultChain` enforces *"default model cannot also appear in fallback_chain"*, but `handleSetDefaultModel` (`web/backend/api/models.go`) still only assigns `cfg.Agents.Defaults.ModelName` and never touches `ModelFallbacks`. That endpoint is still live and still used by three frontend callers — `hooks/use-chat-models.ts:65` (chat page model switcher), `add-model-sheet.tsx:409`, `edit-model-sheet.tsx:392`.

Confirmed by probe against this PR head:

```
POST /api/models/default {"model_name":"backup-a"}  -> 200
RESULT default="backup-a" fallbacks=["backup-a"]
PUT  /api/models/default-chain (that same state)    -> 400
     Validation error: default model "backup-a" cannot also appear in fallback_chain
```

Two consequences: at runtime the first fallback becomes a no-op retry of the model that just failed; and the models page is wedged at 400 until the user happens to open the dialog and remove the entry. The fix is one line — mirror the frontend's draft behaviour in `handleSetDefaultModel`:

```go
cfg.Agents.Defaults.ModelName = req.ModelName
cfg.Agents.Defaults.ModelFallbacks = removeModelName(cfg.Agents.Defaults.ModelFallbacks, req.ModelName)
```

### 2. Unsaved chain drafts are silently discarded by unrelated model actions

`fetchModels` unconditionally overwrites all four draft/baseline states from the server response (`models-page.tsx:82-85`), and it is wired to `EditModelSheet.onSaved`, `AddModelSheet.onSaved`, `DeleteModelDialog.onDeleted` and `CatalogDialog.onModelAdded` (lines 362, 369, 377, 383). So with the sticky "you have unsaved changes" bar on screen, editing an unrelated model's API key wipes the user's reordered chain with no warning.

Either skip the draft reset while `chainDirty`, or split the model refresh from the chain refresh so the two aren't coupled.

---

## Should fix before merge

### 3. `configMu` covers 4 of 13 config writers — false sense of safety

`router.go` adds `configMu sync.Mutex` and `models.go` takes it in add/update/delete/set-default. But `config.go` (x2), `tools.go` (x2), `pico.go` (x2), `oauth.go`, `wecom.go` and `weixin.go` all do the same unsynchronized `LoadConfig -> mutate -> SaveConfig` on the same file. A concurrent `PUT /api/config` still clobbers a model write exactly as before. Either hoist this to a handler-wide config write lock used by every writer, or add a comment stating the scope so the next reader doesn't assume config writes are serialized.

`handleGetDefaultChain` also reads without the lock — lower risk, but inconsistent with the write path it pairs with.

### 4. Model cards mix draft state and server state

`model-card.tsx` uses `isDraftDefault` for the set-default/fallback buttons (lines 60, 79, 197-223) but `model.is_default` for the badge (116, 129), the highlight (103), `canSetDefault` (47) and `deleteDisabled` (87). After drafting a new default, the *old* model keeps the "Default" badge and stays undeletable while the *new* draft default is deletable — and deleting it triggers issue #2 on top. Please pick one source of truth for the card, or render an explicit "pending" treatment for the draft default.

### 5. `summary_plural` never fires — wrong plural syntax for i18next v26

`package.json` pins `i18next: ^26.0.10` and `src/i18n/index.ts` does **not** set `compatibilityJSON: "v3"`, so JSON v4 plural resolution applies: the keys must be `summary_one` / `summary_other`. `summary_plural` is v3 syntax and is dead — English renders "gpt-4o with 3 fallback" for every count.

```jsonc
"summary_one":   "{{model}} with {{count}} fallback",
"summary_other": "{{model}} with {{count}} fallbacks",
```

Note the repo already has this bug in `models.fetch.found_plural` / `fill_plural` — worth a separate cleanup, but let's not add a third instance.

### 6. Only 2 of 5 locales updated

The repo maintains `en`, `zh`, `pt-br`, `bn-in` and `cs`, and all five already carry `models.action.setDefault`. This PR touches only `en.json` and `zh.json`, so three locales will show raw English for the whole `defaultChain` block.

---

## Nits

- **Brittle error classification** — `handleSetDefaultModel` picks 404 vs 400 via `strings.Contains(err.Error(), "not found in model_list")`. Now that validation is centralized, a sentinel (`var errModelNotFound = errors.New(...)` + `errors.Is`) is cheap and won't break when someone rewords the message.
- **Request type reuse** — `handleUpdateDefaultChain` decodes the body into `defaultChainResponse`. A distinct `defaultChainRequest` costs nothing and documents intent.
- **`validateDefaultChainModelSet(..., defaultContext bool)`** — a boolean that swaps six error strings. A small `chainRole` string type (`"default"` / `"fallback"`) reads better at the three call sites.
- **Dead i18n key** — `models.defaultChain.fallbackCount` is added to both locales but never referenced.
- **Leftover async scaffolding** — `handleSetDefault` is now purely local state, yet it is still `async` and still drives `settingDefaultIndex` through a bare `setTimeout(..., 150)` to flash a spinner for a request that no longer happens. Drop both, or make the pending state meaningful.
- **Gateway signature position is a flat global counter** — `computeModelStreamingSignatures` numbers positions across `defaults.primary -> defaults.fallbacks -> routing.light -> every agent's primary/fallbacks`. Reordering `agents.list` with no model change now flips `restart_required`. A scoped key (`"defaults.fb:1"`, `"agent2.fb:0"`) gives the same ordering sensitivity without the cross-scope coupling, and is more readable in a debug dump.
- **Rename sync stops at defaults** — `handleUpdateModel` rewrites `Defaults.ModelName`/`ModelFallbacks` but not `cfg.Agents.List[].Model.Primary`/`.Fallbacks`, which reference the same names (see `gateway.go:516-517`). Pre-existing, and arguably out of scope, but this PR's restart detection reads exactly those fields, so the asymmetry is now visible. Same for `handleDeleteModel`.
- **Duplicate aliases + rename** — this PR's own tests establish that two `ModelList` entries may share a `ModelName`. If one is renamed, `replaceModelName` repoints the default/fallback reference even though the *other* entry still legitimately provides the old alias. Guard with "only rewrite if no remaining entry carries `oldModelName`".

---

## Verdict

Solid feature with real test coverage. **#1 and #2 are user-visible correctness bugs and should be fixed before merge**; #3-#6 are small and worth folding into the same round.

imguoguo (commented):
Re-reviewed at `2a92268c`. Verified in an isolated worktree: `go vet` clean, `go test ./web/backend/api/` fully green (17.9s), and `tsc -b` reports nothing on the files this PR touches.

Thanks for the thorough follow-up — this round is solid.

## Status of the previous round

| # | Issue | Status |
|---|---|---|
| 1 | `POST /api/models/default` bypassed the invariant | Fixed — `removeModelName` added, plus `TestHandleSetDefaultModel_RemovesDefaultFromFallbackChain` |
| 2 | Unsaved drafts silently discarded | Fixed — `refreshModels` / `refreshAfterModelChange` split; dirty state now reconciles the draft instead of resetting it |
| 3 | `configMu` covered 4 of 13 writers | Scope documented rather than widened — acceptable (see nit below on the wording) |
| 4 | Cards mixed draft and server state | Fixed — `isDraftDefault` -> `isDefault` throughout; badge, highlight and delete guard all aligned, and the delete dialog now takes `isDefaultModel` |
| 5 | `summary_plural` wrong plural syntax | Fixed — `summary_one` / `summary_other` |
| 6 | Only 2 of 5 locales updated | Fixed — pt-br, bn-in and cs all filled in |

The nits landed too: dedicated `defaultChainRequest`, scoped signature keys (`defaults.fallback:1`, `agent:2.primary`), the duplicate-alias rename guard via `modelConfigsByNameExceptIndex`, the fake `setTimeout` spinner removed, and the dead `fallbackCount` key dropped. The test refactor (`writeModelChainConfig` / `assertDefaultChain` + six new alias edge cases) is a nice improvement on its own.

## One remaining instance of the same class of bug: `POST /api/models`

`handleAddModel` is the only one of the four model write handlers that does **not** call `normalizeDefaultChainReferences`, and it does not check alias uniqueness. So adding a model that reuses an existing alias on an ASR-only provider still drives the persisted default chain into a state the new validator rejects — the same shape as issue #1 from the last round.

Probed against this head:

```
POST /api/models {"model_name":"primary","provider":"elevenlabs","model":"scribe_v1"}
  -> 200 {"index":2,"status":"ok"}
STORED  default="primary" fallbacks=["backup-b"]
PUT  /api/models/default-chain {"default_model":"primary","fallback_chain":["backup-b"]}
  -> 400 Validation error: model "primary" cannot be used as the default chat model
GET  /api/models/default-chain
  -> 200 {"default_model":"primary","fallback_chain":["backup-b"]}   <- still advertised as valid
```

`add-model-sheet.tsx:185` does guard against duplicates client-side, but `catalog-dialog.tsx:135` calls `addModel({model_name: model.id, ...})` with no such check, and the API does not enforce it either. The fix is to match the other three handlers — one `normalizeDefaultChainReferences(cfg)` before `SaveConfig`.

## Worth a product decision

**An invalid default model now silently wipes the entire fallback chain.** `normalizeDefaultChainReferences` sets `ModelFallbacks = nil` whenever the default fails validation, and `reconcileDefaultChainDraft` mirrors that on the frontend. Editing your default model onto e.g. an ElevenLabs provider therefore discards the whole carefully ordered chain with no warning.

This is self-consistent with `validateDefaultModelChain`'s "default_model is required when fallback_chain is not empty" rule — but it is worth questioning that rule itself: why should losing the default also forfeit the fallback list? If the rule stays, the UI should at least say so explicitly rather than dropping the data silently.

## Nits

- **`handleAddModel` accepts duplicate aliases** — same root as above. `validateIncomingModelConfig` does not check `ModelName` uniqueness, and the only dedup lives in the add sheet. Duplicate aliases are a supported config shape (this PR's own tests rely on it), but the catalog bulk-add path deserves a server-side backstop.
- **Dead key** — `models.action.setDefaultDisabled.setting` went away with `settingDefault` but is still present in all five locale files.
- **Duplicated predicate** — `model-card.tsx:44-45`: `canSetDefault` and `canUseInFallback` are now character-for-character identical, and `!isDefault` is redundant in `canSetDefault` (when `isDefault`, that button is replaced by a static star via the ternary). Collapsing them into one `canUseInDefaultChain` would read better.
- **Misleading comment** — `router.go`: "Other config endpoints still coordinate their own load-modify-save cycles" reads as though the other endpoints have some coordination of their own; they have none. Suggest stating plainly that their load-modify-save cycles are not covered by this lock.
- **`strings.Contains(err.Error(), "not found in model_list")` (models.go:828)** — still there from the last round. Now that validation is centralized, a sentinel + `errors.Is` is cheap, and it won't silently degrade to 400 when someone rewords lines 321/323.
- **`refreshAfterModelChange`'s non-dirty branch duplicates `fetchModels`** almost line for line (~12 lines, differing only in `setLoading`). Worth a parameter or straight reuse.
- **Reordering `agents.list` still flips `restart_required`** — the scope key is `agent:<index>`, so positional sensitivity remains. Probably fine given agent order may carry meaning; noting it for the record.

## Verdict

Both blockers and all four secondary items from the last round are addressed, with tests to match. **One same-class hole remains in `handleAddModel` (a one-line fix)**, and the "invalid default wipes the whole chain" behaviour deserves an explicit product call. Everything else is optional cleanup. LGTM once the add path is covered.

lxowalle (commented):
## Review verdict: Request changes

  I found three user-visible correctness issues that should be addressed before merging.

  ### 1. Valid raw fallback references are silently removed

  The runtime supports fallback references such as `openrouter/fallback-model` even when they are not registered as a `model_name` in `model_list`. However, `normalizeDefaultChainReferences` now treats such
  references as invalid.

  Reproduction:

  1. Configure `model_fallbacks: ["openrouter/fallback-model"]`.
  2. Add an unrelated model through `POST /api/models`.
  3. The request succeeds, but the persisted fallback chain becomes `nil`.

  Add, update, delete, and set-default operations all invoke this normalization, so an unrelated model operation can destroy a valid existing fallback configuration. `PUT /api/models/default-chain` also rejects
  the same runtime-supported reference.

  The validator should follow the runtime resolution rules and preserve valid raw `provider/model` references.

  ### 2. Reordering raw fallbacks does not trigger restart-required

  `computeModelStreamingSignatures` only emits an entry when a reference matches `model_list`. Consequently, reordering two valid raw fallback references produces the same configuration signature, and
  `gateway_restart_required` remains false.

  Each scoped fallback reference should contribute to the signature even when no `model_list` entry matches it. Matched model configuration details can then be appended when available.

  ### 3. Dirty drafts retain a stale persisted baseline

  When `chainDirty` is true, `refreshAfterModelChange` refreshes only the model list and draft state. It does not reload the persisted chain or update the baseline.

  For example:

  1. The saved chain is default `A`, fallback `B`.
  2. Add `C` to the unsaved draft.
  3. Delete `B`; the backend persists default `A` with no `B`.
  4. Click Reset.

  The UI restores the deleted `B` from the stale baseline, hides the save bar, and incorrectly presents that state as saved. Renaming or changing the provider of a referenced model has the same problem.

  The persisted baseline should be refreshed after model writes while the unsaved draft is reconciled separately.

lxowalle:
## Blocking: `/switch model` drops primary-provider inheritance for fallbacks

The cold-start path now lets an uncredentialed fallback on the same protocol inherit the active primary provider. However, the provider map rebuilt by `SwitchModel` skips that inheritance step and calls `populateCandidateProvidersFromCandidates` directly:

https://github.com/sipeed/picoclaw/blob/5996e7718331d27c9091a6755cd93b7d46fa3faf/pkg/agent/agent_command.go#L330-L346

This makes fallback behavior depend on whether the user has switched models during the current process. For example, with two credentialed OpenAI primary aliases and an OpenAI fallback entry that relies on the primary credentials:

```yaml
agents:
  defaults:
    model_name: first
    model_fallbacks: [fallback]

model_list:
  - model_name: first
    provider: openai
    model: first
    api_key: sk-first
  - model_name: second
    provider: openai
    model: second
    api_key: sk-second
  - model_name: fallback
    provider: openai
    model: fallback
```

At startup, `fallback` correctly shares the active primary provider through `inheritPrimaryProviderForCandidates`. After `/switch model to second`, the rebuilt fallback provider instead fails initialization with:

```text
api_key or api_base is required for HTTP-based protocol "openai"
```

and is stored as an `unavailableLLMProvider`, so the fallback can no longer serve requests.

The switch path should mirror the initialization path by inheriting from `nextProvider` before populating the remaining candidates:

```go
inheritPrimaryProviderForCandidates(
    cfg,
    agent.Workspace,
    nextCandidates[0],
    nextCandidates[1:],
    nextProvider,
    nextCandidateProviders,
)
populateCandidateProvidersFromCandidates(
    cfg,
    agent.Workspace,
    nextCandidates[1:],
    nextCandidateProviders,
)
```

lxowalle (approved):

PR #3225, opened 2026-07-04
Support agent-specific runtime overrides

xdatafactor:
Summary:
- allow agents.list entries to define max_tokens, summarization thresholds, and split_on_marker
- apply those per-agent overrides when building AgentInstance
- remove an unused openai_compat import from current main so the touched agent package can compile

Tests:
- go test ./pkg/config
- go test -tags goolm,stdjson ./pkg/agent -run TestNewAgentInstance_UsesAgentOverrides (interrupted after hanging without output after compilation started)

### DIFF
--- a/pkg/agent/instance.go
+++ b/pkg/agent/instance.go
@@ -130,12 +130,17 @@ func NewAgentInstance(
 	sessions := initSessionStore(sessionsDir)
 
 	mcpDiscoveryActive := agentHasDiscoverableMCPServers(cfg, agentMCPServerAllowlist)
+	splitOnMarker := defaults.SplitOnMarker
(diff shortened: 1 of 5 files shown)


afjcjsbx:
Thanks for the PR! While this correctly saves the `summarize_message_threshold` and `summarize_token_percent` overrides on each `AgentInstance`, these values aren't actually being used for non-default agents in the default runtime.
The `legacyContextManager` still relies on `GetDefaultAgent()` to read the history/summary and evaluate if summarization is needed. As a result, a session routed to `support` will incorrectly fall back to using the store and thresholds of the `main` agent.

Also pls accept the CLA 🙏

PR #3226, opened 2026-07-05
fix(tools): stop write_file from coaching destructive overwrite (#3150)

### LINKED ISSUE #3150: [BUG]它给自己整失忆了
##  Environment & Tools
- **PicoClaw Version:** (e.g., v0.1.2 or commit hash)

## 📸 Steps to Reproduce
1. 我建议pico弄个文件夹放外部文件避免污染workspace，然后将data目录的使用规则写入memory
2. 然后pico直接完整覆盖了memory
3. 

Image

ACMYuechen:
## 📝 Description

picoclaw has no dedicated memory-write tool — the agent updates
`memory/MEMORY.md` by pointing the generic file tools at it. `write_file`
framed an existing file purely as "replace", and its overwrite guard even
coached the model toward it:

  `file: … already exists. Set overwrite=true to replace.`

It never mentioned the non-destructive `append_file` / `edit_file` tools, so
the model set `overwrite=true` and wiped earlier memories.

This PR rewords the three texts the model reads at that decision point — the
`write_file` description, the `overwrite` parameter description, and the
overwrite-guard error — to make clear `overwrite=true` replaces the **entire**
file and to steer toward `append_file` / `edit_file` for incremental,
non-destructive updates.

The steer is **conditional on availability**: `append_file` and `edit_file`
register independently — each behind its own enable flag and the per-agent
tool allowlist — so the copy names only the alternatives actually present. A
config that exposes only `write_file` is no longer directed at tools it lacks;
when neither is available the overwrite guard still blocks, just without naming
a tool. Guidance strings only; it applies to every file (not just MEMORY.md)
and leaves the existing overwrite guard in place as a backstop.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue

Fixes #3150

- **Reference URL:** —
- **Reasoning:** The root cause is at the tool layer, not the system/identity
  prompt. When `write_file` is blocked on an existing file, the guard error the
  model reads literally told it how to overwrite (`Set overwrite=true to
  replace`) and omitted the safe alternatives — so the model retried with
  `overwrite=true` and clobbered the file. Rewording the guard + descriptions
  corrects the model at the moment of the decision, for all files, with no
  functional/API change beyond the guidance strings.

  Because `append_file` / `edit_file` are gated by separate enable flags and
  the per-agent allowlist, the guidance is resolved from **actual** availability
  rather than hardcoded: the editors are registered before `write_file`, and
  availability is read via `registry.HasRegistered(...)`, which reflects both
  the enable flag and the allowlist. This keeps the copy correct for restrictive
  agents that expose only `write_file` (it no longer points them at missing
  tools). The overwrite guard (which blocks replacing an existing file unless
  `overwrite=true`) stays as a second layer.

  Covered by tool-level tests (conditional copy for none / one / both
  alternatives) and agent-wiring tests across the enable-flag and allowlist
  dimensions.

## 🧪 Test Environment
- **Hardware:** MacBook Pro (Mac17,2), Apple M5, 24 GB
- **OS:** macOS (Darwin 25.3.0, arm64)
- **Model/Provider:** DeepSeek
- **Channels:** -

## 📸 Evidence (Optional)
<details>

image
The overwrite guard successfully guides picoclaw to use append_file and edit_file.
</details>

## ☑️ Checklist

### DIFF
--- a/pkg/agent/instance.go
+++ b/pkg/agent/instance.go
@@ -103,8 +103,25 @@ func NewAgentInstance(
 			toolsRegistry.Register(tools.NewReadFileBytesTool(workspace, readRestrict, maxReadFileSize, allowReadPaths))
 		}
 	}
+	if cfg.Tools.IsToolEnabled("edit_file") {
+		toolsRegistry.Register(tools.NewEditFileTool(workspace, restrict, allowWritePaths))
+	}
+	if cfg.Tools.IsToolEnabled("append_file") {
+		toolsRegistry.Register(tools.NewAppendFileTool(workspace, restrict, allowWritePaths))
+	}
+	// Build write_file's copy from the registered editors so it steers the agent
+	// to edit_file/append_file only when those tools are actually available.
 	if cfg.Tools.IsToolEnabled("write_file") {
-		toolsRegistry.Register(tools.NewWriteFileTool(workspace, restrict, allowWritePaths))
+		writeTool := tools.NewWriteFileTool(workspace, restrict, allowWritePaths)
+		var altTools []string
+		if toolsRegistry.HasRegistered("append_file") {
+			altTools = append(altTools, "append_file")
+		}
+		if toolsRegistry.HasRegistered("edit_file") {
+			altTools = append(altTools, "edit_file")
+		}
+		writeTool.SetAlternativeTools(altTools)
+		toolsRegistry.Register(writeTool)
 	}
 	if cfg.Tools.IsToolEnabled("list_dir") {
 		toolsRegistry.Register(tools.NewListDirTool(workspace, readRestrict, allowReadPaths))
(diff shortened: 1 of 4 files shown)


afjcjsbx:
Hi, thanks for the PR, just a hint:

`write_file` now always recommends `append_file` and `edit_file`, but those tools are not guaranteed to be available in the same context.

In `pkg/tools/fs/filesystem.go,` the new description, the `overwrite` parameter text, and the guard error all point to `append_file`/`edit_file`. However, in `pkg/agent/instance.go`, the three tools are registered via separate flags.

Consequently, an agent/config that only exposes `write_file` will end up receiving instructions directing it to tools it doesn't actually have, turning the guardrail into a dead end precisely in the most restrictive setups.

I would suggest making the copy conditional based on the actual availability of the tools, or making `append_file` and `edit_file` mandatory dependencies of `write_file`.

ACMYuechen:
Thanks for the review and the clear explanation — that makes sense. I'll work on it and update the PR. 🙏

ACMYuechen:
Thanks for the review! I went with your first suggestion — the write_file copy (description, overwrite param, and the guard error) is now built from the tools actually registered, so it never points at append_file/edit_file when they aren't available. Added tests for the write_file-only case too.

@afjcjsbx  whenever you have a moment, could you take another look? Thanks!

afjcjsbx (approved):
LGTM!

PR #3227, opened 2026-07-06
fix(providers): resolve tool_use name/args from Function on reloaded history

AayushGupta16:
## 📝 Description

Both Anthropic providers (`anthropic_messages` and the SDK-based `anthropic`) emit assistant `tool_use` blocks from `ToolCall.Name` / `ToolCall.Arguments`. Those fields are runtime-only (`json:"-"` in `protocoltypes.ToolCall`), so after chat history round-trips through the session store only `ToolCall.Function` survives. The providers then silently skip every historical tool call while still emitting the matching `tool_result` in the next user message — and the API rejects the request:

```
bad request (400): unexpected `tool_use_id` found in `tool_result` blocks: toolu_01WqUqKC...
```

The result is that **every turn on a session with tool-call history fails** when using either provider with reloaded history. We hit this in production when switching a fleet from `openai_compat` to `anthropic-messages`.

The fix resolves the tool name from `tc.Name` falling back to `tc.Function.Name`, and arguments from `tc.Arguments` falling back to JSON-parsing `tc.Function.Arguments` — the exact pattern the `bedrock` and `openai_compat` providers already use for the same reason (`provider_bedrock.go` ~line 643). Table tests in both providers cover the deserialized-history shape, the runtime shape, and the both-empty skip case, and assert the `tool_use`/`tool_result` pairing stays intact. The tests fail without the fix.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue

None open — discovered in production. Same root cause class as the `json:"-"` fallback already handled in `bedrock` (see `provider_bedrock_test.go` "should fallback to Function fields").

- **Reference URL:** https://platform.claude.com/docs/en/agents-and-tools/tool-use/overview (tool_use/tool_result pairing requirement)
- **Reasoning:** `ToolCall.Name`/`.Arguments` are `json:"-"` (`pkg/providers/protocoltypes/types.go`), so any provider that reads only the runtime fields breaks on deserialized history. `openai_compat` reads `tc.Function.Name` directly and `bedrock` implements this exact fallback; the two Anthropic providers were the remaining gap.

## 🧪 Test Environment
- **Hardware:** GCE e2-small (production fleet), Apple Silicon Mac (dev)
- **OS:** Debian 12, macOS (Darwin 25.5)
- **Model/Provider:** claude-opus-4-8 via `anthropic-messages` (api.anthropic.com); fix also applied to the `anthropic` SDK provider
- **Channels:** Slack (via WebSocket gateway)

## 📸 Evidence (Optional)
<details>

Production trace, same agent session, before/after deploying the fix:

```
07-05 20:14:46 error at llm: failover(format): provider=anthropic-messages model=claude-opus-4-8 status=400:
  {"type":"error","error":{"type":"invalid_request_error","message":"messages.2.content.0: unexpected `tool_use_id` found in `tool_result` blocks: toolu_01WqUqKCfp93xmJ1EfECPGTz..."}}
07-05 20:14:46 turn ended (error, 469ms)
--- fix deployed ---
07-05 21:04:49 llm response (stop)
07-05 21:04:50 turn ended (completed, 4960ms)
```

Also verified against the live API with a harness that performs a real tool round-trip, JSON-marshals/unmarshals the assistant message (session-store shape: `Name:""`, `Function.Name:"get_weather"`), replays it with the matching tool_result — no 400.

Local gates: `go build -tags goolm,stdjson ./...` clean; `go test -tags goolm,stdjson ./pkg/providers/... ./pkg/agent/...` all pass. `make check` passes except the `pkg/tools` shell-sandbox tests, which fail identically on a clean checkout of upstream `main` in this macOS environment (Linux CI should be authoritative there).

</details>

## ☑️ Checklist
- [x] I have updated the documentation accordingly. (No documentation changes required.)

### DIFF
--- a/pkg/providers/anthropic/provider.go
+++ b/pkg/providers/anthropic/provider.go
@@ -181,8 +181,15 @@ func buildParams(
 					blocks = append(blocks, anthropic.NewTextBlock(msg.Content))
 				}
 				for _, tc := range msg.ToolCalls {
+					// Resolve tool name: prefer tc.Name, fallback to tc.Function.Name
+					// (tc.Name/tc.Arguments are json:"-" and may be empty when
+					// history is reloaded from the session store)
+					toolName := tc.Name
+					if toolName == "" && tc.Function != nil {
+						toolName = tc.Function.Name
+					}
 					// Skip tool calls with empty names to avoid API errors
-					if tc.Name == "" {
+					if toolName == "" {
 						continue
 					}
 					args := tc.Arguments
@@ -194,7 +201,7 @@ func buildParams(
 					if args == nil {
 						args = map[string]any{}
 					}
-					blocks = append(blocks, anthropic.NewToolUseBlock(tc.ID, args, tc.Name))
+					blocks = append(blocks, anthropic.NewToolUseBlock(tc.ID, args, toolName))
 				}
 				anthropicMessages = append(anthropicMessages, anthropic.NewAssistantMessage(blocks...))
 			} else {
--- a/pkg/providers/anthropic/provider_test.go
+++ b/pkg/providers/anthropic/provider_test.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"net/http"
 	"net/http/httptest"
+	"reflect"
 	"sync/atomic"
 	"testing"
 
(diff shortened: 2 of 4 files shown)


afjcjsbx (approved):
Thanks for the PR, LGTM!

PR #2192, opened 2026-03-30
fix(provider): anthropic_messages sends system as content blocks with cache_control

whtiehack:
## 📝 Description

Fixes #2191

`anthropic_messages` provider ignored `SystemParts` and sent `system` as a flat string, defeating Anthropic prompt caching.

Now uses content blocks array with per-block `cache_control: ephemeral` on static blocks, matching the SDK-based `anthropic` provider behavior.

## 🗣️ Type of Change
- [x] 🐞 Bug fix (non-breaking change which fixes an issue)

## 🤖 AI Code Generation
- [x] 🛠️ Mostly AI-generated (AI draft, Human verified/modified)

## 🔗 Related Issue

Fixes #2191

## 📚 Technical Context (Skip for Docs)
- **Reference URL:** `pkg/providers/anthropic_messages/provider.go`
- **Reasoning:** The SDK-based `anthropic` provider already sends system as content blocks with `cache_control`. The raw HTTP `anthropic_messages` provider should behave the same way.

## 🧪 Test Environment
- **Hardware:** Mac
- **OS:** macOS
- **Model/Provider:** Anthropic Claude (anthropic_messages provider)
- **Channels:** WeCom

## ☑️ Checklist
- [x] My code/docs follow the style of this project.
- [x] I have performed a self-review of my own changes.
- [x] I have updated the documentation accordingly.

PR #3228, opened 2026-07-06
fix(anthropic-messages): send SystemParts as system blocks with cache_control

### LINKED ISSUE #2191: anthropic_messages provider ignores SystemParts, breaks Anthropic prompt caching
`anthropic_messages` provider sends `system` as a flat string, ignoring `SystemParts` content blocks.

The `anthropic` (SDK) provider correctly uses `SystemParts` with per-block `cache_control: ephemeral`, but `anthropic_messages` (raw HTTP) provider only reads `msg.Content` — the concatenated fallback string.

This means:
- `system` is a string instead of content blocks array
- No `cache_control` markers on static blocks
- Dynamic context (`## Current Time`) mixed in the same string causes the entire system prompt prefix to change every minute, defeating Anthropic prompt caching

**provider.go L176-L281:**
```go
var systemPrompt string
// ...
case "system":
    systemPrompt = msg.Content  // ← ignores msg.SystemParts
// ...
result["system"] = systemPrompt  // ← flat string, no cache_control
```

**Expected:** same behavior as the SDK-based `anthropic` provider — output `system` as a content blocks array with `cache_control: ephemeral` on static blocks.

AayushGupta16:
## 📝 Description

The `anthropic_messages` provider flattens system messages to a single string (`msg.Content`), ignoring `SystemParts`. That makes Anthropic prompt caching impossible to express on this provider: per-block `cache_control` markers have nowhere to go, so cache-aware callers get 0% cache hits even when they structure their system prompt correctly. This is exactly the bug reported in #2191 (confirmed there by @SiYue-ZO), and the fix approach follows the direction of @whtiehack's stale PR #2192 — with tests, and aligned with how the SDK-based `anthropic` provider already handles `SystemParts`.

Two commits (kept separate for review; fine to squash):

1. **`feat(providers): plumb prompt-cache token usage through UsageInfo`** — adds `CacheReadInputTokens` / `CacheCreationInputTokens` to `protocoltypes.UsageInfo` and populates them from the Anthropic response. Without this, callers can't observe whether caching works — it's how we verified the fix.
2. **`fix(anthropic-messages): send SystemParts as system blocks with cache_control`** — when any system message carries `SystemParts`, the request's `system` field is sent in block form with each part's `cache_control` preserved; the flat-string form is kept for callers that don't use `SystemParts` (no behavior change for them).

We run a production fleet on this provider (claude-opus-4-8); before this fix the fleet's cache hit rate was 0% with ~288M uncached input tokens/week. With it (plus follow-up breakpoint tuning we'll propose separately), steady-state calls read the full cached prefix at 0.1× input price.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue

Fixes #2191. Supersedes #2192 (closed by the stale-bot; credit to @whtiehack for the original report and fix direction).

- **Reference URL:** https://platform.claude.com/docs/en/build-with-claude/prompt-caching
- **Reasoning:** Anthropic caching is a prefix match keyed on `cache_control` breakpoints attached to individual blocks. A flattened system string cannot carry them. The SDK-based `anthropic` provider already emits `SystemParts` as blocks with `cache_control`; this brings `anthropic_messages` to parity. Block form is only used when a caller actually provides `SystemParts`, so existing flat-string behavior is unchanged.

## 🧪 Test Environment
- **Hardware:** GCE e2-small (production fleet), Apple Silicon Mac (dev)
- **OS:** Debian 12, macOS (Darwin 25.5)
- **Model/Provider:** claude-opus-4-8 via `anthropic-messages` (api.anthropic.com)
- **Channels:** Slack (via WebSocket gateway)

## 📸 Evidence (Optional)
<details>

Live API verification (4-call harness driving `provider.Chat` with a `SystemParts` block carrying `cache_control`, claude-sonnet-4-6):

```
call1: prompt=3 cache_creation=2484 cache_read=0     <- system prefix written to cache
call2: prompt=3 cache_creation=11   cache_read=2484  <- served from cache at 0.1x
call3: prompt=3 cache_creation=16   cache_read=2495  <- prefix reuse grows
call4: prompt=3 cache_creation=94   cache_read=2511  <- stable across tool-use turns
```

Production (org usage report, week of Jun 28–Jul 4, before the fix): 287.8M uncached input tokens, 0% cache hit rate on this provider. After deploying a build with this fix, per-agent `usage` shows `cache_read_input_tokens` covering ~84% of each call's input.

Local gates: `go build -tags goolm,stdjson ./...` clean; `go test -tags goolm,stdjson ./pkg/providers/... ./pkg/agent/...` all pass (new table tests cover block-form emission, cache_control placement, and the flat-string fallback). `make lint-docs` OK. `make check` passes except the `pkg/tools` shell-sandbox tests, which fail identically on a clean checkout of upstream `main` in this macOS environment (Linux CI should be authoritative there).

</details>

## ☑️ Checklist
- [x] I have updated the documentation accordingly. (No documentation changes required.)

### DIFF
--- a/pkg/agent/pipeline_llm.go
+++ b/pkg/agent/pipeline_llm.go
@@ -564,6 +564,8 @@ func (p *Pipeline) CallLLM(
 		llmResponseFields["prompt_tokens"] = exec.response.Usage.PromptTokens
 		llmResponseFields["completion_tokens"] = exec.response.Usage.CompletionTokens
 		llmResponseFields["total_tokens"] = exec.response.Usage.TotalTokens
+		llmResponseFields["cache_read_input_tokens"] = exec.response.Usage.CacheReadInputTokens
+		llmResponseFields["cache_creation_input_tokens"] = exec.response.Usage.CacheCreationInputTokens
 	}
 	logger.DebugCF("agent", "LLM response", llmResponseFields)
 
--- a/pkg/providers/anthropic/provider.go
+++ b/pkg/providers/anthropic/provider.go
@@ -382,9 +382,16 @@ func parseResponse(resp *anthropic.Message) *LLMResponse {
 		ToolCalls:    toolCalls,
 		FinishReason: finishReason,
 		Usage: &UsageInfo{
-			PromptTokens:     int(resp.Usage.InputTokens),
-			CompletionTokens: int(resp.Usage.OutputTokens),
-			TotalTokens:      int(resp.Usage.InputTokens + resp.Usage.OutputTokens),
+			// Anthropic's input_tokens excludes cache-read/creation tokens;
+			// fold them into the headline counters so PromptTokens reflects
+			// the full prompt size (see UsageInfo docs). The streaming path
+			// funnels through here too: Message.Accumulate copies usage from
+			// message_start and updates cache fields from message_delta.
+			PromptTokens:             int(resp.Usage.InputTokens + resp.Usage.CacheCreationInputTokens + resp.Usage.CacheReadInputTokens),
+			CompletionTokens:         int(resp.Usage.OutputTokens),
+			TotalTokens:              int(resp.Usage.InputTokens + resp.Usage.CacheCreationInputTokens + resp.Usage.CacheReadInputTokens + resp.Usage.OutputTokens),
+			CacheReadInputTokens:     int(resp.Usage.CacheReadInputTokens),
+			CacheCreationInputTokens: int(resp.Usage.CacheCreationInputTokens),
 		},
 	}
 }
--- a/pkg/providers/anthropic/provider_test.go
+++ b/pkg/providers/anthropic/provider_test.go
@@ -167,8 +167,10 @@ func TestProvider_ChatRoundTrip(t *testing.T) {
 				{"type": "text", "text": "Hello! How can I help you?"},
 			},
 			"usage": map[string]any{
-				"input_tokens":  15,
-				"output_tokens": 8,
+				"input_tokens":                15,
+				"output_tokens":               8,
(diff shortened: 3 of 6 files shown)


afjcjsbx:
Hi! thanks for the PR, just two hints:

* **Incomplete total tokens calculation for cached requests (`pkg/providers/anthropic_messages/provider.go` and `pkg/agent/pipeline_llm.go`)**
While the PR adds `CacheReadInputTokens` and `CacheCreationInputTokens`, it still calculates `PromptTokens` and `TotalTokens` using only `input_tokens` and `output_tokens`. With Anthropic, these two fields exclude tokens read/written from the cache. As a result, any provider-agnostic consumers that rely solely on the "headline" counters will continue to underestimate cached requests (this already happens in the common logs within `pkg/agent/pipeline_llm.go`).
Furthermore, this behavior is inconsistent with `pkg/providers/cli/claude_cli_provider.go`, which *does* include cache tokens in the totals. We should make this behavior uniform so observability doesn't remain partial and provider-dependent.
* **Missing cache metrics in SDK-based provider path (`pkg/providers/protocoltypes/types.go` and `pkg/providers/anthropic/provider.go`)**
The new cache fields have been added to the shared `UsageInfo` type, but the SDK-based Anthropic path does not populate them yet, even though the SDK exposes them. This means that switching from `anthropic-messages` to `anthropic/oauth/claude` causes the cache metrics to disappear completely, despite `factory_provider.go` presenting these backends as equivalent variants of the same Anthropic family. If the goal of this PR is to fully plumb prompt-cache token usage through `UsageInfo`, this implementation feels incomplete without updating both paths.

PR #3277, opened 2026-07-20
fix(tools): deferred-tool visibility heal + sliding TTL + SSE tool-call index fix

m4n3z40:
## Problem

Tool-discovery promotions live only in memory: a process restart or TTL expiry (a global counter across all of the agent's sessions) silently removes deferred MCP tools from the request while the session history is full of successful calls to them. The model then tries to call a tool that is not in the tools array and, in streaming, returns reasoning-only/empty completions ("The model returned an empty response") or a short narration with no tool call — deterministically, until a new `tool_search` happens. The model almost never re-searches, because its own history shows direct calls working; a fresh chat recovers only because it follows the discovery nudge.

Verified by A/B replay against the same model with a real session transcript: with the referenced tool defs absent from the request, 6/6 streaming completions came back empty; with them present, 48/48 acted correctly — even with the degraded history kept intact.

## Changes

1. **`ToolRegistry.EnsureVisible`** — re-promotes hidden tools referenced by the outgoing messages (called in `CallLLM` before building provider defs). Any tool the model can see itself using in history is guaranteed callable, so existing stuck sessions heal on their next turn and the state survives restarts and expiry.
2. **`ToolRegistry.TouchTools` (sliding TTL)** — a tool used in a round gets a full TTL again before the tick; only idle tools decay. A long task can no longer outlive its own tools mid-flight. (Allows lowering the configured TTL.)
3. **SSE tool-call assembly fix** (`openai_compat`) — iterate the accumulator map's actual keys in sorted order instead of `0..len-1`: a lone tool call streamed at `index: 1` (or an index gap) was silently dropped, turning a tool-use turn into a bare narration that then conditions the session. The in-progress snapshot loop had the same bug.

## Tests

- `TestParseStreamResponse_ToolCallAtNonZeroIndex` / `_ToolCallIndexGap` / `_SnapshotIncludesNonZeroIndex` (fail on the old code)
- `TestEnsureVisible_RevivesExpiredHiddenTools`, `TestTouchTools_SlidingTTLKeepsUsedToolsAlive`, `TestTouchTools_DoesNotReviveExpired`
- `TestToolCallNamesFromMessages`, `TestDiscoveryPromoteTTL_Default`
- Full suite on touched packages ok (3 known macOS-only ShellTool failures unrelated); golangci-lint 0 issues.

Control-plane follow-up: lower discovery TTL 120→20 + submodule bump.


m4n3z40:
Opened against the wrong base by mistake — this belongs to our fork. Apologies for the noise.

PR #3279, opened 2026-07-21
fix(seahorse): prevent tool-call format leakage into LLM summaries

MrTreasure:
## Related: same symptom triggered by seahorse summaries

I just hit this exact class of bug — tool-call format leaking into user messages — but triggered through a different path: **seahorse's `partsToReadableContent`**.

### The seahorse variant

`partsToReadableContent` (in `pkg/seahorse/store.go`) serializes tool calls as `[tool_use: name, args: {...}]` for FTS5 indexing. This format was also used as a fallback in two summary-generation paths (`formatMessagesForSummary` and `truncateSummary` in `short_compaction.go`). When summaries containing `[tool_use: exec, args: {...}]` were fed back into the LLM context, the model learned to mimic this format in its own responses — outputting mock tool calls as visible text content.

### The pipeline

```
partsToReadableContent -> "[tool_use: exec, args: {...}]"
     -> seahorse summary / truncateSummary
     -> LLM context (as summary text)
     -> LLM mimics format in its content output
     -> User sees raw [tool_use: exec, args: {...}] in chat
```

### Fix

PR incoming at: https://github.com/sipeed/picoclaw/pulls (MrTreasure:fix/seahorse-tool-call-leak)

Three changes:
1. **`short_compaction.go`**: Skip messages with empty content in summary/truncation paths — tool calls aren't conversational content
2. **`store.go`**: Change FTS5 format from `[tool_use: name, args: {...}]` to emoji markers — the bracket-colon pattern is what LLMs latch onto
3. **`agent_utils.go`**: Strip residual `[tool_use: ...]` from outbound messages as a safety net

Same root cause as this issue: LLM context leaking internal serialization format → model mimics it → user sees raw tool-call text. Different trigger (seahorse vs Doubao provider), same outcome.

### DIFF
--- a/pkg/agent/agent_utils.go
+++ b/pkg/agent/agent_utils.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"maps"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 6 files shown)


MrTreasure:
Status update: this fix is still actively maintained. The original issue remains reproducible in long-lived Seahorse sessions: raw tool-call protocol text can enter compacted summaries and later surface to users. I have continued validating the summary/compaction path against current upstream master and am reviewing compatibility with tool-only history entries before the next branch update. Feedback on the preferred serialization behavior is welcome.

MrTreasure:
Updated this PR with the complete fix reproduced from a live DeepSeek/OpenAI-compatible response. The provider returned both native tool_calls and a duplicate `[tool_use: ...]` string in assistant content. The update now sanitizes content, reasoning, and tool feedback immediately after AfterLLM hooks and before publishing, persistence, or Seahorse compaction. The parser handles nested JSON, quoted brackets, legacy tool_user/tool_result variants, and truncated blocks. Seahorse still preserves searchable tool names, arguments, and results using protocol-neutral human-readable text. Rebased onto current upstream main. Verified locally: `go test ./pkg/agent ./pkg/seahorse ./pkg/providers/openai_compat ./pkg/providers/common`.

PR #3282, opened 2026-07-21
feat(nodes): add policy-gated system exec

bogdanovich:
## What changed or was removed

- add opt-in `system.exec.v1` to the slim node companion
- execute direct canonical argv without a shell
- enforce companion-owned executable, working-root, environment, timeout, and output limits before durable acceptance and again before execution
- persist typed timeout/start/wait failures and preserve non-zero process exits as successful typed results
- terminate process groups on Unix and confirm explicit cancellation only after the child has stopped
- keep the capability absent from the advertised catalog unless `system_exec` is explicitly configured

## Why

The node MVP needs useful remote execution without turning pairing or gateway approval into unlimited host authority. This adds the narrow execution primitive described by the node-companion architecture while retaining an independent node-local maximum policy.

## User and developer impact

Operators may opt in with `system_exec.working_roots`, `system_exec.executables`, and an optional environment-name allowlist, plus the existing local command policy. Existing companion configurations and catalogs are unchanged by default. Commands receive only explicitly allowlisted environment variables; shell interpreters are rejected.

## Validation

- `go test -race ./pkg/nodes/... ./cmd/picoclaw-node`
- `go test -tags goolm,stdjson ./...`
- `golangci-lint run --build-tags=goolm,stdjson`
- `GOOS=windows GOARCH=amd64 go test -c ./pkg/nodes/companion`
- `GOOS=windows GOARCH=amd64 go build ./cmd/picoclaw-node`
- `go generate ./...`
- focused subprocess tests cover opt-in advertisement, executable/env denial before ledger acceptance, cwd symlink escape, bounded output, non-zero exit, timeout, and confirmed cancellation

## Later PRs

- operator-facing node configuration and lifecycle/install tooling
- typed service status/log capabilities
- real-process loopback WSS end-to-end coverage and final MVP audit


bogdanovich:
Opened against upstream by CLI repository auto-detection; this ForgeClaw-specific change belongs in the fork.

PR #3297, opened 2026-07-26
fix(security): harden remote prompt and exec boundaries

SiYue-ZO:
## Summary

- keep remote sender and chat metadata in a normalized user-role envelope instead of provider system instructions
- default remote exec to disabled, require independent per-call approval when enabled, and enforce origin policy again at execution time
- migrate configs to schema v4 and add startup warnings for risky remote-exec configurations
- require Telegram allowlists for new Web setups and document the remote-channel security boundary

## Security behavior

Remote turns no longer receive the exec tool unless explicitly enabled. Forged calls, hook-rewritten tool names, missing origin context, and unapproved remote calls are rejected at runtime. Sender-controlled identity fields are normalized and kept out of OpenAI, Azure, Codex, and Anthropic system instructions.


SiYue-ZO:
@imguoguo

PR #3329, opened 2026-08-11
fix(line): warn on inert webhook_host / webhook_port instead of seeding them

### LINKED ISSUE #3328: [BUG] line.settings.webhook_host / webhook_port are never read — they have defaults and docs but no consumer
## Summary

`channel_list.line.settings.webhook_host` and `webhook_port` exist in the config
struct, are given defaults, and are documented — but nothing in the codebase reads
them. Setting either has no effect, and there is no warning to say so.

## Evidence

Declared:

- `pkg/config/config.go:606-607` — `WebhookHost string`, `WebhookPort int`, both with
  `json:` tags and `env:` bindings
- `pkg/config/defaults.go:546-547` — defaults `"webhook_host": "0.0.0.0"`,
  `"webhook_port": 18791`
- `docs/channels/line/README.md` documents them as configurable

Consumed: nowhere. `grep -rn "WebhookPort\|WebhookHost" --include=*.go pkg/channels/`
returns no reads. The LINE channel only implements `WebhookPath()`
(`pkg/channels/line/line.go:126-132`) and is mounted as a handler on the **shared**
HTTP server (`pkg/channels/manager.go:1144`, `m.mux.Handle(wh.WebhookPath(), wh)`).
There is no per-channel listener, so there is no port for the channel to bind.

The port that actually serves the webhook is `gateway.port` (`pkg/config/defaults.go:302`,
default 18790).

## Why this is worth fixing rather than documenting

On a deployed device the config read `"webhook_port": 18791` while the process was
listening on 18790:

```
$ grep webhook_port ~/.picoclaw/config.json
  "webhook_port": 18791
$ netstat -ltn | grep 1879
tcp   0   0 0.0.0.0:18790   0.0.0.0:*   LISTEN
```

Both numbers are plausible, adjacent, and neither the config nor the logs indicate
which one is real. Anyone setting up a reverse proxy or tunnel in front of this reads
the config, points at 18791, and gets a listener that never answers — while every
other signal (channel enabled, process healthy, tunnel connected, credentials valid)
says the system is fine. The failure only surfaces as webhooks silently 404ing.

An inert setting that looks configurable is worse than a missing one: it produces
confident, wrong conclusions.

## Suggested fix

Either:

1. **Remove them** — delete the fields from `LINEConfig`, drop the two lines from
   `defaults.go`, and remove them from the channel docs. Since nothing reads them,
   removal cannot change behaviour. Add a note to the LINE docs saying the webhook is
   served by `gateway.port` on the path from `webhook_path`.

2. **Honour them** — give the LINE channel its own listener when they are set. This is
   more work and probably not wanted, since the shared-server design looks deliberate.

(1) seems right. If backward compatibility with existing config files is a concern,
the keys can stay accepted-and-ignored, but then loading a config that sets them to a
non-default value should log a warning naming `gateway.port` as the real setting.

## Related

The same shape as the skill-description limit (a rule enforced with no reachable
diagnostic) and `gateway.hot_reload` (a change accepted but not applied): the system
reports success while the thing did not take effect.

ex-takashima:
Fixes #3328.

## Problem

`line.settings.webhook_host` / `webhook_port` are declared, defaulted, and env-bound, but nothing reads them. The LINE channel implements `WebhookPath()` and is mounted as a handler on the **shared** gateway HTTP server (`pkg/channels/manager.go`, `m.mux.Handle(wh.WebhookPath(), wh)`) — it has no listener of its own, so there is no port for it to bind. The only other references are the openclaw importer, a test fixture, and a Web UI label.

The webhook is actually served on `gateway.port` (default **18790**), while every fresh config was seeded with `webhook_port: 18791`. Both numbers are plausible and adjacent, and nothing distinguishes them. Anyone putting a reverse proxy or tunnel in front of PicoClaw reads the config, points at 18791, and gets a listener that never answers — while channel-enabled, process-healthy, tunnel-connected, and credentials-valid all report fine. The failure surfaces only as webhooks silently 404ing.

## Change

Option (1) from the issue — stop advertising them — plus the warning the issue suggests for configs that already set them.

- **Defaults** (`pkg/config/defaults.go`): drop both keys from the LINE channel, so new configs no longer advertise a port that is never bound. Only `webhook_path` remains, which is genuinely honoured.
- **Migration** (`pkg/migrate/sources/openclaw/openclaw_config.go`): the importer copied both keys unconditionally, so every migrated `config.json` got an inert `"webhook_host": "", "webhook_port": 0` even when the source never set them. Now carried over only when actually set.
- **Struct** (`pkg/config/config.go`): both fields kept, marked `Deprecated:`, with `omitempty` added. They are what makes a stale config *detectable* — that is the only reason to keep them, and they are no longer written back out.
- **Warning** (`pkg/channels/line/line.go`): when either is set, log one warning at channel startup naming `gateway.host` / `gateway.port` as the real setting, and echo the effective `webhook_path`.

Net effect: fresh and migrated configs are clean, and a config that still sets these says so out loud instead of failing silently.

## Behaviour

Startup with a legacy config carrying both keys:

```
WARN [line] Ignoring LINE webhook bind settings: the webhook is served by the shared gateway HTTP server
  ignored=webhook_host, webhook_port  use_instead=gateway.host / gateway.port  webhook_path=/webhook/line
```

Nothing else changes — the fields were inert before this PR and are inert after it. Deleting the fields outright would also have been load-compatible (unknown keys inside a channel `settings` block are silently ignored, since `Settings` is a `RawNode` decoded with a plain `json.Unmarshal`; only *top-level* unknown fields are a hard error) — but then a config still setting them would go on being ignored in silence, which is the actual complaint in #3328.

## Tests

- `TestInertBindKeys` — table test over unset / host-only / port-only / both
- `TestWebhookPathIgnoresBindSettings` — path resolution unaffected by the bind keys
- `TestDefaultChannels_LINEOmitsInertWebhookBindKeys` — defaults no longer seed them
- `TestLINESettings_InertWebhookBindKeysRemainLoadable` — a config setting them still decodes
- `TestLINESettings_InertWebhookBindKeysNotSerializedWhenUnset` — `omitempty` keeps them out of saved configs
- `TestToStandardConfig_LINEOmitsUnsetWebhookBindKeys` / `TestToStandardConfig_LINEKeepsSetWebhookBindKeys` — migration output, both directions

`go vet` and `go test ./pkg/config/... ./pkg/channels/line/... ./pkg/migrate/...` pass. (`pkg/channels/matrix` fails to build in my container for an unrelated pre-existing reason — missing libolm C headers.)

## Notes for reviewers

**Web UI left alone** — `web/frontend/src/components/channels/channel-forms/generic-form.tsx` still maps `webhook_host` / `webhook_port` to "Webhook listening host/port" descriptions. With the keys gone from defaults and migration, the field only renders for configs that already carry it; removing the entries would also orphan the string in ~20 locale files. Happy to fold that in if you'd prefer.

**Docs need no change** — the issue states these are documented in `docs/channels/line/README.md`, but no LINE README (en/ja/zh/fr/pt-br/vi) mentions `webhook_host`, `webhook_port`, or 18791. The en README already correctly says the shared gateway serves the webhook on 18790.

### DIFF
--- a/pkg/channels/line/line.go
+++ b/pkg/channels/line/line.go
@@ -61,6 +61,14 @@ func NewLINEChannel(
 		return nil, fmt.Errorf("line channel_secret and channel_access_token are required")
 	}
 
+	if keys := inertBindKeys(cfg); len(keys) > 0 {
+		logger.WarnCF("line", "Ignoring LINE webhook bind settings: the webhook is served by the shared gateway HTTP server", map[string]any{
+			"ignored":      strings.Join(keys, ", "),
+			"use_instead":  "gateway.host / gateway.port",
+			"webhook_path": webhookPath(cfg),
+		})
+	}
+
 	client, err := messaging_api.NewMessagingApiAPI(
 		cfg.ChannelAccessToken.String(),
 		messaging_api.WithHTTPClient(&http.Client{Timeout: 30 * time.Second}),
@@ -126,10 +134,29 @@ func (c *LINEChannel) Stop(ctx context.Context) error {
 
 // WebhookPath returns the path for registering on the shared HTTP server.
 func (c *LINEChannel) WebhookPath() string {
-	if c.config.WebhookPath != "" {
-		return c.config.WebhookPath
+	return webhookPath(c.config)
+}
+
+func webhookPath(cfg *config.LINESettings) string {
+	if cfg.WebhookPath != "" {
+		return cfg.WebhookPath
 	}
 	return "/webhook/line"
+}
+
+// inertBindKeys reports which webhook bind settings are set but unused. The
+// LINE channel is mounted as a handler on the shared gateway HTTP server and
+// has no listener of its own, so webhook_host and webhook_port never bind
+// anything. They are still parsed so that existing config files keep loading.
+func inertBindKeys(cfg *config.LINESettings) []string {
+	var keys []string
+	if cfg.WebhookHost != "" {
+		keys = append(keys, "webhook_host")
+	}
+	if cfg.WebhookPort != 0 {
+		keys = append(keys, "webhook_port")
+	}
+	return keys
 }
 
 // ServeHTTP implements http.Handler for the shared HTTP server.
--- a/pkg/channels/line/line.go
+++ b/pkg/channels/line/line.go
@@ -62,11 +62,15 @@ func NewLINEChannel(
 	}
 
 	if keys := inertBindKeys(cfg); len(keys) > 0 {
-		logger.WarnCF("line", "Ignoring LINE webhook bind settings: the webhook is served by the shared gateway HTTP server", map[string]any{
-			"ignored":      strings.Join(keys, ", "),
-			"use_instead":  "gateway.host / gateway.port",
-			"webhook_path": webhookPath(cfg),
-		})
+		logger.WarnCF(
+			"line",
+			"Ignoring LINE webhook bind settings: the webhook is served by the shared gateway server",
+			map[string]any{
+				"ignored":      strings.Join(keys, ", "),
+				"use_instead":  "gateway.host / gateway.port",
+				"webhook_path": webhookPath(cfg),
+			},
+		)
 	}
 
 	client, err := messaging_api.NewMessagingApiAPI(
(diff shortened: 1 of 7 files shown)


ex-takashima:
Still active and waiting for review — the change is complete and all CI checks are green (Linter / Security Check / Tests / Integration Tests). Happy to address any feedback.

PR #3337, opened 2026-08-14
Fix/mcp failure hangs agent loop

### LINKED ISSUE #3269: [BUG]If the MCP server connection fails, the agent loop will hang, causing the Picoclaw chat interface to stop replying to users.
## Quick Summary
If the MCP server connection fails, the agent loop will hang, causing the Picoclaw chat interface to stop replying to users.

##  Environment & Tools
- **PicoClaw Version:** picoclaw nightly (git: 2cf030d2)
- **Go Version:** go1.25.11
- **AI Model & Provider:** Qwen3
- **Operating System:** Android (libpicoclaw-web.so & libpicoclaw.so)
- **Channels:** pico channel

## 📸 Steps to Reproduce
1. In the Picoclaw-Launch web interface, configure an incorrect MCP server (this server is actually unreachable).
2. In the Picoclaw-Launch web interface, click the button to restart the gateway.
3. In the Picoclaw-Launch web chat interface, enter a message and find that Picoclaw no longer replies. 

## ❌ Actual Behavior
Picoclaw no longer replies to messages.

## ✅ Expected Behavior
Picoclaw should reply to messages normally.

## 💬 Additional Context
The problematic code is in agent.go. After MCP initialization fails, the agent loop function returns directly.
func (al *AgentLoop) Run(ctx context.Context) error {
	...
	if err := al.ensureMCPInitialized(ctx); err != nil {
		return err  // Here returns the error directly, causing the entire agent loop to exit.
	}
       ...
}

kuzmichus:
## 📝 Description

Fixes a hang in the agent loop when an MCP server connection fails. Previously, if `ensureMCPInitialized` returned an error (e.g. an unreachable/broken MCP server), `AgentLoop.Run` propagated the error and exited, so the chat interface stopped replying to users entirely until a restart.

This PR makes MCP initialization failure **non-fatal**: the error is logged as a warning and the agent keeps running without MCP tools. The change is applied to every entry point that treated MCP failure as fatal:

- `Run` — keeps the agent loop alive and still processes inbound messages (`pkg/agent/agent.go`).
- `Continue` — no longer drops the claimed session, so queued steering messages are still processed (`pkg/agent/steering.go`).
- `ProcessDirectWithChannel` / `ProcessHeartbeat` — direct and cron/heartbeat calls continue without MCP tools (`pkg/agent/agent_message.go`).

Hooks initialization remains fatal in the direct/heartbeat paths (unchanged behavior), but is also downgraded to a warning inside `Run`/`Continue`, since killing the whole loop there would cause the same hang class.

A regression test (`TestAgentLoopRun_DoesNotExitOnMCPInitFailure`) verifies that with a broken MCP server the loop stays alive, replies to inbound messages, and exits cleanly on context cancel.

## 🗣️ Type of Change

## 🤖 AI Code Generation

## 🔗 Related Issue

Fixes #3269

- **Reference URL:** https://github.com/sipeed/picoclaw/issues/3269
- **Reasoning:** The buggy code returned the MCP init error from `Run`, exiting the agent loop and hanging the chat. `mcpRuntime` already caches the init error (`initOnce`), so a failure does not need to be fatal — the same "log and continue" pattern is already used in `ReloadProviderAndConfig` and the `/mcp` command handlers. A config reload (`mcp.reset()`) already retries initialization after the server is fixed.

## 🧪 Test Environment
- **Hardware:** PC (x86_64)
- **OS:** Linux
- **Model/Provider:** N/A — unit tests with mock provider
- **Channels:** N/A — agent loop (inbound/outbound bus) tested directly

## 📸 Evidence (Optional)

<details>
<summary>Click to view Logs</summary>

Regression test output — MCP failure is logged as a warning, the loop keeps replying:

```
WRN agent agent.go:153 > MCP initialization failed, continuing without MCP tools error="failed to load MCP servers: failed to connect to server broken: ..."
INF agent agent_message.go:137 > Processing message from telegram:user1: hello
INF agent agent_outbound.go:86 > Published outbound response channel=telegram chat_id=chat1 content_len=13
--- PASS: TestAgentLoopRun_DoesNotExitOnMCPInitFailure (0.30s)
```

Full package run: `go test ./pkg/agent/` → `ok`; `make vet` and `gofmt` pass.

</details>

## ☑️ Checklist

### DIFF
--- a/pkg/agent/agent.go
+++ b/pkg/agent/agent.go
@@ -146,10 +146,12 @@ func (al *AgentLoop) Run(ctx context.Context) error {
 	al.running.Store(true)
 
 	if err := al.ensureHooksInitialized(ctx); err != nil {
-		return err
+		logger.WarnCF("agent", "Hooks failed to initialize, continuing without hooks",
+			map[string]any{"error": err.Error()})
 	}
 	if err := al.ensureMCPInitialized(ctx); err != nil {
-		return err
+		logger.WarnCF("agent", "MCP initialization failed, continuing without MCP tools",
+			map[string]any{"error": err.Error()})
 	}
 
 	idleTicker := time.NewTicker(100 * time.Millisecond)
--- a/pkg/agent/agent_mcp_test.go
+++ b/pkg/agent/agent_mcp_test.go
@@ -11,7 +11,9 @@ import (
 	"errors"
 	"strings"
 	"testing"
+	"time"
 
+	"github.com/sipeed/picoclaw/pkg/bus"
 	"github.com/sipeed/picoclaw/pkg/config"
 	"github.com/sipeed/picoclaw/pkg/mcp"
 	agenttools "github.com/sipeed/picoclaw/pkg/tools"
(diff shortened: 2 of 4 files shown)


codeincarnate:
I'd be happy to review this one if it would help move things forward. Very important in my case as MCP servers are essential the use cases I'm targeting. Not unusual at all for servers to go down and having the agent brought down with it is painful 😢

PR #3340, opened 2026-08-17
fix(slack): set FileSize on media upload params

### LINKED ISSUE #3338: [BUG] Slack does not attach image media content
## Quick Summary

Slack media uploads always fail with `file.upload.v2: file size cannot be 0`. `SendMedia` builds `slack.UploadFileParameters` without setting `FileSize`, so the slack-go SDK rejects every upload before any network call.

##  Environment & Tools
- **PicoClaw Version:**  picoclaw 0.3.1 (git: 2cf030d2)
- **Go Version:** 1.25.11
- **AI Model & Provider:** Ollama - GLM-5.2:cloud
- **Operating System:** WSL Ubuntu 22.04
- **Channels:** Slack

## 📸 Steps to Reproduce
1. Ask agent to generate a chart
Image

## ❌ Actual Behavior

The agent generates the chart but cannot send it to slack

Image

## ✅ Expected Behavior

The agent sends generated chart to the slack as a media

### Root cause

`pkg/channels/slack/slack.go:213` builds `slack.UploadFileParameters` without `FileSize`, leaving it at zero-value:

```go
err = c.uploadFileFn(ctx, slack.UploadFileParameters{
    Channel:         channelID,
    ThreadTimestamp: threadTS,
    File:            localPath,
    Filename:        filename,
    Title:           title,
})
```

`slack-go` v0.23.1 rejects this before any network I/O (`files.go:547`):

```go
if params.FileSize == 0 {
    return nil, fmt.Errorf("file.upload.v2: file size cannot be 0")
}
```

The v2 upload flow requires `length` upfront in `files.getUploadURLExternal`. The SDK does **not** derive the size from the `File` path.

### Likely origin

Regression from #2802 (`slack-go` 0.17.3 → 0.23.0), later pinned by #2875. The legacy `files.upload` endpoint didn't require a size; the v2 three-step flow does. The call site was never updated. `slack_test.go:219` doesn't catch it because it mocks `uploadFileFn`, bypassing SDK validation.

### Secondary bug: wrong error classification

`slack.go:221` maps the failure to `channels.ErrTemporary`. A deterministic permanent error triggers 3 useless retries (`manager.go:1782`) and misleads the agent into burning iterations on workarounds.

Logs — 4 identical attempts in 4 seconds:

```
19:41:13 ERR slack slack.go:221 > Failed to upload media error="file.upload.v2: file size cannot be 0" filename=grafico_pizza.png
19:41:14 ERR slack slack.go:221 > Failed to upload media error="file.upload.v2: file size cannot be 0" filename=grafico_pizza.png
19:41:15 ERR slack slack.go:221 > Failed to upload media error="file.upload.v2: file size cannot be 0" filename=grafico_pizza.png
19:41:17 ERR slack slack.go:221 > Failed to upload media error="file.upload.v2: file size cannot be 0" filename=grafico_pizza.png
19:41:17 ERR channels manager.go:1782 > SendMedia failed error="slack send media: temporary failure" channel=slack chat_id=D0B8H010C3V retries=3
```

The file itself is fine — the agent verified it via `exec`:

```
-rw-r--r-- 1 octavio octavio 36515 Aug 16 19:33 grafico_pizza.png
PNG image data, 900 x 900, 8-bit
```

### Suggested fix

```go
info, err := os.Stat(localPath)
if err != nil {
    logger.ErrorCF("slack", "Failed to stat media file", map[string]any{
        "path":  localPath,
        "error": err.Error(),
    })
    continue
}

err = c.uploadFileFn(ctx, slack.UploadFileParameters{
    Channel:         channelID,
    ThreadTimestamp: threadTS,
    File:            localPath,
    FileSize:        int(info.Size()),
    Filename:        filename,
    Title:           title,
})
```

Plus: classify `file.upload.v2:` validation errors as `ErrSendFailed` instead of `ErrTemporary` to avoid pointless retries.

### Scope

Only `pkg/channels/slack/`. `slack_webhook` does not use `UploadFileParameters`.

Happy to open a PR if useful.

octavioturra:
## 📝 Description

SendMedia built slack.UploadFileParameters without FileSize, leaving it at zero-value. slack-go v0.23.1 rejects this before any network call because the files.upload.v2 flow requires the length upfront in files.getUploadURLExternal, and the SDK does not derive it from the File path. Every Slack media upload failed with
"file.upload.v2: file size cannot be 0".

Stat the resolved media file and pass its size. Also classify file.upload.v2 validation errors as ErrSendFailed instead of ErrTemporary, since they are deterministic and the previous mapping triggered three pointless retries per attempt.

Regression from the slack-go 0.17.3 -> 0.23.0 bump (#2802): the legacy files.upload endpoint did not require a size. slack_test.go did not catch it because it mocks uploadFileFn, bypassing SDK validation.

## 🗣️ Type of Change
- [ ] 📖 Documentation update

## 🤖 AI Code Generation

## 🔗 Related Issue

Fixes #3338

- **Reference URL:**
  - slack-go v0.23.1 validation guard: https://github.com/slack-go/slack/blob/v0.23.1/files.go#L547
  - Slack files.getUploadURLExternal (requires `length`): https://api.slack.com/methods/files.getUploadURLExternal
  - Originating dependency bump: #2802
- **Reasoning:** The v2 upload flow is three-step — getUploadURLExternal, PUT to the returned URL, completeUploadExternal. The first step needs the byte length before the file is read, so the SDK cannot infer it from the `File` path and fails fast instead. Passing `os.Stat().Size()` is the minimal fix at the call site; no SDK change or version pin is needed. The error-classification change is included because the two bugs compound: a permanent validation failure mapped to `ErrTemporary` caused the manager to retry three times and reported "temporary failure" to the agent, which then burned several LLM iterations attempting workarounds (base64 encoding, renaming the extension) for a condition that could never resolve on retry.

## 🧪 Test Environment
- **Hardware:** PC (x86_64)
- **OS:** WSL Ubuntu 22.04
- **Model/Provider:** GLM-5.2:cloud via Ollama
- **Channels:** Slack (Socket Mode)

## 📸 Evidence (Optional)
image

### DIFF
--- a/pkg/channels/slack/slack.go
+++ b/pkg/channels/slack/slack.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"strings"
 	"sync"
+	"os"
 
 	"github.com/slack-go/slack"
 	"github.com/slack-go/slack/slackevents"
@@ -210,10 +211,20 @@ func (c *SlackChannel) SendMedia(ctx context.Context, msg bus.OutboundMediaMessa
 			title = filename
 		}
 
+		info, err := os.Stat(localPath)
+		if err != nil {
+			logger.ErrorCF("slack", "Failed to stat media file", map[string]any{
+				"path":  localPath,
+				"error": err.Error(),
+			})
+			continue
+		}
+
 		err = c.uploadFileFn(ctx, slack.UploadFileParameters{
 			Channel:         channelID,
 			ThreadTimestamp: threadTS,
 			File:            localPath,
+			FileSize:        int(info.Size()),
 			Filename:        filename,
 			Title:           title,
 		})


octavioturra:
Any comments or opportunity of approval?

